PR #3, opened 2020-04-28
[REVIEW] migrating cumlHandle_impl (and related dependencies) into raft

teju85:
@dantegd @JohnZed @cjnolet @afender for review.

Notable changes from cuML:
1. a base class called `raft::Allocator` to reduce code duplication between `deviceAllocator` and `hostAllocator`
2. handle is called `raft::handle_t`, but I'm open to suggestions for a better name
3. disabling of logging features, as we need to see how to expose the current spdlog wrapper in cuml, without actually needing users of raft (cugraph and cuml currently) worry about underlying spdlog library.
4. `raft::handle_t` doesn't contain the comms interface. It is better tackled in a separate PR.
5. Enable clang-tidy checks to enforce consistent naming convention.

This PR is still a WIP, but filing it early to solicit feedback about the interface.


seunghwak:
I see multiple styles in naming classes,

e.g.
class Allocator (started with upppercase)
class deviceAllocator (start with_lowercase)
class buffer_base (use _ in the middle)
class handle_t (use _t at the end)

And when do we use which style?

cjnolet:
> raft::handle_t doesn't contain the comms interface. It is better tackled in a separate PR.

This is perfect. I've been planning to move the comms over right after the handle is moved over.

cpp/include/raft/host_buffer.hpp
```diff
@@ -0,0 +1,75 @@
...
+  using const_iterator = typename buffer_base<T, hostAllocator>::const_iterator;
+  using reference = typename buffer_base<T, hostAllocator>::reference;
+  using const_reference =
+    typename buffer_base<T, hostAllocator>::const_reference;
+
+  host_buffer() = delete;
+
+  host_buffer(const host_buffer& other) = delete;
+
+  host_buffer& operator=(const host_buffer& other) = delete;
```
seunghwak, line 59: No support for move constructor and move assignment operator?

And having a constructor (and an assignment operator) taking device_buffer as an input parameter will allow more thrust-like coding style.

e.g.
```
thrust::device_vector<int> d_ints(100, 0);
thrust::host_vector<int> h_ints = d_ints;
```

cpp/include/raft/allocator.hpp
```diff
@@ -0,0 +1,118 @@
...
+ * - Allocations don't need to be zero copy accessible form a device.
+ */
+class hostAllocator : public Allocator {
+};  // class hostAllocator
+
+/** Default cudaMalloc/cudaFree based device allocator */
+class defaultDeviceAllocator : public deviceAllocator {
+ public:
+  void* allocate(std::size_t n, cudaStream_t stream) override {
+    void* ptr = 0;
```
afender, line 90: Should RMM be the default device allocator? RMM could still default to cudaMalloc down the road afaik.

cpp/include/raft/cudart_utils.h
```diff
@@ -0,0 +1,224 @@
...
+void printHostVector(const char* variableName, const T* hostMem,
+                     size_t componentsCount) {
+  printHostVector(variableName, hostMem, componentsCount, std::cout);
+  std::cout.flush();
+}
+
+template <class T, class OutStream>
+void printDevVector(const char* variableName, const T* devMem,
+                    size_t componentsCount, OutStream& out) {
+  T* hostMem = new T[componentsCount];
```
afender, line 209: We also have a couple of helper functions that we should merge here as well (that's really a note to mystelf :) ). Sometimes we have duplicates with slightly different implementations like for printing we use :
`thrust::copy(dev_ptr, dev_ptr + n, std::ostream_iterator<T>(std::cout, " "));`

cpp/include/raft/buffer_base.hpp
```diff
@@ -0,0 +1,183 @@
...
+
+namespace raft {
+
+/**
+ * @brief Base for all RAII-based owning of temporary memory allocations. This
+ *        class should ideally not be used by users directly, but instead via
+ *        the child classes `device_buffer` and `host_buffer`.
+ */
+template <typename T, typename Allocator>
+class buffer_base {
```
seunghwak, line 31: And why do we have this considering that rmm has device_buffer?

cpp/include/raft/handle.hpp
```diff
@@ -0,0 +1,243 @@
...
+  }
+
+  void setHostAllocator(std::shared_ptr<hostAllocator> allocator) {
+    _hostAllocator = allocator;
+  }
+  std::shared_ptr<hostAllocator> getHostAllocator() const {
+    return _hostAllocator;
+  }
+
+  cublasHandle_t getCublasHandle() const {
```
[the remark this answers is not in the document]
afender: Also, I've seen several libs using the singleton design pattern for CUDA libraries' handles. The goal being that one can do a `get<Some>Handle()` without having to carry the main library handle around or repaying the expensive handle creation.

https://github.com/rapidsai/cugraph/blob/905d7ff4d3b1551e1056884b771c12a41bc20d1a/cpp/src/utilities/cusparse_helper.h#L45

cpp/include/raft/cudart_utils.h
```diff
@@ -0,0 +1,224 @@
...
+
+  /** dtor */
+  virtual ~Exception() throw() {}
+
+  /** get the message associated with this exception */
+  virtual const char* what() const throw() { return msg.c_str(); }
+
+ private:
+  /** message associated with this exception */
+  std::string msg;
```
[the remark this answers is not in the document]
seunghwak: And in reviewing a cuGraph PR, @kaatish showed me this 

https://stackoverflow.com/questions/3136594/naming-convention-underscore-in-c-and-c-sharp-variables/20496955#20496955

saying

```
Please DO NOT use UNDERSCORES before any variable name or parameter name in C++!!!

Names beginning with an underscore or a double underscore are RESERVED for the C++ implementers. Names with an underscore are reserved for the library to work.

If you have a read at the C++ Coding Standard, you will see that in the very first page it says:
```

and both Google C++ coding style guide (https://google.github.io/styleguide/cppguide.html#General_Naming_Rules) and Herb Sutter & Andrei Alexandrescu (big names in the C++ world, https://www.drdobbs.com/c-coding-standards/184401863) recommend using trailing _. I think we'd better follow this convention.

dillon-cullinan:
rerun tests

dillon-cullinan:
rerun tests

GPUtester:

cpp/include/raft/allocator.hpp
```diff
@@ -0,0 +1,118 @@
...
+   * asynchronous in the sense that it is only save to use after all work
+   * submitted to the passed in stream prior to the call to allocate has
+   * completed. If the allocation is used before, e.g. in another stream the
+   * behaviour may be undefined.
+   * @todo: Add alignment requirments.
+   *
+   * @param[in] n         number of bytes to allocate
+   * @param[in] stream    stream to issue the possible asynchronous allocation in
+   */
+  virtual void* allocate(std::size_t n, cudaStream_t stream) = 0;
```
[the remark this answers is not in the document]
teju85: This is the so called "bare metal" interface. For RAII interface, one should be using `device_buffer` or `host_buffer` classes.

cpp/include/raft/buffer_base.hpp
```diff
@@ -0,0 +1,183 @@
...
+                                   cudaMemcpyDefault, _stream));
+      }
+      if (nullptr != _data) {
+        _allocator->deallocate(_data, _capacity * sizeof(value_type), _stream);
+      }
+      _data = new_data;
+      _capacity = new_capacity;
+    }
+  }
+
```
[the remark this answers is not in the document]
teju85: We didn't see any such requirement in cuML and thus was not added. Maybe we can think about this.

cpp/include/raft/cudart_utils.h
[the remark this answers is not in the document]
teju85: Please note that this code has been copy-pasted from cuML as-is and this code is one of the oldest components of cuML. So, there are certainly naming convention violations here. We can certainly fix them, but I'd like to do so with a `.clang-tidy` approach. What say?

cpp/include/raft/cudart_utils.h
```diff
@@ -0,0 +1,224 @@
...
+/** helper method to get multi-processor count parameter */
+inline int getMultiProcessorCount() {
+  int devId;
+  CUDA_CHECK(cudaGetDevice(&devId));
+  int mpCount;
+  CUDA_CHECK(
+    cudaDeviceGetAttribute(&mpCount, cudaDevAttrMultiProcessorCount, devId));
+  return mpCount;
+}
+
```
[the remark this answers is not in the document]
teju85: Ah... good point. Let me do that.

cpp/include/raft/handle.hpp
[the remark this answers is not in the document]
teju85: good point @seunghwak . It'll be better to surround the logic inside this method (and related ones below) with a lock.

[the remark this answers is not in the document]
teju85: @afender singleton pattern is a good choice for centrally used resources (eg: logger), but for cuda libaries, they make the design a bit constrained. Moreover, having ability to create multiple handles, give them different streams, is inline with other cuda libraries too. So, it'll make the usage of this library much more intuitive. What say?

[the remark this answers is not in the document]
teju85: @seunghwak by marking these methods as `const`, we call use these methods via a const object of `handle_t` as well. That was the initial motive behind this design in cuML.

cpp/include/raft/host_buffer.hpp
teju85: Good point. This will certainly make our code simpler. This should be easy to add.

cpp/include/raft/allocator.hpp
teju85: The default is setup to be `cudaMalloc/cudaFree` as it comes with the least amount of dependencies. We could certainly add an `rmmDeviceAllocator` *header* in case our users would like to use rmm as their default and also adding that dependency.

cpp/include/raft/cudart_utils.h
teju85: Fair enough. Can you point me to those and I'll try to merge them the best way possible?

cpp/include/raft/buffer_base.hpp
teju85: rmm certainly has a device_buffer, but it doesn't have `host_buffer` (afaik). We need both for best performance.

cpp/include/raft/handle.hpp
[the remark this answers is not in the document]
afender: > @afender singleton pattern is a good choice for centrally used resources (eg: logger), but for cuda libaries, they make the design a bit constrained. Moreover, having ability to create multiple handles, give them different streams, is inline with other cuda libraries too. So, it'll make the usage of this library much more intuitive. What say?

Yup, I'm also fine with this approach. If we expect workflows that create multiple handles per process then it makes more sense.

cpp/include/raft/cudart_utils.h
afender: This was mostly for visibility, it doesn't have to be in this PR unless you would like to reconcile now. 

cuGraph utils are here:
https://github.com/rapidsai/cugraph/blob/branch-0.14/cpp/src/utilities/graph_utils.cuh

cpp/include/raft/buffer_base.hpp
teju85: also, using rmm's device_buffer means we are tied to only rmm, but this interface is generic enough to support any allocator.

seunghwak: rmm supports multiple different allocators, and I was worried about reinventing the wheel and duplicating efforts; and I thought RAPIDS projects are tied to rmm, but this is just my 2 cents :-)

teju85: cuML is NOT tied to rmm :) For eg: one can even use the basic cub caching-allocator too for all allocations done inside cuML and it'll work just fine!

Do you know if rmm supports host-memory (cudaMallocHost) based allocators? AFAIK, this is not the case.

teju85:
@seunghwak @afender @cjnolet @dantegd As discussed offline, I have added clang-tidy support also. However, before I go ahead and fix errors raised by tidy script, I'd like you all to confirm whether the rules mentioned in the `cpp/.clang-tidy` are OK or not. (Disclaimer: this .clang-tidy config file is courtesy xgboost project)

cpp/include/raft/allocator.hpp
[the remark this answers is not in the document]
teju85: I have also explicitly documented this in the doxygen comments for this doc. Let me know what you think.

cpp/include/raft/buffer_base.hpp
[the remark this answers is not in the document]
teju85: I have filed issue #6 for this one.

cpp/include/raft/buffer_base.hpp
seunghwak: > Do you know if rmm supports host-memory (cudaMallocHost) based allocators?

Yeah... AFAIK, rmm currently does not support pinned memory allocation on host.

seunghwak: My biggest concern is that RAPIDS may need to develop support for new features (e.g. something like https://github.com/rapidsai/raft/issues/6) in both rmm and raft as such features will be helpful to both rmm & raft users.

seunghwak:
> @seunghwak @afender @cjnolet @dantegd As discussed offline, I have added clang-tidy support also. However, before I go ahead and fix errors raised by tidy script, I'd like you all to confirm whether the rules mentioned in the `cpp/.clang-tidy` are OK or not. (Disclaimer: this .clang-tidy config file is courtesy xgboost project)

I am new to clang-tidy, but this looks good to me (and in general, I care much more about consistency over specific styles). And now cuDF and cuGraph also adopt clang-format, but I assume raft (and cuML?) are first to adopt clang-tidy, correct? @afender @BradReesWork This will not going to be a small amount of work but at least for newly added code, should cuGraph adopt clang-tidy as well?

teju85:
@seunghwak the .clang-tidy I added here is a much stricter version than the one I added for cuML [here](https://github.com/rapidsai/cuml/pull/1945). We can afford a stricter version for RAFT as it has very little code right now. The plan for cuML is in each version, we'll be progressively making it's .clang-tidy stricter (finally matching the one in RAFT) and accordingly update the code to resolve errors.

cpp/include/raft/cudart_utils.h
```diff
@@ -0,0 +1,224 @@
...
+template <class T>
+void printHostVector(const char* variableName, const T* hostMem,
+                     size_t componentsCount) {
+  printHostVector(variableName, hostMem, componentsCount, std::cout);
+  std::cout.flush();
+}
+
+template <class T, class OutStream>
+void printDevVector(const char* variableName, const T* devMem,
```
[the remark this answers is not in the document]
teju85: Done.

cpp/include/raft/cudart_utils.h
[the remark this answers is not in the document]
teju85: These issues should now be addressed with the help of clang-tidy checks.

cpp/include/raft/allocator.hpp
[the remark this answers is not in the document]
seunghwak:
```
template <typename T>
virtual T* allocate(std:size_t n, cudaStream_t stream) = 0;  // allocate sizeof(T) * n
template <typename T>
virtual void deallocate(T* p, std::size_t n, cudaStream_t stream) = 0;
```

Thinking the code above might be easier as users can call
```
auto p = my_allocator.allocate<int32_t>(n, stream);
```
instead of 
```
auto p = reinterpret_cast<int32_t*>(my_allocator.allocate(sizeof(int32_t) * n, stream));
```

cpp/include/raft/cudart_utils.h
```diff
@@ -0,0 +1,211 @@
...
+//     cudaError_t status = call;                                               \
+//     if (status != cudaSuccess) {                                             \
+//       RAFT_LOG_ERROR("CUDA call='%s' at file=%s line=%d failed with %s ",    \
+//                      #call, __FILE__, __LINE__, cudaGetErrorString(status)); \
+//     }                                                                        \
+//   } while (0)
+
+/** helper method to get max usable shared mem per block parameter */
+inline int get_shared_memory_per_block() {
+  int devId;
```
seunghwak, line 126: We're mostly using lower_case for variable names (except for const ones), but here we're using CamelCase. Are we allowing to mix these two styles, or is there another rule I am missing?

afender: It seems functions are lower_case and internal variables tend to be CamelCase

teju85:
I thought about your point about just directly depending on RMM. @seunghwak you are right. In the long run, it does make sense to just rely on RMM's behavior, especially now that its interface is much richer. Let me update this PR accordingly. Give me another day or so.

cpp/include/raft/cudart_utils.h
teju85: that's my bad. I missed setting up the `LocalVariable*` options for the `readability-identifier-naming` check! Thanks for catching this. Will update soon with the fixes.

teju85: This should be done now.

cpp/include/raft/allocator.hpp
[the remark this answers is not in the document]
teju85: I don't think we can do templated virtual methods in C++, that easily. If know of the right way to do so, please let me know. Also, I was hoping to keep this interface a bit verbose so as to encourage folks to use the better interface of  `device_buffer` and `host_buffer`.

cpp/include/raft/cudart_utils.h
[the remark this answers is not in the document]
teju85: Done

cpp/include/raft/cudart_utils.h
```diff
@@ -0,0 +1,224 @@
...
+#include <string>
+#include <chrono>
+#include <iostream>
+///@todo: enable once logging has been enabled in raft
+//#include "logger.hpp"
+
+namespace raft {
+
+/** base exception class for the whole of raft */
+class Exception : public std::exception {
```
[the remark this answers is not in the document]
teju85: Thanks, I have filed #8 for this, so that we don't forget.

cpp/include/raft/handle.hpp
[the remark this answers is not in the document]
teju85: mutex logic added to all these "get handle" methods. Resolving this discussion.

cpp/include/raft/allocator.hpp
teju85: On a second thought, it does make sense to rely on RMM as our default. Updated the code.

cpp/include/raft/buffer_base.hpp
teju85: It does appear that RMM does not support host_allocator in the way we want it to be, ie. to have support for streams in its interface. So, we are better off for now to have this in raft and later work with rmm to update this interface.

cpp/include/raft/host_buffer.hpp
teju85: copy operator is hard to do as we expect the caller to also pass host allocator object during ctor. Thus, for now, I have only supported the option of doing copy-ctor from device buffer.

cpp/include/raft/cudart_utils.h
teju85: ok... device_utils and printing utilities can be further added in a separate PR. Makes sense. I have filed #9 for this one.

teju85:
Hey @dantegd @cjnolet @JohnZed @afender @seunghwak ,
I think the PR is now ready to be looked at again with all the comments from above addressed. Can you folks check if this now looks good to be merged or not?

cpp/include/raft/allocator.hpp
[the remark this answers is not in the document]
seunghwak: I see, yeah, I didn't know "I don't think we can do templated virtual methods in C++" and I agree with "I was hoping to keep this interface a bit verbose so as to encourage folks to use the better interface of device_buffer and host_buffer"

cpp/include/raft/buffer_base.hpp
```diff
@@ -0,0 +1,183 @@
...
+   *
+   * @return the allocator pointer
+   */
+  std::shared_ptr<Allocator> get_allocator() const { return _allocator; }
+
+ protected:
+  value_type* _data;
+
+ private:
+  size_type _size;
```
seunghwak: Did we switch from trailing _ to preceding _?

I was also using preceding _, but after seeing

https://stackoverflow.com/questions/3136594/naming-convention-underscore-in-c-and-c-sharp-variables/20496955#20496955

saying

```
Please DO NOT use UNDERSCORES before any variable name or parameter name in C++!!!

Names beginning with an underscore or a double underscore are RESERVED for the C++ implementers. Names with an underscore are reserved for the library to work.

If you have a read at the C++ Coding Standard, you will see that in the very first page it says:
```

I switched to trailing _.

(and both Google C++ coding style guide (https://google.github.io/styleguide/cppguide.html#General_Naming_Rules) and Herb Sutter & Andrei Alexandrescu (big names in the C++ world, https://www.drdobbs.com/c-coding-standards/184401863) recommend using trailing _.)

Any reason to go for preceding _ over trailing _ considering all the above recommendations?

teju85: Ouch... another one slipped by me due to simply copy-pasting stuff from cuML! Also a bit surprised that clang-tidy didn't catch this one! Let me see what's going on. Thanks for catching this.

dillon-cullinan:
rerun tests

cpp/include/raft/buffer_base.hpp
teju85: wait a min... this file has been renamed to `cpp/include/mr/buffer_base.hpp`. It has all the correct variable namings. I think you were looking at an older version of this file where the namings were done incorrectly.

teju85:
All I need is another approval from someone from cuML team. @JohnZed mentioned that he's going to take a look at this one soon.

In the meanwhile, if @dantegd can approve this, it'll be great. Asking for more approvals and reviews as I think we need to thoroughly review such a fundamental change to raft!

cpp/include/raft/allocator.hpp
```diff
@@ -0,0 +1,118 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <cuda_runtime.h>
+#include "cudart_utils.h"
+
+namespace raft {
```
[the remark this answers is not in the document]
teju85: cuml's current naming convention are a rather exception than rule. cuML started out much earlier than other projects in RAPIDS. At that time, we did a mistake of not standardizing any of these rules (mostly due to lack of time for thinking about these things)!

If you look at rmm, cudf or cugraph, they too are following the lowercase namespace. Thus, in order to be aligned with the rest of RAPIDS project, even here these are being named in lowercase.

cpp/include/raft/mr/allocator.hpp
```diff
@@ -0,0 +1,68 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <cuda_runtime.h>
+
+namespace raft {
```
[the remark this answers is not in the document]
teju85: I have explained this in the above thread.

cpp/include/raft/handle.hpp
```diff
@@ -0,0 +1,248 @@
...
+    if (!cublas_initialized_) {
+      CUBLAS_CHECK(cublasCreate(&cublas_handle_));
+      cublas_initialized_ = true;
+    }
+    mutex_.unlock();
+    return cublas_handle_;
+  }
+
+  cusolverDnHandle_t get_cusolver_dn_handle() const {
+    mutex_.lock();
```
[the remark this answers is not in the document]
teju85: ah! you're right. I forgot about `lock_guard`. Thanks. Fixed this one.

PR #15, opened 2020-06-03
[REVIEW] Add error check utilities

### LINKED ISSUE #11: [FEA] Exception-based macro and single header file for all error checkers
This was proposed in #7 (https://github.com/rapidsai/raft/pull/7#discussion_r425401511)

seunghwak:
Close #11, #23 

This PR copy-and-paste-and-modify cuDF's https://github.com/rapidsai/cudf/blob/branch-0.15/cpp/include/cudf/utilities/error.hpp

Currently supports

RAFT_EXPECTS, RAFT_FAIL, CUML_EXPECTS, CUML_FAIL, CUGRAPH_EXPECTS, CUGRAPH_FAIL, CUDA_TRY, CURAND_TRY, CUSPARSE_TRY, and NCCL_TRY.

Anything else to add?


cpp/include/raft/error.hpp
```diff
@@ -0,0 +1,329 @@
...
+}
+#undef _CUSPARSE_ERR_TO_STR
+
+inline void throw_cusparse_error(cusparseStatus_t error, const char* file,
+                                 unsigned int line) {
+  throw raft::cusparse_error(
+    std::string{"cuSparse error encountered at: " + std::string{file} + ":" +
+                std::to_string(line) + ": " + std::to_string(error) + " " +
+                cusparse_error_to_string(error)});
+}
```
teju85, line 247: We have this check already defined in https://github.com/rapidsai/raft/blob/branch-0.15/cpp/include/raft/linalg/cusolver_wrappers.h, better to update the existing code there?

seunghwak: > It makes sense to put the basic assertions in here. But putting assertions related to curand/cusparse/nccl in the same header file means that we'll introduce dependencies on these even if the including source doesn't use these components. I'd highly prefer if we could separate these out.

Yeah... so the pros is to keep the related code in one place so it gets easy to enforce consistency in error handling, but the problem you mentioned is also valid (especially NCCL can be problematic if we just compile for single-GPU).

I'm fine with either approach as long as we are consistent and can maintain consistency.

I will move

CUDA related ones to https://github.com/rapidsai/raft/blob/branch-0.15/cpp/include/raft/cudart_utils.h

cuSparse related ones to https://github.com/rapidsai/raft/blob/branch-0.15/cpp/include/raft/sparse/cusparse_wrappers.h

NCCL related ones to https://github.com/rapidsai/raft/blob/branch-0.15/cpp/include/raft/comms/std_comms.hpp

I should addd CUBLAS_TRY and CUSOLVER_TRY to https://github.com/rapidsai/raft/blob/branch-0.15/cpp/include/raft/linalg/cublas_wrappers.h and https://github.com/rapidsai/raft/blob/branch-0.15/cpp/include/raft/linalg/cusolver_wrappers.h

I may drop CURAND related ones for now, but they should be added back when we have curand_wrappers.hpp

And are you guys using XXX_CHECK in RAFT from cuML? If I replace them with XXX_TRY, will this break cuML?

And any thoughts about XXX_TRY vs XXX_CHECK? cuDF is using XXX_TRY, so I am just following cuDF convention but this is undesirable for cuML, I am open to discussion.

afender: +1 for CUBLAS_TRY and CUSOLVER_TRY they will be needed in #12 

I'm fine with XXX_TRY for consistency with cuDF. It also captures the underlying exception mechanism a bit better in the name.

cpp/include/raft/error.hpp
```diff
@@ -0,0 +1,329 @@
...
+                              ":" RAFT_STRINGIFY(__LINE__) ": " reason)
+
+/**
+ * @brief Indicates that an erroneous code path has been taken.
+ *
+ * In host code, throws a `raft::logic_error`.
+ *
+ * @param[in] reason String literal description of the reason
+ */
+#define CUGRAPH_FAIL(reason)                              \
```
afender, line 162: Need the sibling PR in cuGraph to remove utilities/error_utils.h

cpp/include/raft/error.hpp
```diff
@@ -0,0 +1,329 @@
...
+/**
+ * @brief Macro for checking (pre-)conditions that throws an exception when
+ * a condition is violated.
+ *
+ * @param[in] cond Expression that evaluates to true or false
+ * @param[in] reason String literal description of the reason that cond is
+ * expected to be true
+ * @throw raft::logic_error if the condition evaluates to false.
+ */
+#define CUGRAPH_EXPECTS(cond, reason)                         \
```
afender, line 149: Do we really need spearate RAFT/CUGRAPH/CUML EXPECT (or TRY), it seems they are all doing the same thing. Do you see a divergence in the future? I'd be fine using RAFT_EXPECTS in cuGraph.

seunghwak: RAFT_EXPECTS prints a message starting with "RAFT failure at"
CUML(GRAPH)_EXPECTS prints a message starting with "CUML(CUGRAPH) failure at"

So these are not identical and I think this information can be helpful, what do you think?

seunghwak:
After some more investigation,

RAFT's exception throwing mechanism

https://github.com/rapidsai/raft/blob/f48552ee5f0fd52de88c4f192bf1c7446de54b64/cpp/include/raft/cudart_utils.h#L84

provides additional stack trace compare to the one I brought from cuDF (https://github.com/rapidsai/raft/pull/15/files#diff-98265352f57cb794e805742f18ff96efR37)

and this information can be valuable so I think it is better to preserve this mechanism (but I think it is better to create separate exception classes like logic_error or cuda_error inheriting raft::exception similar to cuDF).

I think we should deprecate (and eventually remove) direct use of ASSERT macro (better use RAFT(CUML, or CUGRAPH)_EXPECTS) to be more consistent with the rest of the RAPIDS.

And we should also deprecate XXX_CHECK (and eventually remove) and use XXX_TRY instead.

@teju85 @cjnolet @dantegd @afender What do you guys think?

cpp/include/raft/error.hpp
cjnolet: I would also prefer to keep CUML/CUGRAPH-specifics out of raft as much as possible. If this repository is successful and ends up feeding many different libraries in the RAPIDS ecosystem, I am concerned this approach will become unsustainable. 

What's the intended use for these? For example, will `CUGRAPH_EXPECTS` be getting used in prims like spectral clustering while `CUML_EXPECTS` will get used in dense linear algebra prims?

seunghwak:
> I'm on board with putting the separate `TRY` macros in their respective wrapper headers but I think we should consider the perspective of assuming RAFT prims only know about RAFT and not CUML or CUGRAPH.

OK, makes sense, the initial intention of having separate CUGRAPH_EXPECTS and CUML_EXPECTS was to use them in cuGraph and cuML without the need to redefine error handling macros there (instead of using them inside RAFT, within RAFT, better use RAFT_EXPECTS), but we may better avoid cuGraph & cuML specific stuff to RAFT.

cpp/include/raft/error.hpp
teju85: @seunghwak sorry for the delayed response.

Yes, we are using the *_CHECK macros everywhere inside cuML. So, replacing those will certainly break our codebase. Moreover, the C-style '%' modifiers provides more readable code than the '<<' style C++ syntax (purely my opinion).

teju85:
Hey guys,
I'm a bit undecided about the *_CHECK vs *_TRY (or *_EXPECTS) macros. Let me list out my thoughts:
1. the *_CHECK, THROW and ASSERT macros being used pretty much in the whole of cuML codebase. So, removing them will create a lot of breakages and require widespread changes. This is probably just a one-time thing and I'm, maybe, overthinking? Would appreciate @JohnZed's inputs here.
2. This is an important one, IMO. All these cuML macros provide a nice C-style format specifiers syntax. This helps us to customize error messages, while keeping the code nice and short and readable. Personally, I'd even go along to say that C-style format string based messages are far more readable than C++-style '<<' syntax.
3. There are C++ libraries like 'fmt' to provide format specifiers, but using an external library just for formatting our messages when a simple C-style approach would work, seems like an overkill to me.

I'd love to hear your thoughts about this.

cpp/include/raft/error.hpp
teju85: I too agree with Alex and Corey on this one. Anything specific to cugraph or cuml is better put inside the respective library itself.

cpp/include/raft/error.hpp
teju85: @afender Looking at our above discussion, I think what will end up happening is that cugraph's utilities/error_utils.h will include this error.hpp.

cpp/include/raft/error.hpp
seunghwak: OK, better keep both *_TRY and *_CHECK in the meantime (better pick one eventually).

There can be a long debate in printf-ish vs cout-ish (e.g. https://stackoverflow.com/questions/2872543/printf-vs-cout-in-c).

My biggest concern was the type safety (in the above link) but it seems like most host compilers generates a warning for this and we are OK unless we ignore warnings. This wasn't a case with nvcc and leads to a weird runtime behavior but this is irrelevant here as this is a host side error checking mechanism. We may better keep the C-ish style (and add C++ style in the future if necessary).

cpp/include/raft/error.hpp
seunghwak: Got it :-)

afender:
I believe this PR also addresses https://github.com/rapidsai/raft/issues/23

cpp/include/raft/error.hpp
```diff
@@ -0,0 +1,159 @@
...
+  } while (0)
+
+// FIXME: Need to be replaced with RAFT_EXPECTS
+/** macro to check for a conditional and assert on failure */
+#define ASSERT(check, fmt, ...)              \
+  do {                                       \
+    if (!(check)) THROW(fmt, ##__VA_ARGS__); \
+  } while (0)
+
+#define SET_ERROR_MSG(msg, location_prefix, fmt, ...)                        \
```
teju85, line 118: minor nitpick: `location_prefix` seems to be unused?

teju85: I guess it was supposed to be used in the first `snprintf` call below?

cpp/include/raft/comms/std_comms.hpp
```diff
@@ -44,18 +44,45 @@
...
+    if (ncclSuccess != status) {                                              \
+      std::string msg{};                                                      \
+      SET_ERROR_MSG(msg,                                                      \
+                    "NCCL error encountered at: ", "call='%s', Reason=%d:%s", \
+                    #call, status, ncclGetErrorString(status));               \
+      throw raft::nccl_error(msg);                                            \
+    }                                                                         \
+  } while (0);
+
+/** FIXME: temporary alias for cuML compatibility */
```
teju85, line 79: @seunghwak Currently only std-comms uses this `NCCL_CHECK` macro in cuML. So, feel free to remove this macro and replace any use of `NCCL_CHECK` in raft with `NCCL_TRY`.

cpp/include/raft/error.hpp
seunghwak: Nice catch!!!

cpp/include/raft/comms/std_comms.hpp
seunghwak: Done :-)

seunghwak:
@teju85 I addressed all your last comments :-)

PR #17, opened 2020-06-08
add inline destructor

Iroy30:

### DIFF
--- a/cpp/include/raft/comms/comms.hpp
+++ b/cpp/include/raft/comms/comms.hpp
@@ -89,7 +89,6 @@ constexpr datatype_t get_type<double>() {
 
 class comms_iface {
  public:
-  virtual ~comms_iface();
 
   virtual int get_size() const = 0;
   virtual int get_rank() const = 0;
@@ -322,8 +321,6 @@ class comms_t {
  private:
   std::unique_ptr<comms_iface> impl_;
 };
-
-inline comms_iface::~comms_iface() {}
 
 }  // namespace comms
 }  // namespace raft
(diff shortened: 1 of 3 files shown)


cpp/include/raft/comms/comms.hpp
```diff
@@ -323,7 +323,7 @@ class comms_t {
   std::unique_ptr<comms_iface> impl_;
 };
 
-comms_iface::~comms_iface() {}
+inline comms_iface::~comms_iface() {}
```
seunghwak, line 326: Is there any reason in placing this outside the class definition while we keep the definition of a constructor inside the class definition? If there is no specific reason, we'd better keep this inside the class definition. And we're doing nothing special here, so ~comms_iface() = default; should be better. And are we allowing copy/move constructions/assignments or constructor without an argument? Based on that, better be explicit about either deleting them or marking those as default.

afender: Making it `inline` has the same effect as placing it inside the class definition. 
Although while discussing the fix with Ishika I was also curious about the motivation for adding it outside in the first place.

teju85: Also, if this is just a default dtor, why not just remove it altogether?

cjnolet: Indeed, this should just removed.

Iroy30: Thanks, I was not sure if this was a placeholder for something to come. Either way, I have deleted the same.

PR #25, opened 2020-06-17
[REVIEW] Fix incorrect streams being returned from `get_internal_streams()`

teju85:
Current implementation of `handle_t::get_internal_streams()` used to return 2x the number of internal streams, which was incorrect. This PR fixes this bug and adds a small unit-test to verify the same.


GPUtester:

PR #29, opened 2020-07-08
[REVIEW] Add ceildiv functionality

kaatish:
Adding ceildiv functionality so that it can be reused in cugraph.

### DIFF
--- a/cpp/include/raft/cudart_utils.h
+++ b/cpp/include/raft/cudart_utils.h
@@ -29,16 +29,6 @@
 namespace raft {
 
 /**
- * @brief Provide a ceiling division operation ie. ceil(a / b)
- * @tparam IntType supposed to be only integers for now!
- */
-template <typename T>
-constexpr __host__ __device__ std::enable_if_t<std::is_integral<T>::value, T>
-ceildiv(T a, T b) {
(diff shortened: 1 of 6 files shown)


GPUtester:

cpp/include/raft/cudart_utils.h
```diff
@@ -28,6 +28,17 @@
...
 namespace raft {
 
+/**
+ * @brief Provide a ceiling division operation ie. ceil(a / b)
+ * @tparam IntType supposed to be only integers for now!
+ */
+template <typename T>
+constexpr __host__ __device__
+std::enable_if_t<std::is_integral<T>::value, T>
+ceildiv(T a, T b) {
```
seunghwak, line 38: This looks very similar to 

https://github.com/rapidsai/cudf/blob/branch-0.15/cpp/include/cudf/detail/utilities/integer_utils.hpp
```
template <typename S, typename T>
constexpr inline S div_rounding_up_unsafe(const S& dividend, const T& divisor) noexcept
{
  return (dividend + divisor - 1) / divisor;
}
```

Can we just bring cudf's integer_utils.hpp (we need to put things in raft namespace instead of cudf namespace) and tweak it?

PR #30, opened 2020-07-10
[REVIEW] Add out of place broadcast

kaatish:
Some opg cugraph functions require an out of place broadcast or incur an extra cost of copying data everytime. This PR adds the missing functionality.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - PR #7: Migrating cuml comms -> raft comms_t
 - PR #15: add exception based error handling macros
 - PR #29: Add ceildiv functionality
+- PR #30: Add out of place broadcast
 
 ## Improvements
 - PR #13: Add RMM_INCLUDE and RMM_LIBRARY options to allow linking to non-conda RMM
--- a/cpp/include/raft/comms/comms.hpp
+++ b/cpp/include/raft/comms/comms.hpp
@@ -109,6 +109,10 @@ class comms_iface {
                          datatype_t datatype, op_t op,
                          cudaStream_t stream) const = 0;
 
+  virtual void bcast(const void* sendbuff, void* recvbuff, size_t count,
+                     datatype_t datatype, int root,
+                     cudaStream_t stream) const = 0;
+
   virtual void bcast(void* buff, size_t count, datatype_t datatype, int root,
                      cudaStream_t stream) const = 0;
 
@@ -244,6 +248,23 @@ class comms_t {
   template <typename value_t>
   void bcast(value_t* buff, size_t count, int root, cudaStream_t stream) const {
     impl_->bcast(static_cast<void*>(buff), count, get_type<value_t>(), root,
+                 stream);
+  }
+
+  /**
+   * Broadcast data from one rank to the rest
+   * @tparam value_t datatype of underlying buffers
+   * @param sendbuff buffer to send
+   * @param recvbuff buffer to receive
+   * @param count number of elements if buff
+   * @param root the rank initiating the broadcast
+   * @param stream CUDA stream to synchronize operation
+   */
+  template <typename value_t>
+  void bcast(const value_t* sendbuff, value_t* recvbuff, size_t count, int root,
+             cudaStream_t stream) const {
+    impl_->bcast(static_cast<const void*>(sendbuff),
+                 static_cast<void*>(recvbuff), count, get_type<value_t>(), root,
                  stream);
   }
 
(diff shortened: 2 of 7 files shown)


GPUtester:

seunghwak:
> Some opg cugraph functions require an out of place broadcast or incur an extra cost of copying data everytime. This PR adds the missing functionality.

Could you be more specific about this?

```
thrust::device_vector<T> src_values(100);
thrust::device_vector<T> dst_values(100);
... // update src_values
if (root_node) {
  bcast(src_values.data().get(), ...);
  thrust::copy(src_values.begin(), src_values.end(), dst_values.begin());  // are you considering this copy?
}
else (root_node) {
  bcat(dst_values.data().get(), ...);
}
```

kaatish:
Just a nitpick for your example. As far as I know you cannot have two different pathways for a collective communication. Taken from the [docs](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/collectives.html#) :

> Collective operations have to be called for each rank (hence CUDA device) to form a complete collective operation. Failure to do so will result in other ranks waiting indefinitely.

Anyway, regarding the extra copy, I guess I was mistaken about that. The problem right now is extra memory usage to achieve the same thing.

My use case is that I want to iteratively get a vector from a gpu and reduce it with my custom operator. If I am not mistaken I think with the existing functionality this is what can be done :

```
int count = ...;
thrust::device_vector<T> my_values(count);
thrust::device_vector<T> other_values(count);
thrust::device_vector<T> reduced_values(count);

for (int i = 0; i < total_gpu_count; ++i) {
  T* ptr = nullptr;
  if (i != my_rank) {
    ptr = other_values.data().get();
  } else {
    ptr = my_values.data().get();
  }
  bcast(ptr, count, i, stream);
  if (i != my_rank) {
    //reduced_values += other_values
    vector_add(other_values, reduced_values, reduced_values, custom_op);
  }
}
```

With the functionality I wish to add the same thing can be achieved like this :

```
int count = ...;
thrust::device_vector<T> my_values(count);
thrust::device_vector<T> reduced_values(count);

for (int i = 0; i < total_gpu_count; ++i) {
  bcast(my_values.data().get(), other_values.data().get(), count, i, stream);
  if (i != my_rank) {
    //reduced_values += other_values
    vector_add(other_values, reduced_values, reduced_values, custom_op);
  }
}
```

So I thought that exposing the underlying out of place nccl API might make things easy. What is your opinion about the two approaches of iterative reduction of a vector using custom operator. I am essentially recreating allreduce but with a custom reduction operator.

seunghwak:
> Just a nitpick for your example. As far as I know you cannot have two different pathways for a collective communication. Taken from the [docs](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/collectives.html#) :

Oh, no, the code above is perfectly valid. Every participating processes should call the collectives all together but one can call this in the if statement and other processes can call this in the else case. The collective function has no knowledge about the branch it is called.

If only a subset of processes call a collective, the collective will not return, but

```
if (root_node) {
  bcast(src_values.data().get(), ...);
  thrust::copy(src_values.begin(), src_values.end(), dst_values.begin());  // are you considering this copy?
}
else (root_node) {
  bcat(dst_values.data().get(), ...);
}
```

is perfectly fine. And you don't like if-else, you can do ```bcast(root_node ? src_values.data().get() : dst_values.data().get(), ...)``` as well.

seunghwak:
And in the second example you provided,

```
int count = ...;
thrust::device_vector<T> my_values(count);
thrust::device_vector<T> reduced_values(count);

for (int i = 0; i < total_gpu_count; ++i) {
  bcast(my_values.data().get(), other_values.data().get(), count, i, stream);
  if (i != my_rank) {
    //reduced_values += other_values
    vector_add(other_values, reduced_values, reduced_values, custom_op);
  }
}
```
You're still using other_values but not defining it... It will not compile. Anything I have missed?

kaatish:
I guess that changes things. I'll close this for now and think a bit more over this.

seunghwak:
And what you're actually trying to do is all-reduce. All-reduce is a single reduction + one broadcast, so it will be much faster than doing P broadcasts.

seunghwak:
https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/collectives.html#allreduce

kaatish:
nccl All reduce does not work with bitwise OR. There is an open PR but there is no indication if it would be merged and more likely than not it won't be unless there is a very good justification for it. If upon my experimentation I find that bitwise OR reduction in that PR outperforms the other approaches I am taking in opg bfs then it could be a good motivation for that PR to be merged. In the meanwhile I have to find workarounds to get the same functionality.

Issue #24, opened 2020-06-15
[FEA] Need to integrate device property queries in handle.hpp and cudart_utils.cuh


seunghwak:
**Is your feature request related to a problem? Please describe.**
RAFT handle.hpp stores the entire set of device properties at the beginning to make future queries low cost but cudart_utils.h does not exploit this and individually query device properties again.

https://github.com/rapidsai/raft/blob/2487eb0c12f374729043baa5448c0d309c921e60/cpp/include/raft/handle.hpp#L164

https://github.com/rapidsai/raft/blob/2487eb0c12f374729043baa5448c0d309c921e60/cpp/include/raft/cudart_utils.h#L125
https://github.com/rapidsai/raft/blob/2487eb0c12f374729043baa5448c0d309c921e60/cpp/include/raft/cudart_utils.h#L134

**Describe the solution you'd like**
Better integrate these two; either hanlde_t provides device property query functions or cudart_utils.cuh call internally call handle functions.

seunghwak:
It seems like

inline int get_shared_memory_per_block() and inline int get_multi_processor_count() are not used anywhere (I searched RAFT, cuML, and cuGraph, let me know if I am missing other places).

Just plan to remove these in https://github.com/rapidsai/raft/pull/37

PR #37, opened 2020-07-20
[REVIEW] Clean-up CUDA utilites

seunghwak:
Close https://github.com/rapidsai/raft/issues/24

- Remove device property query functions directly invoking cudaDeviceGetAttribute() (instead of using raft::handle_t's get_device_properties(), and the removed functions are not used in RAFT, cuML, and cuGraph as I checked)
- Add warp_full_mask() function (returning 0xffffffff) next to warp_size() function
- Add utility structs for CUDA grid generation in 1D case (based on cuDF's https://github.com/rapidsai/cudf/blob/branch-0.15/cpp/include/cudf/detail/utilities/cuda.cuh#L36)
- Add utility functions for CUDA atomics (these will be used for the types that are not natively supported by CUDA, based on cuDF's https://github.com/rapidsai/cudf/blob/branch-0.15/cpp/include/cudf/detail/utilities/device_atomics.cuh)
- Removed sm_utils.h (we're CUDA 9+, so no need to generate different shfl, ballot, ... code based on CUDA version)

Let me know if you have any concerns with the updates.


GPUtester:

GPUtester:

cpp/include/raft/cudart_utils.h
```diff
@@ -97,27 +97,96 @@ struct cuda_error : public raft::exception {
...
 }
 
-/** Helper method to get to know warp size in device code */
-constexpr inline int warp_size() { return 32; }
+/**
+ * @brief A kernel grid configuration construction gadget for simple one-dimensional mapping
+ * elements to threads.
+ */
+class grid_1d_thread_t {
+ public:
```
aschaffer, line 112: Perhaps better use an aggregate (`struct`) rather than a `class` with `public:` data members.

cpp/include/raft/cudart_utils.h
```diff
@@ -97,27 +97,96 @@ struct cuda_error : public raft::exception {
...
+    RAFT_EXPECTS(elements_per_thread > 0, "elements_per_thread must be > 0");
+  }
+};
+
+/**
+ * @brief A kernel grid configuration construction gadget for simple one-dimensional mapping
+ * elements to warps.
+ */
+class grid_1d_warp_t {
+ public:
```
aschaffer, line 143: Perhaps better use an aggregate (`struct`) rather than a `class` with `public:` data members.

cpp/include/raft/cudart_utils.h
```diff
@@ -97,27 +97,96 @@ struct cuda_error : public raft::exception {
...
+                 "num_threads_per_block / warp_size() must be > 0");
+  }
+};
+
+/**
+ * @brief A kernel grid configuration construction gadget for simple one-dimensional mapping
+ * elements to blocks.
+ */
+class grid_1d_block_t {
+ public:
```
aschaffer, line 171: Perhaps better use an aggregate (`struct`) rather than a `class` with `public:` data members.

cpp/include/raft/spectral/kmeans.hpp
```diff
@@ -31,10 +31,10 @@
 
 #include <raft/cudart_utils.h>
 #include <raft/linalg/cublas_wrappers.h>
+#include <raft/device_atomics.cuh>
 #include <raft/handle.hpp>
 #include <raft/spectral/matrix_wrappers.hpp>
 #include <raft/spectral/warn_dbg.hpp>
-#include <raft/utils/sm_utils.hpp>
```
aschaffer, line 37: Perhaps a good thing that `sm_utils.hpp` is going away, but just as a CAVEAT, this might break dependencies in `cugraph` (possibly `cuml`, too).

seunghwak: Thanks, for the cuGraph part, https://github.com/rapidsai/cugraph/pull/1021 will address the issue. AFAIK (I have searched for this), cuML does not use sm_utils.h

cpp/include/raft/cudart_utils.h
seunghwak: Yeah... this blindly follows cuDF (https://github.com/rapidsai/cudf/blob/branch-0.15/cpp/include/cudf/detail/utilities/cuda.cuh#L40); cuDF is using class for this.

I guess struct may make slightly more sense but in practice, the difference is pretty minimal (and this has a custom constructor as well... though nothing prevents struct from having a constructor), so I put little more weight on consistency across RAPIDS (this maybe helpful when we embrace RAPIDS-core).

But I am fine with either struct or class... so what do you think, or any other thoughts from anyone else? I don't have a strong opinion on this, so someone has a strong one, I can embrace it.

cpp/include/raft/spectral/kmeans.hpp
seunghwak: Thanks, for cuGraph, https://github.com/rapidsai/cugraph/pull/1021 will address the issue. When I searched cuML, I didn't find any file including sm_utils.hpp from RAFT (Hope cuML folks will correct me if I am wrong).

BradReesWork:
rerun tests

PR #43, opened 2020-07-29
[REVIEW] Fix for bug on cuda toolkit version selection mechanism select

aschaffer:
This fixes the bug of selecting cuda_minors > 0 even for cuda_majors >=11 (which would have bypassed 11.0). This affected SpMV() algorithm selection.

Also added a mechanism (for debugging) to enforce lower than 10.1 versions of SpMV().


GPUtester:

PR #46, opened 2020-08-06
[REVIEW] Fix for cpp file extension issue.

aschaffer:
This addresses the issue of enforcing `nvcc` invocation to inject the definitions of `__CUDACC_VER_MAJOR/MINOR__` pre-processor variables. These are necessary to distinguish proper `cusparse` SpMV calls between toolkit versions >= 10.1, and versions <= 10.0.


GPUtester:

cpp/include/raft/sparse/cusparse_wrappers.h
```diff
@@ -26,6 +26,26 @@
...
+//(1.) CUDA_VER_SELECT aggregates all the CUDA version selection logic;
+//(2.) to enforce a lower version,
+//
+//`#define CUDA_ENFORCE_LOWER
+// #include <raft/sparse/cusparse_wrappers.h>`
+//
+// (i.e., before including this header)
+//
+#define CUDA_VER_SELECT         \
+  (__CUDACC_VER_MAJOR__ > 10 or \
```
JohnZed, line 39: What about using CUSPARSE_VER_MAJOR instead? It would be nice not to require users to only include this in .cu, since .cu files can be slower to compile.

teju85:
@aschaffer It's better to use `CUDART_VERSION` (you'll need include `cuda_runtime_api.h`) macro instead of the ones you are currently using. AFAIK, `CUDACC*` macros are available only when `nvcc` is invoked. Whereas, `cusparse_wrappers.h` is supposed to be called from a pure host `.cpp` code also.

cpp/include/raft/sparse/cusparse_wrappers.h
aschaffer: It looks like `CUSPARSE_VER_*` are not in sync w/ `__CUDACC_VER_*__`. For a CUDA 11.0 build I get:
```pragma message: CUSPARSE_VER_MAJOR=11
                                              ^
#pragma message: CUSPARSE_VER_MINOR=1```

So, the minor is not the same. Which then begs the question if the major might also get out of sync.

Comment (aschaffer) [reply cpp/include/raft/sparse/cusparse_wrappers.h]: Therefore, I've adopted the `CUDART_VERSION` approach.

**Comment by aschaffer** [comment]
> It's better to use `CUDART_VERSION`

Done.

File: cpp/include/raft/sparse/cusparse_wrappers.h [hunk 2]
```diff
@@ -26,6 +26,24 @@
...
+//(1.) CUDA_VER_SELECT aggregates all the CUDA version selection logic;
+//(2.) to enforce a lower version,
+//
+//`#define CUDA_ENFORCE_LOWER
+// #include <raft/sparse/cusparse_wrappers.h>`
+//
+// (i.e., before including this header)
+//
+
+#define CUDA_VER_SELECT (CUDART_VERSION >= 10100)
```
JohnZed, line 39: Maybe a name other than CUDA_VER_SELECT, since SELECT doesn't make it clear we mean "high version," like CUDA_VER_RECENT or CUDA_VER_POST101?

cpp/include/raft/sparse/cusparse_wrappers.h
```diff
@@ -26,6 +26,24 @@
...
+
+#define VALUE_TO_STRING(x) #x
```
JohnZed, line 41: I would make these names more obscure like` `_VALUE` or `_RAFT_PREPROC_VALUE`, since this header will be included in customer codebases, we could easily collide with another use of VALUE in another context.

cpp/include/raft/sparse/cusparse_wrappers.h
aschaffer: So, the meaning meant to be conveyed was a "way to detect the active cuda toolkit version". It might not necessarily be _recent_. A pre-10.1 user should still be able to compile/run these. Just with a different code-path. (I know we don't support 10.0 or below anymore, but it's good to still have a plan B, plus it helps w/ debugging)

cpp/include/raft/sparse/cusparse_wrappers.h
aschaffer: I agree. It's residual code, meaning I checked it in to have it somewhere in a branch in case I need it again for debugging. Not meant to make it in the release. I will remove all of that.

aschaffer: Any objections on removing these "*VALUE*" macros?

cpp/include/raft/sparse/cusparse_wrappers.h
aschaffer: How about `CUDA_VER_10_1_UP` (`POST101` may suggest 10.1 is not included, when in fact it is).

cpp/include/raft/sparse/cusparse_wrappers.h
JohnZed: I'm ok with removing or renaming - either works. Maybe removing is even better

cpp/include/raft/sparse/cusparse_wrappers.h
JohnZed: Sounds good to me. The underscores are definitely clearer than my POST101 suggestion!

aschaffer: Done.

cpp/include/raft/sparse/cusparse_wrappers.h
aschaffer: Done (removed).

aschaffer:
rerun tests.

teju85:
rerun tests

aschaffer:
> 
> 
> @aschaffer we already do the cuda version check in cmake. Do we really need to add macros to check the same in C++ too?

I believe so. The code did not compile without branching on the cuda version in code (via pre-processor macros). I'm not sure how cmake would have addressed coexistance of 2 versions of cupsarse, one of which is not only deprcated, but completely removed. If cmake would have been able to do so, we hadn't seen the build failures, had we? Unless I'm missing the point.

teju85:
Interesting... even if there were multiple CTK's in the system, eventually cmake would have chosen one from [this](https://github.com/rapidsai/raft/blob/branch-0.15/cpp/CMakeLists.txt#L68) line. And that'd then eventually guide the build process to use the right libraries [here](https://github.com/rapidsai/raft/blob/branch-0.15/cpp/CMakeLists.txt#L204). If you have a repro that'd help, Andrei.

PR #48, opened 2020-08-11
[REVIEW] Gtest target names on cmake build

dantegd:


GPUtester:

cpp/cmake/Dependencies.cmake
```diff
@@ -36,15 +36,15 @@ if(BUILD_GTEST)
...
-  add_dependencies(gtest_mainlib googletest)
+  add_dependencies(GTest::GTest googletest)
+  add_dependencies(GTest::Main googletest)
 
 else()
 
   find_package(GTest REQUIRED)
 
-endif(BUILD_TEST)
+endif(BUILD_GTEST)
```
hcho3, line 50: In latest CMake, it is no longer necessary to write out the condition expression in `endif()`.

hcho3: "Per legacy, the else() and endif() commands admit an optional <condition> argument. If used, it must be a verbatim repeat of the argument of the opening if command." https://cmake.org/cmake/help/v3.14/command/if.html

JohnZed: Wow, I hadn't heard... pretty cool!

PR #63, opened 2020-08-29
[REVIEW] Adding MPI comms implementation

cjnolet:
This is strictly used for gtesting MG primitives without dask.


GPUtester:

cpp/include/raft/comms/mpi_comms.hpp
```diff
@@ -0,0 +1,290 @@
...
+      MPI_Error_string(status, mpi_error_string, &mpi_error_string_lenght); \
+      printf("MPI call='%s' at file=%s line=%d failed with %s ", #call,     \
+             __FILE__, __LINE__, mpi_error_string);                         \
+    }                                                                       \
+  } while (0)
+
+namespace raft {
+namespace comms {
+
+static MPI_Datatype get_mpi_datatype(const datatype_t datatype) {
```
seunghwak, line 62: Why's this static?

And this function can be constexpr (so if datatype is available in compile time, this function will be evaluated in compile time).

cpp/include/raft/comms/mpi_comms.hpp
```diff
@@ -0,0 +1,290 @@
...
+      return MPI_FLOAT;
+    case datatype_t::FLOAT64:
+      return MPI_DOUBLE;
+    default:
+      // Execution should never reach here. This takes care of compiler warning.
+      return MPI_DOUBLE;
+  }
+}
+
+static MPI_Op get_mpi_op(const op_t op) {
```
seunghwak, line 86: Same here, why's this static & this function can be constexpr.

cpp/include/raft/comms/mpi_comms.hpp
```diff
@@ -0,0 +1,290 @@
...
+}
+
+class mpi_comms : public comms_iface {
+ public:
+  mpi_comms(MPI_Comm comm, const bool owns_mpi_comm)
+    : _owns_mpi_comm(owns_mpi_comm),
+      _mpi_comm(comm),
+      _size(0),
+      _rank(1),
+      _next_request_id(0) {
```
seunghwak, line 109: I think we agreed to use trailing _ (instead of preceding _) for class private variables.

https://github.com/rapidsai/raft/blob/branch-0.16/cpp/.clang-tidy#L185

And see

https://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier

using preceding _ is generally discouraged.

cpp/include/raft/comms/mpi_comms.hpp
```diff
@@ -0,0 +1,290 @@
...
+  }
+
+  int get_size() const { return _size; }
+
+  int get_rank() const { return _rank; }
+
+  std::unique_ptr<comms_iface> comm_split(int color, int key) const {
+    MPI_Comm new_comm;
+    MPI_TRY(MPI_Comm_split(_mpi_comm, color, key, &new_comm));
+    return std::unique_ptr<comms_iface>(new mpi_comms(new_comm, true));
```
seunghwak, line 139: Just to double check, this works with code requiring NCCL sub-communicators? (as we are initializing NCCL inside mip_comm constructor involving only the processes in the passed mpi_comm object).

cpp/include/raft/comms/mpi_comms.hpp
```diff
@@ -0,0 +1,290 @@
...
+                   array_of_requests[i]);
+      requests.push_back(req_it->second);
+      _free_requests.insert(req_it->first);
+      _requests_in_flight.erase(req_it);
+    }
+    MPI_TRY(MPI_Waitall(requests.size(), requests.data(), MPI_STATUSES_IGNORE));
+  }
+
+  void allreduce(const void* sendbuff, void* recvbuff, size_t count,
+                 datatype_t datatype, op_t op, cudaStream_t stream) const {
```
seunghwak, line 193: This is just FYI,

Unified Collective Communication is coming (UCX is for P2P, UCC supports collectives on top of UCX).

https://www.ucfconsortium.org/projects/ucc/

At some point, we may consider implementing something like

```
collective(address) {
  if (address is device) {
    use NCCL;
  }
  else { // address is host
    // this is a temporary hack, we plan to replace the code below with UCC in the future.
    // communicator may own a small buffer. If the message size is smaller than the buffer size, we can use the buffer to avoid device memory allocation.
    create device_buffer;
    copy_to_device_buffer;
    use NCCL;
    copy back to host;
  }
}
```

I think this will make our code simpler and future ready if we need run collectives on host data.

cpp/include/raft/comms/util.hpp
```diff
@@ -0,0 +1,114 @@
...
+    if (ncclSuccess != status) {                          \
+      printf("NCCL call='%s' failed. Reason:%s\n", #call, \
+             ncclGetErrorString(status));                 \
+    }                                                     \
+  } while (0)
+
+namespace raft {
+namespace comms {
+
+static size_t get_datatype_size(const datatype_t datatype) {
```
seunghwak, line 53: Why's this static? and this can be constexpr.

cpp/include/raft/comms/util.hpp
```diff
@@ -0,0 +1,114 @@
...
+    case datatype_t::FLOAT32:
+      return sizeof(float);
+    case datatype_t::FLOAT64:
+      return sizeof(double);
+    default:
+      RAFT_FAIL("Unsupported datatype.");
+  }
+}
+
+static ncclDataType_t get_nccl_datatype(const datatype_t datatype) {
```
seunghwak, line 76: Why's this static? and this can be constexpr.

cpp/include/raft/comms/util.hpp
```diff
@@ -0,0 +1,114 @@
...
+    case datatype_t::FLOAT32:
+      return ncclFloat;
+    case datatype_t::FLOAT64:
+      return ncclDouble;
+    default:
+      throw "Unsupported";
+  }
+}
+
+static ncclRedOp_t get_nccl_op(const op_t op) {
```
seunghwak, line 99: Why's this static? and this can be constexpr.

cpp/include/raft/comms/mpi_comms.hpp
cjnolet: Yeah, the new `mpi_comms` instance should create the new nccl comm once constructed.

cpp/include/raft/comms/util.hpp
cjnolet: I recall CI was having trouble building these functions w/ `constexpr` (even though we were all able to build them locally) so we agreed to make them static for the time being. Trying again to see.

cpp/include/raft/comms/mpi_comms.hpp
cjnolet: See comment: https://github.com/rapidsai/raft/pull/63#discussion_r479663183

cpp/include/raft/comms/util.hpp
seunghwak: OK, now I remember.... yeah... let's merge this first, and I will take a look for this issue later when I have some time to spare.

cjnolet:
@seunghwak, just FYI in case you haven't noticed yet, changing the static functions to constexpr did pass CI.

seunghwak:
> @seunghwak, just FYI in case you haven't noticed yet, changing the static functions to constexpr did pass CI.

Yeah... I guess adding constexpr may enforce inlining of the function... Need to double check C++ rules about when the function is guaranteed to be inlined when is not...

seunghwak:
https://en.cppreference.com/w/cpp/language/constexpr
```
A constexpr specifier used in a function or static member variable (since C++17) declaration implies inline.
```
Not sure the (since C++17 part) applies to the entire sentence or just the "static member variable" part (I guess the latter but not 100% sure), but it seems like constexpr enforces inlining in at least in C++17 or later. If the "since C++17" applies only to the static member variable part, that we can assume using constexpr enforces inlining even in C++14.

seunghwak:
Another reference.

https://stackoverflow.com/questions/14391272/does-constexpr-imply-inline

```
Yes ([dcl.constexpr], §7.1.5/2 in the C++11 standard): "constexpr functions and constexpr constructors are implicitly inline (7.1.2)."
```

PR #65, opened 2020-09-09
[REVIEW] Moving cuML prims to RAFT

divyegala:
This PR moves cuml prims to RAFT that will break the circular dependency between libcuml and libcumlprims. The changes made in these files are *almost* exactly the same as those mirrored by cuML, with only the header paths updated to reflect how raft does it:

1. Like `#include <common/cudart_utils.h>` became `#include <raft/cudart_utils.h>`
2. "#include "test_utils.h" became "../test_utils.h" (Our prims gtests are all in the same, root level `tests/prims` folder but I took the liberty of dividing it more as `tests/linalg`, `tests/stats`, etc.)

In any case, these changes will not affect when we finally delete prims from cuML. Only thing we would need to do there is also update the include path for these prims (files moved), as they are already in the correct formatting and namespaces.

### DIFF
--- a/cpp/include/raft/cudart_utils.h
+++ b/cpp/include/raft/cudart_utils.h
@@ -256,4 +256,36 @@ void print_device_vector(const char* variable_name, const T* devMem,
 }
 /** @} */
 
+/** cuda malloc */
+template <typename Type>
+void allocate(Type*& ptr, size_t len, bool setZero = false) {
+  CUDA_CHECK(cudaMalloc((void**)&ptr, sizeof(Type) * len));
+  if (setZero) CUDA_CHECK(cudaMemset(ptr, 0, sizeof(Type) * len));
+}
+
+/** helper method to get multi-processor count parameter */
+inline int getMultiProcessorCount() {
+  int devId;
+  CUDA_CHECK(cudaGetDevice(&devId));
+  int mpCount;
+  CUDA_CHECK(
+    cudaDeviceGetAttribute(&mpCount, cudaDevAttrMultiProcessorCount, devId));
+  return mpCount;
+}
+
+/** calculate greatest common divisor of two numbers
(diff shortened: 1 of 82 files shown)


GPUtester:

JohnZed:
We discussed offline that the build changes are new, but can you clarify the other files - are they all _exact_ copies of the recently-updated cuml versions? Or are there additional changes?

divyegala:
@JohnZed updated the description of the PR to clarify

raydouglass:
rerun tests

mike-wendt:
Doesn't look like we're needed for code review, removing  @rapidsai/ops-codeowners; add us back if necessary. Thanks!

cpp/include/raft/cudart_utils.h
```diff
@@ -256,4 +259,107 @@ void print_device_vector(const char* variable_name, const T* devMem,
...
+  return rows * columns;
+}
+
+/** Helper function to check alignment of pointer.
+    * @param ptr the pointer to check
+    * @param alignment to be checked for
+    * @return true if address in bytes is a multiple of alignment
+    */
+template <typename Type>
+bool is_aligned(Type* ptr, size_t alignment) {
```
JohnZed, line 346: It is pretty fun seeing these varying naming conventions together. Will be great to get through the tidy phase!

JohnZed (approved):
Looks straightforward to me! Thanks for making the move.

divyegala:
@JohnZed added the leftover prims, as I mentioned on the call. Reference cuml PR is https://github.com/rapidsai/cuml/pull/3044

afender (approved):

JohnZed:
GIven approvals from both ML and graph, I'll go ahead and merge. Excellent stuff, @divyegala !

PR #73, opened 2020-10-02
[REVIEW] Move DistanceType enum to RAFT from cuML

wphicks:
Transfer DistanceType enum to RAFT for use across cuML, prims, and Cython code


wphicks:
cuML PR: https://github.com/rapidsai/cuml/pull/2918

divyegala (changes requested):
Let's try to run clang-tidy here.. it will have to be a manual one-line run but we can discuss offline on the best way to do this (trying to do the same myself)

dantegd (approved):

PR #81, opened 2020-10-10
[REVIEW] BUG Temporarily expose internal NCCL communicator

seunghwak:
This is a temporary hack to work around a UCX isend/irecv issue, we need to discuss a more proper way to support P2P communication for device data in the future; this hack is a temporary solution and this code should be removed once we have a proper solution.


GPUtester:

afender (commented):
Yea p2p send/receive should really be encapsulated like other NCCL comms
We'll need to refactor this as well as cugraph to get rid of direct nccl accesses.

afender (approved):

BradReesWork (approved):

PR #85, opened 2020-10-22
[WIP] Fix clang tidy errors in raft

teju85:
As per our discussion over slack, here's an initial set of changes showing the kind of things that would have to change in order to get clang-tidy checks to pass in raft. Hope is that this will also trigger any discussions around standardization rules itself, if any of those are questionable/improvable.

I have also taken this opportunity to re-enable .cu files to also be considered for clang-tidy checks in our `run-clang-tidy.py` script. So, we can also fix tidy issues arising out of .cuh files too.

Tagging @mdemoret-nv @divyegala @JohnZed @cjnolet @dantegd from cuML side for any opinions/suggestions/comments.
Tagging @seunghwak @afender @BradReesWork from cuGraph side for any of their opinions/suggestions/comments.

NOTE: this is NOT the exhaustive set of changes to raft! More will come soon, in stages.


GPUtester:

cpp/include/raft/comms/mpi_comms.hpp
```diff
@@ -59,39 +59,39 @@
...
-constexpr MPI_Datatype get_mpi_datatype(const datatype_t datatype) {
+constexpr MPI_Datatype get_mpi_datatype(const DataTypeT datatype) {
   switch (datatype) {
-    case datatype_t::CHAR:
-      return MPI_CHAR;
-    case datatype_t::UINT8:
-      return MPI_UNSIGNED_CHAR;
-    case datatype_t::INT32:
+    case DataTypeT::kChar:
+      return MPI_kChar;
```
JohnZed, line 65: Is this an incorrect sub? This constant comes from MPI, right?

teju85: ouchy! the result of mindless regex-replace. Thanks for catching it John! Will fix it.

cpp/include/raft/comms/util.hpp
```diff
@@ -50,61 +50,61 @@
...
     default:
       throw "Unsupported datatype";
   }
 }
 
-constexpr ncclRedOp_t get_nccl_op(const op_t op) {
+constexpr ncclRedOp_t get_nccl_op(const OpT op) {
   switch (op) {
-    case op_t::SUM:
+    case OpT::kSum:
```
JohnZed, line 101: We have a bunch of old types that end in `_t` due to the old type naming convention. Should these be eliminated? It seems like an arbitrary ending now unless we do it for all types, which... ugh.

teju85: yes, my plan is to update all such types throughout the raft repo, to get those to pass clang-tidy checks.

Unless others have any objections...

teju85:
Here's my process while I was fixing the tidy errors:
1. Try to keep interface-level changes as minimal as possible. This means adding a `// NOLINT` comment against those offenders. I'll file an issue to fix those over the next few releases, thereby spreading the cost of interface changes over time.
2. But that said, I have also deliberately introduced a few interface changes so as to show the reviewers a feeling of how they would look like. If we decide these changes are not worth doing in this PR itself, I'll revert them and instead put a `// NOLINT` tag against those.
3. Unit-tests are an exception since they are local to this repo. So, I have taken the liberty to introduce a base test-fixture class in order to limit the number of tidy errors.

I'd need feedback from the following person(s):
1. cuML side: @JohnZed or @divyegala or @mdemoret-nv or @cjnolet 
2. cuGraph side: @afender or @seunghwak

aschaffer:
I've noticed an issue in the past that turned out to be due to clang-tidy: it seemingly rejected C++ source files that have headers included between 2bl quotes and wanted inclusion via angular brackets, instead. I thought that might be a good thing, initially.

In hindsight, I'm not sure if that's acceptable. As a developer I'd like to have control over _how_ headers are included in C++.

Symptoms first signaled here:
https://nvidia.slack.com/archives/CCE7L5XT3/p1595523810033400
https://nvidia.slack.com/archives/CKTK6TV9T/p1595522462404000

Is that something that was addressed (or discarded as a non-issue)? Thanks.

teju85:
@aschaffer the header file double vs angle-bracket usage standardization is actually not part of clang, but enforced via a separate script [here](https://github.com/rapidsai/raft/blob/branch-0.17/cpp/scripts/include_checker.py).

teju85:
Closing this PR for now.

PR #94, opened 2020-11-11
[REVIEW] Allow generic reductions for the map then reduce op

tfeher:
This PR adds a mapThenReduce prim, analogous to mapThenSumReduce.

### DIFF
--- a/cpp/include/raft/linalg/map_then_reduce.cuh
+++ b/cpp/include/raft/linalg/map_then_reduce.cuh
@@ -24,8 +24,10 @@
 namespace raft {
 namespace linalg {
 
+struct sum_tag {};
+
 template <typename Type, int TPB>
-__device__ void reduce(Type *out, const Type acc) {
+__device__ void reduce(Type *out, const Type acc, sum_tag) {
   typedef cub::BlockReduce<Type, TPB> BlockReduce;
   __shared__ typename BlockReduce::TempStorage temp_storage;
   Type tmp = BlockReduce(temp_storage).Sum(acc);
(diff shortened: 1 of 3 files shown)


divyegala (changes requested):
Just a small suggestion, otherwise LGTM!

cpp/include/raft/linalg/map_then_reduce.cuh
```diff
@@ -76,9 +92,35 @@ void mapThenSumReduceImpl(Type *out, size_t len, MapOp map, cudaStream_t stream,
...
+ * @param map the device-lambda
+ * @param op the reduction device lambda
+ * @param stream cuda-stream where to launch this kernel
+ * @param in the input array
+ * @param args additional input arrays
+ */
+
+template <typename Type, typename MapOp, typename ReduceLambda, int TPB = 256,
+          typename... Args>
+void mapThenReduce(Type *out, size_t len, Type neutral, MapOp map,
```
divyegala, line 119: Maybe we should specialize the templates for input and output differently as `Type1` and `Type2`? Summation or multiplication reduction could easily cause an underflow/overflow, and this will help protect against that

divyegala (commented):
Adding requested change

cpp/include/raft/linalg/map_then_reduce.cuh
tfeher: Thanks Divye, this is a good observation! There is a small issue here: the [CUB block reduce ops](https://nvlabs.github.io/cub/classcub_1_1_block_reduce.html#a089953b3bdfe7c48208632d0cc2ac1fb) that we use under the hood have the same input and output type. So there will be TPB = 256 elements reduced into the same Type1, and the rest of the atomic reductions we can perform with Type2. It probably still make sense to make the suggested change, I will push a commit soon.

tfeher:
Thanks @divyegala for the review, I have updated the ops to allow different output type, and updated the tests.

dantegd:
@divyegala could you take a look and re-review? Thanks!

cpp/include/raft/linalg/map_then_reduce.cuh
```diff
@@ -24,20 +24,34 @@
...
+    raft::myAtomicReduce(out, tmp, op);
+  }
+}
+
+template <typename InType, typename OutType, typename MapOp,
+          typename ReduceLambda, int TPB, typename... Args>
+__global__ void mapThenReduceKernel(OutType *out, size_t len, InType neutral,
+                                    MapOp map, ReduceLambda op,
+                                    const InType *in, Args... args) {
+  InType acc = neutral;
```
divyegala, line 54: Can we not use `OutType` here? So something like:
```
OutType acc = neutral;
acc = map(in[idx], ...)
```
This will ensure the mapping operation casts our input from `InType` to `OutType`, and then CUB reduce will reduce those 256 elements in `OutType` and so will the atomics

divyegala (approved):
Pre-approving in case my suggestion below looks invalid. If it is valid, then great! Thanks, Tamas

cpp/include/raft/linalg/map_then_reduce.cuh
tfeher: Thanks Divye, I think that would work! Let me have a look  with a fresh mind tomorrow morning, and I will make the change afterwards.

tfeher: This would work, I have made the necessary changes. Thanks @divyegala for the suggestion!

tfeher:
All issues are resolved as requested.

dantegd (approved):

PR #95, opened 2020-11-12
[REVIEW] Cholesky rank one update prim

tfeher:
This PR implements rank one update for Cholesky decomposition. 

More info https://en.wikipedia.org/wiki/Cholesky_decomposition#Rank-one_update

This is a building block for LARS.

In addition to the new prim, cuBLAS wrappers for swap and copy are added.

### DIFF
--- a/cpp/include/raft/linalg/cholesky_r1_update.cuh
+++ b/cpp/include/raft/linalg/cholesky_r1_update.cuh
@@ -1,0 +1,199 @@
+/*
+ * Copyright (c) 2020, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


dantegd (approved):
looks great

cpp/include/raft/linalg/cholesky_r1_update.cuh
```diff
@@ -0,0 +1,199 @@
...
+    stream);
+
+  // Check for numeric error: the above sqrt can have negative argument if the
+  // matrix is not positive definite or very ill conditioned.
+  math_t L_22_host;
+  raft::update_host(&L_22_host, L_22, 1, stream);
+  CUDA_CHECK(cudaStreamSynchronize(stream));
+  if (std::isnan(L_22_host)) {
+    THROW("Error during Cholesky rank one update");
+  }
```
teju85, line 196: nitpick: this could be done via `ASSERT` macro for simplicity, right?

teju85 (changes requested):

cpp/include/raft/linalg/cholesky_r1_update.cuh
tfeher: Yes, thanks for pointing this out! I have replaced it with assert.

teju85 (approved):
Changes LGTM.

dantegd:
rerun tests

dantegd:
rerun tests

PR #109, opened 2020-12-01
[REVIEW] Fixing Incorrect Deallocation Size and Count Bugs

mdemoret-nv:
During the development of this PR: https://github.com/rapidsai/rmm/pull/626, I discovered a couple of issues that prevented the new `TrackingMemoryResource` from functioning properly. In some cases, `deallocate()` was being called with the wrong number of bytes or shouldn't have been called at all. This inhibits the `TrackingMemoryResource` from functioning correctly since the allocation bytes and deallocation bytes do not match.

This PR fixes these issues and adds a unit test to check for it moving forward.

### DIFF
--- a/cpp/include/raft/mr/host/allocator.hpp
+++ b/cpp/include/raft/mr/host/allocator.hpp
@@ -43,9 +43,9 @@ class default_allocator : public allocator {
   }
 
   void deallocate(void* p, std::size_t n, cudaStream_t stream) override {
-    ///@todo: enable this once logging is enabled in raft
-    //CUDA_CHECK_NO_THROW(cudaFreeHost(p));
-    CUDA_CHECK(cudaFreeHost(p));
+    //Must call _NO_THROW here is this is called frequently from object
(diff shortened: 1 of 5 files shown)


GPUtester:

cpp/include/raft/mr/host/allocator.hpp
```diff
@@ -43,9 +43,9 @@ class default_allocator : public allocator {
   }
 
   void deallocate(void* p, std::size_t n, cudaStream_t stream) override {
-    ///@todo: enable this once logging is enabled in raft
-    //CUDA_CHECK_NO_THROW(cudaFreeHost(p));
-    CUDA_CHECK(cudaFreeHost(p));
+    //Must call _NO_THROW here is this is called frequently from object
```
dantegd, line 46:
```suggestion
    //Must call _NO_THROW here since this is called frequently from object
```

dantegd (approved):
change lgtm, just left a comment for a comment typo, so pre-approving

PR #114, opened 2020-12-10
[REVIEW] Add gather & gatherv to raft::comms::comms_t

seunghwak:
- Add gather & gatherv to raft::comms::comms_t, raft::comms::std_comms_t, and raft::comms::mpi_comms_t

### DIFF
--- a/cpp/include/raft/comms/comms.hpp
+++ b/cpp/include/raft/comms/comms.hpp
@@ -130,6 +130,15 @@ class comms_iface {
                           const size_t* recvcounts, const size_t* displs,
                           datatype_t datatype, cudaStream_t stream) const = 0;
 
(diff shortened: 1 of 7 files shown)


seunghwak:
This addresses #114

cjnolet (changes requested):
The addition LGTM. Can you also expose a small test through `test.hpp`? Ideally we would also expose it to the Python layer for sanity checks.

seunghwak:
> The addition LGTM. Can you also expose a small test through `test.hpp`? Ideally we would also expose it to the Python layer for sanity checks.

Sure, I will. I will test this locally with cuGraph first, and then, I will add a test suite.

seunghwak:
rerun tests.

afender (approved):
Looks good to me

cjnolet (approved):
LGTM pending CI

PR #116, opened 2020-12-18
Adding additional distances

cjnolet:
Closes https://github.com/rapidsai/raft/issues/118

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -160,7 +160,7 @@ if(CMAKE_COMPILER_IS_GNUCXX)
 endif(CMAKE_COMPILER_IS_GNUCXX)
 
 set(CMAKE_CUDA_FLAGS
-  "${CMAKE_CUDA_FLAGS} -Xcudafe --diag_suppress=unrecognized_gcc_pragma")
+  "${CMAKE_CUDA_FLAGS} -Xcudafe --diag_suppress=unrecognized_gcc_pragma --expt-relaxed-constexpr")
 
 ##############################################################################
 # - dependencies -------------------------------------------------------------
--- a/cpp/include/raft/linalg/distance_type.h
+++ b/cpp/include/raft/linalg/distance_type.h
@@ -19,34 +19,35 @@
 namespace raft {
 namespace distance {
 
-/** enum to tell how to compute euclidean distance */
+/** enum to tell how to compute distance */
 enum DistanceType : unsigned short {
+
   /** evaluate as dist_ij = sum(x_ik^2) + sum(y_ij)^2 - 2*sum(x_ik * y_jk) */
-  EucExpandedL2 = 0,
+  L2Expanded = 0,
   /** same as above, but inside the epilogue, perform square root operation */
-  EucExpandedL2Sqrt = 1,
+  L2SqrtExpanded = 1,
   /** cosine distance */
-  EucExpandedCosine = 2,
+  CosineExpanded = 2,
   /** L1 distance */
-  EucUnexpandedL1 = 3,
+  L1 = 3,
   /** evaluate as dist_ij += (x_ik - y-jk)^2 */
-  EucUnexpandedL2 = 4,
+  L2Unexpanded = 4,
   /** same as above, but inside the epilogue, perform square root operation */
-  EucUnexpandedL2Sqrt = 5,
+  L2SqrtUnexpanded = 5,
   /** basic inner product **/
   InnerProduct = 6,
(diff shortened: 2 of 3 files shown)


cpp/CMakeLists.txt
```diff
@@ -160,7 +160,7 @@ if(CMAKE_COMPILER_IS_GNUCXX)
 endif(CMAKE_COMPILER_IS_GNUCXX)
 
 set(CMAKE_CUDA_FLAGS
-  "${CMAKE_CUDA_FLAGS} -Xcudafe --diag_suppress=unrecognized_gcc_pragma")
+  "${CMAKE_CUDA_FLAGS} -Xcudafe --diag_suppress=unrecognized_gcc_pragma --expt-relaxed-constexpr")
```
JohnZed, line 163: why do you need this additional constexpr flag here?

cpp/include/raft/linalg/distance_type.h
```diff
@@ -35,18 +35,18 @@ enum DistanceType : unsigned short {
   EucUnexpandedL2Sqrt = 5,
   /** basic inner product **/
   InnerProduct = 6,
-  /** ChebyChev (Linf) distance **/
-  ChebyChev = 7,
+  /** Chebyshev (Linf) distance **/
+  EucUnexpandedLinf = 7,
```
JohnZed, line 39: Are there any existing callers using these old names?
I believe @lowener is about to switch over to them so tagging him here so he can see the new names.

cpp/include/raft/linalg/distance_type.h
```diff
@@ -35,18 +35,18 @@ enum DistanceType : unsigned short {
...
   /** Canberra distance **/
-  Canberra = 8
+  EucUnexpandedCanberra = 8,
```
JohnZed, line 41: I'm surprised that Canberra is tagged as being `Euc` (Euclidean?). I would've thought that Euclidean only applies to root of square distance?

JohnZed (commented):
Just a couple quick questions before approving!

cpp/include/raft/linalg/distance_type.h
cjnolet: I thought the same about many of the other distances as well (e.g. L1) and I'm just following the convention already established here. I can change it, but I'd prefer it be consistent (for L1 to be changed). I'm assuming Euclidean in this context to just be derived from the generalized Lp.

cpp/include/raft/linalg/distance_type.h
cjnolet: These names were put here explicitly for this PR so there shouldn't be any existing code using them.

cpp/CMakeLists.txt
cjnolet: This isn't needed for this particular PR but it allows us to call constants from the std libraries directly in CUDA kernels. If not merged with this PR, it will be made when we move the sparse prims over.

cpp/include/raft/linalg/distance_type.h
cjnolet: After offline discussion, opened https://github.com/rapidsai/raft/issues/118

cjnolet:
rerun tests

cpp/include/raft/linalg/distance_type.h
```diff
@@ -35,18 +35,18 @@ enum DistanceType : unsigned short {
   EucUnexpandedL2Sqrt = 5,
   /** basic inner product **/
   InnerProduct = 6,
-  /** ChebyChev (Linf) distance **/
-  ChebyChev = 7,
+  /** Chebyshev (Linf) distance **/
+  UnexpandedLinf = 7,
```
JohnZed, line 39: Sorry to keep the naming discussion ongoing... I'm not sure the unexpanded/expanded prefixes are necessary for the non-euclidean norms where we don't have both versions? E.g. we don't use it for InnerProduct. I'm not sure what an expanded Linf would look like either, so maybe we could omit those?

cjnolet: The expanded/unexpanded just have to do with whether they can be derived from an inner product, not necessarily Euclidean. Linf & L1 don't have expanded forms as far as I know, but the current L1 distance is already marked as `EucUnexpandedL1` so I just kept that naming convention to keep things consistent. I can remove it from the Linf, but since the dense pairwise distance implementation uses the L1 enum, we would need to update that in cuml as well.

JohnZed (approved):

cjnolet:
rerun tests

cjnolet:
rerun tests

PR #119, opened 2021-01-12
Some MST updates for single-linkage / HDBSCAN clustering

cjnolet:
We may not need this for single-linkage / hdbscan clustering. Not sure yet.

### DIFF
--- a/cpp/include/raft/sparse/mst/detail/mst_kernels.cuh
+++ b/cpp/include/raft/sparse/mst/detail/mst_kernels.cuh
@@ -107,7 +107,8 @@ __global__ void min_edge_per_supervertex(
   const vertex_t* color, const vertex_t* color_index, edge_t* new_mst_edge,
   bool* mst_edge, const vertex_t* indices, const weight_t* weights,
   const weight_t* altered_weights, vertex_t* temp_src, vertex_t* temp_dst,
-  weight_t* temp_weights, const weight_t* min_edge_color, const vertex_t v) {
+  weight_t* temp_weights, const weight_t* min_edge_color, const vertex_t v,
+  bool symmetrize_output) {
   auto tid = get_1D_idx<vertex_t>();
   if (tid < v) {
     vertex_t vertex_color_idx = color_index[tid];
@@ -130,7 +131,7 @@ __global__ void min_edge_per_supervertex(
         if (dst_edge_idx != std::numeric_limits<edge_t>::max() &&
             indices[dst_edge_idx] == tid &&
             min_edge_color[dst_color] == altered_weights[dst_edge_idx]) {
-          if (vertex_color < dst_color) {
+          if (symmetrize_output || vertex_color < dst_color) {
             add_edge = true;
           }
         } else {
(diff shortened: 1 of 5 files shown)


cpp/include/raft/sparse/mst/detail/mst_kernels.cuh
```diff
@@ -109,23 +109,38 @@ __global__ void min_edge_per_supervertex(
...
+        auto dst_color = color[color_index[dst]];
+        // vertices added each other
+        // only if destination has found an edge
+        // the edge points back to source
+        // the edge is minimum edge found for dst color
+        if (dst_edge_idx != std::numeric_limits<edge_t>::max() &&
+            indices[dst_edge_idx] == tid &&
+            min_edge_color[dst_color] == altered_weights[dst_edge_idx]) {
+          if (vertex_color < dst_color) {
+            temp_src[tid] = tid;
```
divyegala, line 133: Can you break this logic out with a boolean to avoid code repetition?

cpp/include/raft/sparse/mst/detail/mst_solver_inl.cuh
```diff
@@ -71,6 +71,8 @@ MST_solver<vertex_t, edge_t, weight_t>::MST_solver(
     mst_edge_count(1, 0),
     prev_mst_edge_count(1, 0),
     stream(stream_) {
+  printf("GOT HERE!!!\n");
```
divyegala, line 74: Trailing print

cpp/include/raft/sparse/mst/detail/mst_solver_inl.cuh
```diff
@@ -339,15 +341,16 @@ void MST_solver<vertex_t, edge_t, weight_t>::min_edge_per_supervertex() {
   // the above kernel only adds directed mst edges in the case where
   // a pair of vertices don't pick the same min edge between them
   // so, now we add the reverse edge to make it undirected
-  detail::add_reverse_edge<<<nblocks, nthreads, 0, stream>>>(
-    new_mst_edge_ptr, indices, weights, temp_src_ptr, temp_dst_ptr,
-    temp_weights_ptr, v);
+  //  detail::add_reverse_edge<<<nblocks, nthreads, 0, stream>>>(
```
divyegala, line 344: An extra comment here to document why this kernel launch is commented

divyegala (changes requested):
We can adjust the tests as discussed offline, and this should be good to go!

cpp/include/raft/sparse/mst/detail/mst_kernels.cuh
cjnolet: I believe this was a change you had provided. If you can provide an example of the change you'd like to see, I don't mind adding it to this PR.

cpp/include/raft/sparse/mst/detail/mst_solver_inl.cuh
cjnolet: I had forgotten about this change. @afender, this will probably break cugraph if we remove the symmetry altogether right? Should we make this optional?

divyegala: If this is made optional, then the changes to the above additions will also have to be made optional

cjnolet: Just to make note here- we talked offline about this. @divyegala will push necessary updates directly to this branch (or open a follow-on if necessary). SLHC is dependent upon this change but we can't break cugraph's assumption of symmetric outputs.

PR #120, opened 2021-01-13
[REVIEW] Make NCCL root initialization configurable.

drobison00:
This updates the raft comms implementation to support configurable NCCL root placement, updates scheduler and worker routine logging so that it is visible and usable in a multi-node deployment, and adds additional error handling to the NCCL initialization process to fail early if errors are encountered.

Updates resolve MNMG failures in out of band cuML algorithms where the DASK client does not have the ability to directly communicate with workers.

Unit tests are implemented and passing.

Closes https://github.com/rapidsai/cuml/issues/3261

### DIFF
--- a/python/raft/dask/common/comms.py
+++ b/python/raft/dask/common/comms.py
@@ -377,7 +377,7 @@ def _func_set_scheduler_as_nccl_root(sessionId, verbose, dask_scheduler):
         session_state["nccl_uid"] = nccl.get_unique_id()
 
     if verbose:
-        logger.info(f"Done setting scheduler as NCCL root.")
+        logger.info("Done setting scheduler as NCCL root.")
 
     return session_state["nccl_uid"]
 
@@ -412,7 +412,7 @@ def _func_set_worker_as_nccl_root(sessionId, verbose):
 
     if verbose:
         get_worker().log_event(
-            topic="info", msg=f"Done setting scheduler as NCCL root."
+            topic="info", msg="Done setting scheduler as NCCL root."
         )
 
     return session_state["nccl_uid"]
@@ -492,7 +492,7 @@ def _func_init_nccl(sessionId, uniqueId):
         worker_state(sessionId)["nccl"] = n
     except Exception as e:
         get_worker().log_event(
-            topic="error", msg="An error occurred initializing NCCL!."
+            topic="error", msg=f"An error occurred initializing NCCL: {e}."
         )
         raise
 
--- a/python/raft/dask/common/comms.py
+++ b/python/raft/dask/common/comms.py
@@ -80,10 +80,16 @@ class Comms:
         cluster.close()
     """
 
-    valid_nccl_placements = ('client', 'worker', 'scheduler')
+    valid_nccl_placements = ("client", "worker", "scheduler")
 
-    def __init__(self, comms_p2p=False, client=None, verbose=False,
-                 streams_per_handle=0, nccl_root_location="scheduler"):
+    def __init__(
+        self,
+        comms_p2p=False,
+        client=None,
+        verbose=False,
+        streams_per_handle=0,
+        nccl_root_location="scheduler",
+    ):
         """
         Construct a new CommsContext instance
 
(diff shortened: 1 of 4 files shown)


drobison00:
Rerun tests

drobison00:
Tests currently failing in CI appear to be due to : https://github.com/rapidsai/ucx-py/issues/668

python/raft/dask/common/comms.py
```diff
@@ -90,10 +101,22 @@ def __init__(self, comms_p2p=False, client=None, verbose=False,
...
+        nccl_root_location : string
+                  Indicates where the NCCL's root node should be located.
+                  ['client', 'worker', 'scheduler' (default)]
+
         """
         self.client = client if client is not None else default_client()
+
         self.comms_p2p = comms_p2p
 
+        if nccl_root_location.lower() not in Comms.valid_nccl_placements:
```
cjnolet, line 113: Little nitpick- can we call `lower()` once? Maybe assign to `self` first and then raise the error if `self.nccl_root_location not in Comms.valid_nccl_placements`.

python/raft/dask/common/comms.py
```diff
@@ -149,17 +177,34 @@ def init(self, workers=None):
...
-        self.client.run(_func_init_all,
-                        self.sessionId,
-                        self.uniqueId,
-                        self.comms_p2p,
-                        worker_info,
-                        self.verbose,
-                        self.streams_per_handle,
-                        workers=self.worker_addresses,
-                        wait=True)
+        if self.nccl_root_location == "client":
```
cjnolet, line 180: Can we pull this logic into a function like `create_nccl_uniqueid()`?

python/raft/dask/common/comms.py
```diff
@@ -207,6 +260,38 @@ def local_handle(sessionId):
...
+    dask_scheduler : Dask Scheduler object
+
+    Returns
+    -------
+
+    session state : str
+                    session state associated with sessionId
+    """
+
+    if not hasattr(dask_scheduler, "_raft_comm_state"):
```
cjnolet, line 281: This function looks like it's pretty much a copy of `worker_state`. Could we consolidate these into a function that takes the scheduler/worker instance?

python/raft/dask/common/comms.py
```diff
@@ -241,46 +326,145 @@ def get_ucx():
...
+            msg=f"Setting scheduler as NCCL "
+            f"root for sessionId, '{sessionId}'"
+        )
+
+    if sessionId is None:
+        raise ValueError("sessionId cannot be None.")
+
+    session_state = scheduler_state(
+        sessionId=sessionId, dask_scheduler=dask_scheduler
+    )
```
cjnolet, line 375: Same with `_func_set_worker_as_nccl_root` and `_func_set_scheduler_as_nccl_root`, these look exactly the same with the only difference being the `get_worker` or `dask_scheduler`. I'm also okay just having these separate functions call a shared function like `_func_set_as_nccl_root(sessionId, process_instance, verbose)`

python/raft/test/test_comms.py
```diff
@@ -65,6 +68,47 @@ def func_test_comm_split(sessionId, n_trials):
     return perform_test_comm_split(handle, n_trials)
 
 
+def func_check_uid_on_scheduler(sessionId, uniqueId, dask_scheduler):
+    if not hasattr(dask_scheduler, "_raft_comm_state"):
```
cjnolet, line 72: Same here (with regards to consolidation)

cjnolet (changes requested):
Looks good overall. Just a few minor things to consolidate some of the logic and ease the maintenance burden a little bit.

drobison00:
Rerun tests

dantegd:
rerun tests

cjnolet (approved):
LGTM

dantegd:
@drobison00 this is a breaking change, right? Then cugraph will also need a similar PR to https://github.com/rapidsai/cuml/pull/3386 to upgrade their tagged commit

cc @afender @BradReesWork for vis

drobison00:
@dantegd added a PR to reflect the changes: https://github.com/rapidsai/cugraph/pull/1343

Once the raft update is merged, I can update git tags as well.

PR #126, opened 2021-01-27
[REVIEW] Add brute force KNN

hlinsen:
This PR moves cuML's brute force knn implementation to RAFT.


build.sh
```diff
@@ -18,7 +18,7 @@ ARGS=$*
 # script, and that this script resides in the repo dir!
 REPODIR=$(cd $(dirname $0); pwd)
 
-VALIDARGS="clean cppraft pyraft -v -g --allgpuarch --nvtx --show_depr_warn -h --buildgtest"
+VALIDARGS="clean cppraft pyraft -v -g --allgpuarch --nvtx --show_depr_warn -h --buildgtest --buildfaiss"
```
dantegd, line 21: I think @cjnolet is actually reviewing the code itself, so just asking about this: why default to building faiss statically as opposed to using the conda package? Given that cuML already has the faiss conda package as a dependency, I think we should do the same here

dantegd (changes requested):

build.sh
hlinsen: I'm fine with removing it as well. I added it for consistency since cuml has that option and we also do it for gtest

dantegd: Totally my mistake and for some reason thought it was defaulting to building static faiss by default, so my comment can be disregarded :)

cpp/include/raft/spatial/knn/detail/brute_force_knn.hpp
```diff
@@ -0,0 +1,342 @@
...
+#include <thrust/device_vector.h>
+#include <thrust/iterator/transform_iterator.h>
+#include <iostream>
+#include <raft/handle.hpp>
+#include <set>
+
+#include "processing.hpp"
+
+namespace raft {
+namespace knn {
```
cjnolet, line 23: We should probably also include the `spatial` namespace in the hierarchy here for consistency.

cpp/include/raft/spatial/knn/detail/processing.hpp
```diff
@@ -0,0 +1,190 @@
+/*
+ * Copyright (c)2020, NVIDIA CORPORATION.
```
cjnolet, line 2: We should update all the years here to `2020-2021`

cjnolet (changes requested):
Looks great so far. Also tagging @lowener since he's been working in this area of the code recently.

cpp/include/raft/spatial/knn/detail/brute_force_knn.hpp
hlinsen: Fixed here https://github.com/rapidsai/raft/pull/126/commits/9315752555050877f934522a08709b5dc2daff8a

cpp/include/raft/spatial/knn/detail/processing.hpp
hlinsen: Updated here: https://github.com/rapidsai/raft/pull/126/commits/fe5a378c93b3e1a719b10ec9f72d6a5528e46a07

cjnolet:
rerun tests

cjnolet (approved):
LGTM

PR #125, opened 2021-01-27
[REVIEW] Update check to allow svd solver on CUDA>=10.2

wphicks:
Allow CUDA 10.2 or greater to use svd solver with greater than 46340 rows/columns; eliminate restriction relevant to CUDA 10.1 only

PR #129, opened 2021-01-27
[REVIEW] Eliminate CUDA 10.2 as valid for large svd solving

wphicks:
Do not allow use of svd solver for input with more than 46430 rows until CUDA 11


wphicks:
In #125, I thought that the issue had been resolved in CUDA 10.2, but it appears that I ran my test in the wrong environment. Updating to make CUDA 11 the new minimum. Let's let the tests on https://github.com/rapidsai/cuml/pull/3417 run through to make sure that everything is good before we merge this in.

wphicks:
cuML PR tests passed; I think we're good to go.

wphicks:
@JohnZed could I trouble you for review on this one since you already looked at the other one?

JohnZed (approved):

JohnZed:
@gpucibot merge

PR #131, opened 2021-01-28
[Review] Add idx_t template parameter to matrix helper routines

tfeher:
This PR introduces a template type parameter for the matrix helper functions. The previous int type would overflow while accessing elements of a matrix that is larger than 2^31 elements. 

Related to https://github.com/rapidsai/cuml/issues/3233

A test is added for the copyRows function.

### DIFF
--- a/cpp/include/raft/matrix/matrix.cuh
+++ b/cpp/include/raft/matrix/matrix.cuh
@@ -47,21 +47,21 @@ using namespace std;
  * @param stream cuda stream
  * @param rowMajor whether the matrix has row major layout
  */
-template <typename m_t>
-void copyRows(const m_t *in, int n_rows, int n_cols, m_t *out,
-              const int *indices, int n_rows_indices, cudaStream_t stream,
-              bool rowMajor = false) {
+template <typename m_t, typename idx_array_t = int, typename idx_t = size_t>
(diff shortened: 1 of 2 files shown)


tfeher:
Right now it is the callers responsibility to ensure that idx_t can represent all the offsets. We could additionally define the internal offset variables to be size_t to be on the safe side. Any opinion on this?

JohnZed:
This looks good, @tfeher. To me, it seems fair to let the caller be responsible for having a sufficiently large index type. But if the concern is that nrows and ncols fits in (say) int and nrows * ncols is > max_int, maybe we could assert that to provide a clear error? Like, `if (sizeof(idx_t) < sizeof(size_t)) { assert((size_t)nrows * (size_t)ncols)) < numeric_limits<idx_t>::max()`?

JohnZed (approved):
LGTM. (But see other comment for thoughts on possible safeguards for overflow - can be a follow on change.)

tfeher:
Thanks John for the good suggestion! I think improving the error message can be done in a follow up. Actually, it might make sense to go through the code base and see if there are similar issues elsewhere. As we are pushing the limits on the problem sizes that we can solve we are likely to run into such issues.

JohnZed:
@gpucibot merge

PR #139, opened 2021-02-08
[REVIEW] Moving cuml sparse prims to raft

cjnolet:
Moving cuml sparse prims and gtests to raft. The namespaces have already been adjusted in cuml so this is a simple move for the most part. There are a few places where includes needed to be updated and I Needed to remove cuml debug logs.

PR #140, opened 2021-02-08
[REVIEW] Adding SLHC prims. 

cjnolet:
The dense pairwise distances API in cuML currently relies on CUTLASS and will soon be re-written to remove that dependency. Until then, the SLHC implementation in RAFT cannot depend on that API. In the meantime, I've abstracted that piece out to make it pluggable on the cuML side.

This PR depends on the changes in #139.


cjnolet:
Depends on #119

ajschmidt8:

cpp/include/raft/sparse/hierarchy/detail/agglomerative.cuh
```diff
@@ -0,0 +1,366 @@
...
+                                                  stream, child_size);
+
+  value_idx children_cpy_start = n_edges - child_size;
+  copy_async(label_roots.data(), children + children_cpy_start, child_size,
+             stream);
+
+  thrust::device_ptr<value_idx> t_label_roots =
+    thrust::device_pointer_cast(label_roots.data());
+
+  thrust::sort(thrust::cuda::par.on(stream), t_label_roots,
```
divyegala, line 326: Same comment about exec policy

cpp/include/raft/sparse/hierarchy/detail/agglomerative.cuh
```diff
@@ -0,0 +1,366 @@
...
+               t_label_roots + (child_size), thrust::greater<value_idx>());
+
+  raft::mr::device::buffer<value_idx> tmp_labels(handle.get_device_allocator(),
+                                                 stream, n_vertices);
+
+  // Init labels to -1
+  thrust::device_ptr<value_idx> t_labels =
+    thrust::device_pointer_cast(tmp_labels.data());
+  thrust::fill(thrust::cuda::par.on(stream), t_labels, t_labels + n_vertices,
```
divyegala, line 335: Same comment about exec policy

cpp/include/raft/sparse/hierarchy/detail/agglomerative.cuh
```diff
@@ -0,0 +1,366 @@
...
+               -1);
+
+  // Write labels for cluster roots to "labels"
+  thrust::counting_iterator<uint> first(0);
+  thrust::counting_iterator<uint> last = first + tmp_labels.size();
+
+  auto z_iter = thrust::make_zip_iterator(thrust::make_tuple(
+    first, t_label_roots + (label_roots.size() - n_clusters)));
+
+  thrust::for_each(thrust::cuda::par.on(stream), z_iter, z_iter + n_clusters,
```
divyegala, line 345: Same comment about exec policy

cpp/include/raft/sparse/hierarchy/detail/connectivities.cuh
```diff
@@ -0,0 +1,136 @@
...
+           raft::mr::device::buffer<value_t> &data, int c);
+};
+
+/**
+ * Connectivities specialization to build a knn graph
+ * @tparam value_idx
+ * @tparam value_t
+ */
+template <typename value_idx, typename value_t>
+struct distance_graph_impl<LinkageDistance::KNN_GRAPH, value_idx, value_t> {
```
divyegala, line 54: Is `LinkageDistance::Pairwise` not supported yet? Also fully qualified namespaces to resolve ambiguity when namespace is not a direct child

cpp/include/raft/sparse/hierarchy/detail/connectivities.cuh
```diff
@@ -0,0 +1,136 @@
...
+ * @param[in] n number of columns in X
+ * @param[in] metric distance metric to use
+ * @param[out] indptr indptr array of connectivities graph
+ * @param[out] indices column indices array of connectivities graph
+ * @param[out] data distances array of connectivities graph
+ * @param[out] c constant 'c' used for nearest neighbors-based distances
+ *             which will guarantee k <= log(n) + c
+ */
+template <typename value_idx, typename value_t,
+          hierarchy::LinkageDistance dist_type>
```
divyegala, line 108: Fully qualified namespace

cpp/include/raft/sparse/hierarchy/detail/mst.cuh
```diff
@@ -0,0 +1,188 @@
...
+template <typename value_idx, typename value_t>
+void sort_coo_by_data(value_idx *rows, value_idx *cols, value_t *data,
+                      value_idx nnz, cudaStream_t stream) {
+  thrust::device_ptr<value_idx> t_rows = thrust::device_pointer_cast(rows);
+  thrust::device_ptr<value_idx> t_cols = thrust::device_pointer_cast(cols);
+  thrust::device_ptr<value_t> t_data = thrust::device_pointer_cast(data);
+
+  auto first = thrust::make_zip_iterator(thrust::make_tuple(t_rows, t_cols));
+
+  thrust::sort_by_key(thrust::cuda::par.on(stream), t_data, t_data + nnz,
```
divyegala, line 53: exec policy, and no device pointer casts needed

cpp/include/raft/sparse/hierarchy/detail/mst.cuh
```diff
@@ -0,0 +1,188 @@
...
+
+    RAFT_EXPECTS(
+      mst_coo.n_edges == m - 1,
+      "MST was not able to connect knn graph in a single iteration.");
+  }
+
+  sort_coo_by_data(mst_coo.src.data(), mst_coo.dst.data(),
+                   mst_coo.weights.data(), mst_coo.n_edges, stream);
+
+  // TODO: be nice if we could pass these directly into the MST
```
divyegala, line 175: This is where you want to be able to pass a dense input, right?

cpp/include/raft/sparse/linalg/symmetrize.cuh
```diff
@@ -304,6 +304,92 @@ void from_knn_symmetrize_matrix(
...
+  CUDA_CHECK(cudaMemsetAsync(diff.data(), 0,
+                             ((nnz * 2) + 1) * sizeof(value_idx), stream));
+
+  compute_duplicates_diffs<<<raft::ceildiv(nnz * 2, (size_t)1024), 1024, 0,
+                             stream>>>(symm_rows.data(), symm_cols.data(),
+                                       diff.data(), nnz * 2);
+
+  thrust::device_ptr<value_idx> dev = thrust::device_pointer_cast(diff.data());
+
+  thrust::exclusive_scan(thrust::cuda::par.on(stream), dev, dev + diff.size(),
```
divyegala, line 374: exec policy and device pointer cast

cpp/include/raft/sparse/selection/connect_components.cuh
```diff
@@ -0,0 +1,454 @@
...
+ * @tparam value_idx
+ * @param[in] colors array of components
+ * @param[in] n_rows size of components array
+ * @param[in] stream cuda stream for which to order cuda operations
+ * @return total number of components
+ */
+template <typename value_idx>
+value_idx get_n_components(value_idx *colors, size_t n_rows,
+                           cudaStream_t stream) {
+  thrust::device_ptr<value_idx> t_colors = thrust::device_pointer_cast(colors);
```
divyegala, line 244: exec policy and device ptr cast

cpp/include/raft/sparse/selection/connect_components.cuh
```diff
@@ -0,0 +1,454 @@
...
+   * Create COO array by first computing CSR indptr w/ degrees of each
+   * color followed by COO row/col/val arrays.
+   */
+  // map each component to a separate warp, perform warp reduce by key to find
+  // number of unique components in output.
+
+  count_components_by_color(degrees, components_indptr, nn_components,
+                            n_components, stream);
+
+  thrust::device_ptr<value_idx> t_degrees =
```
divyegala, line 278: exec policy and device ptr cast

cpp/include/raft/sparse/selection/connect_components.cuh
```diff
@@ -0,0 +1,454 @@
...
+  raft::linalg::rowNorm(x_norm.data(), X, n_cols, n_rows, raft::linalg::L2Norm,
+                        true, stream);
+
+  FixConnectivitiesRedOp<value_idx, value_t> red_op(colors, n_rows);
+  raft::distance::fusedL2NN<value_t, cub::KeyValuePair<value_idx, value_t>,
+                            value_idx>(kvp, X, X, x_norm.data(), x_norm.data(),
+                                       n_rows, n_rows, n_cols, workspace.data(),
+                                       red_op, red_op, true, true, stream);
+
+  thrust::device_ptr<cub::KeyValuePair<value_idx, value_t>> t_kvp =
```
divyegala, line 332: exec policy and device ptr cast

cpp/include/raft/sparse/selection/connect_components.cuh
```diff
@@ -0,0 +1,454 @@
...
+ * @param[inout] src_indices array of source vertex indices which will become arg_sort
+ *               indices
+ * @param n_rows number of components in `colors`
+ * @param stream stream for which to order CUDA operations
+ */
+template <typename value_idx, typename value_t>
+void sort_by_color(value_idx *colors, value_idx *nn_colors,
+                   cub::KeyValuePair<value_idx, value_t> *kvp,
+                   value_idx *src_indices, size_t n_rows, cudaStream_t stream) {
+  thrust::device_ptr<value_idx> t_colors = thrust::device_pointer_cast(colors);
```
divyegala, line 358: exec policy and device ptr cast

divyegala (changes requested):
This is my first pass review, where I have not looked over the logic at all yet. What are your thoughts on removing the usage of `raft::mr::device::buffer` to favor `rmm::device_uvector` so we can start having new PRs not have that and eventually move towards removing our allocator completely?

cpp/include/raft/sparse/hierarchy/detail/agglomerative.cuh
```diff
@@ -0,0 +1,366 @@
...
+   *
+   *     1. Initialize "levels" array of size n_leaves * 2
+   *
+   *     2. For each entry in children, write parent
+   *        out for each of the children
+   */
+
+  size_t n_edges = (n_leaves - 1) * 2;
+
+  thrust::device_ptr<value_idx> d_ptr =
```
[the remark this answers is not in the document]
cjnolet: Thrust functions will run the host path when a pointer to device memory is used directly. see: https://docs.nvidia.com/cuda/thrust/index.html#iterators-and-static-dispatching.

Unfortunately, while we want the function signature here to let the caller know we don't intend on modifying `chilren`, I get a compile error when trying to use `device_ptr_cast` on const, so I'm casting the const away temporarily for purposes of the computation. If you know a better way to do this then I'd be willing to try it out.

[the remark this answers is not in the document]
divyegala: That's why I mentioned about using `rmm::exec_policy` which is a device policy and uses RMM's allocators. So I think device ptr's are unneeded. As for const, you can also do `thrust::device_ptr<const value_idx>` (for future purposes)

cpp/include/raft/sparse/hierarchy/detail/connectivities.cuh
cjnolet: The `Pairwise` option lives in cuML at the moment while we await the removal of cutlass from our dense pairwise distances API.

cpp/include/raft/sparse/hierarchy/detail/mst.cuh
cjnolet: No, the unfortunate side effect of the pointers being constructed on the heap by the function being called is that I have no direct control over the memory being used. What if I wanted to hand you a specific chunk of memory to use? This is one of the reasons I'm in favor of handing the function an object that hasn't yet been initialized and allowing the function to call an "initialize" (specifically in the case of sparse data, where the nnz might not be known ahead of time). It allows me to maintain ownership over the object being constructed from the start, and temporarily grants the ability for the called function to mutate that object. My problem with this inversion of ownership is that now it leaves me having to return an object from this function that was created in this function and that propagates to all functions that need to use the resulting data. I'd much rather create an object in the caller and propagate that same object through the call stack  (as we do with the `raft::sparse:COO` everywhere.

cpp/include/raft/sparse/hierarchy/detail/agglomerative.cuh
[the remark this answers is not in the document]
cjnolet: This causes a segfault:
```
  auto thrust_policy = rmm::exec_policy(stream);
  value_idx n_vertices = *(thrust::max_element(thrust_policy,
                                               children, children + n_edges)) +
                         1;
```

This does not:

```
  auto thrust_policy = rmm::exec_policy(stream);
  thrust::device_ptr<const value_idx> d_ptr =
    thrust::device_pointer_cast(children);
  value_idx n_vertices = *(thrust::max_element(thrust_policy,
                                               d_ptr, d_ptr + n_edges)) +
                         1;
```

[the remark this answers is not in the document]
divyegala: That's weird, I have used raw pointers with thrust without any issues. Example here: https://github.com/rapidsai/raft/blob/branch-0.19/cpp/include/raft/sparse/mst/detail/mst_solver_inl.cuh#L86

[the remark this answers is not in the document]
divyegala: Can you try using this with an example that does not use `thrust::max_element`? Just to confirm with our suspicions of a bug in min-max and maybe that causing the deref there to cause a seg-fault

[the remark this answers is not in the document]
cjnolet: This one is not compiling: 
```
template <typename value_idx, typename value_t>
void sort_coo_by_data(value_idx *rows, value_idx *cols, value_t *data,
                      value_idx nnz, cudaStream_t stream) {
  auto thrust_policy = rmm::exec_policy(stream);
  thrust::device_ptr<value_idx> t_rows = thrust::device_pointer_cast(rows);
  thrust::device_ptr<value_idx> t_cols = thrust::device_pointer_cast(cols);
  thrust::device_ptr<value_t> t_data = thrust::device_pointer_cast(data);

  auto first = thrust::make_zip_iterator(thrust::make_tuple(t_rows, t_cols));

  thrust::sort_by_key(thrust_policy, data, data + nnz, first);
}
```

I see the benefit of using the exec_policy but iI'm really failing to see the benefit here when the code works with the pointer casts. Is it just because you're not in love with the extra lines of code? I can't immediately think of a potential pitfall for having these in there.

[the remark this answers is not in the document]
divyegala: Yeah, not a big fan of the extra lines of code for what I have seen to work in other places without an issue. Let's sync up offline on this and I can take a stab at compiling an example by myself. Which in this case, I'd try to have the tuple not be composed of device ptr's either, but no point in me trying to try out stuff to see if it works, when you can move ahead with more important stuff. I can just try it out myself and see if it works and update you on that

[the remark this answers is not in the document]
cjnolet: This doesn't compile either, unfortunately.

```
template <typename value_idx, typename value_t>
void sort_coo_by_data(value_idx *rows, value_idx *cols, value_t *data,
                      value_idx nnz, cudaStream_t stream) {
  auto thrust_policy = rmm::exec_policy(stream);
  auto first = thrust::make_zip_iterator(thrust::make_tuple(rows, cols));

  thrust::sort_by_key(thrust_policy, data, data + nnz, first);
}
```

There are places which have worked without explicit pointer casts. I'm okay making this change in those places and adding the rmm execution policy everywhere but I'm not completely on board w/ setting that as a general rule just for a few extra lines of code that don't seem to have any other adverse affects. 

I'll push what I have if you'd like to play around with it.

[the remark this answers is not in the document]
divyegala: Yep, that's a fair middle ground here for something that *should* work out of the box, but if it's not then we don't waste too much time on it for sure.

cjnolet:
@divyegala, thanks for taking the time to review this! I've pushed the changes based on your review. All of the changes that were using raft device buffer objects should now be using rmm device uvector.

As discussed in the review, I tried to update all the thrust functions per your request but some of them don't seem to like using the thrust policy. I don't believe any of those should be using scratch space, though, so I think we should be okay. If the remaining device pointer casts are still a concern and we can't find a quick way around them, we can always make a Github issue for it and fix it when we figure out the cause. What do you think?

divyegala (approved):
LGTM! Just the one timer left to remove

dantegd (approved):

dantegd:
@gpucibot merge

PR #148, opened 2021-02-11
Streams upgrade in RAFT handle (RMM backend + create handle from parent's pool)

### LINKED ISSUE #141: [FEA] Adopt RMM as the backend for the handle's stream pool
RMM now provides a CUDA stream pool for RAPIDS which makes our custom implementation in RAFT handle redundant.
A linear accessor was added in https://github.com/rapidsai/rmm/pull/696 which makes it compatible with existing usage in cuML.
The next step is to adopt that in RAFT.

### LINKED ISSUE #142: [FEA] Draw a worker stream from a parent handle and set it as the user stream​ in a returned handle
Many functions (including all graph prims) accept a handle and execute on the handle's user stream (accessed by `handle_t.get_stream()​`). This is an issue if the caller is also provided with a single handle but wants to run these functions concurrently. We need to leverage the pool of worker streams of the parent handle without having to change the signature and internals of low-level functions.

One way is to add a `raft::handle_t get_handle_from_pool` feature that draws a worker stream from the parent handle and set it as the user stream​ of the returned handle.

Pass streams explicitly or rewrite prims to access worker streams from the handle

**Additional context**
Needed for https://github.com/rapidsai/cugraph/issues/957

afender:
Close #141 
Close #142

Add:
- RMM pool backends for handle's internal streams
- Feature to get a new handle from an existing handle by specifying a stream in the parent's pool 
- Keep previous accessors so that no change is needed in existing codes
- Additional accessors exposing `rmm::cuda_stream_view` for convenience (comes with `is_default()` and `synchronize()` features from rmm)

Verified that creating a light handle from a parent handle is very inexpensive (microseconds).

Needed for https://github.com/rapidsai/cugraph/issues/957

### DIFF
--- a/cpp/include/raft/handle.hpp
+++ b/cpp/include/raft/handle.hpp
@@ -67,9 +67,22 @@ class handle_t {
       host_allocator_(std::make_shared<mr::host::default_allocator>()) {
     create_resources();
   }
+  handle_t(const handle_t& h) : dev_id_(h.get_device()), num_streams_(0) {}
+  handle_t(const handle_t&& h) : dev_id_(h.get_device()), num_streams_(0) {}
+
+  handle_t& operator=(const handle_t& h) {
+    prop_ = h.get_device_properties();
+    device_prop_initialized_ = true;
+    device_allocator_ = get_device_allocator();
+    host_allocator_ = get_host_allocator();
+    return *this;
+  }
 
   /** Destroys all held-up resources */
-  virtual ~handle_t() { destroy_resources(); }
+  virtual ~handle_t() {
+    std::cout << "dtor" << std::endl;
+    destroy_resources();
+  }
 
   int get_device() const { return dev_id_; }
 
@@ -134,6 +147,14 @@ class handle_t {
       int_streams_vec.push_back(s);
     }
     return int_streams_vec;
+  }
+
+  handle_t get_handle_from_internal_pool(
+    int stream_id, int n_streams = kNumDefaultWorkerStreams) const {
+    handle_t handle(n_streams);
+    handle = *this;
+    handle.set_stream(this->get_internal_stream(stream_id));
+    return handle;
   }
 
   void wait_on_user_stream() const {
--- a/cpp/include/raft/handle.hpp
+++ b/cpp/include/raft/handle.hpp
@@ -86,6 +86,9 @@ class handle_t {
 
   void set_stream(cudaStream_t stream) { user_stream_ = stream; }
   cudaStream_t get_stream() const { return user_stream_; }
+  rmm::cuda_stream_view get_stream_view() const {
+    return rmm::cuda_stream_view(user_stream_);
+  }
 
   void set_device_allocator(std::shared_ptr<mr::device::allocator> allocator) {
     device_allocator_ = allocator;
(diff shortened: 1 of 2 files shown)


cpp/include/raft/handle.hpp
```diff
@@ -62,11 +63,23 @@ class handle_t {
...
+
+  // light copy operator
+  // skip streams, comms, and libs handles
+  handle_t& operator=(const handle_t& h) {
+    prop_ = h.get_device_properties();
+    device_prop_initialized_ = true;
+    device_allocator_ = get_device_allocator();
+    host_allocator_ = get_host_allocator();
+    return *this;
+  }
```
seunghwak, line 82: If you have defined a move constructor, there should be a move operator as well.

And I feel like this is non-idiomatic to use copy/move constructors/operators for partial copies/moves (See the "rulle of five" https://cpppatterns.com/patterns/rule-of-five.html).

And typically, copy constructor can be implemented using { *this = other; } if a copy operator is defined, but here the behaviors of copy constructor & copy operator are different. This can be very surprising to some people.

cpp/include/raft/handle.hpp
```diff
@@ -126,26 +142,42 @@ class handle_t {
...
+    for (int i = 0; i < get_num_internal_streams(); i++) {
+      int_streams_vec.push_back(get_internal_stream(i));
     }
     return int_streams_vec;
   }
 
+  handle_t get_handle_from_internal_pool(
+    int stream_id, int n_streams = kNumDefaultWorkerStreams) const {
+    handle_t handle(n_streams);
+    handle = *this;
```
seunghwak, line 166: If the copy constructor behavior is idiomatic, we can do ```auto handle = *this;``` instead of the above two lines. But the behavior of the copy constructor is non-idiomatic, I assume this is not possible, and I guess this can be quite surprising to many C++ programmers.

cpp/include/raft/handle.hpp
```diff
@@ -126,26 +142,42 @@ class handle_t {
...
-    for (auto s : streams_) {
-      int_streams_vec.push_back(s);
+    for (int i = 0; i < get_num_internal_streams(); i++) {
+      int_streams_vec.push_back(get_internal_stream(i));
     }
     return int_streams_vec;
   }
 
+  handle_t get_handle_from_internal_pool(
+    int stream_id, int n_streams = kNumDefaultWorkerStreams) const {
```
seunghwak, line 164: What happens if kNumDefaultWorkerStreams > 0 (say 10)?

The parent handle will have a stream pool of 10 and the child handle will get its default stream from the parent but will have an additional stream pool with 10 streams?

cpp/include/raft/handle.hpp
afender: Thanks for noticing the move constructor operator. For completeness, I added these really because the compiler was complaining about it for the ROV in `get_handle_from_internal_pool`, but it does not end up being used.
I'll add this in my next commit. 

Answering the copy discussion in the other thread and resolving this one.

cpp/include/raft/handle.hpp
afender: Yes. This is for the case where a prim is given a light handle but need to make concurrent calls to other prims.

cpp/include/raft/handle.hpp
afender: What I really need is a light copy of some fields but not the whole handle. Since the compiler wanted a copy constructor because of the return type I just leveraged this to address what I needed it but yea I agree it's a bit hacky :)
I can probably just add a canonical copy constructor and a `get_light_copy()` feature.

afender: Actually, I did a small redesign I think it is clearer this way and less intrusive on the existing handle design. Let me know how it looks.

seunghwak (approved):

BradReesWork (approved):

cjnolet (approved):
LGTM

BradReesWork:
@gpucibot merge

PR #157, opened 2021-02-22
Add const to KNN handle

hlinsen:
A small update to the header so that we can pass a const handle to `brute_force_knn`


cjnolet (approved):
LGTM

cjnolet:
@gpucibot merge

PR #158, opened 2021-02-23
[REVIEW] Moving optimized L2 1-nearest neighbors implementation from cuml

### LINKED ISSUE #156: [FEA] Move FuseL2NN primitive from cuML into RAFT
This primitive is used to connect knn graphs for SLHC & HDBSCAN and we have decided to prioritize moving it to RAFT (as well as the primitive to connect knn graphs). This will be useful for both cuml and cugraph to build from. 

cc @teju85 @JohnZed @BradReesWork @afender

cjnolet:
Closes #156

### DIFF
--- a/cpp/include/raft/common/device_loads_stores.cuh
+++ b/cpp/include/raft/common/device_loads_stores.cuh
@@ -1,0 +1,120 @@
+/*
+ * Copyright (c) 2021, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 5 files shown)


cpp/include/raft/common/device_loads_stores.cuh
```diff
@@ -0,0 +1,120 @@
...
+namespace raft {
+
+/**
+ * @defgroup SmemStores Shared memory store operations
+ * @{
+ * @brief Stores to shared memory (both vectorized and non-vectorized forms)
+ * @param[out] addr shared memory address
+ * @param[in]  x    data to be stored at this address
+ */
+DI void sts(float* addr, const float& x) { *addr = x; }
```
divyegala, line 30: Functions with the same implementation in this file can be templated out

divyegala (approved):
LGTM. Very small comment since this is supposed to be an almost 1:1 move

teju85 (approved):
Changes LGTM. Thanks Corey for migrating this out to raft.

PR #168, opened 2021-03-08
[REVIEW] Map operation

viclafargue:
This PR adds a map operation. This can be useful when unary and binary operations are not enough.


cjnolet (approved):
LGTM. If I understand the usage correctly, I would use this if I wanted to perform anything greater than a binary element-wise operation, right?

viclafargue:
> LGTM. If I understand the usage correctly, I would use this if I wanted to perform anything greater than a binary element-wise operation, right?

Thanks for the review and merge! Exactly, that will be helpful in case you need more than two parameters.

PR #172, opened 2021-03-16
[REVIEW] Porting over recent updates to distance prim from cuml

cjnolet:
The sparse distance primitive was recently removed from cuml but Cosine and Jaccard were updated to support empty rows shortly before being removed. This PR ports those changes to the RAFT versions once and for all.


lowener (approved):

JohnZed (approved):
Lovin' those parens! ;)

JohnZed:
@gpucibot merge

PR #178, opened 2021-03-22
[REVIEW] Fixing RAFT CI & a few small updates for SLHC Python wrapper

cjnolet:


cpp/test/sparse/linkage.cu
```diff
@@ -46,9 +46,110 @@ struct LinkageInputs {
...
+                                    const T* secondClusterArray, uint64_t size,
+                                    uint64_t* a, uint64_t* b) {
+  //calculating the indices of pairs of datapoints compared by the current thread
+  uint64_t j = threadIdx.x + blockIdx.x * blockDim.x;
+  uint64_t i = threadIdx.y + blockIdx.y * blockDim.y;
+
+  //thread-local variables to count a and b
+  uint64_t myA = 0, myB = 0;
+
+  if (i < size && j < size && j < i) {
```
viclafargue, line 68: Isn't it `j <= i` here?

cpp/test/sparse/linkage.cu
```diff
@@ -46,9 +46,110 @@ struct LinkageInputs {
...
+    else if (firstClusterArray[i] != firstClusterArray[j] &&
+             secondClusterArray[i] != secondClusterArray[j]) {
+      ++myB;
+    }
+  }
+
+  //specialize blockReduce for a 2D block of 1024 threads of type uint64_t
+  typedef cub::BlockReduce<uint64_t, BLOCK_DIM_X,
+                           cub::BLOCK_REDUCE_WARP_REDUCTIONS, BLOCK_DIM_Y>
+    BlockReduce;
```
viclafargue, line 85: Maybe an `else if (threadIdx.x != 0 || threadIdx.y != 0) { return; }` here would avoid reducing zeroes when j > i.

viclafargue (approved):
Thanks for working on this fix Corey. This looks good! Just have two little question/comment.

viclafargue:
@rapidsai/ops One of the builds on this PR seemed to have timed out during testing : `Build timed out (after 40 minutes). Marking the build as failed.`. Would it be possible to increase the allocated time?

cpp/test/sparse/linkage.cu
cjnolet: This is literally a copy/paste from the adjusted rand score in cuml and only needed for this gtest at the moment. I admit I didn't exhaustively review the implementation since we should have tests for it in cuml. If you think these improvements are worth pursuing, they should be done directly in cuml.

cjnolet:
@viclafargue thanks for the fast review! I tagged you offline about the timeout. Our tests are very small and none of them should be timing out.

cjnolet:
rerun tests

afender (approved):

dantegd:
@gpucibot merge

cjnolet:
rerun tests

cjnolet:
rerun tests

PR #180, opened 2021-03-24
[REVIEW] Moving labels prims from cuml

cjnolet:
This is needed by the `fix connectivites` primitive that connects an otherwise unconnected knn graph by adding edges that connect components. Unfortunately, the MST primitive is not expected to provide the component colors from a monotonically increasing set so we need the `make_monotonic` function to do so. 

WIthout this, the memory becomes an otherwise sparse data structure represented in dense form and the footprint grows with the number of samples (e.g. with 1M samples, it's possible for the resulting colorings to be 0 and 999999 even if there are only two components). The fix connectivities primitive doesn't even need to revert the mapping either, once the colors are mapped so they are drawn from a monotonically increasing set, the memory footprint will be on the order of the actual number of components, which is expected to shrink significantly even on a single iteration of the MST (just because of the nature of using nearest neighbors, even if each neighbor started as their own component, they would connect across components just by the nature of being neighbors).

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -252,6 +252,8 @@ if(BUILD_RAFT_TESTS)
     test/distance/fused_l2_nn.cu
     test/handle.cpp
     test/integer_utils.cpp
+    test/label/label.cu
+    test/label/merge_labels.cu
     test/lap/lap.cu
     test/linalg/add.cu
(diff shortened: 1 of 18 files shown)


cjnolet:
@tfeher, if you have some time, it would be great to have your review on this. TLDR; this is really just a direct copy from cuml with namespaces renamed.

Issue #179, opened 2021-03-23
[FEA] Move `label/classlabels.cuh` from cuml to RAFT


cjnolet:
The MSF currently returns a colors array with colors that are not always likely to be drawn from a monotonically increasing set.

The algorithm to connect a knn graph currently uses the colors array as-is, but the algorithm would be more efficient and save a lot of memory if the array sizes didn't depend on the max label but instead on the number of unique labels. This can be accomplished easily by re-labeling the colors array so they are drawn from a monotonically increasing set.

divyegala:
This was done in #181

PR #181, opened 2021-03-24
[REVIEW] Scaling knn graph fix connectivities algorithm

cjnolet:
This PR reduces unecessary memory usage while connecting a KNN graph by normalizing colors so they are drawn from a monotonically increasing set.

### DIFF
--- a/cpp/include/raft/label/classlabels.cuh
+++ b/cpp/include/raft/label/classlabels.cuh
@@ -33,7 +33,7 @@ namespace label {
  * The y array is assumed to store class labels. The unique values are selected
  * from this array.
  *
- * \tparam math_t numeric type of the arrays with class labels
+ * \tparam value_t numeric type of the arrays with class labels
  * \param [in] y device array of labels, size [n]
  * \param [in] n number of labels
  * \param [out] y_unique device array of unique labels, unallocated on entry,
@@ -42,12 +42,12 @@ namespace label {
  * \param [in] stream cuda stream
  * \param [in] allocator device allocator
  */
-template <typename math_t>
-void getUniquelabels(math_t *y, size_t n, math_t **y_unique, int *n_unique,
+template <typename value_t>
+void getUniquelabels(value_t *y, size_t n, value_t **y_unique, int *n_unique,
                      cudaStream_t stream,
                      std::shared_ptr<raft::mr::device::allocator> allocator) {
-  raft::mr::device::buffer<math_t> y2(allocator, stream, n);
-  raft::mr::device::buffer<math_t> y3(allocator, stream, n);
+  raft::mr::device::buffer<value_t> y2(allocator, stream, n);
+  raft::mr::device::buffer<value_t> y3(allocator, stream, n);
   raft::mr::device::buffer<int> d_num_selected(allocator, stream, 1);
   size_t bytes = 0;
   size_t bytes2 = 0;
@@ -68,7 +68,8 @@ void getUniquelabels(math_t *y, size_t n, math_t **y_unique, int *n_unique,
   CUDA_CHECK(cudaStreamSynchronize(stream));
 
   // Copy unique classes to output
-  *y_unique = (math_t *)allocator->allocate(*n_unique * sizeof(math_t), stream);
+  *y_unique =
+    (value_t *)allocator->allocate(*n_unique * sizeof(value_t), stream);
   raft::copy(*y_unique, y3.data(), *n_unique, stream);
 }
 
@@ -78,7 +79,7 @@ void getUniquelabels(math_t *y, size_t n, math_t **y_unique, int *n_unique,
  * The output labels will have values +/-1:
  * y_out = (y == y_unique[idx]) ? +1 : -1;
  *
- * The output type currently is set to math_t, but for SVM in principle we are
+ * The output type currently is set to value_t, but for SVM in principle we are
  * free to choose other type for y_out (it should represent +/-1, and it is used
  * in floating point arithmetics).
  *
@@ -90,15 +91,15 @@ void getUniquelabels(math_t *y, size_t n, math_t **y_unique, int *n_unique,
  * \param [in] idx index of unique label that should be labeled as 1
  * \param [in] stream cuda stream
  */
-template <typename math_t>
-void getOvrlabels(math_t *y, int n, math_t *y_unique, int n_classes,
-                  math_t *y_out, int idx, cudaStream_t stream) {
+template <typename value_t>
+void getOvrlabels(value_t *y, int n, value_t *y_unique, int n_classes,
+                  value_t *y_out, int idx, cudaStream_t stream) {
   ASSERT(idx < n_classes,
          "Parameter idx should not be larger than the number "
          "of classes");
   raft::linalg::unaryOp(
     y_out, y, n,
-    [idx, y_unique] __device__(math_t y) {
+    [idx, y_unique] __device__(value_t y) {
       return y == y_unique[idx] ? +1 : -1;
     },
     stream);
--- a/cpp/include/raft/label/classlabels.cuh
+++ b/cpp/include/raft/label/classlabels.cuh
@@ -19,10 +19,10 @@
 #include <cub/cub.cuh>
 
 #include <raft/cudart_utils.h>
-#include <raft/mr/device/allocator.hpp>
-#include <raft/mr/device/buffer.hpp>
 #include <raft/cuda_utils.cuh>
 #include <raft/linalg/unary_op.cuh>
+#include <raft/mr/device/allocator.hpp>
+#include <raft/mr/device/buffer.hpp>
 
 namespace raft {
 namespace label {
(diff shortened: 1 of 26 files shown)


cjnolet:
rerun tests

cjnolet:
rerun tests

cpp/include/raft/label/classlabels.cuh
```diff
@@ -0,0 +1,193 @@
...
+ * \param [in] y device array of labels, size [n]
+ * \param [in] n number of labels
+ * \param [out] y_unique device array of unique labels, unallocated on entry,
+ *   on exit it has size [n_unique]
+ * \param [out] n_unique number of unique labels
+ * \param [in] stream cuda stream
+ * \param [in] allocator device allocator
+ */
+template <typename math_t>
+void getUniquelabels(math_t *y, size_t n, math_t **y_unique, int *n_unique,
```
divyegala, line 46: closes #179, I presume?

cpp/include/raft/label/classlabels.cuh
```diff
@@ -0,0 +1,193 @@
...
+ *
+ * \param [in] y device array if input labels, size [n]
+ * \param [in] n number of labels
+ * \param [in] y_unique device array of unique labels, size [n_classes]
+ * \param [in] n_classes number of unique labels
+ * \param [out] y_out device array of output labels
+ * \param [in] idx index of unique label that should be labeled as 1
+ * \param [in] stream cuda stream
+ */
+template <typename math_t>
```
divyegala, line 93: typename consistency with functions below

cpp/include/raft/label/classlabels.cuh
```diff
@@ -0,0 +1,193 @@
...
+void make_monotonic(Type *out, Type *in, size_t N, cudaStream_t stream,
+                    Lambda filter_op,
+                    std::shared_ptr<raft::mr::device::allocator> allocator,
+                    bool zero_based = false) {
+  static const size_t TPB_X = 256;
+
+  dim3 blocks(raft::ceildiv(N, TPB_X));
+  dim3 threads(TPB_X);
+
+  Type *map_ids;
```
divyegala, line 157: This should be a 0 sized device buffer that is passed by reference and resized later

cpp/include/raft/label/classlabels.cuh
```diff
@@ -0,0 +1,193 @@
...
+   * which items in the array to map.
+   * @param out the output monotonic array
+   * @param in input label array
+   * @param N number of elements in the input array
+   * @param stream cuda stream to use
+   * @param filter_op an optional function for specifying which values
+   * should have monotonically increasing labels applied to them.
+   */
+template <typename Type, typename Lambda>
+void make_monotonic(Type *out, Type *in, size_t N, cudaStream_t stream,
```
divyegala, line 148: `raft::handle` to not have allocator on the signature

cpp/include/raft/label/classlabels.cuh
```diff
@@ -0,0 +1,193 @@
...
+                    Lambda filter_op,
```
divyegala, line 149: default lambda for `filter_op` so that the definition below can go away?

cpp/include/raft/label/merge_labels.cuh
```diff
@@ -0,0 +1,146 @@
...
+ *  E={(A[i], B[i]) | M[i]=1} and make this step faster */
+template <typename Index_, int TPB_X = 256>
+__global__ void __launch_bounds__(TPB_X)
+  propagate_label_kernel(const Index_* __restrict__ labels_a,
+                         const Index_* __restrict__ labels_b,
+                         Index_* __restrict__ R, const bool* __restrict__ mask,
+                         bool* __restrict__ m, Index_ N) {
+  Index_ tid = threadIdx.x + blockIdx.x * TPB_X;
+  if (tid < N) {
+    if (__ldg((char*)mask + tid)) {
```
divyegala, line 41: Can you tell why this is necessary if the pointers are already tagged with `const` and `__restrict__`?

cpp/include/raft/label/merge_labels.cuh
```diff
@@ -0,0 +1,146 @@
...
+__global__ void __launch_bounds__(TPB_X)
+  reassign_label_kernel(Index_* __restrict__ labels_a,
+                        const Index_* __restrict__ labels_b,
+                        const Index_* __restrict__ R, Index_ N,
+                        Index_ MAX_LABEL) {
+  Index_ tid = threadIdx.x + blockIdx.x * TPB_X;
+  if (tid < N) {
+    // Note: labels are from 1 to N
+    Index_ la = labels_a[tid];
+    Index_ lb = __ldg(labels_b + tid);
```
divyegala, line 73: Same question as above

cpp/include/raft/linalg/init.h
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * Fills the output as out[i] = i.
+ *
+ * \param [out] out device array, size [end-start]
+ * \param [in] start of the range
+ * \param [in] end of range (exclusive)
+ * \param [in] stream cuda stream
+ */
+template <typename T>
+void range(T *out, int start, int end, cudaStream_t stream) {
```
divyegala, line 39: this can be replaced entirely with `thrust::sequence`

cpp/include/raft/sparse/op/reduce.cuh
```diff
@@ -0,0 +1,149 @@
...
+
+template <typename value_idx>
+__global__ void compute_duplicates_diffs_kernel(const value_idx *rows,
+                                                const value_idx *cols,
+                                                value_idx *diff, size_t nnz) {
+  size_t tid = blockDim.x * blockIdx.x + threadIdx.x;
+  if (tid >= nnz) return;
+
+  value_idx d = 1;
+  if (tid == 0 || (rows[tid - 1] == rows[tid] && cols[tid - 1] == cols[tid]))
```
divyegala, line 56: small nitpick because the formatting is a little ugly: braces?

divyegala: Also is the safeguard here always valid that the second condition will only be accessed when `tid >= 1`? Theory suggests it should but just confirming

divyegala: Thirdly, why is the first entry marked as duplicate?

cpp/include/raft/sparse/op/sort.h
```diff
@@ -55,37 +67,15 @@ namespace op {
...
-  raft::mr::device::buffer<T> vals_sorted(d_alloc, stream, nnz);
-
-  CUSPARSE_CHECK(raft::sparse::cusparsegthr<T>(
-    handle, nnz, vals, vals_sorted.data(), d_P.data(), stream));
-
-  CUDA_CHECK(cudaStreamSynchronize(stream));
-
-  raft::copy(vals, vals_sorted.data(), nnz, stream);
+  auto coo_indices = thrust::make_zip_iterator(thrust::make_tuple(rows, cols));
+  auto coo_vals = thrust::make_zip_iterator(thrust::make_tuple(vals));
```
divyegala, line 74: Is this necessary?

divyegala (changes requested):
Looks great and mostly small comments/questions! Happy to iterate on this today and get it in before the weekend

cpp/include/raft/label/classlabels.cuh
cjnolet: I'm not sure if I completely agree w/ this one but I'd like to minimize deltas here as much as possible since we're making the move before changing the code in cuml.

cpp/include/raft/label/classlabels.cuh
cjnolet: If we are going to be replacing the allocator w/ the handle, we should be removing the explicit stream as well, as we discussed previously. I don't think we should be changing function signatures in this PR though since we haven't changed over cuml to use these methods (I added a argument w/ a default just so I didn't break cuml when we move it over).

cpp/include/raft/label/merge_labels.cuh
cjnolet: This code is copied from cuml. I didn't write it.

cpp/include/raft/label/classlabels.cuh
cjnolet: Trying to minimize deltas. I needed to copy this stuff over for SLHC. We should move it "formally" in a future PR.

cpp/include/raft/linalg/init.h
cjnolet: Minimizing deltas.

divyegala (approved):
LGTM, since we are trying to get this code copied over from cuML for SLHC to make 0.19. May target cleanup in 0.20. Thanks, Corey!

cpp/include/raft/sparse/op/reduce.cuh
cjnolet: As discussed offline, create #185

dantegd (approved):

dantegd:
@gpucibot merge

PR #189, opened 2021-03-30
Adjust Hellinger pairwise distance to vaoid NaNs

lowener:
This change will fix NaNs that arise in Hellinger pairwise distance when the input to sqrt is negative.
These kind of inputs can happen even when the inputs are normalized row-wise to 1 due to accumulation of rounding approximation.

### DIFF
--- a/cpp/include/raft/sparse/distance/l2_distance.cuh
+++ b/cpp/include/raft/sparse/distance/l2_distance.cuh
@@ -270,7 +270,11 @@ class hellinger_expanded_distances_t : public distances_t<value_t> {
 
     raft::linalg::unaryOp<value_t>(
       out_dists, out_dists, config_->a_nrows * config_->b_nrows,
-      [=] __device__(value_t input) { return sqrt(1 - input); },
+      [=] __device__(value_t input) {
+        // Adjust to replace NaN in sqrt with 0 if input to sqrt is negative
+        value_t rectifier = (1 - input) > 0;
+        return rectifier * sqrt(1 - input);
+      },
       config_->stream);
   }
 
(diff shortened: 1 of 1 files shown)


cpp/include/raft/sparse/distance/l2_distance.cuh
```diff
@@ -270,7 +270,11 @@ class hellinger_expanded_distances_t : public distances_t<value_t> {
 
     raft::linalg::unaryOp<value_t>(
       out_dists, out_dists, config_->a_nrows * config_->b_nrows,
-      [=] __device__(value_t input) { return sqrt(1 - input); },
+      [=] __device__(value_t input) {
+        // Adjust to replace NaN in sqrt with 0 if input to sqrt is negative
+        value_t rectifier = (1 - input) > 0;
```
divyegala, line 275: This should be bool. Just combine this with next line?

divyegala (approved):
Pre-approving since it's a small requested change

cjnolet:
We should just run an L1 norm on the input here to guarantee it's positive.  (that's what UMAP does, which provides this with their API).

cjnolet (changes requested):
These are supposed to be probabilities but we should just run an L1 normalization on the input here to guarantee it's positive. (that's what UMAP does, which provides this with their API).

lowener:
Even when the inputs are L1 normalized this can arise. Running `csr_row_normalize_l1` at the start of this function to enforce it doesn't solve these negative values in sqrt. These negative values are often negative epsilon that comes from either the first sqrt or the inner product.

A solution could be to run csr_row_normalize_l1 and keep this rectifier for negative epsilon, or to add an `abs` function call before sqrt.

cjnolet:
I looked at your failed Pytest asserion in cuml. It looks like Hellinger is running into a similar problem that other expanded metrics had been facing where the diagonal (which should be 0's) has a nan. This is likely because of the subtle numerical instabilities w/ expanded euclidean-based metrics (catastrophic cancellations). 

One thing we can do is zero out the diagonal of the resulting matrix using the `initializeDiagonalMatrix` primitive.

cjnolet (approved):
LGTM

cjnolet:
@gpucibot merge

PR #204, opened 2021-04-12
[REVIEW] Adding assert for MST edge count in SLHC

cjnolet:
We've seen a few failure in cuML's SLHC pytests where the dendrogram didn't have the correct number of nodes even though the edges are connected. This is an indication that either there might not be the correct number of edges in the MST or there might be duplicate edges. Unfortunately, I've been unable to reproduce the pytest failure locally so I've added an assertion to verify the number of expected edges in the MST to, hopefully, give us an indicator of what's going wrong.

### DIFF
--- a/cpp/include/raft/sparse/hierarchy/detail/mst.cuh
+++ b/cpp/include/raft/sparse/hierarchy/detail/mst.cuh
@@ -201,6 +201,12 @@ void build_sorted_mst(const raft::handle_t &handle, const value_t *X,
                " or increase 'max_iter'",
                max_iter);
 
+  RAFT_EXPECTS(mst_coo.n_edges == m - 1,
+               "n_edges should be %d but was %d. This"
+               "could be an indication of duplicate edges returned from the"
+               "MST or symmetrization stage.",
+               m - 1, mst_coo.n_edges);
+
   sort_coo_by_data(mst_coo.src.data(), mst_coo.dst.data(),
                    mst_coo.weights.data(), mst_coo.n_edges, stream);
 


cpp/include/raft/sparse/hierarchy/detail/mst.cuh
```diff
@@ -201,6 +201,12 @@ void build_sorted_mst(const raft::handle_t &handle, const value_t *X,
                " or increase 'max_iter'",
                max_iter);
 
+  RAFT_EXPECTS(mst_coo.n_edges == m - 1,
```
afender, line 204: I think you meant `2m-2 `

afender (approved):
If that helps we have similar checks in cugraph C++
https://github.com/rapidsai/cugraph/blob/36371e98ba06800cfb3b341d68a38c06114086d5/cpp/tests/tree/mst_test.cu#L139

As well as a full output comparison against Nx 
https://github.com/rapidsai/cugraph/blob/36371e98ba06800cfb3b341d68a38c06114086d5/python/cugraph/tests/test_minimum_spanning_tree.py#L64

cpp/include/raft/sparse/hierarchy/detail/mst.cuh
divyegala: It's a de-symmetrized output, or supposed to be at least

cjnolet: In the case of SLHC, we don't symmetrize the output. Divye had added an argument to turn off symmetrized output in 0.19.

afender: Based on the wording I think I should clarify that symmetrize was not a post-processing step. I haven't looked at the code I don't think one can just skip safely skip this as you may run into propagation issues

divyegala: The propagation part isn't affected. `symmetrize=false` only adds the lower colored src->dst pair to the output that is copied as a COO, but that part is not used in the color propagation code

cjnolet: @afender, thanks for the quick review on this one! My apologies for the delayed reply. This week, I'm hoping to do some validation on this in order to root cause https://github.com/rapidsai/cuml/issues/3741. Hopefully I'll get some more clues as to what's going on.

PR #205, opened 2021-04-12
Include missing headers in host allocator

wphicks:
Include `cudart_utils.h` in `mr/host/allocator.hpp` in order to access `CUDA_CHECK` and `CUDA_CHECK_NO_THROW` macros
Include `cstddef` where `std::size_t` is referenced

### DIFF
--- a/cpp/include/raft/mr/device/allocator.hpp
+++ b/cpp/include/raft/mr/device/allocator.hpp
@@ -16,6 +16,7 @@
 
 #pragma once
 
+#include <cstddef>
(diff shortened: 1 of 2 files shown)


dantegd (approved):
lgtm

wphicks:
@dantegd Could I trouble you for a quick re-review? It looks like we were also missing `<cstddef>` to pick up `std::size_t` in both of raft's allocation headers, so I added that here as well.

dantegd (approved):

dantegd:
@gpucibot merge

PR #207, opened 2021-04-19
[WIP] Sparse semiring cleanup + hash table and batching strategies

cjnolet:
This branch includes several new features and optimizations:

1. Introduces a hash table strategy to sparsify the vector in the coo spmv shared memory
2. Adds cucollections to build (necessary for item 1)
2. Adds a batching strategy for rows with nnz too large to fit into shared memory
3. Removes the need for the cusparse csrgemm
5. Uses raft handle in `distances_config_t` rather than accepting each resource explicitly
6. Removes the naive CSR semiring code

This PR is also required to merge #261, which introduces the remaining distances


cpp/include/raft/sparse/distance/l2_distance.cuh
```diff
@@ -108,6 +148,59 @@ void compute_l2(value_t *out, const value_idx *Q_coo_rows,
...
+
+  compute_row_sum_kernel<<<raft::ceildiv(Q_nnz, tpb), tpb, 0, stream>>>(
+    Q_norms.data(), Q_coo_rows, Q_data, Q_nnz);
+  compute_row_sum_kernel<<<raft::ceildiv(R_nnz, tpb), tpb, 0, stream>>>(
+    R_norms.data(), R_coo_rows, R_data, R_nnz);
+
+  value_t n_cols_div = 1.0 / n_cols;
+  raft::linalg::unaryOp<value_t>(
+    Q_sq_norms.data(), Q_sq_norms.data(), m,
+    [=] __device__(value_t input) { return input * n_cols_div; }, stream);
```
divyegala, line 189: Extract repeating lambdas into one lambda. Also, these kernels could be combined, right? Using `thrust::zip_iterator` and `thrust::transform`, only one kernel launch will be needed to do these 4 things

divyegala (commented):
Initial review looks good to me (excepting strategies part which I will clean up soon)! Only one comment and one question - the tests for the new distances remain to be added, right?

cpp/include/raft/sparse/distance/l2_distance.cuh
cjnolet: I'm not sure I understand what you are saying here. I'm using our unary op kernel to perform an element-wise operation. I  generally prefer this over thrust::transforms just because I like the layer of indirection in a consistent API (for all I care, we could just call thrust underneath). However, I'm not sure I agree with your statement about a single kernel launch to compute row-wise norms and combine them with a dot product (e.g. I'm not sure I'd want a single kernel to do all of that). I might be misunderstanding what you mean, though.

cjnolet: Ah, my mistake. The changelog w/ your comment wasn't showing me all the unary op calls below. I don't mind combining those.

cjnolet:
@dantegd, I've updated this PR for branch 21.06 but I'm not familiar with the new package management. I'm getting the following build error, though I do see cuco in the build directory. Is there something we need to add to the gtest build?

```
In file included from /home/cjnolet/workspace/raft/cpp/include/raft/sparse/distance/coo_spmv.cuh:20,
                 from /home/cjnolet/workspace/raft/cpp/test/sparse/dist_coo_spmv.cu:28:
/home/cjnolet/workspace/raft/cpp/include/raft/sparse/distance/coo_spmv_strategies/hash_strategy.cuh:21:10: fatal error: cuco/static_map.cuh: No such file or directory
   21 | #include <cuco/static_map.cuh>
      |          ^~~~~~~~~~~~~~~~~~~~~
```

cjnolet:
Just marking this here so I don't forget- It looks like the current cuco API is slightly different than it was when the `hash_table` in this PR was written. I see two differences so far that are causing compile errors:
1. The `slot_type` was removed from the `static_map::device_mutable_view` and it's unclear to me which one should be used in its place. I tried using from `pair_atomic_type` from `static_map`, which seems to remove this compile error but I'm still not sure if it's the correct one to use
2. `make_from_uninitialized_slots` seems to have been removed from `static_map` altogether and it's not clear to me what should be used in its place.
3. A seed argument was added to a constructor and I'm assuming this might need to be incorporated into cucollections. 

These are the obvious errors I see at the top of the trace. I'm hoping fixing these might fix everything else? I'm not sure yet, but I think maybe we should push this change to 21.08.

cjnolet:
Closing this as it's replaced by #269

PR #210, opened 2021-04-21
`debug_utils.cuh` in RAFT

divyegala:
This PR introduces `debug_utils.cuh` in RAFT, which aims to provide primitives for standard testing/debugging. The first primitive to be provided by this PR is to check whether `nans` exist in a given device vector


cjnolet (changes requested):
Looks good. Just one little request. 

I also have just one additional (minor) nitpick: since this is such a trivial utility function, it might look a cleaner to remove the `isnan_functor` and convert it into a lambda function.

PR #211, opened 2021-04-22
Support for thrust 1.12

ChuckHastings:
Tried building cugraph with thrust 1.12.  It appears that  thrust 1.12 headers changed and we need to explicitly include thrust/host_vector.h in order to use thrust::host_vector.


dantegd:
@gpucibot merge

PR #213, opened 2021-04-26
[REVIEW] Random Ball Cover Algorithm for 2D Haversine/Euclidean

cjnolet:
This PR is a proof of concept to use the triangle inequality to prune the tree of   exhaustive distance computations into something smaller, such as on the order of  where c is called an expansion constant, based on the dimensionality. 

This should (hopefully) be able to benefit both sparse and dense k-nearest neighbors and all algorithms that use them, hopefully providing a significant speedup for our sparse semirings primitive when only the k-nearest neighbors are desired. 

The goal here is to construct a tree out of the random ball cover algorithm such that we can utilize it in algorithms which would otherwise be able to make efficient use of a ball tree. However, there are additional challenges to this algorithm on the GPU, such as being able to batch the tree lookups.


cjnolet:
Tentative plan for the random ball cover PoC:

- [ ] 1. Build PoC for Haversine w/ n_cols hardcoded to 2 and verify perf 
- [ ] 2. If 1 is successful, abstract n_cols to any size
- [ ] 3. If 2 is successful, attempt to improve performance further with a tree-based structure
- [ ] 4. If 2 is successful, attempt to speed up the sparse knn

ajschmidt8:

cjnolet:
Wanted to provide a link to the paper by Laurence Cayton that describes this algorithm in detail: https://www.lcayton.com/rbc.pdf.

cjnolet:
I've found this algorithm also has much less of a dependence on the intrinsic dimensionality than I had originally thought. It's dependent upon the actual distance metric being used and they prove the bounds of the algorithm with an expansion constant, c, which is 2^d for Euclidean. Random Ball Cover's asymptotic bound is c^{3/2} while cover tree is c^12. There's a huge gap there- since the algorithm's boundary is `sqrt(n) * c^{3/2}`. It's not quite log(n) as some other algorithms but what it loses in the log(n) branching of a binary tree it makes up for in its parallelism ability.

cjnolet:
I'm establishing a C++ API w/ an index class called `BallCoverIndex` and the following functions:
```c++
//  1. rbc_build_index() - populate a BallCoverIndex
//  2. rbc_knn_query() - given a populated index, perform query against different query array
//  3. rbc_all_knn_query() - populate a BallCoverIndex and query against training data
//  4. rbc_eps_neigh() - given a populated index, perform query against different query array
//  5. rbc_all_eps_neigh() - populate a BallCoverIndex and query against training data
```

cjnolet:
This still has some minor cleanup coming but is ready for early reviews when someone has a moment.

cjnolet:
@teju85 @dantegd this PR still has some minor cleanup coming but it's ready for early reviews

cjnolet:
Wanted to also post a great paper that experimented w/ RBC in higher dimensions and shows that the additional pruning conditions introduced in this PR lower the number of distances significantly in higher dimensions. It's generally expected that returns will diminish with extremely high k (100+) and as the number of intrinsic dimensions approach 1k.

Something to note is that over 40% of the distances were pruned on MNIST (784 dims... likely less intrinsic dims) with small(-ish) k (<10) and over 20% of the distances were pruned for moderate k (like 75). The effect of k is exponential in the number of dimensions. 

https://arxiv.org/pdf/1801.05055.pdf

divyegala (commented):
I did not review any of the faiss parts because that's outside my understanding. Rest of the code looks great, very minor suggestions. Thanks!

ChuckHastings (approved):
LGTM

cjnolet:
rerun tests

cpp/include/raft/spatial/knn/detail/ball_cover/registers.cuh
```diff
@@ -0,0 +1,514 @@
...
+__global__ void perform_post_filter_registers(
+  const value_t *X, value_int n_cols, const value_idx *R_knn_inds,
+  const value_t *R_knn_dists, const value_t *R_radius, const value_t *landmarks,
+  int n_landmarks, int bitset_size, int k, distance_func dfunc,
+  uint32_t *output, float weight = 1.0) {
+  // allocate array of size n_landmarks / 32 ints
+  extern __shared__ uint32_t smem[];
+
+  // Start with all bits on
+  for (int i = threadIdx.x; i < bitset_size; i += tpb) smem[i] = 0xffffffff;
```
wphicks, line 71:
```suggestion
  for (int i = threadIdx.x; i < bitset_size; i += tpb){ smem[i] = 0xffffffff; }
```

cpp/include/raft/spatial/knn/detail/ball_cover/registers.cuh
```diff
@@ -0,0 +1,514 @@
...
+                         heap.warpKTop * cur_candidate_dist) /
+                          heap.warpKTopRDist;
+        z = isnan(z) ? 0.0 : z;
+        // If lower bound on distance could possibly be in
+        // the closest k neighbors, compute it and add to k-select
+        value_t dist = std::numeric_limits<value_t>::max();
+        if (z <= heap.warpKTop) {
+          const value_t *y_ptr = X_index + (n_cols * cur_candidate_ind);
+          value_t local_y_ptr[col_q];
+          for (int j = 0; j < n_cols; j++) local_y_ptr[j] = y_ptr[j];
```
wphicks, line 203:
```suggestion
          for (int j = 0; j < n_cols; ++j){ local_y_ptr[j] = y_ptr[j]; }
```

cpp/include/raft/spatial/knn/detail/ball_cover/registers.cuh
```diff
@@ -0,0 +1,514 @@
...
+                         abs(heap.warpKTopRDist - cur_candidate_dist) -
+                       heap.warpKTop * cur_candidate_dist) /
+                        heap.warpKTopRDist;
+
+      z = isnan(z) ? 0.0 : z;
+      value_t dist = std::numeric_limits<value_t>::max();
+      if (i < k || z <= heap.warpKTop) {
+        const value_t *y_ptr = X_index + (n_cols * cur_candidate_ind);
+        value_t local_y_ptr[col_q];
+        for (int j = 0; j < n_cols; j++) local_y_ptr[j] = y_ptr[j];
```
wphicks, line 353:
```suggestion
        for (int j = 0; j < n_cols; ++j){ local_y_ptr[j] = y_ptr[j]; }
```

cpp/include/raft/spatial/knn/detail/ball_cover/registers.cuh
```diff
@@ -0,0 +1,514 @@
...
+
+      heap.add(dist, cur_candidate_dist, cur_candidate_ind);
+    }
+
+    if (i < R_size) {
+      value_idx cur_candidate_ind = R_1nn_cols[R_start_offset + i];
+      value_t cur_candidate_dist = R_1nn_dists[R_start_offset + i];
+      value_t z = heap.warpKTopRDist == 0.0
+                    ? 0.0
+                    : (abs(heap.warpKTop - heap.warpKTopRDist) *
```
wphicks, line 366: Include `<stdlib.h>` or (better yet) use `std::abs` from `<cmath>`.

cpp/include/raft/spatial/knn/detail/ball_cover/registers.cuh
```diff
@@ -0,0 +1,514 @@
...
+                         abs(heap.warpKTopRDist - cur_candidate_dist) -
+                       heap.warpKTop * cur_candidate_dist) /
+                        heap.warpKTopRDist;
+
+      z = isnan(z) ? 0.0 : z;
+      value_t dist = std::numeric_limits<value_t>::max();
+      if (i < k || z <= heap.warpKTop) {
+        const value_t *y_ptr = X_index + (n_cols * cur_candidate_ind);
+        value_t local_y_ptr[col_q];
+        for (int j = 0; j < n_cols; j++) local_y_ptr[j] = y_ptr[j];
```
wphicks, line 376: I won't keep pointing it out, but in general I think it would be good to put braces around single-line `for` and  `if` clauses to reduce the likelihood of future developers accidentally pushing code that should be in the clause out of it. Also, don't forget to use pre-increment unless we're making use of the output value.

cpp/include/raft/spatial/knn/detail/warp_select_faiss.cuh
```diff
@@ -0,0 +1,739 @@
...
+
+    // kb is always second in the list, so we needn't use our lane
+    // in this comparison
+    if (FullMerge) {
+      bool swapb = Dir ? Comp::lt(kb, otherKa) : Comp::gt(kb, otherKa);
+      assign(swapb, kb, otherKa);
+      assign(swapb, vb.key, otherVa.key);
+      assign(swapb, vb.value, otherVa.value);
+
+    } else {
```
wphicks, line 485: Any reason to have this empty else clause here as opposed to just putting a comment below?

cpp/test/spatial/spatial_data.h
```diff
@@ -0,0 +1,1674 @@
+#include <vector>
+
+namespace raft {
+namespace spatial {
+std::vector<float> spatial_data = {
```
wphicks, line 5: Could you add a comment just saying where these data come from, whether they're randomly generated or brought from some other source?

wphicks (commented):
Looking good! A few items in-line.

cpp/include/raft/spatial/knn/ball_cover_common.h
```diff
@@ -0,0 +1,97 @@
...
+  bool is_index_trained() const { return index_trained; };
+
+  // This should only be set by internal functions
+  void set_index_trained() { index_trained = true; }
+
+  const raft::handle_t &handle;
+
+  const value_int m;
+  const value_int n;
+  const int n_landmarks;
```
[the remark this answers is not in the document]
cjnolet: I do prefer to stick with a 32-bit type where appropriate so I avoid using a `size_t` unless it's a type I fear will eventually need to be promoted. `n_landmarks` is almost always guaranteed to be smalle enough that we shouldn't need 64-bit for it. In this case, I could use a `uint32_t`. I had originally been using a value less than 1 to denote that we should be setting to `sqrt(n)` and anything 1 or greater to allow the user to set explicitly. I had removed that in this PR but would like to revisit in the higher dims case, since that makes the algorithm more similar to the FAISS IVF `nlists` parameter.

cpp/include/raft/spatial/knn/detail/ball_cover.cuh
```diff
@@ -0,0 +1,342 @@
...
+void k_closest_landmarks(const raft::handle_t &handle,
+                         BallCoverIndex<value_idx, value_t> &index,
+                         const value_t *query_pts, value_int n_query_pts, int k,
+                         value_idx *R_knn_inds, value_t *R_knn_dists) {
+  std::vector<value_t *> input = {index.get_R()};
+  std::vector<int> sizes = {index.n_landmarks};
+
+  brute_force_knn_impl<int, int64_t>(
+    input, sizes, index.n, const_cast<value_t *>(query_pts), n_query_pts,
+    R_knn_inds, R_knn_dists, k, handle.get_stream(), nullptr, 0, (bool)true,
```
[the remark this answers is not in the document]
cjnolet: IIRC, this is a silly artifact from type templating which is meant to avoid compiler confusion about `nvbool`. It's pretty harmless. I was trying to keep from having to explicitly assign the value to a variable. I can try other ways to solve this if you know any.

cpp/include/raft/spatial/knn/detail/ball_cover/registers.cuh
cjnolet: This is in a kernel so this is actually using the cuda fastmath intinsics (which are used implicitly within the cuda runtime API).

cpp/include/raft/spatial/knn/detail/warp_select_faiss.cuh
cjnolet: This code is actually taken directly from FAISS and modified. I tried to limit the modifications so that we can more easily track them and update if necessary. I'm also hoping the high dims PR might be able to remove this altogether.

cpp/include/raft/spatial/knn/detail/ball_cover/registers.cuh
cjnolet: > Also, don't forget to use pre-increment unless we're making use of the output value.

Is this necessary in a for loop?

cpp/include/raft/spatial/knn/detail/warp_select_faiss.cuh
wphicks: Great! That's more than sufficient reason to leave it as is.

cpp/include/raft/spatial/knn/detail/ball_cover/registers.cuh
cjnolet: Ah, I guess it can save the need to explicitly store the value for one iteration in memory.

wphicks: "Necessary" is a strong word :stuck_out_tongue:. Godbolt reports that gcc compiles them to the same assembly, but in general, I think it's preferable to be explicit about when you do and don't want the temporary that would implicitly get created with post-increment without optimization. If I'm using post-increment, it's a signal to future developers that yes, I really did mean to use that return value and if I didn't make use of it, it's a mistake.

cpp/include/raft/spatial/knn/detail/ball_cover/registers.cuh
wphicks: D'oh! Quite right...

cpp/include/raft/spatial/knn/detail/ball_cover.cuh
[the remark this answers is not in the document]
wphicks: Oh! Gotcha. In that case, I'd just do a `static_cast` and call it a day.

cpp/include/raft/spatial/knn/ball_cover_common.h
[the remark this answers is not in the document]
wphicks: Yeah, I'd say `std::uint32_t` is a fine choice here. In general, using `std::optional` is probably preferable to using negative values for special-case signaling if this gets revisited down the line.

wphicks (approved):
All looks great! I'm pre-approving, but I noticed that [cpp/test/spatial/ball_cover.cu](https://github.com/rapidsai/raft/pull/213/files/59c74885511c109d08f533855f6461d7183ee555..2850c790b996d7f4d2cd9ec6e2e3b891753ead38#diff-9a48cd603f16967da47b16e198cf11f2e1c1dc7f866627b3a53ec73cd6aef912), is the only file where we still have fixed width types without a corresponding include. Can we get a quick `#include <cstdint>` and `s/uint32_t/std::uint32_t/g`?

dantegd (approved):

cjnolet:
rerun tests

cjnolet:
@gpucibot merge

PR #215, opened 2021-05-01
Fix Nan on Hellinger

lowener:
This push will fix NaNs that might appear on Hellinger.
The result of the operation 0 * NaN was Nan, so this fix will correct it by having zero in the sqrt.


cjnolet (approved):
LGTM

cjnolet:
@gpucibot merge

PR #223, opened 2021-05-10
Loss of Precision in MST weight alteration

divyegala:


cpp/include/raft/sparse/mst/detail/mst_solver_inl.cuh
```diff
@@ -116,7 +122,9 @@ MST_solver<vertex_t, edge_t, weight_t>::solve() {
   timer0 = duration_us(stop - start);
 #endif
 
-  Graph_COO<vertex_t, edge_t, weight_t> mst_result(2 * v - 2, stream);
+  auto n_expected_edges = symmetrize_output ? 2 * v - 2 : v - 1;
```
afender, line 125: This is not a valid expected number of edges for the spanning forest case.

cpp/include/raft/sparse/mst/detail/mst_solver_inl.cuh
```diff
@@ -152,7 +160,12 @@ MST_solver<vertex_t, edge_t, weight_t>::solve() {
     timer3 += duration_us(stop - start);
 #endif
 
-    if (prev_mst_edge_count[0] == mst_edge_count[0]) {
+    auto curr_mst_edge_count = mst_edge_count[0];
+    RAFT_EXPECTS(curr_mst_edge_count <= n_expected_edges,
```
afender, line 164: This drops minimum spanning forest support

cpp/include/raft/sparse/hierarchy/detail/mst.cuh
```diff
@@ -116,7 +116,7 @@ void connect_knn_graph(const raft::handle_t &handle, const value_t *X,
 
   // On the second call, we hand the MST the original colors
   // and the new set of edges and let it restart the optimization process
-  auto new_mst = raft::mst::mst<value_idx, value_idx, value_t>(
+  auto new_mst = raft::mst::mst<value_idx, value_idx, value_t, double>(
```
afender, line 119: Using 64b is only helpful if `max_gap` is large compared to `max_weight`. Always using 64b may unnecessarily slow down the execution.

afender (changes requested):
Dynamically selecting 64b only when needed would be a good improvement to consider

cpp/include/raft/sparse/mst/detail/mst_solver_inl.cuh
afender: It should be seen as an upper bound

cpp/include/raft/sparse/hierarchy/detail/mst.cuh
divyegala: This is specifically to enable HDBSCAN (and remove errors in SLHC). While dynamic weight selection may help, MST is far from being the bottleneck for HDBSCAN, and we are trying to make HDBSCAN for this release so it's the path of least resistance for now

afender (approved):

afender:
Thanks @divyegala !

divyegala:
rerun tests

cjnolet (approved):
LGTM

cjnolet:
@gpucibot merge

PR #224, opened 2021-05-10
[REVIEW] error.hpp contains unnecessary extra headers

ChuckHastings:
Got a bunch of warnings compiling cugraph.  This lead to the observation that error.hpp contains 4 headers that don't appear to be necessary.

This PR removes those extraneous includes.


dantegd:
@gpucibot merge

PR #230, opened 2021-05-18
Make sure we use CalVer when checking out rapids-cmake

robertmaynard:
Make sure to use CalVer when we clone `rapids-cmake`


dantegd (approved):

dantegd:
@gpucibot merge

dantegd:
@gpucibot merge

divyegala (approved):
LGTM

PR #248, opened 2021-06-02
cuco git tag update (again)

seunghwak:
The previous version unintentionally disallowed floating point value types and cuCollection fixed this in the most recent version. Update git tag accordingly.


BradReesWork (approved):

BradReesWork:
@gpucibot merge

PR #258, opened 2021-06-08
Update get_rmm.cmake to better support CalVer

harrism:
Updates raft's get_rmm.cmake logic to work like cuML and cuGraph, so that it works better with CalVer. This helps when building RAPIDS libraries locally and depending on a local RMM (e.g. with rapids-compose).

PR #259, opened 2021-06-08
Pass stream to device_scalar::value() calls.

harrism:
In comms/test.hpp `device_scalar::value` was not being passed an explicit stream, which means that the default stream was being synced. rapidsai/rmm#789 will remove the default from this parameter, and would have therefore broken the RAFT build. So this PR fixes the oversynchronization and ensures RAFT will build after the RMM PR is merged.

Note this PR includes the cmake changes from #258 (just so I could build locally). Once #258 is merged this PR's changes will be simplified.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,11 +1,11 @@
 #cpp code owners
-cpp/               @divyegala @rapidsai/cuml-cpp-codeowners @rapidsai/cugraph-cpp-codeowners
+cpp/               @rapidsai/cuml-cpp-codeowners @rapidsai/cugraph-cpp-codeowners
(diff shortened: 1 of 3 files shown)


dantegd (approved):
Changes look good to me, same comment about merging 21.08 so CI passes

harrism:
> Changes look good to me, same comment about merging 21.08 so CI passes

Done.

ChuckHastings (approved):

harrism:
@gpucibot merge

harrism:
I don't think I can merge.

ajschmidt8:
@gpucibot merge

PR #260, opened 2021-06-08
Pin `dask` versions in CI

ajschmidt8:
Similarly to https://github.com/rapidsai/cuml/pull/3937/files, this PR pins the version of `dask` and `distributed` that's used in CI.

### DIFF
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -59,8 +59,8 @@ gpuci_conda_retry install -c conda-forge -c rapidsai -c rapidsai-nightly -c nvid
 # Install the master version of dask, distributed, and dask-ml
 gpuci_logger "Install the master version of dask and distributed"
 set -x
-pip install "git+https://github.com/dask/distributed.git" --upgrade --no-deps
-pip install "git+https://github.com/dask/dask.git" --upgrade --no-deps
+pip install "git+https://github.com/dask/distributed.git@2021.05.1" --upgrade --no-deps
+pip install "git+https://github.com/dask/dask.git@2021.05.1" --upgrade --no-deps
 set +x
 
 
--- a/ci/local/old-gpubuild.sh
+++ b/ci/local/old-gpubuild.sh
@@ -80,10 +80,10 @@ if [ "$RUN_CUGRAPH_LIBCUGRAPH_TESTS" = "ON" ] || [ "$RUN_CUGRAPH_PYTHON_TESTS" =
 fi
 
 # Install the master version of dask, distributed, and dask-ml
-gpuci_logger "pip install git+https://github.com/dask/distributed.git --upgrade --no-deps"
-pip install "git+https://github.com/dask/distributed.git" --upgrade --no-deps
-gpuci_logger "pip install git+https://github.com/dask/dask.git --upgrade --no-deps"
-pip install "git+https://github.com/dask/dask.git" --upgrade --no-deps
+set -x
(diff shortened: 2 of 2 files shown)


dantegd (approved):

ci/gpu/build.sh
```diff
@@ -59,8 +59,8 @@ gpuci_conda_retry install -c conda-forge -c rapidsai -c rapidsai-nightly -c nvid
 # Install the master version of dask, distributed, and dask-ml
 gpuci_logger "Install the master version of dask and distributed"
 set -x
-pip install "git+https://github.com/dask/distributed.git" --upgrade --no-deps
-pip install "git+https://github.com/dask/dask.git" --upgrade --no-deps
+pip install "git+https://github.com/dask/distributed.git@2021.05.1" --upgrade --no-deps
+pip install "git+https://github.com/dask/dask.git@2021.05.1" --upgrade --no-deps
```
pentschev, line 63:
```suggestion
```

pentschev: We shouldn't need to install again, it's already installed from conda.

ci/local/old-gpubuild.sh
```diff
@@ -80,10 +80,10 @@ if [ "$RUN_CUGRAPH_LIBCUGRAPH_TESTS" = "ON" ] || [ "$RUN_CUGRAPH_PYTHON_TESTS" =
...
 
 # Install the master version of dask, distributed, and dask-ml
-gpuci_logger "pip install git+https://github.com/dask/distributed.git --upgrade --no-deps"
-pip install "git+https://github.com/dask/distributed.git" --upgrade --no-deps
-gpuci_logger "pip install git+https://github.com/dask/dask.git --upgrade --no-deps"
-pip install "git+https://github.com/dask/dask.git" --upgrade --no-deps
+set -x
+pip install "git+https://github.com/dask/distributed.git@2021.05.1" --upgrade --no-deps
+pip install "git+https://github.com/dask/dask.git@2021.05.1" --upgrade --no-deps
+set +x
```
pentschev, line 86:
```suggestion
```

pentschev (changes requested):
Picking from conda should be enough.

ci/gpu/build.sh
jakirkham: The reason it is being `pip install`ed is to test against `main`

ci/local/old-gpubuild.sh
jakirkham: Same comment as above

pentschev (approved):
Unblocking (or not, because I don't have owner permissions in RAFT), and leaving my previous comments as suggestions. Pinning or removing `pip install` should have the same effect.

ci/gpu/build.sh
pentschev: Yes, but we can't test against `main` in branch-21.06 anymore, only branch-21.08 should do so.

jakirkham: Fair enough. Either should be fine

The primary thing is making sure we don't forget to add this back to `branch-21.08` (with `main` being used)

pentschev: Yeah, we should follow-up with a PR immediately after this lands, as we did for other libraries yesterday.

jakirkham (approved):

dillon-cullinan (approved):

raydouglass (approved):

PR #261, opened 2021-06-08
[REVIEW] Remaining sparse semiring distances

cjnolet:
This PR is intended to be merged after #207 (hash table strategy) has been merged.

This PR introduces the following distances:
- Hamming
- Jensen-Shannon
- Russell-Rao
- KL-Divergence
- Correlation

Most of the changes here are from #207 and will be reviewed in that PR. The only files that need to be reviewed for this PR are `sparse/distance/l2_distance.cuh`, `sparse/distance/bin_distance.cuh`, `sparse/distance/lp_distances.cuh`, and their corresponding gtests: `test/sparse/distance.cuh`


cpp/cmake/thirdparty/get_cuco.cmake
```diff
@@ -19,8 +19,8 @@ function(find_and_configure_cuco VERSION)
     rapids_cpm_find(cuco ${VERSION}
       GLOBAL_TARGETS cuco cuco::cuco
       CPM_ARGS
-        GIT_REPOSITORY https://github.com/NVIDIA/cuCollections.git
-        GIT_TAG        8fce363b40254f4170d97eaedcb12fc8e84e1b74
+        GIT_REPOSITORY https://github.com/divyegala/cuCollections.git
+        GIT_TAG        view-initialization-ctor
```
cjnolet, line 23: This should get updated when #269 is merged

cpp/include/raft/sparse/distance/l2_distance.cuh
```diff
@@ -116,6 +145,55 @@ void compute_l2(value_t *out, const value_idx *Q_coo_rows,
...
+                  const value_idx *R_coo_rows, const value_t *R_data,
+                  value_idx R_nnz, value_idx m, value_idx n, value_idx n_cols,
+                  std::shared_ptr<raft::mr::device::allocator> alloc,
+                  cudaStream_t stream) {
+  // sum_sq for std dev
+  raft::mr::device::buffer<value_t> Q_sq_norms(alloc, stream, m);
+  raft::mr::device::buffer<value_t> R_sq_norms(alloc, stream, n);
+
+  // sum for mean
+  raft::mr::device::buffer<value_t> Q_norms(alloc, stream, m);
```
dantegd, line 170: If it wouldn't be too many changes, could you use rmm uvectors in general instead? (just adding a single comment to avoid repeated ones)

dantegd (changes requested):
Just had one quick comment/question about using RMM directly, the code for the distances looks good!

cpp/include/raft/sparse/distance/l2_distance.cuh
cjnolet: Yeah I don't think that would be too much to do for this PR.

cjnolet:
rerun tests

cjnolet:
@dantegd, I replaced the usage of the raft device buffer w/ the rmm device uvector everywhere. This is ready to be merged.

dantegd (approved):

dantegd:
@gpucibot merge

PR #263, opened 2021-06-08
 Use nested include in destination of install headers to avoid docker permission issues

dantegd:


robertmaynard:
This will stop raft.hpp from being installed as it isn't in the raft subdirectory

robertmaynard (approved):

divyegala (approved):
LGTM

dantegd:
@gpucibot merge

PR #265, opened 2021-06-08
[REVIEW] Move FAISS ANN wrappers to RAFT

cjnolet:


cpp/include/raft/spatial/knn/detail/ivf_pq_ann.cuh
```diff
@@ -0,0 +1,209 @@
+/*
+ * Copyright (c) 2021, NVIDIA CORPORATION.
```
viclafargue, line 2: Why is this file named `ivf_pq_ann.cuh`? Wouldn't something like `ann.cuh` or `ann_implem.cuh` be better description since it implements `IVF`, `IVFPQ` and `IVFSQ`?

viclafargue (approved):
Thanks for working on this. I'm sure the `cuGraph` team or others might be interested in using it! Just one comment/question. Otherwise, it looks good to me!

cpp/include/raft/spatial/knn/detail/ivf_pq_ann.cuh
cjnolet: Great point! I wanted to future-proof our API a little bit because we might be exposing other (potentially non-FAISS) implementations (such as graph, space partitioning or hash-based) in the future in addition to the quantization methods we're exposing here. Maybe we should rename the file to something like `ann_quantized.cuh`? I'm also inclined to suffix the file with `_faiss` (same for the brute force impl file) just so we can be more clear about the dependencies here. In the worst case, that should allow us to more easily separate out the header files in the future so those requiring the FAISS dependency become more clear to the callers. For now I think we can keep the header files the way they are since all of our dense nearest neighbors API requires FAISS. What do you think?

viclafargue: Understood. Yes, both should do it. However, if we call this `ann_faiss.cuh`, other files would probably have to be called `ann_<lib>.cuh` which might be a bit confusing as to what the specific methods in these files are. That's why it may be preferable to call it `ann_<method>.cuh`. But again both are fine to me.

cjnolet: My preference here would be to call it `ann_quantized_faiss.cuh`, actually, so it sounds like we're on the same page.

dantegd (approved):
lgtm except it's missing an update for the header name change:

```
In file included from ../test/sparse/knn.cu:22:
../include/raft/sparse/selection/knn.cuh:36:10: fatal error: raft/spatial/knn/detail/brute_force_knn.cuh: No such file or directory
   36 | #include <raft/spatial/knn/detail/brute_force_knn.cuh>
      |     
```

BradReesWork (approved):

cjnolet:
@gpucibot merge

PR #269, opened 2021-06-09
Sparse semirings cleanup + hash table & batching strategies

divyegala:
This branch includes several new features and optimizations:

1. Introduces a hash table strategy to sparsify the vector in the coo spmv shared memory
2. Adds a batching strategy for rows with nnz too large to fit into shared memory
3. Removes the need for the cusparse csrgemm
4. Uses raft handle in distances_config_t rather than accepting each resource explicitly
5. Removes the naive CSR semiring code

This PR is also required to merge #261, which introduces the remaining distances


cpp/cmake/thirdparty/get_cuco.cmake
```diff
@@ -19,8 +19,8 @@ function(find_and_configure_cuco VERSION)
     rapids_cpm_find(cuco ${VERSION}
       GLOBAL_TARGETS cuco cuco::cuco
       CPM_ARGS
-        GIT_REPOSITORY https://github.com/NVIDIA/cuCollections.git
-        GIT_TAG        0b672bbde7c85a79df4d7ca5f82e15e5b4a57700
+        GIT_REPOSITORY https://github.com/divyegala/cuCollections.git
+        GIT_TAG        view-initialization-ctor
```
cjnolet, line 23: Reminder to revert this

cpp/include/raft/cudart_utils.h
```diff
@@ -267,12 +267,12 @@ void allocate(Type*& ptr, size_t len, bool setZero = false) {
 }
 
 /** helper method to get max usable shared mem per block parameter */
-inline int getSharedMemPerBlock() {
+__host__ __device__ inline int getSharedMemPerBlock() {
   int devId;
-  CUDA_CHECK(cudaGetDevice(&devId));
+  cudaGetDevice(&devId);
```
cjnolet, line 272: Any reason we can't query this on host and pass it in? That would allow us to keep the checks here and not need to query it for each individual block.

cpp/include/raft/sparse/distance/lp_distance.cuh
```diff
@@ -42,51 +41,35 @@ namespace distance {
...
-
-  } else {
-    // TODO: Find max nnz and set smem based on this value.
-    // Ref: https://github.com/rapidsai/cuml/issues/3371
-    generalized_csr_pairwise_semiring<value_idx, value_t>(
-      out_dists, *config_, product_func, accum_func);
-  }
+   *
+   *  - if n_cols < available smem, just use dense conversion for rows of A
+   *  - if n_cols > available smem but max nnz < available smem, use hashing
```
cjnolet, line 50: Do we need to update this?

cpp/include/raft/sparse/distance/utils.cuh
```diff
@@ -0,0 +1,114 @@
...
+    value_idx count = __syncthreads_count(predicate);
+
+    if (threadIdx.x == 0) atomicAdd(out_count, count);
+  }
+
+  if (threadIdx.x == 0) atomicMax(out, agg);
+}
+
+template <typename value_idx, bool count_greater_than>
+inline std::pair<value_idx, value_idx> max_degree(
```
cjnolet, line 64: Can we do a brief doxygen for this?

cpp/test/CMakeLists.txt
```diff
@@ -53,7 +53,7 @@ add_executable(test_raft
     test/matrix/matrix.cu
     test/mr/device/buffer.cpp
     test/mr/host/buffer.cpp
-    test/mst.cu
+    # test/mst.cu
```
cjnolet, line 56: Should this and the knn test below be re-enabled?

cpp/test/sparse/dist_coo_spmv.cu
```diff
@@ -171,48 +202,56 @@ class SparseDistanceCOOSPMVTest
...
+    CUDA_CHECK(cudaFree(out_dists_hash));
+    CUDA_CHECK(cudaFree(out_dists_chunk));
     CUDA_CHECK(cudaFree(out_dists_ref));
   }
 
   void compare() {
     raft::print_device_vector("expected: ", out_dists_ref,
                               params.out_dists_ref_h.size(), std::cout);
-    raft::print_device_vector("out_dists: ", out_dists,
+    raft::print_device_vector("out_dists_chunk: ", out_dists_chunk,
```
cjnolet, line 231: If we must keep the prints, we should probably only do them when the results don't match (I admit I'm guilty of keeping prints around in tests).

cpp/test/sparse/distance.cu
```diff
@@ -105,21 +79,20 @@ class SparseDistanceTest
...
     int out_size = dist_config.a_nrows * dist_config.b_nrows;
 
     allocate(out_dists, out_size);
 
     pairwiseDistance(out_dists, dist_config, params.metric, params.metric_arg);
 
-    CUDA_CHECK(cudaStreamSynchronize(stream));
+    CUDA_CHECK(cudaStreamSynchronize(handle.get_stream()));
+
+    raft::print_device_vector("dists", out_dists, out_size, std::cout);
```
cjnolet, line 91: We should probably remove this as well.

cpp/test/sparse/knn.cu
```diff
@@ -109,25 +76,22 @@ class SparseKNNTest
...
-      nnz, n_rows, params.n_cols, out_indices, out_dists, k, cusparseHandle,
-      alloc, stream, params.batch_size_index, params.batch_size_query,
-      params.metric);
+      nnz, n_rows, params.n_cols, out_indices, out_dists, k, handle,
+      params.batch_size_index, params.batch_size_query, params.metric);
+
+    CUDA_CHECK(cudaStreamSynchronize(handle.get_stream()));
 
-    CUDA_CHECK(cudaStreamSynchronize(stream));
+    printf("Done.");
```
cjnolet, line 84: Debug print

cpp/test/sparse/dist_coo_spmv.cu
```diff
@@ -64,29 +64,51 @@ class SparseDistanceCOOSPMVTest
   : public ::testing::TestWithParam<
       SparseDistanceCOOSPMVInputs<value_idx, value_t>> {
  public:
+  SparseDistanceCOOSPMVTest() : dist_config(handle) {}
```
cjnolet, line 67: Are we running each test on a couple different chunk sizes?

cjnolet (changes requested):
Overall, the hash and chunking strategies look great. Mostly minor things. I'm ready to finally get this in!

cpp/include/raft/cudart_utils.h
divyegala: Ya, I didn't make this change either. If you remember, I had asked you a couple of weeks ago about this as well. Happy to revert it to just a host function

cpp/include/raft/sparse/distance/utils.cuh
divyegala: I don't even think I'm using this as a pair of returned values anymore. I'll revert it to minimal function and add doxygen

cpp/test/sparse/dist_coo_spmv.cu
divyegala: Hmmm yeah, we should do that. I'll add a test for chunk size of 3

cpp/include/raft/cudart_utils.h
cjnolet: Interesting, yeah I don't recall making this change either. I suppose it's possible I could have made it early on when I was trying something out, but I had tried reverting tthe change at one point and it broke something that I didn't believe I had written. No biggie either way, but it would probably be best to just revert this if we can.

cpp/test/sparse/distance.cu
cjnolet, line 91: We should either remove this print or use a conditional to print only when the test fails.

cjnolet (changes requested):
Almost good to go. Just one more trivial thing.

cpp/test/sparse/distance.cu
divyegala: I think removing seems better, I'll do it

cjnolet (approved):
LGTM

cjnolet:
@gpucibot merge

PR #273, opened 2021-06-11
Add lds and sts inline ptx instructions to force vector instruction generation

mdoijade:
Adds inline ptx assembly for lds & sts instructions for float, float2, float4, double, double2.
This ensures that compiler doesn't mistakenly generate non-vectorized instructions whenever we need it to generate vectorized version.
Also this ensures that we always generate non-generic ld/st instructions eliminating compiler from generating generic ld/st instructions.
These functions now requires the given shmem pointer should be aligned by the vector length, like for float4 lds/sts shmem pointer should be aligned by 16 bytes else it might silently fail or can also give runtime error.


teju85:
@mdoijade did you get a chance to check whether this change causes any perf regressions, especially in the distance prims? (asking because AFAIK, they are only ones who are using these methods in their kernels)

teju85:
rerun tests

teju85 (approved):
Changes LGTM

mdoijade:
@teju85 can we merge this PR now? btw the above failure in "gpuCI/raft/gpu/cuda/11.0/python/3.7/ubuntu16.04" is CI docker installation issue.

teju85:
@dantegd can we get some help in getting the CI passed for this PR, please?

teju85:
rerun tests

mdoijade:
@teju85 can we merge this now?

teju85:
@gpucibot merge

PR #275, opened 2021-06-14
expose epsilon parameter to allow precision to to be specified

ChuckHastings:
Per user request, this PR exposes the epsilon value that controls the precision where the Hungarian algorithm determines that a value has been reduced to sufficiently close to 0.

### DIFF
--- a/cpp/include/raft/lap/lap_kernels.cuh
+++ b/cpp/include/raft/lap/lap_kernels.cuh
@@ -53,7 +53,7 @@ bool __device__ near_zero(weight_t w, weight_t epsilon) {
 }
 
 template <>
-bool __device__ near_zero<int64_t>(int64_t w, int64_t epsilon) {
+bool __device__ near_zero<int32_t>(int32_t w, int32_t epsilon) {
   return (w == 0);
 }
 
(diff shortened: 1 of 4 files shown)


cjnolet (approved):
LGTM

cpp/include/raft/lap/lap_kernels.cuh
```diff
@@ -48,20 +48,18 @@ const int AUGMENT{4};
...
+bool __device__ near_zero(weight_t w, weight_t epsilon) {
+  return ((w > -epsilon) && (w < epsilon));
 }
 
 template <>
-bool __device__ near_zero<float>(float w) {
-  const float epsilon{1e-6};
-  return ((w > -epsilon) && (w < epsilon));
+bool __device__ near_zero<int32_t>(int32_t w, int32_t epsilon) {
+  return (w == 0);
```
seunghwak, line 57: Why aren't we using epsilon here?

seunghwak (approved):
I have one question, but besides the question, looks good to me.

cpp/include/raft/lap/lap_kernels.cuh
ChuckHastings: For integer weights we can do an exact comparison to zero.  For floating point values we need to do an approximate comparison.

cjnolet:
@gpucibot merge

PR #277, opened 2021-06-16
Add a default for epsilon in LAP constructor

LeviBarnes:
I encountered a fail when building cugraph against this latest version of raft. This change adds a default for the new epsilon parameter.

### DIFF
--- a/cpp/include/raft/lap/lap.cuh
+++ b/cpp/include/raft/lap/lap.cuh
@@ -60,7 +60,7 @@ class LinearAssignmentProblem {
 
  public:
   LinearAssignmentProblem(raft::handle_t const &handle, vertex_t size,
-                          vertex_t batchsize, weight_t epsilon)
+                          vertex_t batchsize, weight_t epsilon=1e-6)
     : handle_(handle),
       size_(size),
       batchsize_(batchsize),


cpp/include/raft/lap/lap.cuh
```diff
@@ -60,7 +60,7 @@ class LinearAssignmentProblem {
 
  public:
   LinearAssignmentProblem(raft::handle_t const &handle, vertex_t size,
-                          vertex_t batchsize, weight_t epsilon)
+                          vertex_t batchsize, weight_t epsilon=1e-6)
```
ChuckHastings, line 63: Note... this will provide a warning if weight_t is not float/double.  Cugraph builds a version where the weight_t is n int32_t.

There is a PR in the cugraph pipeline (https://github.com/rapidsai/cugraph/pull/1673) that will modify cugraph to specify this parameter in all cases.  Hopefully that will make this PR unnecessary.

ChuckHastings:
@LeviBarnes - Suggesting closing this.  The linked cugraph PR has been merged, which should resolve the issue that triggered this problem.

The solution in this PR won't work for non-floating point types.

PR #278, opened 2021-06-22
Use the 21.08 branch of rapids-cmake as rmm requires it

robertmaynard:
Now that `rmm` uses `rapids-cmake` we need to update to the `21.08` branch to get the new `rapids_cmake_write_version_file` function

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -19,7 +19,7 @@ include(FetchContent)
 FetchContent_Declare(
   rapids-cmake
   GIT_REPOSITORY https://github.com/rapidsai/rapids-cmake.git
-  GIT_TAG        origin/branch-21.06
+  GIT_TAG        origin/branch-21.08
   )
 FetchContent_MakeAvailable(rapids-cmake)
 include(rapids-cmake)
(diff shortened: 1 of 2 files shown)


dantegd:
Before RMM's migration to rapids-cmake RAFT's CMake was just relying on RMM's for all of thrust's build needs, but now this is running into 

```
CMake Error at /opt/conda/envs/rapids/lib/cmake/rmm/rmm-config.cmake:122 (thrust_create_target):
  Unknown CMake command "thrust_create_target".
```

Does that mean that we need a get_thrust in raft (like [cuML's](https://github.com/rapidsai/cuml/blob/branch-21.08/cpp/cmake/thirdparty/get_thrust.cmake))?

robertmaynard:
> Before RMM's migration to rapids-cmake RAFT's CMake was just relying on RMM's for all of thrust's build needs, but now this is running into
> 
> ```
> CMake Error at /opt/conda/envs/rapids/lib/cmake/rmm/rmm-config.cmake:122 (thrust_create_target):
>   Unknown CMake command "thrust_create_target".
> ```
> 
> Does that mean that we need a get_thrust in raft (like [cuML's](https://github.com/rapidsai/cuml/blob/branch-21.08/cpp/cmake/thirdparty/get_thrust.cmake))?

Yes that is the easiest way to make sure we pull a version of Thrust of the sufficient version that has that CMake function. I have updated this PR to do this.

dantegd (approved):

dantegd:
@gpucibot merge

PR #287, opened 2021-07-07
[FIX] Always add faiss library alias if it's missing

trxcllnt:
Always add the `FAISS::FAISS` library target alias if it doesn't exist. This can happen if cuML is built and installs FAISS before cuGraph or vice-versa.

Related PRs:
https://github.com/rapidsai/cuml/pull/4028
https://github.com/rapidsai/cugraph/pull/1694

Note: We can probably remove the `get_faiss.cmake` file in cuML and cuGraph since they both should get it from RAFT.


cjnolet (approved):
LGTM

trxcllnt:
rerun tests

trxcllnt:
rerun tests

dantegd (approved):

dantegd:
@gpucibot merge

PR #288, opened 2021-07-07
Temporarily revert #276

wphicks:
This reverts commit 82061e00200ee0f53c2242cbbdd4d8cd636ed310. PR 276 appears to have resulted in a substantial increase in compilation time, causing timeouts in CI for cuML PR 4029. These changes should be reinstated once the cause of the increased compilation time has been determined. Note that this also temporarily installs an old version of `clang-format` in order to unblock changes which will allow us to support version 11.


cjnolet (approved):
LGTM

teju85:
IIRC, @mdoijade was trying to reduce compilation time (due to adding these metrics) in https://github.com/rapidsai/cuml/pull/3990?

dantegd (changes requested):
Leaving a request changes review to avoid accidental merge, this PR should not be necessary after cuml/3990

PR #291, opened 2021-07-12
One cudaStream_t instance per raft::handle_t

### LINKED ISSUE #115: [FEA] upgrade CUDA stream pool
We should consider leveraging [RMM CUDA stream pool](https://github.com/rapidsai/rmm/pull/659) to replace the [custom implementation in RAFT ](https://github.com/rapidsai/raft/blob/4593c21455d26434ed06a36c4b962bb88cee2522/cpp/include/raft/handle.hpp#L130)

An important aspect that I think should be reconciled regardless is the API to acquire a stream from the pool.

### LINKED ISSUE #293: [FEA] Update `raft::handle_t` to be coupled with one `cudaStream_t` instance
The idea is to create a "light" handle that does not do any resource initialization except on request. The changes to the handle will be as follows:

1. Copy and move constructors/assignment operators are now deleted, since it's unsafe to be sure about safely copying or moving the number of resources held by the handle
2. The only constructor will now be `raft::handle_t(rmm::cuda_stream_view = {}, const rmm::cuda_stream_pool& = {0}`. This means both the main stream and stream pool are now not owned by the handle, except for NULL stream and 0 sized pool
3. The above ensures that the handle is always created with a light copy of a pre-existing stream and stream pool. It also ensures that no stream or stream pool will be deleted by mistake in the scope of the handle. This was not confirmed by the previous `handle.set_stream()` method
4. The idea behind `rmm::cuda_stream_pool` being created by the developer and not the handle allows for the safe creation and deletion of streams (which are a light resource) by the scope which they are bound. This will enable several programming patterns:

Previous pattern No. 1:
```
void foo() {
  // work on main stream
  // code block requiring multiple streams
  // continue on main stream
}
```

New pattern No. 1:
```
// refactor code-block with multiple streams
void multiple_stream_foo() {
    // create a stream pool with as many streams as needed at this point
    rmm::cuda_stream_pool stream_pool(n_streams);
    // code block requiring multiple streams
    // pool deleted and synced at end if needed
}

void foo() {
    // work on main stream
    multiple_stream_foo();
    // continue work on main stream
}
```

Previous pattern No. 2:
```
raft::handle_t h(5);
for (int i = 0; i < 5; i++)
    // confusing extra stream parameter on prim
    prim(h, ..., h.get_internal_stream(i));
```

New pattern No. 2:
```
rmm::cuda_stream_pool stream_pool(5);
raft::handle_t heavy(rmm::cuda_stream_default, stream_pool)
for (int i = 0; i < 5; i++)
    // create light handle with stream from pool
   // similar to thrust API for example, each API has a main stream
    prim(raft::handle_t(heavy.get_stream_pool().get_stream()), ...);
```
5. New APIs called `raft::handle_t::get_stream()` and `raft::handle_t::get_stream_pool()`
6. 3 new APIs for stream sync (previous ones were not syncing main stream correctly): 
        `raft::handle_t::sync_stream()` -> syncs main stream
        `raft::handle_t::sync_stream_pool()` -> syncs stream pool if size > 0
        `raft::handle_t::wait_stream_pool_on_stream()` -> make stream pool wait on main stream to finish its latest task
7. Remove `raft::handle_t::set_stream()` method since a new light raft handle can be created instead, without needing to update resources of previous handle. Also removed equivalent method from cython
8. CUDA math libraries to use `setStream*` function to correctly set their handles on main stream. Previously, this was not being done
9. Related updates to cython/python API
10. New cython `Handle` constructor can accept `stream=None` and `n_streams=0` in its constructor

divyegala:
closes #293 
closes #115 

This PR also updates the cython build to `std=c++17`.


cpp/include/raft/handle.hpp
```diff
@@ -148,40 +143,33 @@ class handle_t {
...
-      int_streams_vec.push_back(get_internal_stream(i));
-    }
-    return int_streams_vec;
-  }
+  /**
+   * @brief synchronize main stream on the handle
+   */
+  void sync_stream() const { stream_view_.synchronize(); }
 
-  void wait_on_user_stream() const {
```
divyegala, line 174: To specifically point why I removed this method, it does nothing if `internal_pools` is a vector of size 0

divyegala:
Depends on https://github.com/rapidsai/rmm/pull/818

cpp/include/raft/handle.hpp
```diff
@@ -121,6 +113,7 @@ class handle_t {
     std::lock_guard<std::mutex> _(mutex_);
     if (!cublas_initialized_) {
       CUBLAS_CHECK(cublasCreate(&cublas_handle_));
+      CUBLAS_CHECK(cublasSetStream(cublas_handle, stream_view));
```
ChuckHastings, line 116: `stream_view_` ?

divyegala: Oh, typo! Thanks!

ChuckHastings (approved):

cpp/include/raft/handle.hpp
```diff
@@ -52,56 +53,47 @@ class handle_t {
...
 
-  void set_stream(cudaStream_t stream) { user_stream_ = stream; }
-  cudaStream_t get_stream() const { return user_stream_; }
-  rmm::cuda_stream_view get_stream_view() const {
-    return rmm::cuda_stream_view(user_stream_);
-  }
+  /**
+   * @brief returns main stream on the handle
+   */
+  const rmm::cuda_stream_view& get_stream() const { return stream_view_; }
```
seunghwak, line 91: Isn't this redundant to 
```
  rmm::cuda_stream_view get_stream_view() const {
    return rmm::cuda_stream_view(user_stream_);
  }
```
https://github.com/rapidsai/raft/blob/branch-21.08/cpp/include/raft/handle.hpp#L102

cpp/include/raft/handle.hpp
```diff
@@ -148,40 +143,33 @@ class handle_t {
...
+  void sync_stream() const { stream_view_.synchronize(); }
 
-  void wait_on_user_stream() const {
-    CUDA_CHECK(cudaEventRecord(event_, user_stream_));
-    for (int i = 0; i < get_num_internal_streams(); i++) {
-      CUDA_CHECK(cudaStreamWaitEvent(get_internal_stream(i), event_, 0));
+  /**
+   * @brief synchronize the stream pool on the handle
+   */
+  void sync_stream_pool() const {
```
seunghwak, line 160: Can we synchronize with a subset of the streams in the pool?

cpp/include/raft/handle.hpp
```diff
@@ -52,56 +53,47 @@ class handle_t {
...
+  handle_t& operator=(const handle_t&) = delete;
+  handle_t(handle_t&&) = delete;
+  handle_t& operator=(handle_t&&) = delete;
+
   /**
-   * @brief Construct a handle with the specified number of worker streams
+   * @brief Construct a handle with a stream view and stream pool
    *
-   * @param[in] n_streams number worker streams to be created
+   * @param[in] stream default NULL stream
```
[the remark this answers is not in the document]
divyegala: Can I word this better? I mean to say that the default value is NULL

cpp/include/raft/handle.hpp
divyegala: Yes, but now the previous function is redundant. The new one returns a `rmm::cuda_stream_view` by default, but with https://github.com/rapidsai/rmm/pull/818 it is implicitly convertible to `cudaStream_t`, thus improving default behavior without breaking existing code

cpp/include/raft/handle.hpp
divyegala: Sure, what parameter argument could we use here? An `std::vector` of stream indices?

cpp/include/raft/handle.hpp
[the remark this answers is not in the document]
seunghwak: So, what happens if a users calls ```handle_t(non_null_stream)```? I assume this code is valid, and I guess the valid description of this input parameter is something like "the default stream (which has the default value of nullptr if unspecified)" than the current "default NULL stream" (the latter implies ```stream``` should be NULL).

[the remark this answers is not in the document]
divyegala: That indeed does sound more clear, I'll update the description. Thank you!

cpp/include/raft/handle.hpp
seunghwak: Yup. std::vector sounds like a reasonable choice.

cpp/test/handle.cpp
```diff
@@ -24,65 +24,41 @@ namespace raft {
...
   ASSERT_NE(nullptr, h.get_cusolver_dn_handle());
   ASSERT_NE(nullptr, h.get_cusolver_sp_handle());
   ASSERT_NE(nullptr, h.get_cusparse_handle());
 }
 
 TEST(Raft, Handle) {
-  handle_t h(4);
-  ASSERT_EQ(4, h.get_num_internal_streams());
+  // test stream pool creation
+  rmm::cuda_stream_pool stream_pool{4};
```
wphicks, line 38: To avoid a "magic" number in the assertion below, could we declare the pool size as a variable and reuse it as we do with `n_streams` at line 53?

wphicks (approved):

cpp/test/handle.cpp
divyegala: Done, thanks!

divyegala:
@cjnolet whenever you get a chance to review, now that internal streams don't exist as a vector anymore, the brute force KNN detail API had to be updated accordingly. I think it will look cleaner if that API also just accepted the handle instead. If you agree, I'll go ahead and change that API

cpp/test/linalg/cholesky_r1.cu
```diff
@@ -49,35 +47,35 @@ class CholeskyR1Test : public ::testing::Test {
     // Initializing in CUBLAS_FILL_MODE_LOWER, because that has larger workspace
     // requirements.
     raft::linalg::choleskyRank1Update(handle, L.data(), n_rows, n_rows, nullptr,
-                                      &n_bytes, CUBLAS_FILL_MODE_LOWER, stream);
+                                      &n_bytes, CUBLAS_FILL_MODE_LOWER,
+                                      handle.get_stream());
     Lwork = std::max(Lwork * sizeof(math_t), (size_t)n_bytes);
-    workspace.resize(Lwork, stream);
+    workspace.resize(Lwork, handle.get_stream());
```
cjnolet, line 53: Very very nice. I'm glad this one came full-circle. Much cleaner and safer.

python/raft/common/handle.pyx
```diff
@@ -50,51 +51,36 @@ cdef class Handle:
...
-    # python world cannot access to this raw object directly, hence use
-    # 'size_t'!
-    cdef size_t h
-
-    # not using __dict__ unless we need it to keep this Extension as lean as
-    # possible
-    cdef int n_streams
-
-    def __cinit__(self, n_streams=0):
+    def __cinit__(self, stream=None, n_streams=0):
```
cjnolet, line 54: Eventually it would also be nice to either have the Python handle accept other library resources explicitly or to interact directly w/ or wrap the cupy-maintained resources. Currently, we just use the default stream in cupy because there's no sharing/interaction between the resources held by cupy and cuml.

cjnolet: Interested how / if cugraph is also handling this (cc @Iroy30)

python/raft/common/handle.pyx
```diff
@@ -50,51 +51,36 @@ cdef class Handle:
...
-        cdef size_t s = <size_t>stream.getStream()
-        cdef handle_t* h_ = <handle_t*>self.h
-        h_.set_stream(<_Stream>s)
-
-    def sync(self):
+        self.stream_pool.reset(new cuda_stream_pool(n_streams))
+
+        cdef cuda_stream_view c_stream
+        if stream is None:
+            # this constructor will construct a "main" handle on null stream
```
cjnolet, line 60: While I can't imagine too many cases where we require asynchronicity in the Python layer currently, I would think we'd want to follow the same pattern that we're following in C++ and create a stream if it doesn't exist so that we aren't blocking off the entire GPU, right? For example, if someone is using multiple processes to submit jobs to the same GPU, the kernel launches won't be asynchronous wrt each other. Or was this discussed already and I'm just late to the game?

cpp/include/raft/handle.hpp
```diff
@@ -52,56 +53,47 @@ class handle_t {
...
+  handle_t& operator=(const handle_t&) = delete;
+  handle_t(handle_t&&) = delete;
+  handle_t& operator=(handle_t&&) = delete;
+
   /**
-   * @brief Construct a handle with the specified number of worker streams
+   * @brief Construct a handle with a stream view and stream pool
    *
-   * @param[in] n_streams number worker streams to be created
+   * @param[in] stream the default stream (which has the default value of nullptr if unspecified)
```
cjnolet, line 66: The use of the term `default_stream` might be confusing me a bit. We had originally been using the CUDA default (e.g. device synchronous) stream everywhere and had to go through quite a clean up job to replace it with an explicit stream everywhere. Is this the CUDA device-synchronous default stream which is set in the kernels when stream == nullptr or does the stream view somehow make that into an explicit stream? Or has our ruling changed on the topic of using the device synchronous default stream within raft/cuml? 

If it is the case that we plan to allow the device synchronous default stream, we probably should update the cuml docs. Also it's worth making sure that we continue to account for a stream in the code everywhere applicable. Tagging @teju85 just to make sure there weren't any other reasons for that change initially.

cjnolet (changes requested):
Looks great, just mostly a question about the use of `default_stream`

cpp/include/raft/handle.hpp
divyegala: At some point, this came back I believe because `raft::handle_t` was creating a handle with `stream == nullptr` even before updates by this PR

python/raft/common/handle.pyx
divyegala: This is a very, very nice catch. We weren't accounting for this up until now either. This doc should shed better light on this https://docs.nvidia.com/cuda/cuda-runtime-api/stream-sync-behavior.html.

In that case, we shouldn't use https://github.com/rapidsai/rmm/blob/2179ffd5dbc2847c00d2685c010b345f366f2816/include/rmm/cuda_stream_view.hpp#L114 or https://github.com/rapidsai/rmm/blob/2179ffd5dbc2847c00d2685c010b345f366f2816/include/rmm/cuda_stream_view.hpp#L119

We should instead use this as default: https://github.com/rapidsai/rmm/blob/2179ffd5dbc2847c00d2685c010b345f366f2816/include/rmm/cuda_stream_view.hpp#L124

Let me know if this solution looks correct

seunghwak (approved):
Approving the C++ part of this PR.

python/raft/common/handle.pyx
divyegala: Yeah, cupy has a stream setter as well, which acts like a singleton I believe. You can set it with the pointer of the stream that you own. We could set it with the stream used by the handle, I'm not sure how to correctly context manage it though. Maybe it should be context managed by cuML or cuGraph when they get the handle and want to run cupy stuff.

https://docs.cupy.dev/en/stable/reference/generated/cupy.cuda.ExternalStream.html

tfeher (changes requested):
In general the PR looks good to me. I think that the title could be improved: "couple raft::handle_t with a single main cuda stream". 

Removing handle.set_stream() will affect the cuML unit test, but I see that it is already getting addressed in PR https://github.com/rapidsai/cuml/pull/4071. Please also update the developer guide with the modification on stream usage as part of https://github.com/rapidsai/cuml/pull/4071.

cjnolet (approved):
LGTM

divyegala:
Hi all, I have requested a re-review from all involved as these are the changes that I have applied:

1. Merge Victor's changes
2. Account for changes in RMM
3. Enable `per-thread default stream` flag

cpp/include/raft/spatial/knn/detail/knn_brute_force_faiss.cuh
```diff
@@ -261,15 +263,18 @@ void brute_force_knn_impl(std::vector<float *> &input,
...
 
   for (size_t i = 0; i < input.size(); i++) {
     float *out_d_ptr = out_D + (i * k * n);
     IdxType *out_i_ptr = out_I + (i * k * n);
 
-    cudaStream_t stream =
-      raft::select_stream(userStream, internalStreams, n_int_streams, i);
+    auto stream = handle.is_stream_pool_initialized()
+                    ? handle.get_stream_from_stream_pool()
+                    : handle.get_stream();
```
tfeher, line 277: The user wants to select another stream from the pool, or fall back to the user stream. This pattern also appears [here](https://github.com/rapidsai/cuml/blob/067344041b1563b19301e2e69240a56605a67997/cpp/src_prims/linalg/lstsq.cuh#L254-L260), and I think it is a generally useful pattern.

I suggest that we add a small helper to the handle that does this (not sure about the right naming):
```c++
rmm::cuda_stream_view raft::handle::get_next_stream() {
  return handle.is_stream_pool_initialized()
                    ? handle.get_stream_from_stream_pool()
                    : handle.get_stream();
}
```

tfeher (changes requested):
Hi @divyegala, I had another look and in general it looks good. I have one suggestion for an additional helper function for the handle. Please consider whether it would make sense to include that, or if there are any objections against it.

cpp/include/raft/spatial/knn/detail/knn_brute_force_faiss.cuh
divyegala: I like the idea! Maybe `get_any_worker_stream`? cc @cjnolet for ideas

cjnolet: With my limited knowledge of RMM's stream pool API, the better name depend on the expected behavior, IMO. For example, if get_stream_from_stream_pool` is acting like a thread pool and returning any available stream, then I think `get_any_worker_stream` might be the better name because it doesn't imply anything about order. if `get_stream_from_stream_pool` always returns the next stream in the pool, I would be inclined to prefer `get_next_stream()`. 

Though neither of these seem to imply anything about the fact that the default stream on the handle is used if there's no stream is configured, which makes me think that maybe a more concise name could be something like `get_usable_stream()` or `get_next_usable_stream()`

divyegala: `get_stream_from_stream_pool` always returns the next stream in the pool. In that case, I will use `get_next_usable_stream` and clarify the expected behavior in the docs

divyegala:
@tfeher could you re-review?

ajschmidt8 (approved):

divyegala:
@gpucibot merge

PR #296, opened 2021-07-20
[REVIEW] Fix support for different input and output types in linalg::reduce

### LINKED ISSUE #165: [BUG] raft::linalg::reduce won't compile when InType != OutType
**Describe the bug**
I am reducing a boolean matrix to an integer vector (cf code sample in the section below).
I only need `coalescedReduction`, but the `reduce` function instantiates `stridedReduction` too and this is where the problem happens, at [line 150 of strided_reduction.cuh](https://github.com/rapidsai/raft/blob/b055cf862a599fd45537d21a309edd8a6e06da4c/cpp/include/raft/linalg/strided_reduction.cuh#L150): the `if` statement is not determined at compile time (it's not a C++17 `if constexpr`) so it attempts to compile `stridedSummationKernel` even if the input and output types are different, in which case it can't compile.

Note that at the moment I solved the problem for my use case by calling directly `coalescedReduction` instead of `reduce`.

This is approximately what my code looks like:
```cuda
bool* in;
int* out;
raft::linalg::reduce(
  out, in, N, N, 0, true, true, stream, false,
  [] __device__(bool value, int idx) {
    return static_cast<int>(value);
  },
  raft::Sum<int>(), raft::Nop<int>());
```

Compilation error:
```
include/raft/linalg/strided_reduction.cuh(152): error: no instance of function template "raft::linalg::stridedSummationKernel" matches the argument list
argument types are: (int *, const __nv_bool *, int, int, int, lambda [](__nv_bool, int)->int)
detected during: instantiation of "void raft::linalg::stridedReduction(OutType *, const InType *, IdxType, IdxType, OutType, cudaStream_t, __nv_bool, MainLambda, ReduceLambda, FinalLambda) [with InType=__nv_bool, OutType=int, IdxType=int, MainLambda=lambda [](__nv_bool, int)->int, ReduceLambda=raft::Sum<int>, FinalLambda=raft::Nop<int, int>]"
```

**Expected behavior**
The reduction should support different input and output types as suggested by the template parameters and the docs.

Nyrio:
Closes #165 

Uses a C++17 `if constexpr` to discard at compile time a code path that doesn't support different input and output types, and adds a test for such a case (the test won't compile without that `constexpr` keyword).

### DIFF
--- a/cpp/include/raft/linalg/strided_reduction.cuh
+++ b/cpp/include/raft/linalg/strided_reduction.cuh
@@ -147,8 +147,8 @@ void stridedReduction(OutType *dots, const InType *data, IdxType D, IdxType N,
 
   ///@todo: this complication should go away once we have eliminated the need
   /// for atomics in stridedKernel (redesign for this is already underway)
-  if (std::is_same<ReduceLambda, raft::Sum<OutType>>::value &&
-      std::is_same<InType, OutType>::value)
+  if constexpr (std::is_same<ReduceLambda, raft::Sum<OutType>>::value &&
+                std::is_same<InType, OutType>::value)
     stridedSummationKernel<InType>
       <<<nblks, thrds, shmemSize, stream>>>(dots, data, D, N, init, main_op);
   else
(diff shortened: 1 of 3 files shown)


cjnolet (approved):
LGTM

BradReesWork (approved):

BradReesWork:
@gpucibot merge

PR #306, opened 2021-07-30
Add Hamming, Jensen-Shannon, KL-Divergence, Russell rao and Correlation distance metrics support

mdoijade:
This PR introduces the following distances:
- Hamming
- Jensen-Shannon
- Russell-Rao
- KL-Divergence
- Correlation
with unit tests for each of them.

### DIFF
--- a/cpp/include/raft/distance/correlation.cuh
+++ b/cpp/include/raft/distance/correlation.cuh
@@ -1,0 +1,243 @@
+/*
+ * Copyright (c) 2021, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+#include <raft/distance/pairwise_distance_base.cuh>
+#include <raft/linalg/reduce.cuh>
+#include <raft/cuda_utils.cuh>
+
+namespace raft {
+namespace distance {
+
+
+/**
+ * @brief the Correlation distance matrix:
+ *
+ * @tparam DataT          input data-type (for A and B matrices)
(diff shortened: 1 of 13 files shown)


mdoijade:
@cjnolet @teju85 for help with review.

teju85:
@cjnolet Mahesh found that in some metrics there's a huge perf gain if we used the fast-math instrinsics. Thus, we were discussing about introducing "fast" versions of some of these metrics (eg: `jensen-shannon` and `jensen-shannon-fast`) in our list of metrics currently supported. What are your thoughts? BTW, if this sounds good, it'll be addressed in a separate PR.

mdoijade:
@cjnolet ping for help with review.

cjnolet:
@teju85 @mdoijade, 

So the proposal is to have two separate versions, since the faster one is an approximation, and make it optional for the user to select it? That sounds fine to me.

cjnolet (approved):
LGTM

cjnolet:
@teju85, 

I notice this is marked w/ the `breaking` label. Are there any updates that need to be made on the cuML side before this can be merged?

dantegd:
@cjnolet @mdoijade is a corresponding PR in cuML like this [one](https://github.com/rapidsai/cuml/pull/3990/files) needed?

mdoijade:
> 
> 
> @teju85 @mdoijade,
> 
> So the proposal is to have two separate versions, since the faster one is an approximation, and make it optional for the user to select it? That sounds fine to me.

@cjnolet yes that is the proposed plan to have additional distance metrics for jensen-shannon & kl-divergence as `fast` versions. I see a perf difference of almost 2-3x when using `fast log()` for these distance metrics.

mdoijade:
> 
> 
> @cjnolet @mdoijade is a corresponding PR in cuML like this [one](https://github.com/rapidsai/cuml/pull/3990/files) needed?

@dantegd though merging this raft PR shouldn't break the cuML build, I am working on exposing these distance metrics on cuML cpp/python interface. that PR should be ready in a day or two.

BradReesWork (approved):

GPUtester:

dantegd:
add to allowlist

mdoijade:
@teju85 can you rerun the tests? seems like the some CI error causing failure.

teju85:
rerun tests

teju85:
> I notice this is marked w/ the breaking label. Are there any updates that need to be made on the cuML side before this can be merged?

Ah I missed your question, @cjnolet . This was an oversight from me. This is not a breaking change. I've rectified it.

PR #299, opened 2021-07-27
Warnings are errors

harrism:
This PR fixes current RAFT C++/CUDA compilation warnings and turns on -Wall to treat warnings as errors.

Fixes #225
Fixes #289

PR #307, opened 2021-08-03
Enable CUDA device code warnings as errors

harrism:
Adds `-Werror=all-warnings` NVCC flag to ensure all CUDA device code warnings are treated as errors. Only enabled on CUDA 11.2+ because CUDA 11.0 has PTXAS warnings that go away in newer CUDA versions.

Missed this in #299.


dantegd (approved):

dantegd:
@gpucibot merge

PR #315, opened 2021-08-14
Temporarily pin RMM while refactor removes deprecated calls

dantegd:
This will unblock cuGraph's build.


BradReesWork (approved):

dantegd:
@gpucibot merge

PR #320, opened 2021-08-24
Update with rapids cmake new features

robertmaynard:
This combines some general CMake style cleanup and brings new rapids-cmake features to RAFT including:

- Usage of `rapids_cmake_install_lib_dir` to make sure we install raft correctly on non-debain based distro's ( lib64 ), while also handling conda installation requirements ( always lib no matter the distro )
- Usage of `rapids_cpm` pre-configured pacakges
- Removal of early termination before `rapids_cpm_find` since a better solution now exists ( https://github.com/rapidsai/rapids-cmake/issues/49 )

### DIFF
--- a/cpp/cmake/thirdparty/get_rmm.cmake
+++ b/cpp/cmake/thirdparty/get_rmm.cmake
@@ -16,6 +16,12 @@
 
 function(find_and_configure_rmm VERSION)
 
+    # No way to override the version rapids-cmake pulls
+    # so we need to manually checkout when needing a diverged
+    # version
+    # include(${rapids-cmake-dir}/cpm/rmm.cmake)
+    # rapids_cpm_rmm()
+
     if(${VERSION} MATCHES [=[([0-9]+)\.([0-9]+)\.([0-9]+)]=])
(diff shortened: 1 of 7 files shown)


cpp/cmake/thirdparty/get_rmm.cmake
```diff
@@ -16,9 +16,11 @@
...
 function(find_and_configure_rmm VERSION)
 
-    if(TARGET rmm::rmm)
-        return()
-    endif()
+    # No way to override the version rapids-cmake pulls
+    # so we need to manually checkout when needing a diverged
+    # version
+    # include(${rapids-cmake-dir}/cpm/rmm.cmake)
+    # rapids_cpm_rmm()
```
dantegd, line 23: We can remove the pinned rmm now (I'm doing it in #322 as well to run CI correctly )

dantegd (changes requested):
Changes look great, just a comment about the RMM pin

cpp/cmake/thirdparty/get_rmm.cmake
robertmaynard: done

dantegd (approved):

dantegd:
@gpucibot merge

PR #324, opened 2021-09-09
Fused L2 (unexpanded) kNN kernel for NN <= 64, without using temporary gmem to store intermediate distances

mdoijade:
benchmarking with cuml python interface kNN datasets (it claims to generate gaussian distribution) tried till 200k x 128 database/query vectors.
found some different behavior on my small GPU GP107 vs on GA102(Tesla A40)
on GP107 fused L2 kNN is slower on larger datasets
on GA102 fused L2 kNN is always faster like approx **1.15x-1.5x** for all datasets I tried (except 200k x 128).

I will also have L2 expanded version of fused L2 kNN in a separate PR due to which on larger dimension like > 128 distance computation from fused L2 kNN won't become bottleneck.

There is scope to optimize the distance computation in fused L2 kNN as there is no usage of vectorized LDG/STS in it.

Overall it looks that fused L2 kNN is better on GPUs with decent compute power but not on small old GPUs like GP107.

& benchmarking with cuml cpp kNN regression tests the performance is 
On A30 (GA100) , For NN == 64, resultant Dist matrix 1M x 1M,
Fused L2 kNN = 11550ms
FAISS kNN = 23933 ms.
**Overall 2.07x faster**
And for NN == 32, it is **1.43x faster**
runtimes for NN == 32,
Fused L2 kNN = 11198ms
FAISS kNN = 16124ms


mdoijade:
@teju85 @cjnolet @dantegd for help with review.

cjnolet:
@mdoijade I do appreciate your patience as I'm still working through the review for this PR. 

> benchmarking with cuml python interface kNN datasets (it claims to generate gaussian distribution) tried till 200k x 128 database/query vectors.

Are you referring to the `make_blobs` generator here? I guess even though this still scales quadratically, the distribution of the data points can still be relevant to the k-select because it will determine the order in which points get ignored or the k-select needs to sort/merge, right?

mdoijade:
> 
> 
> @mdoijade I do appreciate your patience as I'm still working through the review for this PR.
> 
> > benchmarking with cuml python interface kNN datasets (it claims to generate gaussian distribution) tried till 200k x 128 database/query vectors.
> 
> Are you referring to the `make_blobs` generator here? I guess even though this still scales quadratically, the distribution of the data points can still be relevant to the k-select because it will determine the order in which points get ignored or the k-select needs to sort/merge, right?

right I am referring to `make_blobs`. yes dataset generated by it seems fine to me.

cjnolet (changes requested):
Looks great! Really the only thing I see is that new files should contain only the current year. 

Until now, the gtests for the brute-force knn in RAFT haven't been extensive because the existing prim just a proxy for the FAISS bfknn call. Now that we are doing more of the work in RAFT we probably want to beef up the gtest a little bit, at least enough to feel confident that consumers will see expected results under different inputs and neighborhood sides. We don't need to hold up this PR do that, though. On that same note, the cuml pytests should be verified against this implementation at the very least, if you have not done so already.

cjnolet:
rerun tests

ChuckHastings (approved):

mdoijade:
> Until now, the gtests for the brute-force knn in RAFT haven't been extensive because the existing prim just a proxy for the FAISS bfknn call. Now that we are doing more of the work in RAFT we probably want to beef up the gtest a little bit, at least enough to feel confident that consumers will see expected results under different inputs and neighborhood sides. We don't need to hold up this PR do that, though. On that same note, the cuml pytests should be verified against this implementation at the very least, if you have not done so already.

I mainly verified these changes with cuML cpp & python kNN tests, mainly KNNRegression & KNNClassify cpp tests, and nearest_neighbors.py on pytests side.
I will add a more thorough test case in RAFT as well, as this kernel belongs in RAFT, thanks for pointing this out!
I have triggered cuML CI tests with these raft changes. - https://github.com/rapidsai/cuml/pull/4212

cjnolet:
@gpucibot merge

cjnolet (approved):
LGTM

PR #325, opened 2021-09-09
Fix `matrixVectorOp` to verify promoted pointer type is still aligned to vectorized load boundary

viclafargue:
Fix for https://github.com/rapidsai/cuml/issues/3965
The function did not check for the memory alignment of the pointer provided.

### DIFF
--- a/cpp/include/raft/linalg/matrix_vector_op.cuh
+++ b/cpp/include/raft/linalg/matrix_vector_op.cuh
@@ -73,6 +73,12 @@ void matrixVectorOpImpl(Type *out, const Type *matrix, const Type *vec,
 
 /**
  * @brief Operations for all the columns or rows with a given vector.
+ * Caution : Threads process multiple elements to speed up processing. These
+ * are loaded in a single read thanks to type promotion. Faster processing
+ * would thus only be enabled when adresses are optimally aligned for it.
+ * Note : the function will also check that the size of the window of accesses
+ * is a multiple of the number of elements processed by a thread in order to
+ * enable faster processing
  * @tparam Type the matrix/vector type
  * @tparam Lambda a device function which represents a binary operator
  * @tparam IdxType Integer type used to for addressing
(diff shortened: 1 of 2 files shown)


cjnolet:
@viclafargue, RMM should be aligning allocations by default, which is why this has not been an issue for cuml algorithms. Before RMM, we did manually align allocations, but since RMM does this for us, we shouldn't have to explicitly check this in the algorithms (since the number of places this might need to be done could be exhaustive). 

Hopefully the solution proposed  [here](https://github.com/rapidsai/cuml/issues/3965#issuecomment-916272893)  solves the problem for the user.

viclafargue:
Thanks for finding a solution. That would solve the problem on the Python side. Maybe these should even be done undercover by default when cuML is imported? However, do we have the guarantee that pointers provided to the native API are always memory aligned?

cjnolet:
> However, do we have the guarantee that pointers provided to the native API are always memory aligned?
The assumption here is that users will be using RAPIDS and so we should be able to expect that the RAPIDS memory manager is being used to allocate any pointers that are handed to these prims.

viclafargue:
This PR is not necessary anymore.

cpp/include/raft/linalg/matrix_vector_op.cuh
```diff
@@ -93,17 +93,23 @@ void matrixVectorOp(Type *out, const Type *matrix, const Type *vec, IdxType D,
                     IdxType N, bool rowMajor, bool bcastAlongRows, Lambda op,
                     cudaStream_t stream) {
   IdxType stride = rowMajor ? D : N;
-  size_t bytes = stride * sizeof(Type);
-  if (16 / sizeof(Type) && bytes % 16 == 0) {
+  size_t stride_bytes = stride * sizeof(Type);
```
cjnolet, line 96: I think this PR looks good to merge for 21.10 but we should add a nice comment here which summarizes these low-level details and justifies the need for the additional check on pointer alignment. Before knowing myself that the primitive was promoting the type to optimize the reads, I was under the impression the alignment issue was happening from the inputs. It would also be helpful if we added some additional documentation to the `TxN_t` struct as well. It currently states this: `Obviously, it's caller's responsibility to take care of pointer alignment!` but it would be helpful to add a sentence or two justifying why (the description of the vectorized op itself isn't bad but it still doesn't make particularly clear why the alignment might sometimes be needed).

cjnolet:
I've reopened this pull request after chatting w/ @viclafargue and now that I have a better understanding of how this could be causing the issue in https://github.com/rapidsai/cuml/issues/4199. 

Specifically, the `matrixVectorOp` attempts to make the application of vector elements to the matrix more efficient by loading multiple vector elements into a single CUDA thread. This alone won't cause an issue, but the kernel is further optimizing the reads by loading the vector elements into the read-only L1 cache and, insodoing, might promote the datatype to minimize the number of reads (e.g. when 2x 8byte elements need to be read, it will use 1x 16 byte datatype so that it can perform a single read). It's this promotion of the datatype that assumes the pointer is aligned to the type being promoted. Thus, the pointer alignment needs to be explicitly checked when this promotion occurs. This is an internal assumption, specific to the `matrixVectorOp`.

cjnolet (changes requested):

cjnolet (approved):
LGTM

cjnolet:
@gpucibot merge

PR #326, opened 2021-09-10
[REVIEW] RBC in high dims

cjnolet:
This PR adds higher dim support for RBC.


cpp/include/raft/selection/col_wise_sort.cuh
```diff
@@ -0,0 +1,346 @@
...
+            // @ToDo: Figure out dynamic shared memory for block sort kernel - better for volta and beyond
+            // int currDevice = 0, smemLimit = 0;
+            // CUDA_CHECK(cudaGetDevice(&currDevice));
+            // CUDA_CHECK(cudaDeviceGetAttribute(&smemLimit, cudaDevAttrMaxSharedMemoryPerBlock, currDevice));
+            // size_t maxElementsForBlockSort = smemLimit / perElementSmemUsage;
+
+            // for 48KB smem/block, can fit in 6144 4byte key-value pair
+            // assuming key-value sort for now - smem computation will change for value only sort
+            // dtype being size of key-value pair
+            std::map<size_t, int> dtypeToColumnMap = {{4, 12288},   // short + short
```
wphicks, line 204: It's not immediately obvious to me how often this will be called, but if performance is a concern, we should make this a switch statement (or if we expect the number of cases to grow, a sorted vector) rather than using a `std::map`. For small maps like this, the cache-unfriendly accesses in the underlying RB-tree add up surprisingly quickly. In Treelite, we had a very similar `std::map` usage that didn't seem like it would be a problem, but profiling showed that those map accesses added a shocking amount to runtime.

cpp/include/raft/selection/col_wise_sort.cuh
```diff
@@ -0,0 +1,346 @@
...
+                                                      {12, 6144},   // double + int/float
+                                                      {16, 6144}};  // double + double
+
+            if (dtypeToColumnMap.count(perElementSmemUsage) != 0 &&
+                n_columns <= dtypeToColumnMap[perElementSmemUsage]) {
+                // more elements per thread --> more register pressure
+                // 512(blockSize) * 8 elements per thread = 71 register / thread
+
+                // instantiate some kernel combinations
+                if (n_columns <= 512)
```
wphicks, line 215: I always get a little anxious about omitting braces in single-line ifs like this. It's too easy for another developer to come along later and accidentally introduce a bug by forgetting that the clauses need to be single-line. Can we add braces to these?

cpp/include/raft/selection/col_wise_sort.cuh
```diff
@@ -0,0 +1,346 @@
...
+
+                    // for segment offsets
+                    workspaceSize += raft::alignTo(sizeof(int) * (size_t)numSegments, memAlignWidth);
+                } else {
+                    size_t workspaceOffset = 0;
+
+                    if (!sortedKeys) {
+                        sortedKeys      = reinterpret_cast<InType*>(workspacePtr);
+                        workspaceOffset = raft::alignTo(sizeof(InType) * (size_t)totalElements, memAlignWidth);
+                        workspacePtr    = (void*)((size_t)workspacePtr + workspaceOffset);
```
wphicks, line 266: Any particular reason for a C-style cast as opposed to a `reinterpret_cast` here?

cpp/include/raft/selection/col_wise_sort.cuh
```diff
@@ -0,0 +1,346 @@
...
+                    }
+
+                    OutType* dValuesIn = reinterpret_cast<OutType*>(workspacePtr);
+                    workspaceOffset    = raft::alignTo(sizeof(OutType) * (size_t)n_columns, memAlignWidth);
+                    workspacePtr       = (void*)((size_t)workspacePtr + workspaceOffset);
+
+                    // layout idx
+                    CUDA_CHECK(layoutIdx(dValuesIn, 1, n_columns, stream));
+
+                    for (int i = 0; i < n_rows; i++) {
```
wphicks, line 329: Any reason for `i` to be an `int` as opposed to a `std::size_t`? It looks like `i` is an index, and keeping it as a `std::size_t` can help avoid unexpected signed/unsigned comparison issues.

Also, we should probably default to using `++i` unless we're using the output.

cpp/include/raft/sparse/linalg/degree.cuh
```diff
@@ -43,11 +43,11 @@ namespace linalg {
...
  * @param results array to place results
  */
-template <int TPB_X = 64>
-__global__ void coo_degree_kernel(const int *rows, int nnz, int *results) {
+template <int TPB_X = 64, typename T = int>
+__global__ void coo_degree_kernel(const T *rows, int nnz, T *results) {
   int row = (blockIdx.x * TPB_X) + threadIdx.x;
   if (row < nnz) {
-    raft::myAtomicAdd(results + rows[row], 1);
+    atomicAdd(results + rows[row], (T)1);
```
wphicks, line 50: I won't keep pointing it out, but in general if we have a C-style cast we should probably explicitly justify its usage in comments and avoid it wherever we can.

cpp/include/raft/spatial/knn/ball_cover.hpp
```diff
@@ -0,0 +1,133 @@
...
+#include "detail/ball_cover.cuh"
+#include "detail/ball_cover/common.cuh"
+
+namespace raft {
+namespace spatial {
+namespace knn {
+
+template <typename value_idx = int64_t, typename value_t,
+          typename value_int = int>
+inline void rbc_build_index(const raft::handle_t &handle,
```
wphicks, line 30: Does this need to be `inline`?

cpp/include/raft/spatial/knn/ball_cover.hpp
```diff
@@ -0,0 +1,133 @@
...
+                              detail::EuclideanFunc(), perform_post_filtering,
+                              weight);
+
+    // TODO: Call sqrt on distances
+  }
+
+  index.set_index_trained();
+}
+
+template <typename value_idx = int64_t, typename value_t,
```
wphicks, line 95: I'd give slight preference to `std::int64_t` and `#include <cstdint>` over `int64_t` and `#include <stdint.h>`. Basically the same argument as for `std::size_t`: using a C type should be an indication to future developers that we are interfacing with C code, and C++ types should otherwise be the default.

cpp/include/raft/spatial/knn/detail/ann_quantized_faiss.cuh
```diff
@@ -17,12 +17,14 @@
...
 
 #include "../ann_common.h"
+#include "knn_brute_force_faiss.cuh"
 
 #include "common_faiss.h"
 #include "processing.hpp"
 
 #include <raft/cudart_utils.h>
 #include <raft/cuda_utils.cuh>
+#include "processing.hpp"
```
wphicks, line 27: Double include (line 23)

cpp/test/spatial/ball_cover.cu
```diff
@@ -0,0 +1,199 @@
...
+#include "spatial_data.h"
+
+#include <thrust/transform.h>
+#include <rmm/exec_policy.hpp>
+
+namespace raft {
+namespace spatial {
+namespace knn {
+
+using namespace std;
```
wphicks, line 36: It's probably best to avoid this even in tests. We only do it in a couple other test files, and I'd be in favor of not adding any more.

cpp/test/spatial/spatial_data.h
```diff
@@ -0,0 +1,1674 @@
+#include <vector>
+
+namespace raft {
+namespace spatial {
+std::vector<float> spatial_data = {
```
wphicks, line 5: Could you just add a comment saying where these data came from? If they're randomly generated, that's cool, but it would probably be good for future folks to know.

wphicks (commented):
This looks great! Really great documentation throughout. The things that I saw a few times throughout that I think we might want to generally touch up if possible are:

- Avoiding C-style casts
- Using C++ versions of integer types by default
- Using braces for single-line `if` and `for` clauses

Besides that, there were just a few random things I called out in line.

Really like the implementation of this! There's a lot going on, but the logic is very clear and explicit throughout.

cjnolet:
@wphicks, I apologize for not providing the pull request link directly! Now I feel badly because this PR is still a WIP (and i just realized the target branch was still set to 21.10. Doh!). The RBC PR that we have slated to 21.10 is this one: https://github.com/rapidsai/raft/pull/213.

wphicks:
Haha! Oh no! I'm so sorry. I should have looked more carefully myself. Well... it was a good warmup. I'll go work on the real thing now. :)

PR #328, opened 2021-09-14
Add broadcast with const input iterator

seunghwak:
The current ```bcast``` function takes a single ```value_t*``` pointer (MPI style) for both input (if root) and output (if non-root).

This does not compile if we have ```const value_t*``` pointer for input. This PR adds a ```bcast``` function that takes separate ```const value_t*``` input and ```value_t```` output pointers (NCCL style).


cjnolet (approved):
LGTM.

cjnolet:
@gpucibot merge

PR #329, opened 2021-09-17
Accounting for rmm::cuda_stream_pool not having a constructor for 0 streams

divyegala:
Comes from this PR https://github.com/rapidsai/rmm/pull/873

### DIFF
--- a/cpp/test/handle.cpp
+++ b/cpp/test/handle.cpp
@@ -24,7 +24,6 @@ namespace raft {
 
 TEST(Raft, HandleDefault) {
   handle_t h;
-  ASSERT_EQ(0, h.get_num_internal_streams());
   ASSERT_EQ(0, h.get_device());
   ASSERT_EQ(nullptr, h.get_stream());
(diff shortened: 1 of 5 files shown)


ChuckHastings (approved):

dantegd (approved):

cpp/test/handle.cpp
```diff
@@ -64,18 +62,6 @@ TEST(Raft, GetHandleFromPool) {
   ASSERT_EQ(parent.get_device(), child.get_device());
 }
 
-TEST(Raft, GetHandleFromPoolPerf) {
```
divyegala, line 67: I had to remove this test because for some reason, the destructor for `parent` was being called while in the for loop. Since this is just a temporary fix, and my handle refactor will take care of the copy constructor for `raft::handle_t`, we should be fine

ChuckHastings (approved):

dantegd (approved):

cjnolet:
@gpucibot merge

PR #339, opened 2021-09-27
[REVIEW] Add Fused L2 Expanded KNN kernel

mdoijade:
-- adds fused L2 expanded kNN kernel, this is faster by at least 20-25% on higher dimensions (D >= 128) than L2 unexpanded version.
-- also on smaller dimension (D <=32) L2 expanded is always faster by 10-15%
 -- slight improvement in updateSortedWarpQ device function by reducing redundant instruction.
-- Fix incorrect output for NN >32 case when taking prod-cons knn merge path, this was caught in HDBSCAN pytest.


ci/prtest.config
```diff
@@ -1,6 +1,6 @@
 RUN_CUGRAPH_LIBCUGRAPH_TESTS=OFF
 RUN_CUGRAPH_PYTHON_TESTS=OFF
 
-RUN_CUML_LIBCUML_TESTS=OFF
-RUN_CUML_PRIMS_TESTS=OFF
-RUN_CUML_PYTHON_TESTS=OFF
+RUN_CUML_LIBCUML_TESTS=ON
+RUN_CUML_PRIMS_TESTS=ON
+RUN_CUML_PYTHON_TESTS=ON
```
ajschmidt8, line 6: The `ops` team is tagged as a `codeowner` for this file, but I'm not sure what the preferred value is for these variables. I'll defer to @dantegd for this change since it seems he's touched this file last.

mdoijade:
@cjnolet @teju85 @dantegd  I think this PR is now in review state, please help with review whenever possible.

ci/prtest.config
cjnolet: I don't believe this file is even being used currently but we should probably revert these changes so we're not introducing unneeded changes into this PR

cpp/include/raft/spatial/knn/detail/fused_l2_knn.cuh
```diff
@@ -300,6 +327,16 @@ __global__ __launch_bounds__(Policy::Nthreads, 2) void fusedL2kNN(
     [numOfNN, sqrt, m, n, ldd, out_dists, out_inds] __device__(
       AccT acc[Policy::AccRowsPerTh][Policy::AccColsPerTh], DataT * regxn,
       DataT * regyn, IdxT gridStrideX, IdxT gridStrideY) {
+      if (useNorms) {
+#pragma unroll
```
cjnolet, line 331: Aside from the template instantiations for each distance, is it possible that the unrolling of these loops could also be having a drastic increase on the compile time?

cpp/include/raft/spatial/knn/detail/knn_brute_force_faiss.cuh
```diff
@@ -271,44 +272,93 @@ void brute_force_knn_impl(std::vector<float *> &input,
...
-        args.vectorsRowMajor = rowMajorIndex;
-        args.numVectors = sizes[i];
-        args.queries = search_items;
-        args.queriesRowMajor = rowMajorQuery;
-        args.numQueries = n;
-        args.outDistances = out_d_ptr;
-        args.outIndices = out_i_ptr;
-
-        /**
+    if (k <= 64 && rowMajorQuery == rowMajorIndex && rowMajorQuery == true &&
```
cjnolet, line 275: The conditionals here are getting quite involved. It would be easier to read and maintain this if we moved everything inside this initial `if` block into a separate function and invoked that directly.

cpp/include/raft/spatial/knn/knn.hpp
```diff
@@ -116,7 +116,7 @@ inline void brute_force_knn(
   std::vector<int> &sizes, int D, float *search_items, int n, int64_t *res_I,
   float *res_D, int k, bool rowMajorIndex = true, bool rowMajorQuery = true,
   std::vector<int64_t> *translations = nullptr,
-  distance::DistanceType metric = distance::DistanceType::L2Unexpanded,
+  distance::DistanceType metric = distance::DistanceType::L2Expanded,
```
cjnolet, line 119: This here is a potentially breaking change to libraries downstream. I know the expanded form can have quite a signifciant speedup from the exhausted unexpanded form but we'll want to at least verify this doesn't break cuml or euler before changing this. Another option could be to get this PR merged in sooner without this change and then make the change in a different PR once the downstream libraries have been tested.

cpp/test/spatial/ball_cover.cu
```diff
@@ -80,40 +78,6 @@ uint32_t count_discrepancies(value_idx *actual_idx, value_idx *expected_idx,
   return result;
 }
 
-template <typename value_t>
```
cjnolet, line 83: Thanks for reverting this! Though we no longer need to invoke the fused knn directly, I do still benefit to keeping the additional helper function so it's more simple to change the bfknn call across all the gtests in the future.

cpp/test/spatial/ball_cover.cu
```diff
@@ -226,11 +198,16 @@ class BallCoverAllKNNTest : public ::testing::TestWithParam<BallCoverInputs> {
...
 
+    cudaStream_t *int_streams = nullptr;
+    std::vector<int64_t> *translations = nullptr;
+
     std::vector<float *> input_vec = {d_train_inputs.data()};
     std::vector<uint32_t> sizes_vec = {n};
 
-    compute_bfknn(handle, d_train_inputs.data(), d_train_inputs.data(), n, d, k,
-                  metric, d_ref_D.data(), d_ref_I.data());
+    raft::spatial::knn::detail::brute_force_knn_impl<uint32_t, int64_t>(
```
cjnolet, line 207: Now that we have a knn that's not just proxying down to faiss, we should be gtesting it accordingly, similar to what's being done w/ the haversine and ball cover gtests. It's also important going forward because RAFT is beginning to get used by more projects and thus the impact of breaking tests is more than just cuml.

My suggestion is to test `l2_unexpanded_knn` and `l2_expanded_knn` directly in the gtests and then we can test the `brute_force_knn` more generally.

cpp/include/raft/spatial/knn/knn.hpp
mdoijade: I'll be verifying this fork with cuML PR for verifying these changes in CI

cpp/test/spatial/ball_cover.cu
mdoijade: certainly a rigorous tests within RAFT are needed for them, I'll add gtests for them separately, now instead of `l2_unexpanded_knn` and `l2_expanded_knn` we have single entry `fusedL2Knn` function for both of them.
For these kernel I relied on cuML cpp knn tests & pytests so far which is no longer correct as you rightly mention it.

GPUtester:

cjnolet:
add to allowlist

cpp/test/spatial/ball_cover.cu
mdoijade: a new test `tests\spatial\fused_l2_knn.cu` is added which does testing for both L2 exp/unexp cases which compares its output with faiss bfknn call. 
I've also polished the fp32 atomicMax device function which I believe is more faster than atomicCAS based version and also takes care of NaNs.

mdoijade: Apologies for the delay in updating this PR with the unit test

ChuckHastings (approved):
Shouldn't affect cugraph

mdoijade:
@cjnolet can this PR get merged?

mdoijade:
@cjnolet I see this PR is marked for v22.02 so you'll be auto merging it or there is additional action required from my side?

cjnolet:
@mdoijade, I scraped through the PRs a couple weeks ago and aligned them to expected releases. Looking back through my reivew, I'm really happy for the new tests but it looks like there are still a couple (very minor) things to address.

mdoijade:
> 
> 
> @mdoijade, I scraped through the PRs a couple weeks ago and aligned them to expected releases. Looking back through my reivew, I'm really happy for the new tests but it looks like there are still a couple (very minor) things to address.

@cjnolet  I believe now I have addressed all the points in this PR, the build failure is coming from `../test/sparse/dist_coo_spmv.cu`  in the CI, but locally I cannot repro this build failure.

cjnolet (approved):
LGTM

cjnolet:
@gpucibot merge

PR #349, opened 2021-10-01
Use 64 bit CuSolver API for Eigen decomposition

lowener:
This PR replace the legacy call to `cusolverDnsyevj` with the 64-bit version.
It also improves the indexing types used, from int to size_t when possible.

### DIFF
--- a/cpp/include/raft/linalg/cusolver_wrappers.h
+++ b/cpp/include/raft/linalg/cusolver_wrappers.h
@@ -719,5 +719,82 @@ inline cusolverStatus_t cusolverSpcsrqrsvBatched(  // NOLINT
 }
 /** @} */
 
+/**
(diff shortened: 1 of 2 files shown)


lowener:
I just noticed that `DnXsyevd` was introduced in CUDA 11.1.0 in Sept. 2020. So the CI with CUDA version 11.0 will always fail.

cjnolet:
@lowener, 

> I just noticed that DnXsyevd was introduced in CUDA 11.1.0 in Sept. 2020. So the CI with CUDA version 11.0 will always fail.

Indeed, I remember discussing this very detail w/ @teju85 not too long ago. A good workaround here could be to be check the CUDA version and have the wrapper call the old version in 11.1.

cjnolet (changes requested):

cjnolet:
@gpucibot merge

PR #356, opened 2021-10-06
Hiding implementation details for random, stats, and matrix

divyegala:


cpp/include/raft/random/rng.cuh
```diff
@@ -291,7 +249,8 @@ class Rng {
    */
   template <typename Type, typename LenType = int>
   void fill(Type *ptr, LenType len, Type val, cudaStream_t stream) {
```
cjnolet, line 251: It looks like there's quite abit of implementation details for the `Rng` type, including CUDA details. Also, if the hidden impls of the random generator kernels need access to the `Rng` class then the details API is going to be including this file and vice versa. One thing we can do is rename this to something like `RngImpl` and move it to `detail` and then expose it to the public API under the `Rng` name.

cpp/include/raft/stats/mean.cuh
```diff
@@ -90,13 +53,13 @@ void mean(Type *mu, const Type *data, IdxType D, IdxType N, bool sample,
     dim3 grid(raft::ceildiv(N, (IdxType)RowsPerBlk),
               raft::ceildiv(D, (IdxType)ColsPerBlk));
     CUDA_CHECK(cudaMemsetAsync(mu, 0, sizeof(Type) * D, stream));
-    meanKernelRowMajor<Type, IdxType, TPB, ColsPerBlk>
+    detail::meanKernelRowMajor<Type, IdxType, TPB, ColsPerBlk>
```
cjnolet, line 56: Same here wrt invoking kernels from public API. Just proxy to the function in the details API.

cjnolet (changes requested):
Looks like a great start. We want to make sure any classes which are being used internally and contain kernel-specific things are also being hidden behind detail and exposed through a proxy in the public API.

cpp/include/raft/matrix/math.hpp
```diff
@@ -338,51 +317,7 @@ __global__ void argmaxKernel(const T *d_in, int D, int N, T *argmax) {
...
-  }
-  __syncthreads();
-
-  if (need_sign_flip) {
-    for (int i = threadIdx.x; i < D; i += TPB) {
-      int idx = rowStart + i;
-      d_in[idx] = -d_in[idx];
-    }
-  }
+  detail::argmax(in, n_rows, n_cols, out, stream);
```
cjnolet, line 320: much better!

cjnolet (approved):
Changes look great!

cpp/include/raft/random/rng.hpp
```diff
@@ -0,0 +1,376 @@
+/*
```
dantegd, line 1: cugraph uses rng.cuh in a couple of places, so leaving a quick review so we don’t merge this before that side is ready (if it already is, just let me know and glad to approve/merge)

dantegd (changes requested):

dantegd (approved):

dantegd:
@gpucibot merge

PR #363, opened 2021-10-20
Use FAISS with RMM

viclafargue:
Answers https://github.com/rapidsai/cuml/issues/2821


cpp/include/raft/spatial/knn/faiss_mr.hpp
```diff
@@ -0,0 +1,333 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
```
cjnolet, line 7: Thanks for opening the issue w/ the FAISS project: https://github.com/facebookresearch/faiss/issues/2097. 

We also need to reference that issue here with a comment that explains why we've temporarily copied this code and how we can remove the copied code once that issue is resolved.

cjnolet (changes requested):
The changes are looking much better. I also want to make sure we clearly describe the path to removing the copied code and adapting it to their API once that issue is resolved. Once that's done, this PR will be ready to go.

viclafargue:
Opened an issue on FAISS to avoid code duplication : facebookresearch/faiss#2097.

Using a custom memory manager in FAISS currently requires overriding methods of `GpuResources` and `GpuResourcesProvider` classes. Most of the logic however, is consistent with existing `StandardGpuResourcesImpl` and `StandardGpuResources` classes causing unnecessary and difficulty maintainable rewrites.

It would be possible to prevent consequent rewrites thanks to updates to the `StandardGpuResources.h` header file. Indeed the ideal situation would be to inherit the logic of these classes and only override the `allocMemory` and `deallocMemory` methods.

The necessary changes would be the following :
1. Promote attributes from private to protected
2. Use `allocMemory` in `initializeForDevice` to prevent `cudaHostAlloc` call
3. (optionally) Remove unnamed namespace to allow the reuse of the `allocsToString` function

cjnolet:
@viclafargue, I looked back over my last comment and realized it was fairly ambiguous. I think having the comment on this PR is very useful as well but we should also have the comment in the code, preferably at the top of `faiss_mr.hpp`

cjnolet (approved):
LGTM

cjnolet:
@viclafargue it's been a little while so just wanted to check the status on this PR. IIRC, we are waiting on cuml side to pass successfuly, right?

viclafargue:
> @viclafargue it's been a little while so just wanted to check the status on this PR. IIRC, we are waiting on cuml side to pass successfuly, right?

Thanks for reminding me of this. Yes, I just solved the merge conflicts.

cjnolet:
@gpucibot merge

PR #370, opened 2021-11-04
Port libcudacxx patch from cudf

dantegd:
Porting @robertmaynard's patch for libcudacxx patch for CUDA 11.5 from cuDF, this should fix RAFT in 11.5 and in conjunction with https://github.com/rapidsai/cuml/pull/4327 also unblocks cuML

### DIFF
--- a/cpp/cmake/thirdparty/get_libcudacxx.cmake
+++ b/cpp/cmake/thirdparty/get_libcudacxx.cmake
@@ -1,0 +1,30 @@
+# =============================================================================
+# Copyright (c) 2020-2021, NVIDIA CORPORATION.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+# in compliance with the License. You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software distributed under the License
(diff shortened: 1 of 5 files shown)


cjnolet (approved):
LGTM

cjnolet:
rerun tests

cpp/cmake/thirdparty/get_libcudacxx.cmake
```diff
@@ -0,0 +1,30 @@
...
+# This function finds libcudacxx and sets any additional necessary environment variables.
+function(find_and_configure_libcudacxx)
+  include(${rapids-cmake-dir}/cpm/libcudacxx.cmake)
+
+  rapids_cpm_libcudacxx(
+    BUILD_EXPORT_SET raft-exports INSTALL_EXPORT_SET raft-exports PATCH_COMMAND patch
+    --reject-file=- -p1 -N < ${RAFT_SOURCE_DIR}/cmake/libcudacxx.patch || true
+  )
+
+  set(LIBCUDACXX_INCLUDE_DIR
```
robertmaynard, line 24: These lines should be dropped as `LIBCUDACXX_INCLUDE_DIR` is not used in RAFT, and since we don't check `libcudacxx_ADDED` the value stored would be incorrect

robertmaynard (changes requested):

robertmaynard (approved):

dantegd:
@gpucibot merge

raydouglass (approved):

PR #372, opened 2021-11-09
Fix coalesced access checks in matrix_vector_op

achirkin:
One of the conditions in [`test_aligned_access`](https://github.com/rapidsai/raft/blob/branch-21.12/cpp/include/raft/linalg/matrix_vector_op.cuh#L106) in `linalg/matrix_vector_op.cuh` was incorrect (`ptr % elem_size` should be zero, not otherwise). Due to that typo, `matrixVectorOp` function was never using vectorized load/store instructions.
This PR fixes the problem while also adding a new helper struct to simplify such checks in future.


achirkin:
I've noticed this problem while investigating that normalizing the dataset using this op takes major fraction of time in the least squares algorithm (lstsq/eig). Just fixing that reduces the typical exec time of `matrixVectorOp` by a factor of x1.5-x2 in my tests. However, a highly specialized handwritten-kernel version reduces the time by a factor of x8-x10. So I will try to see if there is a room for improvements here.

dantegd:
rerun tests

achirkin:
return tests

achirkin:
rerun tests

cpp/include/raft/pow2_utils.cuh
```diff
@@ -0,0 +1,85 @@
...
+
+namespace raft {
+
+/**
+ * @brief Fast arithmetics and alignment checks for power-of-two values known at compile time.
+ *
+ * @tparam Value a compile-time value representable as a power-of-two.
+ */
+template <auto Value>
+struct Pow2 {
```
cjnolet, line 29: I know it seems tedius, but since we're adding these new functions here as separate utiliities from `matrix_vector_op`, we should add a couple small unit tests specifically for them.

cjnolet:
rerun tests

cjnolet:
rerun tests

cpp/include/raft/pow2_utils.cuh
achirkin: Sure, sorry for that

cjnolet:
@gpucibot merge

PR #375, opened 2021-11-15
Fixing bad host->device copy

cjnolet:
A host vector in haversine knn test was being resized to `n` rather `n*d` before `n*d` elements were copied from it to a device vector. Not sure if this is causing the following issue in the test, but it's not correct nonetheless:

```
[----------] 1 test from HaversineKNNTestF
[ RUN      ] HaversineKNNTestF.Fit
unknown file: Failure
C++ exception with description "std::bad_alloc: CUDA error at: _deps/rmm-src/include/rmm/mr/device/cuda_memory_resource.hpp:70: cudaErrorMemoryAllocation out of memory" thrown in the test fixture's constructor.
[  FAILED  ] HaversineKNNTestF.Fit (0 ms)
[----------] 1 test from HaversineKNNTestF (0 ms total)
```

Since it was happening on host, the only thing I can think of is that somehow there is some host memory being used to determine the amount of device memory to allocate, which is grabbing a garbage value from somewhere.

### DIFF
--- a/cpp/test/spatial/haversine.cu
+++ b/cpp/test/spatial/haversine.cu
@@ -77,9 +77,6 @@ class HaversineKNNTest : public ::testing::Test {
     raft::update_device<value_idx>(d_ref_I.data(), h_res_I.data(), n * n,
                                    stream);
 
-    std::vector<value_t *> input_vec = {d_train_inputs.data()};
-    std::vector<value_idx> sizes_vec = {n};
-
     raft::spatial::knn::detail::haversine_knn(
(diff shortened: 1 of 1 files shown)


cjnolet:
rerun tests

divyegala (approved):
LGTM

cjnolet:
@achirkin, absolutely I can remove those 2 lines. I'm actually surprised it wasn't tripping an error since those two variables are never used. Might be an indication that a flag needs to be enabled during compile.

BradReesWork (approved):

cjnolet:
@gpucibot merge

Issue #128, opened 2021-01-27
[BUG] RAFT header conflicts


bryevdv:
**Describe the bug**
RAFT headers do not play nicely with some other RAPIDS libraries because of duplicated names

in rapids-js, just doing 
```
include <cugraph/graph.hpp>
```
results in:

```
./../../.cache/cpm/raft/39a4ee56e5ca66c2be3eb2b78767b9d0d46759cc/cpp/include/raft/cudart_utils.h:52: error: "CUDA_TRY" redefined [-Werror]
   52 | #define CUDA_TRY(call)                                                        \
      | 
In file included from ../../../cudf/src/node_cudf/utilities/error.hpp:20,
                 from ../../../cudf/src/node_cudf/scalar.hpp:17,
                 from ../../../cudf/src/node_cudf/column.hpp:17,
                 from ../../src/node_cugraph/graph_coo.hpp:17,
                 from ../../src/graph_coo.cpp:15:
../../../.cache/cpm/cudf/6bbbff12b33dfcdeef951b16babffd1472c03bdb/cpp/include/cudf/utilities/error.hpp:102: note: this is the location of the previous definition
  102 | #define CUDA_TRY(call)                                            \
      | 
```

Currently to mitigate we have an `internal/graph.hpp` header to do this:

```
#ifdef CUDA_TRY
#undef CUDA_TRY
#endif
#ifdef CHECK_CUDA
#undef CHECK_CUDA
#endif
#include <cugraph/graph.hpp>
#ifdef CHECK_CUDA
#undef CHECK_CUDA
#endif
#ifdef CUDA_TRY
#undef CUDA_TRY
#endif
```

But it would be better if these RAPIDS tools could coordinate to avoid these clashes. 

cc @trxcllnt

bryevdv:
This still seems worth addressing.

AjayThorve:
+1 Facing similar issues while building raft with cuml, for `node-rapids` bindings.

cjnolet:
@bryevdv @AjayThorve @trxclln

tIs the problem here is related to conflicting names which are defined in multiple projects or are these conflicts defined only in RAFT but having problems because somehow the RAFT headers are being installed in different locations? If it's the former, I would recommend we centralize the conflicting names into RAFT (potentially prefixed w/ a `RAFT` specifier) to remove duplication. If the latter, we might want to consider consolidating the RAFT header installs. Any thoughts?

bryevdv:
cc @trxcllnt 

[...]

PR #378, opened 2021-11-15
Removing conflict w/ CUDA_CHECK

cjnolet:
Closes https://github.com/rapidsai/raft/issues/128

### DIFF
--- a/cpp/include/raft/cudart_utils.h
+++ b/cpp/include/raft/cudart_utils.h
@@ -38,10 +38,10 @@ namespace raft {
 /**
  * @brief Exception thrown when a CUDA error is encountered.
  */
-struct cuda_error : public raft::exception {
-  explicit cuda_error(char const* const message) : raft::exception(message) {}
-  explicit cuda_error(std::string const& message) : raft::exception(message) {}
-};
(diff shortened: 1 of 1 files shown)


divyegala (approved):
LGTM

cpp/include/raft/cudart_utils.h
```diff
@@ -101,6 +105,15 @@ struct cuda_error : public raft::exception {
...
     }                                                                     \
   } while (0)
+#endif
+
+/**
+ * Alias to raft scope for now.
+ * TODO: Rename original implementations in 22.04 to fix
+ * https://github.com/rapidsai/raft/issues/128
+ */
+#define RAFT_CUDA_CHECK(call) CUDA_CHECK(call)
```
ChuckHastings, line 115: I would have done this in reverse.  That is, I would have conditionally defined `CUDA_CHECK` as a call to `RAFT_CUDA_CHECK` which would make the next change smaller.  But this way is fine too.

ChuckHastings (approved):

cjnolet:
@gpucibot merge

Issue #330, opened 2021-09-17
[FEA] All primitives should be hiding implementation details and exposing only public API 


cjnolet:
While the primitives in cuml only needed to worry about cuml algorithms as consumers, hiding implementation details wasn't particularly important. However, now that the primitives in raft are beginning to find more consumers downstream, we should start doing this across the board for all primitives. There are several benefits to doing this an the only drawback that I can think of is that it's going to require an initial refactor to clean up the current primitives (which I propose we do incrementally)

Benefits: 
1. The contract between RAFT and the consumers will be more concrete so we can change implementations with more agility so long as it doesn't create large changes to the contract (or expectations of the behavior). 
2. The public APIs become easier for users to discover while digging around the code
3. Documentation of the public APIs becomes more manageable and it becomes much more straightford to build and maintain doxygen API docs that can be published with the rest of the RAPIDS docs. 
4. It becomes much more straightforward which prims belong to which packages and, reiterating from no.1, it's much more straightforward for users which functions are being used. 
5. The include files for  the public API can be consistently `hpp`, which puts less burden on consumers to know which one to use. I think we can maintain backwards compatibility with this for awhile, maybe with a warning, and then cut out the `cuh` files from the public API altogether at some point. 

Tagging the usual suspects for thoughts. @teju85 @divyegala @ChuckHastings @tfeher @wphicks My proposal would be that we start doing this for all new prims as of 21.12 and then iteratively move the implementation details for existing prims (note that we should be able to do this initially without breaking the public APIs at all, so long as the consumers are all actually using public APIs downstream).

divyegala:
@cjnolet thanks for bringing this up. I agree with each and every proposed point here! The only thing that I would like to do differently is that I would like to specify a standard contract, as you mention, for every RAFT public primitive that we have right now. This would involve things like (and likely more):
1. Getting this PR https://github.com/rapidsai/raft/pull/294 merged for type safety, multiple input types, etc
2. Removing extra stream argument 
3. Replacing unreadable booleans on the API with clear enum classes

Once we have a contract ready for all existing prims, we can enforce it for new prims. But going backwards would mean having to refactor for all prims that come in at the same time. Thus, I propose that we implement all your proposals for existing primitives first, and let that be the standard that needs to be met for any new primitive that comes into RAFT.

cjnolet:
@divyegala,

[...]

PR #383, opened 2021-11-17
LinAlg impl in detail

divyegala:
Addresses #330


ajschmidt8 (approved):

cpp/include/raft/distance/detail/distance.cuh
```diff
@@ -31,6 +30,7 @@
 #include <raft/distance/detail/l1.cuh>
 #include <raft/distance/detail/minkowski.cuh>
 #include <raft/distance/detail/russell_rao.cuh>
+#include <raft/linalg/distance_type.hpp>
```
cjnolet, line 33: We should use this opportunity to move `distance_type` into `raft/distance`.

cpp/include/raft/handle.hpp
```diff
@@ -32,10 +32,10 @@
 ///@todo: enable once we have migrated cuml-comms layer too
 //#include <common/cuml_comms_int.hpp>
 
-#include <raft/linalg/cublas_wrappers.h>
-#include <raft/linalg/cusolver_wrappers.h>
 #include <raft/sparse/cusparse_wrappers.h>
 #include <raft/comms/comms.hpp>
+#include <raft/linalg/cublas_wrappers.hpp>
```
cjnolet, line 37: These should be in detail. They aren't intended to be public and are only meant to be used as a facade for the actual cublas/cusolver/cusparse/etc... calls.

cpp/include/raft/linalg/binary_op.hpp
```diff
@@ -0,0 +1,54 @@
+/*
```
cjnolet, line 1: The years need to be updated everywhere.

cpp/include/raft/linalg/eig.hpp
```diff
@@ -0,0 +1,114 @@
...
+           cudaStream_t stream)
+{
+  detail::eigDC(handle, in, n_rows, n_cols, eig_vectors, eig_vals, stream);
+}
+
+using detail::COPY_INPUT;
+using detail::EigVecMemUsage;
+using detail::OVERWRITE_INPUT;
+
+#if CUDART_VERSION >= 10010
```
cjnolet, line 53: We should be able to drop this now since RAPIDS is always >=CTK 11

cjnolet (changes requested):
I've only gone through it quickly in a first pass so far but changes look great for the most part. Just saw a couple smaller things initially.

cpp/include/raft/handle.hpp
divyegala: Oh yep, you are totally right - missed that one my bad

cjnolet: Jus to update this conversation so it's not lost over the holidays, the intention here is to wrap the standard cuda library functions so we centralize all the calls but to provide RAFT public API functions which are completely agnostic of the standard cuda libraries underneath.

As discussed in https://github.com/rapidsai/cuml/pull/4446, for a short time, we might have a few places in the cuml solver code which is calling down to raft's `linalg::detail` namespace. These will need to be updated to use the public API, once it's available (and we do that in a follow-on PR).

cjnolet (changes requested):
This is a lot of files so I might need to do a second pass over them. So far, it's looking really good, though. The couple things I've found are very minor.

cpp/include/raft/spectral/spectral_util.hpp
```diff
@@ -18,6 +18,7 @@
 
 #include <raft/cudart_utils.h>
 #include <raft/handle.hpp>
+#include <raft/linalg/detail/cublas_wrappers.hpp>
```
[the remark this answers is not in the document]
divyegala: I have not worked with `spectral` in this PR. Should this not be addressed when we work with that package?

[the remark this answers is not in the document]
cjnolet: Ah, yeah I overlooked the packaging during review. We can disregard this one for now.

cpp/include/raft/distance/distance_type.hpp
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, NVIDIA CORPORATION.
+ * Copyright (c) 2021, NVIDIA CORPORATION.
  *
```
cjnolet, line 3: Stray copyright that needs updating

cpp/include/raft/linalg/detail/eig.hpp
```diff
@@ -16,17 +16,20 @@
 
 #pragma once
 
+#include "cusolver_wrappers.hpp"
```
cjnolet, line 19: I think this is duplicated.

cpp/include/raft/linalg/detail/lanczos.hpp
```diff
@@ -0,0 +1,1400 @@
...
+
+  // -------------------------------------------------------
+  // Compute second Lanczos vector
+  // -------------------------------------------------------
+  if (*iter <= 0) {
+    *iter = 1;
+
+    // Apply matrix
+    if (shift != 0)
+      CUDA_TRY(cudaMemcpyAsync(lanczosVecs_dev + n,
```
cjnolet, line 126: We need to be mindful that we update these as well while we're moving this stuff over. `RAFT_CUDA_TRY`, `RAFT_CUBLAS_TRY`, etc...

cjnolet (changes requested):
Sorry for the delayed in the review- this is a massive set of changes to go through. Mostly small things at this point. This is almost ready to go. Will look through cuml side next.

cpp/include/raft/linalg/detail/map.cuh
```diff
@@ -0,0 +1,48 @@
...
+}
+
+template <typename InType, typename OutType, typename MapOp, int TPB, typename... Args>
+void mapImpl(
+  OutType* out, size_t len, MapOp map, cudaStream_t stream, const InType* in, Args... args)
+{
+  const int nblks = raft::ceildiv(len, (size_t)TPB);
+  mapKernel<InType, OutType, MapOp, TPB, Args...>
+    <<<nblks, TPB, 0, stream>>>(out, len, map, in, args...);
+  CUDA_CHECK(cudaPeekAtLastError());
```
cjnolet, line 43: Need another rename from `CUDA_CHECK` -> `RAFT_CUDA_TRY`

cpp/include/raft/linalg/detail/map_then_reduce.cuh
```diff
@@ -88,70 +89,9 @@ void mapThenReduceImpl(OutType* out,
   const int nblks = raft::ceildiv(len, (size_t)TPB);
   mapThenReduceKernel<InType, OutType, MapOp, ReduceLambda, TPB, Args...>
     <<<nblks, TPB, 0, stream>>>(out, len, neutral, map, op, in, args...);
-  RAFT_CUDA_TRY(cudaPeekAtLastError());
+  CUDA_CHECK(cudaPeekAtLastError());
```
cjnolet, line 92: Here too

cpp/include/raft/linalg/detail/subtract.cuh
```diff
@@ -0,0 +1,69 @@
...
+                       const math_t* singleScalarDev,
+                       IdxType len,
+                       cudaStream_t stream)
+{
+  // Just for the note - there is no way to express such operation with cuBLAS in effective way
+  // https://stackoverflow.com/questions/14051064/add-scalar-to-vector-in-blas-cublas-cuda
+  const IdxType nblks = raft::ceildiv(len, (IdxType)TPB);
+  subtract_dev_scalar_kernel<math_t>
+    <<<nblks, TPB, 0, stream>>>(outDev, inDev, singleScalarDev, len);
+  CUDA_CHECK(cudaPeekAtLastError());
```
cjnolet, line 64: Here too

cjnolet (approved):

cjnolet (approved):
LGTM

cjnolet:
@gpucibot merge

PR #387, opened 2021-11-18
DBG Change cusolver to static

dantegd:


robertmaynard (commented):
We need to do the following so we don't break devs that have only cuda-toolkit from conda

```cmake
set(cusolver_target CUDA::cusolver_static)
if(NOT TARGET CUDA::cusolver_static)
  message(STATUS "FALLING BACK TO SHARED CUDA::cusolver")
  set(cusolver_target CUDA::cusolver)
endif()
```
And link to  `${cusolver_target}`

PR #393, opened 2021-11-22
Ensure that the correct cusolver version exists when running tests

robertmaynard:
cusolver changed SOVERSION's during the CUDA 11.X release cycle, and therefore we need to ensure that we have the same ABI as compiled against when running tests.


ajschmidt8:
@robertmaynard, should this target `branch-21.12`?

robertmaynard:
> @robertmaynard, should this target `branch-21.12`?

Yes it should, Updated

ajschmidt8:
CI passed. admin merged to `branch-21.12`

PR #397, opened 2021-11-22
[REVIEW] Adding dev conda environment files. 

cjnolet:
With the recent update of the default clang-format version breaking the style checks in a few projects, I realized other projects are explicitly pinning this version while RAFT is not. Now that RAFT is beginning to act more independently of other projects, it's time for it to provide its own dev conda environment files. This will also get the directories in place for the conda recipes, which will be coming in the next couple versions.


raydouglass (approved):

cjnolet:
@gpucibot merge

PR #399, opened 2021-11-23
[REVIEW] Span implementation.

trivialfis:
The implementation is largely ported from xgboost with some simplification and cleanups.

The one in XGBoost was modeled after the core guideline support library instead of std, which
was still a draft back then.  The one in this PR uses a plain pointer as the iterator and doesn't
have bound check in release mode.  ISO span uses `size_type` for index operator, this PR defines
a template parameter to allow indexing using other integer types.

* Implement host_span
* Implement device_span

### DIFF
--- a/cpp/include/raft/common/span.hpp
+++ b/cpp/include/raft/common/span.hpp
@@ -1,0 +1,325 @@
+/*
+ * Copyright (c) 2021, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <cassert>
+#include <cinttypes>  // size_t
+#include <cstddef>    // std::byte
+#include <limits>     // numeric_limits
+#include <type_traits>
+
+#include <thrust/functional.h>
+#include <thrust/iterator/reverse_iterator.h>
+
+namespace raft::common {
+
+constexpr std::size_t dynamic_extent = std::numeric_limits<std::size_t>::max();
+
+template <class ElementType, bool is_device, std::size_t Extent>
+class span;
+
+namespace detail {
+/*!
+ * The extent E of the span returned by subspan is determined as follows:
+ *
+ *   - If Count is not dynamic_extent, Count;
+ *   - Otherwise, if Extent is not dynamic_extent, Extent - Offset;
+ *   - Otherwise, dynamic_extent.
+ */
+template <std::size_t Extent, std::size_t Offset, std::size_t Count>
+struct extent_value_t
+  : public std::integral_constant<
+      std::size_t,
+      Count != dynamic_extent ? Count : (Extent != dynamic_extent ? Extent - Offset : Extent)> {
+};
+
+/*!
+ * If N is dynamic_extent, the extent of the returned span E is also
+ * dynamic_extent; otherwise it is std::size_t(sizeof(T)) * N.
+ */
+template <typename T, std::size_t Extent>
+struct extent_as_bytes_value_t
+  : public std::integral_constant<std::size_t,
+                                  Extent == dynamic_extent ? Extent : sizeof(T) * Extent> {
+};
+
+template <std::size_t From, std::size_t To>
+struct is_allowed_extent_conversion_t
+  : public std::integral_constant<bool,
+                                  From == To || From == dynamic_extent || To == dynamic_extent> {
+};
+
+template <class From, class To>
+struct is_allowed_element_type_conversion_t
+  : public std::integral_constant<bool, std::is_convertible<From (*)[], To (*)[]>::value> {
+};
+
+template <class T>
+struct is_span_oracle_t : std::false_type {
+};
+
+template <class T, bool is_device, std::size_t Extent>
+struct is_span_oracle_t<span<T, is_device, Extent>> : std::true_type {
+};
+
+template <class T>
+struct is_span_t : public is_span_oracle_t<typename std::remove_cv<T>::type> {
+};
+
+template <class InputIt1, class InputIt2, class Compare>
+__host__ __device__ constexpr auto lexicographical_compare(InputIt1 first1,
+                                                           InputIt1 last1,
+                                                           InputIt2 first2,
+                                                           InputIt2 last2) -> bool
+{
+  Compare comp;
+  for (; first1 != last1 && first2 != last2; ++first1, ++first2) {
+    if (comp(*first1, *first2)) { return true; }
+    if (comp(*first2, *first1)) { return false; }
+  }
+  return first1 == last1 && first2 != last2;
+}
+}  // namespace detail
+
+/**
+ * \brief The span class defined in ISO C++20.  Iterator is defined as plain pointer and
+ *        most of the methods have bound check on debug build.
+ */
+template <typename T, bool is_device, std::size_t Extent = dynamic_extent>
+class span {
+ public:
+  using element_type    = T;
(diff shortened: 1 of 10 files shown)


seunghwak:
Any reason to adopt the XGBoost based one instead of the cudf one (https://github.com/rapidsai/cudf/blob/branch-22.02/cpp/include/cudf/utilities/span.hpp)?

AFAIK, this cudf based implementation will eventually go to libcu++.

trivialfis:
Hi, we haven't decided which implementation to use.  The one from xgboost is closer to `std::span` defined in c++ 20.  This PR is just to push the discussion.

seunghwak:
> Hi, we haven't decided which implementation to use. The one from xgboost is closer to `std::span` defined in c++ 20. This PR is just to push the discussion.

I see. I guess the benefit of the cudf one is that it has host_span & device_span so allows us to specify whether the pointer is for host memory or device memory. I guess this is beneficial unless CUDA's unified memory model advances to the level that this distinction is unnecessary.

trivialfis:
Yup, having safeguards is certainly useful.  But the one in cudf doesn't support static extend https://github.com/rapidsai/cudf/blob/7ac8aac8d9174ecfcb06f56199c641bf3d869961/cpp/include/cudf/utilities/span.hpp#L42 .  If we were to adopt it static extend must be integrated (not sure what's the best way yet).

trivialfis:
Added a template parameter to distinguish host and device span.

trivialfis:
Made lint changes to some other files.

hcho3:
Any progress on this one? TreeExplainer in cuML (rapidsai/cuml#4473) can benefit from a Span implementation. (Right now, it has its own small Span class.) Or is this going to be part of the upcoming MDSpan (#437) ?

trivialfis:
One can use mdspan to define span equivalence, but the underlying procedures are lot more complicated.

hcho3:
Got it. In that case, TreeExplainer could potentially use mdspan instead.

cpp/include/raft/common/span.hpp
```diff
@@ -0,0 +1,325 @@
...
+    return {data() + _offset, _count == dynamic_extent ? size() - _offset : _count};
+  }
+
+ private:
+  size_type size_{0};
+  pointer data_{nullptr};
+};
+
+template <typename T, size_t extent = dynamic_extent>
+using host_span = span<T, false, extent>;
```
trivialfis, line 259: Should I rename it to `host_span_t`?

trivialfis: also, should I move the file to top include dir?

cjnolet: I'm going back and forth about whether we should move the mdspan/mdarray into common or most this to the top-level. What do you think?

cpp/include/raft/common/span.hpp
```diff
@@ -0,0 +1,325 @@
...
+  constexpr span(element_type (&arr)[N]) noexcept : size_(N), data_(&arr[0])
+  {
+  }
+
+  template <class U,
+            std::size_t OtherExtent,
+            class = typename std::enable_if<
+              detail::is_allowed_element_type_conversion_t<U, T>::value &&
+              detail::is_allowed_extent_conversion_t<OtherExtent, Extent>::value>>
+  constexpr span(const span<U, is_device, OtherExtent>& _other) noexcept
```
cjnolet, line 142: The public API for this will need some comprehensive doxygen docs. It would also be very useful to include some very brief and simple usage examples (same for mdspan/mdarray).

cjnolet (changes requested):
Very minor things at this point but wanted to provide the review for you so we can get both span and mdspan into 22.04.

cpp/include/raft/common/span.hpp
divyegala: I would actually not want to name it `host_span_t`, because I think only STL types should use that `_t` extension. `raft::host_span` seems okay to me personally

divyegala (commented):
Looks great to me. Just one comment, and love the test coverage. Just another question - why so many direct calls to cudart in the tests?

cpp/include/raft/common/span.hpp
trivialfis: I will try to document it briefly, but it's part of c++ standard with CUDA additions.  So for details, I think it's best to refer to cppreference or similar services.  If something is inconsistent with the c++ standard we should fix it.

trivialfis:
> I would actually not want to name it host_span_t, because I think only STL types should use that _t extension. raft::host_span seems okay to me personally

Got it, I will remove all the postfix for span and mdspan.

> why so many direct calls to cudart in the tests?

My apologies, it was ported from xgboost, which supports enumerating devices.  Will remove those calls.

cpp/include/raft/common/span.hpp
```diff
@@ -0,0 +1,325 @@
...
+namespace detail {
+/*!
+ * The extent E of the span returned by subspan is determined as follows:
+ *
+ *   - If Count is not dynamic_extent, Count;
+ *   - Otherwise, if Extent is not dynamic_extent, Extent - Offset;
+ *   - Otherwise, dynamic_extent.
+ */
+template <std::size_t Extent, std::size_t Offset, std::size_t Count>
+struct extent_value_t
```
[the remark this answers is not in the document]
trivialfis: @cjnolet We have some integration work to do.  For instance, the `dynamic_extent` in this PR is duplicated with the one in `stdex`.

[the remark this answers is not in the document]
trivialfis: We can polish after having the base implementation and make a list of additional features we want, like conversion between span and 1-dim mdspan, padding for arrays, resize for arrays.

cpp/include/raft/common/span.hpp
```diff
@@ -0,0 +1,325 @@
+/*
+ * Copyright (c) 2021, NVIDIA CORPORATION.
```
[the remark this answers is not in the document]
trivialfis: Done.

trivialfis:
Changed base to 22.04.

cpp/include/raft/common/span.hpp
```diff
@@ -0,0 +1,325 @@
...
+#include <cassert>
+#include <cinttypes>  // size_t
+#include <cstddef>    // std::byte
+#include <limits>     // numeric_limits
+#include <type_traits>
+
+#include <thrust/functional.h>
+#include <thrust/iterator/reverse_iterator.h>
+
+namespace raft::common {
```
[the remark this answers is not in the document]
trivialfis: Done.

cpp/include/raft/common/span.hpp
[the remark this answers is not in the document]
trivialfis: Opened 2 issues:
- https://github.com/rapidsai/raft/issues/479
- https://github.com/rapidsai/raft/issues/478

cjnolet (approved):
@gpucibot merge

BradReesWork (approved):

trivialfis:
Hey, is there anything else I need to do for this PR? Maybe retrigger the CI?

cjnolet:
@gpucibot merge

cpp/include/raft/span.hpp
```diff
@@ -0,0 +1,283 @@
...
+
+  constexpr auto subspan(size_type _offset, size_type _count = dynamic_extent) const
+    -> span<element_type, is_device, dynamic_extent>
+  {
+    assert((_count == dynamic_extent) ? (_offset <= size()) : (_offset + _count <= size()));
+    return {data() + _offset, _count == dynamic_extent ? size() - _offset : _count};
+  }
+
+ private:
+  size_type size_{0};
```
jrhemstad, line 200: This implementation is missing a principle characteristic of `span` where a span with a static extent has only a single pointer data member. This is usually accomplished by doing something like:
```
template <typename T, std::size_t Extent>
struct span_storage{
   span_storage(T* p, std::size_t /*unused*/) : data_{p} {}
   constexpr std::size_t size(){ return Extent; };
   T* data_ = nullptr;
   // no size member
};
template <typename T>
struct span_storage<T, dynamic_extent>{
   span_storage(T* p, std::size_t s) : data_{p}, size_{s} {}
   constexpr std::size_t size() { return size_; };
private:
   T* data_ = nullptr;
   std::size_t size_;
};
...

template <typename T, std::size_t Extent = dynamic_extent>
struct span{
   ...
private:
   span_storage<T, Extent> storage_;
}
```

This further guarantees the compiler will optimize according to the static extent.

Issue #308, opened 2021-08-03
[BUG] raft::allocate bypasses RMM


harrism:
**Describe the bug**
```
/** cuda malloc */
template <typename Type>
void allocate(Type*& ptr, size_t len, bool setZero = false) {
  CUDA_CHECK(cudaMalloc((void**)&ptr, sizeof(Type) * len));
  if (setZero) CUDA_CHECK(cudaMemset(ptr, 0, sizeof(Type) * len));
}
```

Surprised to see that this quite commonly called function (in RAFT and cuML) calls directly into cudaMalloc, making memory pool use impossible.

dantegd:
Not sure where you found that allocate that indeed it shouldn’t be around unless it’s for a test perhaps?  But in general the allocate that is used by raft (and cuML) is here https://github.com/rapidsai/raft/blob/947e22f65c2fd44f33f0f27c738bd74a1a1a69b4/cpp/include/raft/mr/device/allocator.hpp#L43 (which is being replaced to use rmm directly) so if something is using that allocate you mention it is a big and should be removed. 

Cc @viclafargue since you’re doing the raft rmm refactor, it’d be good to double check nothing in the main code is using this cudamalloc

viclafargue:
Yes, with the RAFT refactor, `raft::allocate` should now use RMM :
https://github.com/rapidsai/raft/blob/7725a809ab87fe929c9a8e3b4221d8ab675104d3/cpp/include/raft/cudart_utils.h#L268-L277

The new design of `raft::allocate`, `raft::deallocate`, and `raft::deallocate_all` is still in progress though.

harrism:
@dantegd it's in `<raft/cudart_utils.h>`

And it's called *all over the place*, not just in tests. Just search the codebase for `raft::allocate(` and you will see.

harrism:
Moreover, very little code should need to call `memory_resource::allocate`: only thrust allocators. Everything else should use containers: `device_buffer` for untyped or byte data, `device_uvector` and `device_vector` for typed data. No raw pointers.  Searching the codebase for the string `allocate(` should turn up very few results. 

As an example, if you search for this string in libcudf it comes up 10 times, 8 of which are inside allocators. And I think all 10 will be gone from libcudf in the near future.

dantegd:
[...]

PR #400, opened 2021-11-24
[WIP] Remove RAFT memory management

viclafargue:
Answers #308.
Requires the appropriate changes in `cuML` and `cuGraph` before merging.


cjnolet:
@viclafargue, it looks like there's just a couple includes for `allocator` remaining in the code and this is ready to go?

viclafargue:
Postponing to 22.04, as discussed offline.

cjnolet:
rerun tests

cpp/include/raft/mr/host/allocator.hpp
```diff
@@ -1,61 +0,0 @@
...
- * further to the ones listed in `Allocator`:
- * - Allocations don't need to be zero copy accessible form a device.
- */
-class allocator : public base_allocator {
-};
-
-/** Default cudaMallocHost/cudaFreeHost based host allocator */
-class default_allocator : public allocator {
- public:
-  void* allocate(std::size_t n, cudaStream_t stream) override
```
cjnolet, line 44: I had tried to remove this at one point and realized cugraph is still using it. We should also open a PR in cugraph to move the use of host allocator/buffer to something equivalent.

seunghwak: I just searched cugraph/cpp and could not find code calling this. We need to double check but we may have removed the code using this function.

cjnolet: @seunghwak I think you are right, I'm no longer seeing it in cugraph either.

cjnolet (approved):

cjnolet:
@gpucibot merge

PR #403, opened 2021-12-01
[REVIEW] Reverting default knn distance to L2Unexpanded for now.

cjnolet:
I believe the new expanded knn changes might need some further investigation. Before we started using the fused l2 knn, both unexpanded and expanded variants mapped to the same (expanded) distance computation in FAISS and the cuml tests passed consistently. For some reason, the new expanded fused l2 knn version is causing test failures within cuml. We should figure out if this is from assertions which are too tight or from a possible bug.


cjnolet:
@gpucibot merge

dantegd (approved):

PR #411, opened 2021-12-09
fix nan issues in L2 expanded sqrt KNN distances

mdoijade:
-  fix nans introduced in l2 expanded sqrt  KNN distances due to very small negative 0s produced by fusedL2knn due to floating point computation uncertainty. 
- re-enable fused l2 knn.


cjnolet (approved):
LGTM. Thanks, Mahesh!

cjnolet:
@gpucibot merge

PR #416, opened 2021-12-10
NVTX range helpers

achirkin:
Moves the helpers for NVTX ranges from cuml with minor additions.

Also adds the helpers to couple places in the codebase - as a way of testing the new feature.


cpp/include/raft/common/nvtx.hpp
```diff
@@ -0,0 +1,119 @@
...
+  When NVTX is enabled, push a named nvtx range and pop it at the end of the enclosing code block.
+
+  This macro initializes a dummy AUTO_RANGE variable on the stack,
+  which pushes the range in its constructor and pops it in the destructor.
+*/
+#ifdef NVTX_ENABLED
+#define RAFT_USING_RANGE(...) raft::common::AUTO_RANGE _AUTO_RANGE_##__LINE__(__VA_ARGS__)
+#else
+#define RAFT_USING_RANGE(...) (void)0
+#endif
```
achirkin, line 116: Tagging @wphicks , @cjnolet to determine the faith of the macro (continuing discussion https://github.com/rapidsai/cuml/pull/4436).

A quick recap; I added the macro in the first place because:
  1. It looks neater *to me* to use `RAFT_USING_RANGE("nvtx label")` instead of `AUTO_RANGE someName("nvtx label")`
  2. It generates a unique name, so I'm sure the compiler wouldn't generate any warnings about name shadowing in nested blocks.
  3. With NVTX disabled, I'm sure it compiles to absolutely nothing and I'm sure the compiler won't give an unused variable warning
 
Which isn't compelling to @wphicks, because we should avoid adding new macros when possible, and the compiler is smart enough to not do things mentioned in (3).

cpp/include/raft/common/detail/nvtx.cuh
```diff
@@ -0,0 +1,202 @@
...
+  /** value */
+  static constexpr float V = 0.85f;
+  /** golden ratio */
+  static constexpr float Phi = 1.61803f;
+  /** inverse golden ratio */
+  static constexpr float InvPhi = 1.f / Phi;
+};
+
+std::unordered_map<std::string, uint32_t> ColorGenState::allColors;
+std::mutex ColorGenState::mapMutex;
```
achirkin, line 50: I've just realized that the implementation copied from cuml's `nvtx.cu` is really meant to be in a source file rather than in the header. What shall we do about it in currently header-only `raft`?

cpp/include/raft/common/detail/nvtx.cuh
```diff
@@ -0,0 +1,202 @@
...
+  }
+  float h = rand() * 1.f / RAND_MAX;
+  h += ColorGenState::InvPhi;
+  if (h >= 1.f) h -= 1.f;
+  auto rgb = hsv2rgb(h, ColorGenState::S, ColorGenState::V);
+  if (!tag.empty()) { ColorGenState::allColors[tag] = rgb; }
+  return rgb;
+}
+
+static inline nvtxDomainHandle_t domain = nvtxDomainCreateA("raft");
```
achirkin, line 139: Previously in cuml, this code was only used in cuml, so the domain name was fixed. Now, it can be used in raft and in any of its dependencies. But we don't provide a way to set/change the NVTX domain. 

Any suggestions how we could improve that? Preferably, without adding an extra argument to each helper?

cpp/include/raft/common/nvtx.hpp
```diff
@@ -0,0 +1,119 @@
...
+   */
+  template <typename... Args>
+  AUTO_RANGE(const char* format, Args... args) : streamMaybe(std::nullopt)
+  {
+    PUSH_RANGE(format, args...);
+  }
+
+  ~AUTO_RANGE()
+  {
+    if (streamMaybe.has_value())
```
wphicks, line 98: Don't forget brackets here.

cpp/include/raft/common/nvtx.hpp
```diff
@@ -0,0 +1,119 @@
...
+inline void POP_RANGE() { detail::popRange(); }
+
+/**
+ * @brief Synchronize CUDA stream and pop the latest nvtx range
+ * @param stream stream to synchronize
+ */
+inline void POP_RANGE(rmm::cuda_stream_view stream) { detail::popRange(stream); }
+
+/** Push a named nvtx range that would be popped at the end of the object lifetime. */
+class AUTO_RANGE {
```
wphicks, line 57: Might we use a more descriptive name? Maybe `NVTXRange` or `ScopedNVTXRange` if we're feeling verbose?

cpp/include/raft/common/nvtx.hpp
```diff
@@ -0,0 +1,119 @@
...
+  template <typename... Args>
+  AUTO_RANGE(const char* format, Args... args) : streamMaybe(std::nullopt)
+  {
+    PUSH_RANGE(format, args...);
+  }
+
+  ~AUTO_RANGE()
+  {
+    if (streamMaybe.has_value())
+      POP_RANGE(streamMaybe.value());
```
wphicks, line 99: Given that you've already checked that `streamMaybe` has a value immediately beforehand, it is safe to use the non-error-checked version of value retrieval:
```suggestion
      POP_RANGE(*streamMaybe);
```

wphicks (changes requested):

cpp/include/raft/common/nvtx.hpp
achirkin: Sure. Does your comment also apply to PUSH/POP versions?

cpp/include/raft/common/nvtx.hpp
achirkin: Ok! Though I find the `.value()` version somewhat more readable.

cpp/include/raft/common/nvtx.hpp
wphicks: I'm less concerned about those because it's in line with how people usually name their NVTX helpers, but yep I think it would be slightly nicer to have more descriptive names for them as well if it's not too much trouble.

cpp/include/raft/common/detail/nvtx.cuh
achirkin: To avoid overcomplicating this, I just changed the domain name to a more neutral "application".

cpp/include/raft/common/nvtx.hpp
```diff
@@ -0,0 +1,125 @@
...
+}
+
+/**
+ * @brief Synchronize CUDA stream and push a named nvtx range
+ * @param format range name format (accepts printf-style arguments)
+ * @param args the arguments for the printf-style formatting
+ * @param stream stream to synchronize
+ */
+template <typename... Args>
+inline void PUSH_NVTX_RANGE(rmm::cuda_stream_view stream, const char* format, Args... args)
```
cjnolet, line 42: Nitpick: I'd prefer to keep caps for constants and macros. Any reason we can't/shouldn't use lower-case for these?

cpp/include/raft/common/nvtx.hpp
```diff
@@ -0,0 +1,125 @@
...
+  When NVTX is enabled, push a named nvtx range and pop it at the end of the enclosing code block.
+
+  This macro initializes a dummy NvtxRange variable on the stack,
+  which pushes the range in its constructor and pops it in the destructor.
+*/
+#ifdef NVTX_ENABLED
+#define _RAFT_USING_NVTX_RANGE_CAT(a, b)       a##b
+#define _RAFT_USING_NVTX_RANGE_UNIQUE_NAME(ln) _RAFT_USING_NVTX_RANGE_CAT(_NvtxRange_, ln)
+#define RAFT_USING_NVTX_RANGE(...) \
+  raft::common::NvtxRange _RAFT_USING_NVTX_RANGE_UNIQUE_NAME(__LINE__)(__VA_ARGS__)
```
cjnolet, line 119: The use of the line information and variable args here is a compelling reason enough for me to be in favor of using a macro here. This is also the reason we use macros for logging. I could go either way with the `NVTX_ENABLED` variable but I'm inclined to use this pattern since it's familiar and used throughout the RAPIDS c/c++ stack. @wphicks does exposing the macro as a wrapper just to capture line / args info seem like a reasonable compromise here? 

I do like the use of the macros as a direct wrapper around a C++ function, which should make it easier to maintain.

cpp/test/nvtx.cpp
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright (c) 2019-2021, NVIDIA CORPORATION.
```
cjnolet, line 2: Same here wrt year

cjnolet (changes requested):
I'm really happy to host this in RAFT and hopeful other consumers might also be able to make use of it. Mostly minor things and my thoughts on the macro usage.

cpp/include/raft/common/nvtx.hpp
achirkin: Note, I use the line info here only to give a unique name to the variable, it's not passed along to the constructor. Also the arguments are passed as-is, so it's really only about the convenience of using the macro vs the clarity of declaring an `NvtxRange` variable of explicitly.

achirkin: (Just to clarify, I use these three macros here instead of one to make sure `__LINE__` is expanded before it's concatenated with `##`)

cpp/include/raft/common/nvtx.hpp
achirkin: It was in caps in cuml. I suppose, there was some logic or legacy behind this. Maybe to make it stand out of everything else to have a better chance of not forgetting to match push/pop directives?.. Also here in raft the caps may scare off developers and compel them to use the safer `NvtxRange` version :)

Though, I don't mind changing it to the camel case if you prefer that.

cpp/include/raft/common/nvtx.hpp
wphicks: @cjnolet I think given that the line info isn't really exposed in the reported NVTX info, I'd still prefer to avoid it (unless someone sees some other benefit that I'm not?).

wphicks: Also please feel free to just overrule me if there's consensus in favor of the macro. I'm trying to reduce our macro usage wherever possible, but this is a _great_ contribution with or without. So, let's get it in one way or another! :smile:

cpp/include/raft/common/nvtx.hpp
cjnolet: Thanks. I actually prefer snake case here and I believe there was a general consensus to use it across raft (Thejaswi had created a PR at once point to enforce this but it hasn't yet been merged).

cjnolet: That said, there are enough places in raft that are still using camel case but my preference is to start using snake case where possible in the public bits until we do our batch rename (after we establish the public API)

cpp/include/raft/common/nvtx.hpp
cjnolet: If the line info is never shown to the user, I do agree with Will here that I don’t see a point to using a macro over just invoking the function directly.

cjnolet: With that said, in practice, I'm also somewhat impartial here, so @achirkin, if you prefer to use the macro, I'm okay with keeping it in here.

cjnolet:
Also tagging some folks from cugraph for their thoughts about the macro

Cc @seunghwak @ChuckHastings @BradReesWork @afender

cpp/include/raft/common/nvtx.hpp
achirkin: Wow, I didn't really know that! sure, I'll change that to snake

cpp/include/raft/common/detail/nvtx.hpp
```diff
@@ -0,0 +1,198 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <rmm/cuda_stream_view.hpp>
+
+namespace raft::common::detail {
```
cjnolet, line 21: Let's use the raft::common::nvtx namespace (and raft::common::nvtx::detail) to better separate this package.

cjnolet (changes requested):
Changes look great, I just found one more very small thing as I was doing my final review.

achirkin:
As @teju85 suggested, I probably should also remove the versions of the helpers that synchronize cuda streams. The reason is that modern nsys profiler can map the ranges on gpu streams well, but addition of the lots of synchronize calls may bias the profiler/performance overview. @cjnolet, @wphicks do you have any objections to that?

cjnolet:
If this is the case with all the ctk versions supported by rapids (Eg ctk 11+) then it's fine with me. Probably better not to modify the performance while we are trying to measure the performance.

wphicks:
No objections here. Makes sense!

wphicks (approved):
Since no one else has a strong opinion on the macro issue, I say it's better to be biased toward approving and we should move ahead. Let's take care of the stream synchronization issue and Corey's last note on namespacing, then this looks good as far as I'm concerned.

achirkin:
All done, thanks for the input!

Note, I also added a template parameter `Domain`, which should allow an easy way to customize the NVTX domain message in downstream projects/applications.

One creates a domain by declaring a struct (anywhere):
https://github.com/rapidsai/raft/blob/d8357b2a9b0d739f9cb2837569f67c0a5fc65f03/cpp/include/raft/common/nvtx.hpp#L32-L34

And this is how to use it:
https://github.com/rapidsai/raft/blob/d8357b2a9b0d739f9cb2837569f67c0a5fc65f03/cpp/include/raft/linalg/svd.cuh#L67-L68

cjnolet (approved):
The changes look great, Artem. Thanks for moving this to raft and improving it!

cjnolet:
Just a little side note: the usage links are great but it would also be nice to have usage examples right in the doxygen.

achirkin:
rerun tests

cjnolet:
@gpucibot merge

PR #418, opened 2021-12-10
[DRAFT] Support knn search with 32-bit indices

zbjornson:
Supported by FAISS since 1.6.4.

### DIFF
--- a/cpp/include/raft/sparse/selection/detail/knn_graph.cuh
+++ b/cpp/include/raft/sparse/selection/detail/knn_graph.cuh
@@ -33,6 +33,7 @@
 #include <thrust/sort.h>
 
 #include <algorithm>
+#include <cstdint>
 #include <limits>
 
 namespace raft {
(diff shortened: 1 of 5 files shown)


cjnolet (changes requested):
@zbjornson PR looks good so far but this will need a PR opened in cuml just to make sure nothing broke. We're working to get to the point where we won't need to do this anymore, but unfortunately it's still necessary (as you've seen with recent changes). 

You only need to modify the raft fork and branch in `cpp/cmake/get_raft.cmake` in cuml.

zbjornson:
@cjnolet I'm hopefully a few minutes away from opening one, if I can figure out a weird linker error. :)

zbjornson:
@cjnolet does this need to be totally non-breaking (for cuML at least)? There's one implicit conversion in cuML that needs to be changed to work with this PR as-is. I think could add a specific int32 overload here to avoid that, but the implicit conversion in cuML is narrowing and should be fixed anyway. https://github.com/rapidsai/cuml/pull/4443/commits/7c397513a14d2a99f4eb90bbb267d40c8b237cb0

cjnolet:
For now, we want to provide the change in raft and fix cuml at the same time so that the raft change doesn't break cuml once merged. Then we will merge the raft side and update your cuml PR to point to the release branch of raft, then merge the cuml side. 

(Raft will eventually become more independent from its consumers but this is the process in the meantime)

cjnolet:
@zbjornson the KNN APIs should now be able to support 32-bit indices. Any reason you think we should keep this issue open or can it be closed now?

cjnolet:
I'm going to close this for now. Please feel free to open back up if this issue has not been addressed.

PR #420, opened 2021-12-13
[REVIEW] [FEA] error macros: determining buffer size instead of fixed 2048 chars

### LINKED ISSUE #419: [FEA] Handling error messages with more than 2048 characters
The error handling macros `THROW` and `SET_ERROR_MSG` use fixed-sized character buffers.
This can cause weird issues, especially since the length of the message may depend on things that are not trivial to control for the user of the macros, such as filenames.

Do not use any fixed-length buffer in error handling.

Even with a fixed-size buffer, we could simply check if the message fits, and throw an exception in case it doesn't or print something out. I don't really like this solution since it's easy enough to address the underlying issue.

MatthiasKohl:
Determining buffer size for error macros instead of using fixed size of 2048 bytes.
I ran all tests (which use these macros extensively) as well as tested without CUDA driver to see that output looks as expected.

Closes #419.


cjnolet (approved):
The changes look great. It would also be great if we could find a way to test this through our automated gtests.

MatthiasKohl:
That's a good point and no, unfortunately I didn't test with cuml. I can do that after vacation. Should I still add specific tests for these changes as well?

cjnolet:
@MatthiasKohl, 

A couple very trivial gtest cases would be great. The goal is to get to a point where we can modify RAFT without having to test whether all the clients break.

MatthiasKohl:
@cjnolet I did add specific tests for this, checking that the error messages are as expected.

BradReesWork (approved):

cjnolet:
@gpucibot merge

PR #428, opened 2021-12-20
Simplify raft component CMake logic, and allow compilation without FAISS

robertmaynard:
The previous logic didn't allow for developers to build raft without FAISS as `raft_nn` was always required to be built.


cpp/CMakeLists.txt
```diff
@@ -149,47 +135,67 @@ set(RAFT_LINK_LIBRARIES
...
 
 set(RAFT_COMPILE_DEFINITIONS
         $<$<BOOL:${NVTX}>:NVTX_ENABLED>
         )
-
 target_compile_definitions(raft INTERFACE ${RAFT_COMPILE_DEFINITIONS})
-target_compile_definitions(raft_distance PRIVATE ${RAFT_COMPILE_DEFINITIONS})
-target_compile_definitions(raft_nn PRIVATE ${RAFT_COMPILE_DEFINITIONS})
-
-target_compile_options(raft_distance
```
cjnolet, line 165: When I add these back, the final packages seem to compile successfully. Are these required in general?

cpp/CMakeLists.txt
```diff
@@ -106,39 +111,23 @@ include(cmake/thirdparty/get_rmm.cmake)
 include(cmake/thirdparty/get_libcudacxx.cmake)
 include(cmake/thirdparty/get_cuco.cmake)
 
-if(BUILD_TESTS)
+if(BUILD_TESTS OR RAFT_ENABLE_NN_COMPONENT)
```
cjnolet, line 114: WIth this PR, I notice FAISS is always building from source now, rather than linking to an existing shared lib if it exists.

cjnolet (approved):
LGTM. Tested locally

cjnolet:
rerun tests

cjnolet:
rerun tests

robertmaynard:
Thanks for fixing the build issues around the faiss name @cjnolet

cjnolet:
@robertmaynard, after discussing this w/ @rlratzel yesterday, I decided to make some additional changes [here](https://github.com/rapidsai/raft/pull/431). Please feel free to merge them into this PR as well. 

The changes allow for the individual components to pass their dependencies so consumers can pick them up downstream. By default, RAFT will continue to operate in "header-only" mode and will install all the headers but when `RAFT_ENABLE_NN_COMPONENT` is enabled, the faiss dependency (and include directories) will pass through. Eventually, we'd like to install the headers for individual packages within their component targets, rather than always installing them all. We have some work to do in the public API before we can fully separate the headers for individual components though.

divyegala:
@robertmaynard could we ensure that we can build `cumlprims_mg` with these changes? cc @dantegd @cjnolet

cpp/CMakeLists.txt
```diff
@@ -49,13 +49,19 @@ option(DISABLE_DEPRECATION_WARNINGS "Disable depreaction warnings " ON)
...
+
 message(VERBOSE "RAFT: Build RAFT unit-tests: ${BUILD_TESTS}")
 message(VERBOSE "RAFT: Enable detection of conda environment for dependencies: ${DETECT_CONDA_ENV}")
 message(VERBOSE "RAFT: Disable depreaction warnings " ${DISABLE_DEPRECATION_WARNINGS})
 message(VERBOSE "RAFT: Disable OpenMP: ${DISABLE_OPENMP}")
 message(VERBOSE "RAFT: Enable kernel resource usage info: ${CUDA_ENABLE_KERNELINFO}")
 message(VERBOSE "RAFT: Enable lineinfo in nvcc: ${CUDA_ENABLE_LINEINFO}")
 message(VERBOSE "RAFT: Enable nvtx markers: ${NVTX}")
+message(VERBOSE "RAFT: Enable raft::nn target: ${RAFT_ENABLE_NN_COMPONENT}")
+message(VERBOSE "RAFT: Enable raft::distance target: ${RAFT_ENABLE_DISTANCE_COMPONENT}")
```
dantegd, line 64: This variables seem to be undefined? They don't appear in the log, and seem to not be defined anywhere as far as I can see?

cpp/CMakeLists.txt
```diff
@@ -100,96 +106,129 @@ endif()
 # add third party dependencies using CPM
 rapids_cpm_init()
 
-# thrust and libcudacxx need to be before cuco!
-include(cmake/thirdparty/get_thrust.cmake)
 include(cmake/thirdparty/get_rmm.cmake)
 include(cmake/thirdparty/get_libcudacxx.cmake)
 include(cmake/thirdparty/get_cuco.cmake)
+# CPMFindPackage(faiss) so it's available for raft consumers to use
+include(cmake/thirdparty/get_faiss.cmake)
```
dantegd, line 113: I'm a little confused (after being pinged about this by @divyegala ), but wouldn't moving this out of the conditional, make it so that every consumer of raft calls cpm find faiss? how does cmake know to not do the get_faiss.cmake logic when not linking to `nn` or `distance` components?

dantegd (changes requested):
Had a couple of questions

cpp/CMakeLists.txt
divyegala: For context, we saw this happen in a consumer library which was neither linking to `nn` or `distance` but started building FAISS by default

robertmaynard: You are correct when building from source currently `get_faiss` will be called.

CMake doesn't have any mechanisim to allow lazy/deferred find/cpm calls based on a targets usage. So we can't delay looking for `faiss` based on the usage of `raft::nn`.

Since `raft::nn` can be used without any compilation, and requires faiss we are in an ambigous situation.

What I have done is add a `RAFT_ENABLE_NN_DEPENDENCIES` option whose default value is the same as `RAFT_COMPILE_LIBRARIES`. CPM consumers of raft ( cuml/cugraph ) which don't want to use the `raft::nn` can set `RAFT_ENABLE_NN_DEPENDENCIES` to OFF in the CPM options to disable faiss searches.

cpp/CMakeLists.txt
robertmaynard: Removed, they are left over from a previous iteration.

cpp/CMakeLists.txt
divyegala: Can confirm this option works well! Thanks

cjnolet: @robertmaynard, you had mentioned (offline) that we should be able to do something like `find_package(raft COMPONENTS nn)` in order to bring in the FAISS dependency (cuml would use), while `find_package(raft)` won't search for FAISS at all (which cugraph would use). Is this still something we are able to do in order to simplify the dependency process for consuming libraries?

cjnolet: Perfect!

dantegd (approved):

cpp/CMakeLists.txt
```diff
@@ -49,6 +49,11 @@ option(DISABLE_DEPRECATION_WARNINGS "Disable depreaction warnings " ON)
 option(DISABLE_OPENMP "Disable OpenMP" OFF)
 option(NVTX "Enable nvtx markers" OFF)
 
+option(RAFT_COMPILE_LIBRARIES "Enable building raft shared library instantiations" ON)
+option(RAFT_ENABLE_NN_DEPENDENCIES "Search for raft::nn dependencies like faiss" ${RAFT_COMPILE_LIBRARIES})
+include(CMakeDependentOption)
+cmake_dependent_option(RAFT_USE_FAISS_STATIC "Build and statically link the FAISS library for nearest neighbors search on GPU" ON RAFT_COMPILE_LIBRARIES OFF)
```
cjnolet, line 55: I opened a [PR](https://github.com/rapidsai/cuml/pull/4476) w/ cuml to integrate the changes in this PR but it looks like it's trying to build FAISS from source every time now, rather than using the shared lib that's already installed w/ conda. 

I also tried removing the explicit call to `get_faiss` and linking  of the `FAISS::FAISS` targets in cuml but it didn't seem to make a difference.

cjnolet (changes requested):

cpp/CMakeLists.txt
robertmaynard: Your PR doesn't change the version of RAFT it is using to this PR.

cjnolet: Ah, good catch. I've been building locally w/ `CPM_raft_SOURCE`, though. Let me update the PR as well.

cjnolet:
Linking #432

cpp/CMakeLists.txt
cjnolet: [Here](https://gpuci.gpuopenanalytics.com/job/rapidsai/job/gpuci/job/cuml/job/prb/job/cuml-cpu-cuda-build-arm64/CUDA=11.5/518/console) it looks like it's building FAISS from source in CI in cuml.

robertmaynard: This PR changes the name of the searched package to `faiss` ( lowercase ) to follow modern CMake patterns, while your cuml PR keeps the existing behavior of using `FAISS`.

This for CPM means they are distinctly different projects. When we look at the build logs we see the following:
```
-- CPM: using local package FAISS@1.7.0
-- CPM: raft: adding package faiss@1.7.0 (bde7c0027191f29c9dadafe4f6e68ca0ee31fb30)
```

So we can see that `FAISS` was found locally, and `faiss` is going to be built from source. 
In general we wouldn't expect RAFT search again for `faiss` but given the mismatch on names it does so. 

As for why RAFT is building `faiss` from source, the cuml `get_raft.cmake` project file currently enables `CPM_DOWNLOAD_ALL` which forces raft and all dependencies to be built from source and not search for a local copy.

So what can you do?

1. Update cuml to search for `faiss` instead of `FAISS`
2. Don't have `get_raft` in cuml always trying to build `raft` from source. Since this will also stop the benifits of `raft::nn` and `raft::distance`

cjnolet: I've verified (at least locally) that changing the faiss target (and `GLOBAL_TARGETS` for faiss to `faiss::faiss`) works when both with and without `CPM_raft_SOURCE`. 

We're working towards item number 2. conda packaging will make that much easier to implement and maintain.

cpp/CMakeLists.txt
```diff
@@ -100,96 +105,136 @@ endif()
...
 
-target_include_directories(raft_distance PUBLIC "$<BUILD_INTERFACE:${RAFT_SOURCE_DIR}/include>"
+target_include_directories(raft INTERFACE
+        "$<BUILD_INTERFACE:${RAFT_SOURCE_DIR}/include>"
         "$<INSTALL_INTERFACE:include>")
 
-target_include_directories(raft_nn PUBLIC "$<BUILD_INTERFACE:${RAFT_SOURCE_DIR}/include>"
-        "$<INSTALL_INTERFACE:include>")
+target_link_libraries(raft INTERFACE
+  CUDA::cublas
```
trivialfis, line 130: I think some of these libraries might be `PUBLIC` instead of `INTERFACE`.

robertmaynard: The `raft` target is an INTERFACE library and therefore has no compilation/source files. This means that it can only have `INTERFACE` properties as only the consumers of `raft` need these dependencies.

trivialfis: Got it, thanks for the explanation.

ajschmidt8 (approved):

cpp/CMakeLists.txt
```diff
@@ -100,96 +105,141 @@ endif()
 # add third party dependencies using CPM
 rapids_cpm_init()
 
-# thrust and libcudacxx need to be before cuco!
-include(cmake/thirdparty/get_thrust.cmake)
+# cuco before rmm to properly configure thrust/cub
```
robertmaynard, line 108: @cjnolet We should restore the call to `include(cmake/thirdparty/get_thrust.cmake)` as the first line before cuco or rmm.

If we have cuco as the first entry it will bring in thrust 1.12 which we don't want. By having our `thrust` first that will fetch 1.15 and will stop cuco from fetching 1.12

cjnolet: @robertmaynard, I need some help here then. 

Without this change, I get a thrust error when building cuml (locally) about the incorrect cub verison being picked up (1.12). With this change, cuml builds (locally) successfully and I believe it's pulling in thrust 1.15 because I would have expected to get a cub build error in HDBSCAN if it were not.

cjnolet: @trxcllnt also opened a PR to cuco to update the thrust version to 1.15, which we're using in this PR: https://github.com/NVIDIA/cuCollections/pull/134

robertmaynard: yes for rmm it uses the version specified by `versions.json` in rapids-cmake

cpp/cmake/versions.json
```diff
@@ -0,0 +1,9 @@
+{
```
robertmaynard, line 1: I am not seeing a package_override load of this file, so it will do nothing.

cjnolet:
@gpucibot merge

cjnolet:
@gpucibot merge

PR #434, opened 2021-12-22
[REVIEW] Improvents to RNG

vinaydes:
With this change proposing number of improvements to the Random Number Generator (RNG) class of RAFT

- Remove inferior TAPS and KISS99 generators
- Fix and optimize bounded integer generation
- Fix and optimize uniform floating point numbers (float/double)
- Add faster PCG based generator
- Provide device API for generator with all supported probability distributions


GPUtester:

cjnolet:
add to allowlist

teju85:
Not a comment, but a question @vinaydes 
I have a case where I want to generate uniformly distributed bit-vectors, which are stored in an array of `uint32_t`. Can I use the `next_u32` methods found in the `rng_impl.cuh` to accomplish this?

vinaydes:
@teju85 If you want each bit to have 50% chance of being either 1 or 0, then `next_u32` (or overloaded next with uint32_t data type) is correct.

teju85:
Yes, that's correct. I'd like to have them 1/0 with 50% probability. Thanks for the confirmation.

MatthiasKohl (commented):
General comments from me:

cpp/test/random/rng.cu:
1. The comments mention sigma being multiplied by 4 (line 101) or 3-4 (line 185), but then it actually seems to multiply by 10 (line 102 and 244-245, 289-290). Is this because we actually have the variance instead of standard deviation? But in this case, why is it done with both mean and variance values?

cpp/include/raft/random/rng.hpp:
1. I believe all members of the main `Rng` class are public at the moment. I think it would make sense to make them protected. However, I also think like we should allow users to manually set `offset` and `seed`, as well as provide a simple helper function to update the `offset` as needed: the arguments would be the total number of RNG calls and the type of RNG call (some will use the underlying generator twice or more). As of now, what the PR does is to simply reset the seed using the mt19937 generator. This is doable but not ideal since generators on GPU should kind of mimic CPU generators where there is a difference between offsetting (using the same sequence but at a later point in the period) and re-seeding (starting an entirely new sequence).

cpp/include/raft/random/rng_impl.cuh:
1. I see that you're using the "lemire" method for bounded integer generation. Does it make sense to add the swift method here? It seems like it performs well, especially with PCG, although I don't mind keeping both methods around since it may depend on exact use cases.

MatthiasKohl:
@vinaydes I'm currently changing the usage of RNG in cugraph-ops from the (previous) internal one to this new interface.
I'm still unsure how this is actually supposed to be used.

So let's say I have a kernel which should use the device interface:
```
#include <raft/random/rng.hpp>
__global__ void kernel(uint64_t seed, uint64_t subsequence, uint64_t offset) {
   auto gen = raft::random::PCGenerator(seed, subsequence + threadIdx.x, offset);
   gen.next(...);
}

kernel<<<...>>>(rng.seed, rng._base_subsequence, rng.offset);
rng.advance(...);
```

First of all, is this how one is supposed to use the new interface for device-side generation?

If this is the case, I have several issues with it:
1. All kernels wanting to generate random numbers now have to have 3 additional arguments, and the `offset` isn't even used.
2. When launching the kernel, I need to access the `_base_subsequence` : why name it this way, it makes it look as if this was a private member ?
3. I still have to know that I need to add any offset for a given call to `subsequence` within the kernel. This means that the whole `advance` method seems a bit useless since I have to know how it actually works internally.

I don't want to delay this PR too much again, but to me, this just doesn't feel right and I also don't want to refactor cugraph-ops entirely again just because we'll change this interface in a month in another PR.

I propose the following functions to simplify working with the new Rng class in the setting above.

```
/* allows to pass state to kernel in a single argument */
struct RngState {
  uint64_t seed, subsequence, offset /* possibly remove offset here to avoid passing unnecessary stuff */;
};

/* called when launching kernel, should be a method on Rng IMO */
RngState get_rng_state(const Rng& r) {
  return RngState{r.seed, r._base_subsequence, r.offset /* possibly not needed*/ };
}

/* called within kernel, aware of the fact that advance actually works with subsequence / _base_subsequence behind the scenes */
template <typename GenT>
__device__ __force_inline__ GenT get_generator(const RngState& state, uint64_t offset) {
  return GenT(state.seed, state.subsequence + offset, state.offset /* use 0 here if offset doesn't exist in state */);
}
```

Does this make sense to you?

vinaydes:
Thanks for the comments Matthias.
I will remove the prefix underscore from `_base_sequence`. It is mistakenly leftover.
In the interface/API you are proposing, there would still be `advance` called because somewhere the state needs to be modified. Or are you proposing `offset` argument to `get_generator` as the only variable holding state across the kernels?

MatthiasKohl:
I edited the comment above, the `offset` in `get_generator` is the offset of a given thread (renamed it `thread_offset`) on top of the "base" offset / subsequence which distinguishes that thread from other threads.
In some cases, we want the same offset for different threads (let's say we want to generate the same number based on input data or an entire warp/block should generate the same numbers).

One of the reasons I'm proposing this `get_generator` function is precisely because all the different offsets / subsequences are so confusing. Plus, the fact that we add this `thread_offset` to the `subsequence` / `base_subsequence` rather than the `offset` is something that should be hidden from users of the API: they should simply call `get_generator` (on device) and `advance` (on host) with the right arguments, and should get something that makes sense.

teju85:
What if we did something like this?

```c++
struct rng_state {
  uint64_t seed, subsequence, offset;
};

struct PhiloxGenerator {
  /**
  * @param[in] seq_id  represents a unique stream of random numbers. Eg: if seq_id == blockIdx.x, then all threads in the block will generate the same numbers everytime, across all threads
  */
  DI PhiloxGenerator(const rng_state& s, uint64_t seq_id);
};

struct PCGenerator {
  DI PCGenerator(const rng_state& s, uint64_t seq_id);
};

struct Rng {
  rng_state& get_state() { return state; }
  const rng_state& get_state() const { return state; }
  void advance(uint64_t num_uniq_generators_used) { get_state().subsequence += num_uniq_generators_used; }
  // rest of Rng class...
 private:
  rng_state state;
};
```

Now, if someone wants to use the generate random numbers inside a kernel...
```c++
__global__ void kernel(rng_state& state, ...) {
  // if you want to generate unique numbers for all threads
  PhiloxGenerator phg{state, threadIdx.x + blockIdx.x * blockDim.x};
  // if you want to generate same numbers for all threads in a block
  PhiloxGenerator phg{state, blockIdx.x};
}

void some_func(Rng& r, ...) {
  kernel<<<blks,threads>>>(r.get_state(), ...);
  r.advance(blks * threads);   // if every thread generated a unique stream of numbers
  r.advance(blks);   // if every block consumed a unique stream of numbers
}
```

MatthiasKohl:
I agree with @teju85 , but I wasn't sure whether we can change the interface of the constructors.
If we can, that would be a better option.

Another thing I wanted to mention here since I noticed it when porting over code: many of the `custom_next` functions don't use the `idx` and `stride` arguments at all.
In this case, I would write it like this:
```
template <typename GenType, typename OutType, typename LenType = int>
DI void custom_next(
  GenType& gen, OutType* val, InvariantDistParams<OutType> params, LenType /*idx*/ = 0, LenType /*stride*/ = 0)
{
  *val = params.const_val;
}
```
This allows calling it without having to provide dummy inputs for `idx` and `stride`

vinaydes:
I'll make these changes.

cpp/include/raft/random/rng.hpp
```diff
@@ -53,16 +63,16 @@ using detail::Kiss99Generator;
 template <typename Type>
 DI void box_muller_transform(Type& val1, Type& val2, Type sigma1, Type mu1, Type sigma2, Type mu2)
```
cjnolet, line 64: Will be nice to see these moved into a device-specific API (and maybe even a device namespace) once we get to https://github.com/rapidsai/raft/issues/443.

cpp/include/raft/random/rng.hpp
```diff
@@ -298,8 +301,8 @@ class Rng : public detail::RngImpl {
    * @param sigma the sigma
    * @param stream stream where to launch the kernel
    */
-  template <typename Type, typename LenType = int>
-  void rayleigh(Type* ptr, LenType len, Type sigma, cudaStream_t stream)
+  template <typename OutType, typename LenType = int>
```
cjnolet, line 304: Any particular reason for the name change?

cpp/test/random/rng.cu
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018-2020, NVIDIA CORPORATION.
+ * Copyright (c) 2018-2021, NVIDIA CORPORATION.
```
cjnolet, line 2: These need to be updated to 2022 now.

cpp/include/raft/random/rng.hpp
```diff
@@ -21,21 +21,31 @@
 namespace raft {
 namespace random {
```
cjnolet, line 22: Are we planning to work the `rng_state` class into this PR? If not, it would be worth capturing some of the follow-on changes and discussion in an issue just so they don't get lost.

cpp/include/raft/random/detail/rng_impl.cuh
```diff
@@ -37,10 +32,99 @@ namespace detail {
...
   GenPhilox = 0,
-  /** LFSR taps generator */
-  GenTaps,
-  /** kiss99 generator (currently the fastest) */
-  GenKiss99
+  /** Permuted Congruential Generator */
+  GenPC
+};
+
+template <typename OutType>
```
cjnolet, line 39: Since this is c++ code, we might want to consider adding some constructors to each of these to make it easier for users to construct these. Also, I don't believe the documentation propagates through the `using` clause, unfortunately, so it might be worth exposing these through an `rng_common.hpp` header which can be included both in the `detail` and public namespaces.

cjnolet (changes requested):
Changes look good. Since there's going to be follow-on changes which are likely to be API breaking, I'm mostly looking at the basics here. Before this is merged, a cuml PR needs to be opened and pinned to this branch so we can fix cuml in concert w/ this merge. 

We're working to make RAFT independent of its consumers and stabilize its public API for 22.04 so that we can be very explicit about API changes. If we want to get this merged quickly, we'll want to stabilize the public bits in the follow-on so that in the future we can deprecate any breaking API changes.

cpp/include/raft/random/rng.hpp
vinaydes: Just for consistency with other functions.

MatthiasKohl:
@vinaydes when further testing with this, I noticed a subtle but important change between philox and PCG:
for floating point numbers, it seems like your PCG-based implementation will produce numbers in `[0, 1[` whereas curand philox does `]0, 1]`.
I honestly don't care that much but I think it would be nice that it would be consistent with curand.
However, where it really matters is in the `box_muller_transform`: you simply pass the value to `raft::myLog` which means that if it was 0, then `raft::myLog` will give `-inf` which will then lead to infinite outputs.
This causes tests in cugraph-ops to fail.
It would be great to add an epsilon in box muller or to update the PCG-based implementation such that it produces floating point in `]0, 1]` which may be a bit hard though (I honestly don't know how to do this trivially).

cjnolet:
@vinaydes @MatthiasKohl I know we're looking to get this merged ASAP. Do you guys have an idea of how close we are to doing this? @vinaydes has this been tested against cuml yet to verify API/functional compatibility?

vinaydes:
> @vinaydes @MatthiasKohl I know we're looking to get this merged ASAP. Do you guys have an idea of how close we are to doing this? @vinaydes has this been tested against cuml yet to verify API/functional compatibility?

@cjnolet I am newrly done with cuML PR and all the other items discussed in conversations above. I'll push the changes and cuML PR soon (on Monday if not before)

vinaydes:
I have tried to summarize status of all the action items came up with regard to this PR. Some of the items have to deferred in order to get this PR merged sooner.

**From Matthias** 
1. Getter/setter methods for `Rng` class
	- **Not done**. We now have introduced `RngState`, do we still need getter/setter methods? Perhaps to `RngState`?
2. Introduce `RngState`
	- **Done**
3. Change the name for `_base_subsequence`
	- **Done**. Renamed to `subsequence` and is inside `RngState`
4. Improvements to advance method
	- **Done**. See `advance()` in `rng_impl.cuh`
5.  Add default values for `stride` and `idx`
	- **Done**
6. Address discrepancy between Philox and PCG `next_float()` call.  [Details](https://github.com/rapidsai/raft/pull/434#issuecomment-1021526827)
	- **WIP** I have modified the code of PCG `next_float` and `next_double` to produce float in range `(0, 1]`. However bunch of tests are failing after this change. Once I fix the issue, I'll push the changes
7. For `cpp/test/random/rng.cu`: The comments mention sigma being multiplied by 4 (line 101) or 3-4 (line 185), but then it actually seems to multiply by 10 (line 102 and 244-245, 289-290). Is this because we actually have the variance instead of standard deviation? But in this case, why is it done with both mean and variance values?
	- **Done/Deferred**. The test seems to work even with `num_sigma = 4`, I had to change the seed for some of the tests to make them work. This will require some more work as I am not sure how correctness is checked here therefore I have added it to deferred items list.
8.  Add swift method
	- **Deferred**

**From Corey**

9. Open corresponding cuML PR : https://github.com/rapidsai/cuml/pull/4562
	- **Done**
10. Are we planning to work the `rng_state` class into this PR? If not, it would be worth capturing some of the follow-on changes and discussion in an issue just so they don't get lost.
	- **Done**. `RngState` is introduced and deferred items are being tracked [here](https://github.com/rapidsai/raft/issues/493)
11. Update copyright year
	- **Done**
12. Add constructor for distribution parameter classes.
	- **Deferred**
13. Introduce `rng_common.hpp` for things that are common to `random` and `detail` namespace. [Details](https://github.com/rapidsai/raft/pull/434#discussion_r791100546)
	- **Deferred**
14. Move box Muller transform to device specific namespace
	- **Deferred**


**Bonus**
1. I came across a [bug](https://github.com/rapidsai/raft/issues/491) in eigen value/vector test, which should be fixed with this PR.

@MatthiasKohl @cjnolet Let me know if I have missed any point.

MatthiasKohl:
@vinaydes Thank you very much for the detailed overview of items above!

A few minor comments from my side, all of which can be deferred to another PR:
1. We can probably remove `offset` from the `RngState` struct. The point of the struct is that it allows to abstract away how exactly we access/create the Rng class in the device function. Right now, if we call something like `Rng::normal` many times in a row, `offset` will never be updated and always be `0`. Thus, I don't see a reason why we would should spend the additional constant/register memory and send the parameter from host to device. If someone wants to use all three parameters, they can still construct the device classes using the other constructor, like you do in `fillKernel`.
2. The constructors for the device classes using `RngState` should accept a thread-based offset. This is different to the `offset` in the `RngState` class because that thread-based offset must be added to the state in a way consistent with `advance` later on the host (see for example how you do it in `fillKernel`). Otherwise, users still have to know how the internals of the state work, i.e. adding the thread-based offset to subsequence, then updating the subsequence in `advance` by how many we had, and they must not forget about it!

teju85:
@cjnolet can we get a final review and merge this PR soon, please? cuML RF and cugraph-ops sampling are blocked on this.

cjnolet:
@teju85 this PR looks fine to go in but it is blocked by https://github.com/rapidsai/cuml/pull/4562 since it's a breaking change on the cuml side.

teju85:
Ah got it. Thanks Corey. That PR still seems to have build issues @vinaydes ?

vinaydes:
>Ah got it. Thanks Corey. That PR still seems to have build issues @vinaydes ?

Working to fix it. It is a tiny change.

cjnolet:
@vinaydes thw cuml build will be broken until this PR is merged: https://github.com/rapidsai/cuml/pull/4548. It might help to merge that into your branch RNG branch for now.

vinaydes:
Right, I saw the breaking changes and for now reverted merge of 22.04 locally. I'll use 4548 once I fix the build issue on older commit. Thanks.

cjnolet (approved):
LGTM. Just making a note to merge after the corresponding cuml side passes CI.

vinaydes:
@MatthiasKohl Addressed your comments with latest push. I am deferring the task of reconciliation of `next_float` / `next_double` since it will require some more work.

MatthiasKohl:
Thanks, looks good to me!

cjnolet:
@gpucibot merge

PR #436, opened 2021-12-22
Fix comms memory leak

### LINKED ISSUE #435: [BUG] Memory leak in comms
**Describe the bug**
comms_iface has no virtual destructor
(https://github.com/rapidsai/raft/blob/branch-22.02/cpp/include/raft/comms/comms.hpp#L91).
This means std_comms or mpi_comms destructor will not be called when comms_t's destructor is called and `std::unique_ptr<comms_iface> impl_;` gets destructed (https://github.com/rapidsai/raft/blob/branch-22.02/cpp/include/raft/comms/comms.hpp#L608).

Also,

initialize_mpi_comms with the given `MPI_Comm comm` calls mpi_comms with `owns_mpi_comm = true`
(https://github.com/rapidsai/raft/blob/branch-22.02/cpp/include/raft/comms/mpi_comms.hpp#L442).
cuGraph calls `initialize_mpi_comms(&handle, MPI_COMM_WORLD)` and as `owns_mpi_comms` is set to true, mpi_comms destructor will call MPI_Comm_free. This is erroneous.

**Additional context**
`virtual ~comms_iface() {}` should be added to `class comms_iface`.
`initialize_mpi_comms` should call `mpi_comms` with `owns_mpi_comms = false` (and caller should be responsible for destroying `comm`).

seunghwak:
Close #435 

- [x] Add a virtual destructor to comms_iface (to invoke std_comms/mpi_comms destructors when std_comms/mpi_comms objects are deleted through a comms_iface pointer).
- [x] Set `owns_mpi_comms` to false in `initialize_mpi_comms` as this takes an already initialized `comm` as an input parameter so `comm` may better be destroyed by the caller (and no need to destroy if `comm` is MPI_COMM_WORLD, just calling MPI_Finalize will be sufficient).

### DIFF
--- a/cpp/include/raft/comms/comms.hpp
+++ b/cpp/include/raft/comms/comms.hpp
@@ -90,6 +90,8 @@ constexpr datatype_t get_type<double>()
 
 class comms_iface {
  public:
+  virtual ~comms_iface() {}
+
   virtual int get_size() const = 0;
   virtual int get_rank() const = 0;
 
--- a/cpp/include/raft/comms/mpi_comms.hpp
+++ b/cpp/include/raft/comms/mpi_comms.hpp
@@ -439,7 +439,7 @@ class mpi_comms : public comms_iface {
 inline void initialize_mpi_comms(handle_t* handle, MPI_Comm comm)
 {
   auto communicator =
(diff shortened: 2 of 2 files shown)


cjnolet (approved):
LGTM

ChuckHastings (approved):

cjnolet:
@gpucibot merge

PR #438, opened 2022-01-07
Public apis for remainder of matrix and stats

divyegala:
At first, when I created public APIs for `matrix` and `stats` packages, I let implementation be in the public API if it was a simple thrust call or a call to another RAFT API. However, after further discussion, it's cleaner consistency to have the implementation always be in `detail`. I also added missing docs in this PR.

### DIFF
--- a/cpp/include/raft/matrix/detail/math.cuh
+++ b/cpp/include/raft/matrix/detail/math.cuh
@@ -16,12 +16,351 @@
 
 #pragma once
 
+#include <raft/handle.hpp>
+
 #include <cub/cub.cuh>
 #include <raft/cuda_utils.cuh>
+#include <raft/linalg/binary_op.cuh>
+#include <raft/linalg/map_then_reduce.cuh>
+#include <raft/linalg/matrix_vector_op.cuh>
+#include <raft/linalg/unary_op.cuh>
+#include <rmm/device_scalar.hpp>
+#include <rmm/device_uvector.hpp>
 
 namespace raft {
 namespace matrix {
 namespace detail {
+
+template <typename math_t>
+void power(math_t* in, math_t* out, math_t scalar, int len, cudaStream_t stream)
+{
+  auto d_src  = in;
+  auto d_dest = out;
+
+  raft::linalg::binaryOp(
+    d_dest,
+    d_src,
+    d_src,
+    len,
+    [=] __device__(math_t a, math_t b) { return scalar * a * b; },
+    stream);
+}
+
+template <typename math_t>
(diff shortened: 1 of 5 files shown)


cjnolet (approved):
The changes look great! Ready to go, pending successful CI.

cjnolet (changes requested):
These seem like harmless changes and shouldn't affect the public API, however when I attempted to build cuml w/ the changes, I get the following compile error:

```
/share/workspace/rapids_projects/raft/cpp/include/raft/matrix/matrix.hpp: In instantiation of ‘m_t raft::matrix::getL2Norm(const raft::handle_t&, m_t*, idx_t, cudaStream_t) [with m_t = float; idx_t = int; cudaStream_t = CUstream_st*]’:
/share/workspace/rapids_projects/raft/cpp/include/raft/linalg/svd.cuh:366:33:   required from ‘bool raft::linalg::evaluateSVDByL2Norm(const raft::handle_t&, math_t*, math_t*, math_t*, math_t*, int, int, int, math_t, cudaStream_t) [with math_t = float; cudaStream_t = CUstream_st*]’
/share/workspace/rapids_projects/cuml/cpp/test/prims/rsvd.cu:269:289:   required from here
/share/workspace/rapids_projects/raft/cpp/include/raft/matrix/matrix.hpp:224:1: error: no return statement in function returning non-void [-Werror=return-type]
  224 | }
      | ^
/share/workspace/rapids_projects/raft/cpp/include/raft/matrix/matrix.hpp: In instantiation of ‘m_t raft::matrix::getL2Norm(const raft::handle_t&, m_t*, idx_t, cudaStream_t) [with m_t = double; idx_t = int; cudaStream_t = CUstream_st*]’:
/share/workspace/rapids_projects/raft/cpp/include/raft/linalg/svd.cuh:366:33:   required from ‘bool raft::linalg::evaluateSVDByL2Norm(const raft::handle_t&, math_t*, math_t*, math_t*, math_t*, int, int, int, math_t, cudaStream_t) [with math_t = double; cudaStream_t = CUstream_st*]’
/share/workspace/rapids_projects/cuml/cpp/test/prims/rsvd.cu:286:289:   required from here
/share/workspace/rapids_projects/raft/cpp/include/raft/matrix/matrix.hpp:224:1: error: no return statement in function returning non-void [-Werror=return-type]
```

Can you open a PR to cuml to test these changes? (Hopefully we won't have to do this for too much longer).

cjnolet (approved):
LGTM

cjnolet:
I verified locally that cuml builds successfully w/ these changes

cjnolet:
@gpucibot merge

PR #442, opened 2022-01-13
Updates to Python and C++ Docs

### LINKED ISSUE #2: Start doc for RAFT "principals"
Just a simple outline of principals we plan to follow in RAFT to help the ML and graph teams coordinate. 

- header only
- comms
- priorities for different versions (eg. what will be moved and when)
- abstraction paradigms (eg. resource management, etc...)

cjnolet:
As we're beginning to hide the implementation details and establisht he public API, it's time to start putting a little focus on the presentation of our docs. Eventually, we want to make sure our different packages are easy to navigate and find. We will be updating this more as we continue to stabilize the public API. 

The docs can be built by running `./build.sh docs` from the repository root. 

Closes #2


cpp/doxygen/main_page.md
```diff
@@ -0,0 +1,13 @@
+# libraft
+
+RAFT is an analytics engine, containing building-blocks for rapid composition of RAPIDS Analytics. These building-blocks include shared representations, mathematical computational primitives, and utilities that accelerate building analytics and data science algorithms in the RAPIDS ecosystem. Both the C++ and Python components can be included in consuming libraries, providing building-blocks for both dense and sparse matrix formats in the following general categories:
```
divyegala, line 3: Can we maybe use a different term on the third "building-blocks"? Feels like a little repetitive term, we could use "representations"?

divyegala (approved):
Just one small comment, looks great otherwise

cpp/doxygen/main_page.md
JohnZed: Maybe not referring to it as an "engine" which sounds more like a server or executable to me? Library, framework, collection, building blocks, those are all great terms.

ajschmidt8 (commented):
If these docs will eventually get published on our site, we'll need to add the `breathe` dependency to the `rapids-doc-env` below. @cjnolet, can you open a PR for that?

- https://github.com/rapidsai/integration/blob/3986715992ec3087f7e38e8b3dbf69035366874c/conda/recipes/rapids-doc-env/meta.yaml#L36

cjnolet:
Absolutely, @ajschmidt8, I will do that.

ajschmidt8 (approved):
Approving `ops-codeowner` changes now that https://github.com/rapidsai/integration/pull/415 is merged.

BradReesWork (approved):

cjnolet:
@gpucibot merge

PR #446, opened 2022-01-18
Remove libcudacxx patch needed for nvcc 11.4

robertmaynard:
The `libcudacxx.patch` was required to fix issues with libcudacxx 1.6 and incorrect detection of the arm nvcc 11.4 compiler.

As rapids-cmake has moved to libcudacxx 1.7 this patch is not needed, and should be removed.


cjnolet:
@gpucibot merge

PR #449, opened 2022-01-19
Move `random` package of cuML to RAFT

divyegala:
This includes:

- [ ] make_arima
- [x] make_blobs
- [x] make_regression
- [x] multi_variable_gaussian
- [x] permute


cpp/include/raft/random/detail/curand_wrappers.hpp
```diff
@@ -0,0 +1,55 @@
...
+    ASSERT(status == CURAND_STATUS_SUCCESS, "FAIL: curand-call='%s'. Reason:%d\n", #call, status); \
+  } while (0)
+
+/**
+ * @defgroup normal curand normal random number generation operations
+ * @{
+ */
+template <typename T>
+curandStatus_t curandGenerateNormal(
+  curandGenerator_t generator, T* outputPtr, size_t n, T mean, T stddev);
```
cjnolet, line 37: While we're at this, we should consider adding a TODO here to replace any consumers of this function w/ our new RNG API.

cpp/include/raft/random/detail/make_blobs.cuh
```diff
@@ -0,0 +1,179 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <raft/cuda_utils.cuh>
+#include <raft/cudart_utils.h>
+#include <raft/linalg/unary_op.cuh>
+#include <raft/random/permute.hpp>
```
cjnolet, line 22: I'm wondering if there's any includes we might be able to remove here just for light cleanup.

cjnolet (changes requested):
Looking great so far. Just providing some early feedback. I can review more deeply if you'd like after the cuml side is ready.

cjnolet (approved):

cjnolet:
@gpucibot merge

PR #451, opened 2022-01-19
Replace RMM CUDA Python bindings with those provided  by CUDA-Python

shwina:
As a follow up to https://github.com/rapidsai/rmm/pull/930, fix RAFT to rely on CUDA Python directly rather than custom  CUDA bindings that RMM provided.

### DIFF
--- a/python/raft/common/cuda.pyx
+++ b/python/raft/common/cuda.pyx
@@ -19,7 +19,7 @@
 # cython: embedsignature = True
 # cython: language_level = 3
 
-from cuda.ccudart cimport (
+from cuda.ccudart cimport(
     cudaStream_t,
     cudaError_t,
     cudaSuccess,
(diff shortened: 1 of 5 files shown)


python/raft/common/cuda.pyx
```diff
@@ -47,26 +58,22 @@ cdef class Stream:
...
+    # If we store cudaStream_t directly, this always leads to the following error:
+    #   "Cannot convert Python object to 'cudaStream_t'"
     # I was unable to find a good solution to this in reasonable time. Also,
     # since cudaStream_t is a pointer anyways, storing it as an integer should
     # be just fine (although, that certainly is ugly and hacky!).
-    cdef size_t s
 
     def __cinit__(self):
-        if self.s != 0:
-            return
```
shwina, line 59: It's not trivial to initialize a `Stream` outside of this constructor. Is handling this case critical for something?

cjnolet: If we aren't accepting a stream directly in the constructor, I can't see why this would be needed.

cjnolet (approved):
LGTM

cjnolet:
@gpucibot merge

Ethyling (approved):

PR #472, opened 2022-02-01
single-pass raft::stats::meanvar

achirkin:
Add a helper to compute column-wise mean and variance for a matrix in a single pass.

Implementation and tests are to a large extent copy-pasted from `raft::stats::mean` and `raft::stats::stddev`.
It is to be used in cuml's `ML::GLM::preProcessData`, where it finishes ~1.8 faster compared to running mean+var separately (on column-major data).


cpp/include/raft/stats/meanvar.hpp
```diff
@@ -0,0 +1,53 @@
...
+#pragma once
+
+#include "detail/meanvar.cuh"
+
+namespace raft::stats {
+
+/**
+ * @brief Compute mean and variance for each column of a given matrix.
+ *
+ * The operation is performed in a single sweep.
```
cjnolet, line 26: To help users who may not immediately see the benefit, can you briefly mention the performance implication / expected speedup of fusing these two operations in a single kernel?

cjnolet (changes requested):
Looks great, just one very minor thing

cjnolet (approved):
LGTM

achirkin:
rerun tests

cjnolet:
rerun tests

cjnolet:
@gpucibot merge

PR #473, opened 2022-02-01
Adding destructor for std comms and using nccl allreduce for barrier in mpi comms

### LINKED ISSUE #281: [BUG] std comms to clean up contents of requests_in_flight upon failure
Currently, these are only cleaned up from assumed successful execution. Since failure is always catastrophic, we have not noticed any issues from this before but it would be correct to clean up the requests in ‘requests_in_flight’

cc @robertmaynard

cjnolet:
Closes #281


cpp/include/raft/comms/detail/mpi_comms.hpp
```diff
@@ -104,8 +105,14 @@ constexpr MPI_Op get_mpi_op(const op_t op)
 
 class mpi_comms : public comms_iface {
  public:
-  mpi_comms(MPI_Comm comm, const bool owns_mpi_comm)
-    : owns_mpi_comm_(owns_mpi_comm), mpi_comm_(comm), size_(0), rank_(1), next_request_id_(0)
+  mpi_comms(MPI_Comm comm, const bool owns_mpi_comm, cudaStream_t stream)
```
seunghwak, line 108: Shouldn't we better use rmm::cuda_stream_view instead of cuda_stream_t?

We're using rmm::cuda_stream_view elsewhere (https://github.com/rapidsai/raft/blob/branch-22.04/cpp/include/raft/handle.hpp#L137) and better be consistent unless there is a strong reason to mix the two.

cpp/include/raft/comms/detail/mpi_comms.hpp
```diff
@@ -119,8 +126,11 @@ class mpi_comms : public comms_iface {
 
     // initializing NCCL
     RAFT_NCCL_TRY(ncclCommInitRank(&nccl_comm_, size_, id, rank_));
+    initialize();
   }
 
+  void initialize() { sendbuff_ = status_.data(); }
```
seunghwak, line 132: May better set status_ to 0 to avoid status_ holding garbage. Unlikely to affect any program behavior, but I guess this is a safer practice (especially considering that performance overhead is negligible).

And if we use rmm::device_scalar instead of rmm::device_uvector, we can simply call set_value_to_zero_async (https://github.com/rapidsai/rmm/blob/branch-22.04/include/rmm/device_scalar.hpp#L215).

cpp/include/raft/comms/detail/mpi_comms.hpp
```diff
@@ -139,7 +149,15 @@ class mpi_comms : public comms_iface {
     return std::unique_ptr<comms_iface>(new mpi_comms(new_comm, true));
   }
 
-  void barrier() const { RAFT_MPI_TRY(MPI_Barrier(mpi_comm_)); }
+  void barrier() const
+  {
+    RAFT_CUDA_TRY(cudaMemsetAsync(sendbuff_, 1, sizeof(int), stream_));
```
seunghwak, line 154: As we're not using allreduce output, this line is unnecessary.

cpp/include/raft/comms/detail/mpi_comms.hpp
```diff
@@ -428,6 +446,10 @@ class mpi_comms : public comms_iface {
   bool owns_mpi_comm_;
   MPI_Comm mpi_comm_;
 
+  cudaStream_t stream_;
+  int* sendbuff_;
```
seunghwak, line 450: I slightly prefer int32_t over int (as we use datatype_t::INT32).

cpp/include/raft/comms/detail/std_comms.hpp
```diff
@@ -103,12 +103,14 @@ class std_comms : public comms_iface {
...
-  void initialize()
+  ~std_comms()
   {
-    sendbuff_ = status_.data();
-    recvbuff_ = status_.data() + 1;
+    requests_in_flight_.clear();
+    free_requests_.clear();
   }
 
+  void initialize() { sendbuff_ = status_.data(); }
```
seunghwak, line 112: May better set status_ to 0 to avoid status_ holding garbage. Unlikely to affect any program behavior, but I guess this is a safer practice (especially considering that performance overhead is negligible).

And if we use rmm::device_scalar instead of rmm::device_uvector, we can simply call set_value_to_zero_async (https://github.com/rapidsai/rmm/blob/branch-22.04/include/rmm/device_scalar.hpp#L215).

cpp/include/raft/comms/detail/std_comms.hpp
```diff
@@ -180,9 +182,8 @@ class std_comms : public comms_iface {
   void barrier() const
   {
     RAFT_CUDA_TRY(cudaMemsetAsync(sendbuff_, 1, sizeof(int), stream_));
```
seunghwak, line 184: As we're not using allreduce output, this line is unnecessary.

cpp/include/raft/comms/detail/mpi_comms.hpp
seunghwak: And do we really need `sendbuf_`? Just use `status_.data()` instead?

cpp/include/raft/comms/detail/std_comms.hpp
```diff
@@ -536,7 +537,7 @@ class std_comms : public comms_iface {
   ncclComm_t nccl_comm_;
   cudaStream_t stream_;
 
-  int *sendbuff_, *recvbuff_;
+  int* sendbuff_;
```
seunghwak, line 540: I slightly prefer int32_t over int (as we use datatype_t::INT32).
And do we really need sendbuf_? Just use status_.data() instead?

cpp/include/raft/comms/mpi_comms.hpp
```diff
@@ -24,10 +24,10 @@ namespace comms {
 
 using mpi_comms = detail::mpi_comms;
 
-inline void initialize_mpi_comms(handle_t* handle, MPI_Comm comm)
+inline void initialize_mpi_comms(handle_t* handle, MPI_Comm comm, cudaStream_t stream)
```
seunghwak, line 27: Shouldn't we better use rmm::cuda_stream_view instead of cuda_stream_t?

We're using rmm::cuda_stream_view elsewhere (https://github.com/rapidsai/raft/blob/branch-22.04/cpp/include/raft/handle.hpp#L137) and better be consistent unless there is a strong reason to mix the two.

seunghwak (approved):
I have several suggestions about cosmetic issues, but this looks great to me in general.

Approving to not be a bottleneck (and all my complaints really don't affect program behavior).

cjnolet:
rerun tests

cjnolet:
@gpucibot merge

PR #476, opened 2022-02-01
[FIX] Define PTDS via `-D` to fix cache misses in sccache

trxcllnt:
Define PTDS via `-DCUDA_API_PER_THREAD_DEFAULT_STREAM` because sccache doesn't recognize the `--default-stream per-thread` arg yet.


ajschmidt8 (approved):

cjnolet (approved):

robertmaynard:
@trxcllnt Can you file a bug on sccache for this flag, so we know when it is safe to switch back to it

ajschmidt8:
@gpucibot merge

trxcllnt:
https://github.com/mozilla/sccache/issues/1105

PR #483, opened 2022-02-02
Add CMake `install` rule for tests

ajschmidt8:
This PR adds a CMake `install` rule for test targets. This step is a prerequisite to being able to package tests in their own `conda` package, which will enable us to deprecate _Project Flash_.


robertmaynard (approved):

ajschmidt8:
rerun tests

cjnolet:
@gpucibot merge

PR #494, opened 2022-02-09
Enabling ccache in cmake

cjnolet:


robertmaynard:
I prefer the approach where we expect the user to have specified this information via one of the following approaches:
-  via the `cmake` command line
- by having the ccache compiler shims first on the path ( aka so that `g++` is `ccache` )
- Having the equivalantly named enviornment variables specified

These approaches allow each project to not care about ccache / sccache and avoid developers needing to know project specific settings.

cjnolet:
@robertmaynard, for your first bullet point, are you suggesting we just set this in the `cmake` command-line?

```
 -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_CUDA_COMPILER_LAUNCHER=ccache
```

PR #508, opened 2022-02-11
Adjusting conda packaging to remove duplicate dependencies

cjnolet:


conda/recipes/libraft_distance/meta.yaml
```diff
@@ -45,12 +45,15 @@ requirements:
     - ucx-proc=*=gpu
     - gtest=1.10.0
     - gmock
+    - rmm {{ minor_version }}
```
ajschmidt8, line 48: I'm not very familiar with `raft`, but does `rmm` need to be listed under the `host` requirements here? My understanding is that the `host` requirements are used during the building/linking process. Since `rmm` is a Python package, is it necessary for the `raft` build process? Should this be changed to `librmm`? Or alternatively is it extraneous and therefore be removed entirely? This question applies to all the recipes in this PR that have `rmm` in the `host` section

cjnolet: @ajschmidt8 since librmm is actually a build dependency specified in RAFT's cmake, it's not needed as a host dependency, but I've added it anyways just to make it intentional and explicit. 

The reason librmm needs to be specified as a runtime dependency is because the conda packages currently include the rmm headers, but they will remove the headers if they are already part of another conda package (Eg librmm). This is the case with the thrust and cuco headers as well but we currently don't have any other conda packages for them so we kind of have to include them in RAFT for the time being.

Ethyling (approved):

cjnolet:
@gpucibot merge

Issue #493, opened 2022-02-08
[FEA] List of work items deferred from RNG improvements PR 434


vinaydes:
Following is the list of work items that could not be included in the RNG improvement PR #434:
- Revisit  `cpp/test/random/rng.cu`: [Details](https://github.com/rapidsai/raft/pull/434#pullrequestreview-842938169).
- Add Swift method of bounded random integer generation.
- Move Box Muller transform to device specific namespace.
- Add constructors for distribution parameter classes, such as `UniformDistParams`, `NormalIntDistParams`.

vinaydes:
Item "1. Revisit ...", addressed in PR #513

PR #513, opened 2022-02-17
[WIP] RNG test fixes and improvements

vinaydes:
- Fixes the failures of RNG tests and issues mentioned [here](https://github.com/rapidsai/raft/issues/493).
- Reference standard deviation calculations: https://gist.github.com/vinaydes/cee04f50ff7e3365759603d39b7e079b
- Additionally fixes issue of Rng throwing NaNs and Infs
- Need to add parameter validation for each distribution

@MatthiasKohl @teju85 Please take a look.

### DIFF
--- a/cpp/include/raft/random/detail/rng_impl.cuh
+++ b/cpp/include/raft/random/detail/rng_impl.cuh
@@ -487,7 +487,7 @@ struct PhiloxGenerator {
   {
     float ret;
     uint32_t val = next_u32() >> 8;
-    ret          = static_cast<float>(val) / (1U << 24);
+    ret          = static_cast<float>(val) / float(uint32_t(1) << 24);
     return ret;
   }
 
@@ -495,7 +495,7 @@ struct PhiloxGenerator {
   {
     double ret;
     uint64_t val = next_u64() >> 11;
-    ret          = static_cast<double>(val) / (1LU << 53);
+    ret          = static_cast<double>(val) / double(uint64_t(1) << 53);
     return ret;
   }
 
--- a/cpp/include/raft/random/detail/rng_impl.cuh
+++ b/cpp/include/raft/random/detail/rng_impl.cuh
@@ -394,6 +394,9 @@ DI void custom_next(GenType& gen,
   constexpr OutType one     = (OutType)1.0;
   constexpr OutType two     = (OutType)2.0;
   constexpr OutType oneHalf = (OutType)0.5;
+
+  // The <= comparison here means, number of samples going in `if` branch are more by 1 than `else`
+  // branch. However it does not matter as for 0.5 both branches evaluate to same result.
   if (res <= oneHalf) {
     out = params.mu + params.scale * raft::myLog(two * res);
   } else {
--- a/cpp/include/raft/random/detail/rng_impl.cuh
+++ b/cpp/include/raft/random/detail/rng_impl.cuh
@@ -496,12 +496,14 @@ struct PhiloxGenerator {
     return ret;
   }
 
-  DI void next(float& ret) {
-    //ret = curand_uniform(&(this->philox_state));
+  DI void next(float& ret)
+  {
+    // ret = curand_uniform(&(this->philox_state));
     ret = next_float();
   }
 
-  DI void next(double& ret) {
+  DI void next(double& ret)
+  {
     // ret = curand_uniform_double(&(this->philox_state));
     ret = next_double();
   }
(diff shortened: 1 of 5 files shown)


vinaydes:
I'll take a look at failing test `Rng.MeanError`.

cpp/include/raft/random/detail/rng_impl.cuh
```diff
@@ -294,7 +294,11 @@ DI void custom_next(
   GenType& gen, OutType* val, GumbelDistParams<OutType> params, LenType idx = 0, LenType stride = 0)
 {
   OutType res = 0;
-  gen.next(res);
+
+  do {
+    gen.next(res);
+  } while (res == OutType(1.0));
```
MatthiasKohl, line 300: I'm surprised we're not checking for `0.0` here whereas in Laplace, we are checking for that case.
If `0.0` is a possible output of the RNG, then clearly it will lead to `Inf` here in the first `log`, right?

vinaydes: I ran the code several times with randomized seeds and Gumbel never failed for `0.0` value. I think I'll put the check anyway.

MatthiasKohl: I mean the question really is whether `0.0` is a possible output of the RNG.
If it is, then for a strictly positive `beta`, the final value should be `-Inf`.
If `0.0` is not a possible output of the RNG then we shouldn't do the redundant check, but we shouldn't do it in Laplace either IMO.

MatthiasKohl: Otherwise, you can simply test with offsets going from `0 .. 2^25` which should guarantee that you hit each output number once (unless the RNG is biased), and see what you get in each case.
I think it's important that we do this at least once with both philox and PCG just to check which output range we are actually getting (regardless of what we think we should be getting).

vinaydes: As per design `0,0` should be possible with PCG at least. I'll do the exhaustive search as you suggested and confirm the behavior.

vinaydes:
Summary of latest 3 commits:
1. Make uniform generation same for both PCG and cuRAND - `[0.0, 1.0)`
2. Change seed of GEMM unit test since it was generating too small floating point values to compare correctly
3. Fix Inf/NaN occurrences by adding 0.0 checks at multiple places

MatthiasKohl (commented):
I went over all distributions again (not just the ones you changed). Things I noticed:
1. For Bernoulli / ScaledBernoulli, I'm almost sure the calculation is simply wrong: the probability parameter usually represents chance of success, but with `res > params.prob` this would mean that for a probability of 40% we get success 60% of the time. So I guess here probability represents non-success. But, if `params.prob` is 0, we can still get non-success if `res == 0`. So I'm almost certain that it should actually be `res < params.prob`: For `params.prob == 1`, we always get success, for `params.prob == 0`, we always get non-success and everything in-between should work fine.
2. In Laplace, there is something similar to the above, but it is correct in the code: I would still add a comment that since `res` is in `[1, 2^24 - 1] / 2^24`, we will have `2^23` numbers in the first branch and `2^23 - 1` numbers in the second one. The special value `2^23 / 2^24 == oneHalf` would give the same result anyway in both branches, so it doesn't matter where it goes. This is probably not obvious at first sight for someone going over this code.

Apart from that, everything else looks good to me.

cpp/include/raft/random/detail/rng_impl.cuh
```diff
@@ -451,8 +480,33 @@ struct PhiloxGenerator {
     return ret;
   }
 
-  DI void next(float& ret) { ret = curand_uniform(&(this->philox_state)); }
-  DI void next(double& ret) { ret = curand_uniform_double(&(this->philox_state)); }
+  DI float next_float()
+  {
+    float ret;
+    uint32_t val = next_u32() >> 8;
+    ret          = static_cast<float>(val) / (1U << 24);
```
MatthiasKohl, line 487: Minor nitpick: there is an implicit conversion from `unsigned int` to `float` here, which I would make explicit: `static_cast<float>(1U << 24)` or `float(1U << 24)`.
This is similar in the double case. Additionally, in double I would use `long long unsigned` since `long` is not guaranteed to have 64 bits in all C++ implementations: `static_cast<double>(1LLU << 53)`

vinaydes: Good catch, I'll make these changes. Let's use `uint64_t` to be absolutely sure.

MatthiasKohl (approved):
Changes LGTM from my side

cjnolet (approved):
LGTM

cjnolet:
@gpucibot merge

PR #514, opened 2022-02-17
Unify weighted mean code

lowener:
Needed for https://github.com/rapidsai/cuml/pull/4428


cpp/include/raft/stats/detail/weighted_mean.cuh
```diff
@@ -18,14 +18,17 @@
 
 #include <raft/cudart_utils.h>
 #include <raft/linalg/coalesced_reduction.hpp>
+#include <raft/linalg/reduce.cuh>
```
cjnolet, line 21: These are going to have `hpp` extensions now.

cpp/include/raft/stats/weighted_mean.hpp
```diff
@@ -40,7 +41,35 @@ void rowWeightedMean(
...
+ * @param data the input matrix
+ * @param weights per-sample weight
+ * @param D number of columns of data
+ * @param N number of rows of data
+ * @param row_major input matrix is row-major or not
+ * @param along_rows whether to reduce along rows or columns
+ * @param stream cuda stream to launch work on
+ */
+template <typename Type>
+void rowSampleWeightedMean(Type* mu,
```
cjnolet, line 58: We need a gtest for this as well

cjnolet (changes requested):

lowener:
I reworked this PR to unify and clean the weighted mean code:
- I added `weightedMean` which is more general than the col and row functions.
- I fixed `colWeightedMean` documentation which was wrong
- I added a template for IndexType.
- I added a test for the cases not covered by `rowWeightedMean` and `colWeightedMean` (when the input is col-major)

cpp/test/stats/weighted_mean.cu
```diff
@@ -171,29 +172,86 @@ class ColWeightedMeanTest : public ::testing::TestWithParam<WeightedMeanInputs<T
   thrust::device_vector<T> din, dweights, dexp, dact;
 };
 
+template <typename T>
+class WeightedMeanTest : public ::testing::TestWithParam<WeightedMeanInputs<T>> {
+ protected:
+  void SetUp() override
+  {
+    params = ::testing::TestWithParam<WeightedMeanInputs<T>>::GetParam();
+    raft::random::Rng r(params.seed);
```
cjnolet, line 181: Just for future reference, we now have `make_blobs` within RAFT!

cpp/test/stats/weighted_mean.cu
```diff
@@ -171,29 +172,86 @@ class ColWeightedMeanTest : public ::testing::TestWithParam<WeightedMeanInputs<T
...
+    thrust::host_vector<T> hexp(mean_size);
+
+    // compute naive result & copy to GPU
+    if (params.along_rows)
+      naiveRowWeightedMean(hexp.data(), hin.data(), hweights.data(), rows, cols, false);
+    else
+      naiveColWeightedMean(hexp.data(), hin.data(), hweights.data(), rows, cols, false);
+    dexp = hexp;
+
+    // compute ml-prims result
```
cjnolet, line 209: We should remove references to ml-prims since it's now legacy.

cpp/test/stats/weighted_mean.cu
```diff
@@ -171,29 +172,86 @@ class ColWeightedMeanTest : public ::testing::TestWithParam<WeightedMeanInputs<T
...
+    weightedMean(dact.data().get(),
+                 din.data().get(),
+                 dweights.data().get(),
+                 cols,
+                 rows,
+                 false,
```
cjnolet, line 215: We should add a param for `row_major` as well so we can test that path.

cjnolet (changes requested):
I really like the consolidated version. Overall it looks great.

cpp/test/stats/weighted_mean.cu
lowener: The `row_major=true` path is already tested through `colWeightedMean` and `rowWeightedMean` so no need to test it twice. I added a comment for the future to make it explicit.

cjnolet: We should test each option in the public API so that the interface isn't making assumptions about the implementation details (e.g. someone changing this in the future might not realize that assumption.)

cjnolet (approved):
LGTM pending CI. Thanks Mickael!

cjnolet:
@gpucibot merge

PR #516, opened 2022-02-18
raft-config is idempotent no matter RAFT_COMPILE_LIBRARIES value

robertmaynard:
This allows us to correctly handle raft with and without compiled libraries.


cpp/CMakeLists.txt
```diff
@@ -267,7 +267,7 @@ install(TARGETS raft
         DESTINATION ${lib_dir}
         EXPORT raft-exports)
 install(TARGETS raft_distance
-        DESTINATION ${lib_dir}
+        DESTINATION $lib_dir}
```
cjnolet, line 270: Do we need the curly brace here?

cjnolet (approved):
LGTM

cjnolet:
@gpucibot merge

PR #521, opened 2022-02-21
Use shfl_xor in warpReduce for broadcast

akifcorduk:
Some applications might need the reduction result in all lanes of the warp. To prevent executing additional shuffle operations after the reduction, it is better to have an option to do the shuffle during reduction. The implementation uses `shfl_xor` to keep the valid results across results at each step of the reduction. For >SM_80 devices we can consider implicit reduction operations [B.21. Warp Reduce Functions](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#warp-reduce-functions)

### DIFF
--- a/cpp/include/raft/cuda_utils.cuh
+++ b/cpp/include/raft/cuda_utils.cuh
@@ -652,18 +652,20 @@ DI T shfl_xor(T val, int laneMask, int width = WarpSize, uint32_t mask = 0xfffff
 /**
  * @brief Warp-level sum reduction
  * @param val input value
- * @return only the lane0 will contain valid reduced result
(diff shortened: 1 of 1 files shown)


teju85 (changes requested):
Thanks @akifcorduk for the PR. Like we discussed in a separate thread, it is simpler to just replace `shfl` with `shfl_xor`. It will not change the existing codes and will avoid the need for an extra broadcast for the cases which require all lanes to obtain the reduced result.

teju85 (approved):
Changes LGTM.

cjnolet:
@gpucibot merge

PR #524, opened 2022-02-22
Moving device functions to cuh files and deprecating hpp

cjnolet:
For consistency, we had originally scraped through the primitive functions and used the `hpp` extension across the public API. However, it was brought to my attention more recently that this is confusing when considering the larger scope of the project- which also contains many host-only APIs that don't require a cuda-enabled compiler. 

However, as we're gaining more consumers, we need to start being more careful about making breaking changes to the public APIs and their header files. For this reason, I'm opting to copy the existing `hpp` files into `cuh` files, deprecating the hpp files, and using `#define` w/ conditionals to make sure the contents from only one file get defined even if both are included (for example, when a user includes `filea.hpp` but raft internally includes `filea.cuh`. This should allow us to set a version where we can make an announcement to remove the offending `hpp` files and give ample notice before the breaking change is made.


cjnolet:
I wrote a little script to automatically update includes of downstream repositories so they match the extensions in RAFT.

```python
import sys
import glob
import re
import shutil

if __name__ == "__main__":

    script, raft_include_path, source_path = sys.argv

    p = re.compile(r'.*\#include \<(.*)\.(.*)\>')

    header_map = {}
    files_globbed = []

    types = ("/**/*.cuh", "/**/*.hpp", "/**/*.h")

    for files in types:
      files_globbed.extend([f for f in glob.glob(raft_include_path + files, recursive=True)])

    for f in files_globbed:
      fname, fext = f.split(".")
      header_map[fname.replace(raft_include_path, "")] = fext

    types_to_correct = ("/**/*.cuh", "/**/*.hpp", "/**/*.h", "/**/*.cu", "/**/*.cpp", "/**/*.c")

    files_to_correct = []
    for files in types_to_correct:
      files_to_correct.extend([f for f in glob.glob(source_path + files, recursive=True)])

    for f in files_to_correct:
        
      with open(f, 'r') as fp:
        with open("__tmp", "w") as tfp:
          lines = fp.readlines()
          for line in lines:
            match_br = p.match(line)

            if match_br is None:
                tfp.write(line)
            else:
              fname = match_br.group(1)
              fext = match_br.group(2)

              if fname in header_map and header_map[fname] != fext:
                  # p.sub(r'\2', header_map[fname])
                  replacement = "#include <%s.%s>" % (fname, header_map[fname])
                  new_line = p.sub(replacement, line)
                  tfp.write(new_line)
                  print("fname=%s, fext=%s, header_map=%s, new=%s" % (fname, fext, header_map[fname], new_line))
              else:
                  tfp.write(line)

      shutil.move("__tmp", f)
```

Invoke the script w/ 2 arguments: `python match_raft_headers.py <raft_include_path> <consuming_repo_source_path>`

cjnolet:
rerun tests

msadang (approved):
All checks have passed.  Numerous small changes of year and file extension make up the majority of the changes.  Looks good to me.

dantegd (approved):

cjnolet:
@gpucibot merge

Issue #528, opened 2022-02-25
[BUG] `compute-sanitizer` errors in gtests


cjnolet:
I noticed some stange failures in the gtests for the [3d rbc changes](https://github.com/rapidsai/raft/pull/527) when running all of the gtests that I don't see when just running the ball cover tests. Running `compute-sanitizer` in the gtests is tripping some different memory errors. I'm creating this issue as a tracker and will put the different memory errors in comments.

cjnolet:
The Hangarian tests look like they are failing. I also verified they are failing in 22.02:

```
========= COMPUTE-SANITIZER
Running main() from ../googletest/src/gtest_main.cc
Note: Google Test filter = Raft.Hung*
[==========] Running 6 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 6 tests from Raft
[ RUN      ] Raft.HungarianIntFloat
========= Invalid __global__ write of size 4 bytes
=========     at 0xdd0 in void raft::lap::detail::kernel_dualUpdate_1<int, float>(T2 *, const T2 *, const int *, int, T1, T2)
=========     by thread (2,0,0) in block (0,0,0)
=========     Address 0x7fb04a880a08 is out of bounds
=========     and is 1 bytes after the nearest allocation at 0x7fb04a880a00 of size 8 bytes
=========     Saved host backtrace up to driver entry point at kernel launch time
=========     Host Frame: [0x23adbc]
=========                in /lib/x86_64-linux-gnu/libcuda.so
=========     Host Frame: [0x141cc]
```

cjnolet:
I'm seeing an error in the interruptible tests as well. cc @achirkin 

```
========= COMPUTE-SANITIZER
Running main() from ../googletest/src/gtest_main.cc
Note: Google Test filter = Raft.Inter*
[==========] Running 4 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 4 tests from Raft
[ RUN      ] Raft.InterruptibleBasic
[       OK ] Raft.InterruptibleBasic (0 ms)
[ RUN      ] Raft.InterruptibleRepeatedGetToken
[       OK ] Raft.InterruptibleRepeatedGetToken (0 ms)
[ RUN      ] Raft.InterruptibleDelayedInit
[       OK ] Raft.InterruptibleDelayedInit (0 ms)
[ RUN      ] Raft.InterruptibleOpenMP
[...]

PR #531, opened 2022-02-26
Allocate sufficient memory for Hungarian if number of batches > 1

ChuckHastings:
Addresses Hungarian bug described in #528.

The `dualUpdate` method was originally using an array of size one which was eventually changed to a scalar.  It really needs to be an array of size SP (number of subproblems in Date/Nagi nomenclature, number of batches as integrated into raft).


cjnolet (approved):
LGTM

cjnolet:
@gpucibot merge

PR #532, opened 2022-02-28
find_package(raft) can now be called multiple times safely

robertmaynard:
Previously it would try to add the `raft::Thrust` target multiple times, causing a CMake error as the target already existed.

This is required as find_package(X) can be invoked any number of times and will execute the `X-config.cmake` each time.


cjnolet (approved):

robertmaynard:
@gpucibot merge

PR #540, opened 2022-03-02
Pylibraft for pairwise distances

### LINKED ISSUE #546: [FEA] Make UCX / NCCL dependencies optional
Currently, UCX and NCCL dependencies are required both when building RAFT from source and installing the conda packages. This presents a challenge for users who just want to invoke a couple functions and don't intend to use RAFT's mnmg tooling. We should figure out a good way to isolate these dependencies, maybe by creating an additional (optional) component like we've done w/ the distances and nearest neighbors components.

cjnolet:
The goal w/ the initial pylibraft package is to 1) create one end to end API (distances) and 2) show how we would reuse specializations from the existing shared libs in order to expose primitives through Python using minimal dependencies (only rmm, numpy, and scipy, for example).

Closes #546


cjnolet:
rerun tests

ci/gpu/build.sh
```diff
@@ -73,6 +75,9 @@ pip install "git+https://github.com/dask/distributed.git@main" --upgrade --no-de
 pip install "git+https://github.com/dask/dask.git@main" --upgrade --no-deps
 set +x
 
+# Install pre-built conda packages from previous CI step
+gpuci_logger "Install libraft conda packages from CPU job"
+gpuci_mamba_retry install --use-local "${LIBRAFT_CONDA_PACKAGES}/libraft*.bz2"
```
ajschmidt8, line 80: This line could potentially cause issues. From the `conda install` [docs here](https://docs.conda.io/projects/conda/en/latest/commands/install.html):

> conda can also be called with a list of explicit conda package filenames (e.g. ./lxml-3.2.0-py27_0.tar.bz2). Using conda in this mode implies the --no-deps option, and should likewise be used with great caution.

So the implication of calling `conda install` on an explicit filename is that it will only install the package itself and not any of the package's runtime dependencies.

The solution to this problem is to install the package using the `--channel`/`-c` flag with a value that points to the local `conda` channel. So something like this:

```sh
export LIBRAFT_CONDA_PACKAGES="$WORKSPACE/ci/artifacts/raft/cpu/.conda-bld/" # notice there is no `linux-64` here
gpuci_mamba_retry install -c "${LIBRAFT_CONDA_PACKAGES}" libraft
```

This will tell `conda install` to do install `libraft` from the channel on the local filesystem and to also install `libraft`'s dependencies as it usually would.

ci/local/README.md
```diff
@@ -23,7 +23,7 @@ where:
 ```
 
 Example Usage:
-`bash build.sh -r ~/rapids/raft -i gpuci/rapidsai-base:cuda9.2-ubuntu16.04-gcc5-py3.6`
+`bash build.sh -r ~/rapids/raft -i gpuci/rapidsai-base:cuda11.5-ubuntu21.04-py3.8`
```
ajschmidt8, line 26:
```suggestion
`bash build.sh -r ~/rapids/raft -i gpuci/rapidsai-base:cuda11.5-ubuntu20.04-py3.8`
```

I think this is a typo. We don't have any `ubuntu21.04` images.

cjnolet:
rerun tests

cjnolet:
rerun tests

cpp/test/CMakeLists.txt
```diff
@@ -138,44 +138,42 @@ add_executable(test_raft
...
-  GTest::gtest_main
-  Threads::Threads
-  $<TARGET_NAME_IF_EXISTS:OpenMP::OpenMP_CXX>
-  $<TARGET_NAME_IF_EXISTS:conda_env>
-)
+        PRIVATE
+        raft::raft
+        raft::distance
+        raft::nn
+        faiss::faiss
```
rlratzel, line 167: Does this mean RAFT tests also require that FAISS is installed? I didn't see that mentioned in the `BUILD.md` or `README.md`. (I suppose the section on tests in `BUILD.md` is a good place to add this and other requirements for building and running tests)

BUILD.md
```diff
@@ -16,43 +17,61 @@
...
-To remain flexible, the individual shared libraries have their own flags and multiple can be used (though currently only the `nn` and `distance` packages contain shared libraries):
+Individual shared libraries have their own flags and multiple can be used (though currently only the `nn` and `distance` packages contain shared libraries):
 ```bash
 ./build.sh libraft --compile-nn --compile-dist
 ```
 
-###<a id="gtests"></a>Googletests
+Add the `--install` flag to the above example to also install the shared libraries into `$INSTALL_PREFIX/lib`.
+
+###<a id="gtests"></a>Tests
```
rlratzel, line 70:
```suggestion
### <a id="gtests"></a>Tests
```
(same for other section headers missing a space since I think they're not rendering right)

cpp/CMakeLists.txt
```diff
@@ -55,6 +55,13 @@ option(RAFT_COMPILE_LIBRARIES "Enable building raft shared library instantiation
 option(RAFT_COMPILE_NN_LIBRARY "Enable building raft nearest neighbors shared library instantiations" OFF)
 option(RAFT_COMPILE_DIST_LIBRARY "Enable building raft distant shared library instantiations" OFF)
 option(RAFT_ENABLE_NN_DEPENDENCIES "Search for raft::nn dependencies like faiss" ${RAFT_COMPILE_LIBRARIES})
+
+option(RAFT_ENABLE_cuco_DEPENDENCY "Enable cuCollections dependency" ON)
```
dantegd, line 59: When is it recommended to use this flag? It doesn't seem to be used currently in a few places (like line 150) to disable cuco::cuco

python/pylibraft/pylibraft/distance/pairwise_distance.pyx
```diff
@@ -0,0 +1,135 @@
...
+
+def distance(X, Y, dists, metric="euclidean"):
+    """
+    Compute pairwise distances between X and Y
+
+    Parameters
+    ----------
+
+    X : CUDA array interface matrix shape (m, k)
+    Y : CUDA array interface matrix shape (n, k)
```
dantegd, line 81:
```suggestion
    X : CUDA array interface compliant matrix of shape (m, k)
    Y : CUDA array interface compliant matrix of shape (n, k)
```

python/pylibraft/pylibraft/distance/pairwise_distance.pyx
```diff
@@ -0,0 +1,135 @@
...
+    dists : Writable CUDA array interface matrix shape (m, n)
+    metric : string denoting the metric type
+    """
+
+    # TODO: Validate inputs, shapes, etc...
```
dantegd, line 86: Discussion for beyond this PR: since these are quite low level APIs (at least as they are right now), it might be worth to make the validation an option that can be disabled

python/pylibraft/pylibraft/distance/pairwise_distance.pyx
```diff
@@ -0,0 +1,135 @@
...
+    n = y_cai["shape"][0]
+    k = x_cai["shape"][1]
+
+    x_ptr = <uintptr_t>x_cai["data"][0]
+    y_ptr = <uintptr_t>y_cai["data"][0]
+    d_ptr = <uintptr_t>dists_cai["data"][0]
+
+    cdef handle_t *h = new handle_t()
+
+    # TODO: Support single and double precision
```
dantegd, line 101: This is already in the code, no?

python/pylibraft/pylibraft/distance/pairwise_distance.pyx
```diff
@@ -0,0 +1,135 @@
...
+    if x_dt == np.float32:
+        pairwise_distance(deref(h),
+                          <float*> x_ptr,
+                          <float*> y_ptr,
+                          <float*> d_ptr,
+                          <int>m,
+                          <int>n,
+                          <int>k,
+                          <DistanceType>distance_type,
+                          <bool>True, <float>0.0)
```
dantegd, line 123:
```suggestion
                          <bool>True, 
                          <float>0.0)
```

python/pylibraft/pylibraft/distance/pairwise_distance.pyx
```diff
@@ -0,0 +1,135 @@
...
+    elif x_dt == np.float64:
+        pairwise_distance(deref(h),
+                          <double*> x_ptr,
+                          <double*> y_ptr,
+                          <double*> d_ptr,
+                          <int>m,
+                          <int>n,
+                          <int>k,
+                          <DistanceType>distance_type,
+                          <bool>True, <float>0.0)
```
dantegd, line 133:
```suggestion
                          <bool>True, 
                          <float>0.0)
```

python/pylibraft/pylibraft/test/test_distance.py
```diff
@@ -0,0 +1,81 @@
...
+    def copy_to_host(self):
+        return np.frombuffer(self.device_buffer_.tobytes(),
+                             dtype=self.ndarray_.dtype,
+                             like=self.ndarray_)\
+            .astype(self.ndarray_.dtype)\
+            .reshape(self.ndarray_.shape)
+
+
+@pytest.mark.parametrize("n_rows", [10, 100])
+@pytest.mark.parametrize("n_cols", [10, 100])
```
dantegd, line 49:
```suggestion
@pytest.mark.parametrize("n_rows", [100])
@pytest.mark.parametrize("n_cols", [100])
```

The case of 10 doesn't test anything that the 100 does, right? We can avoid over parametrization in RAFT a priori 🙂

python/pylibraft/setup.cfg
```diff
@@ -0,0 +1,58 @@
...
+    buck-out
+    build
+    dist
+    __init__.py
+
+[options]
+packages = find:
+install_requires =
+    numpy
+    numba>=0.49
```
dantegd, line 57: Is numba required?

README.md
```diff
@@ -23,12 +23,9 @@ The algorithms in RAFT span the following general categories:
...
 
-RAFT also provides a Python library that is currently limited to
-1. a python wrapper around the `raft::handle_t` for managing cuda library resources
-2. definitions for using `raft::handle_t` directly in cython
-3. tools for building multi-node multi-GPU algorithms that leverage [Dask](https://dask.org/)
-
-The Python API is being improved to wrap the algorithms and primitives from the categories above.
+RAFT also provides 2 Python libraries:
+- `pylibraft` - cython wrappers around RAFT algorithms and primitives.
+- `pyraft` - reusable infrastructure for building analytics, such as tools for building multi-node multi-GPU algorithms that leverage [Dask](https://dask.org/).
```
dantegd, line 28: Might be good to specify exactly what is contained in each of these, it's a change even for existing functionality that seems to have moved like interruptible execution so it'd be very helpful

dantegd (changes requested):

BUILD.md
cjnolet: The auto generated section titles do some weird things with different characters which was non-obvious. I prefer the html also for maintainability because it makes very explicit the link which corresponds to the title.

BUILD.md
```diff
@@ -95,72 +113,101 @@ RAFT's cmake has the following configurable flags available:.
...
-
+Currently, shared libraries are provided for the `libraft-nn` and `libraft-distance` components. The `libraft-nn` component depends upon [FAISS](https://github.com/facebookresearch/faiss) and the `RAFT_ENABLE_NN_DEPENDENCIES` option will build it from source if it is not already installed.
 
 ### <a id="python"></a>Python
 
 Conda environment scripts are provided for installing the necessary dependencies for building and using the Python APIs. It is preferred to use `mamba`, as it provides significant speedup over `conda`. The following example will install create and install dependencies for a CUDA 11.5 conda environment:
 
 ```bash
-conda env create --name raft_env -f conda/environments/raft_dev_cuda11.5.yml
+mamba env create --name raft_env -f conda/environments/raft_dev_cuda11.5.yml
```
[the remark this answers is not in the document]
cjnolet: Isn't `-n` the same as `--name`? Or are you suggesting changing `raft_env` to `dev_env`?

BUILD.md
```diff
@@ -95,72 +113,101 @@ RAFT's cmake has the following configurable flags available:.
...
 ```
 
 ## <a id="use_raft"></a>Using RAFT in downstream projects
 
+There are two different strategies for including RAFT in downstream projects, depending on whether or not the required dependencies are already installed and available on the `lib` and `include` paths. 
+
 ### <a id="cxx_integration"></a>C++ header-only integration using cmake
 
-Use RAFT in cmake projects with `find_package(raft)` for header-only operation and the `raft::raft` target will be available for configuring linking and `RAFT_INCLUDE_DIR` will be available for includes. Note that if any packages are used which require downstream dependencies, such as the `libraft-nn` package requiring FAISS, these dependencies will have be installed and configured in cmake independently.
+When the needed [build dependencies](#required_depenencies) are already satisfied, RAFT can be trivially integrated into downstream projects by cloning the repository and adding `cpp/include` from RAFT to the include path:
```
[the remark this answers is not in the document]
cjnolet: > Not sure the wording trivial applies to CMake dependencies in general

The word `trivial` here is not implying anything about cmake dependencies and cmake doesn't necessarily even have to be used to get the dependencies. It's a header-only library so this statement is noting that the most trivial way to install RAFT if you already have the needed dependencies is to just clone it and point to the include directory. This is what spark-rapids-ml is doing currently because the only dependencies they need are the CTK libs.

README.md
```diff
@@ -80,34 +77,56 @@ auto metric = raft::distance::DistanceType::L2SqrtExpanded;
...
+The example below demonstrates computing the pairwise Euclidean distances between cupy arrays.
+```python
+import cupy as cp
+
+from pylibraft.distance import pairwise_distance
+
+n_samples = 5000
+n_features = 50
+
+input = cp.random.random_sample((n_samples, n_features), dtype=cp.float32)
```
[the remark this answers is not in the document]
cjnolet: Good catch!

README.md
```diff
@@ -80,34 +77,56 @@ auto metric = raft::distance::DistanceType::L2SqrtExpanded;
...
+output = cp.empty((n_samples, n_samples), dtype=cp.float32)
+
+pairwise_distance(input, input, output, "euclidean")
```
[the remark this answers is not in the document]
cjnolet: > Might be good to add that this is a low level API that is prioritizing efficiency and simplicity over being pythonic? 

This is a good point. I actually think it might make sense to expose a very lightweight CAI wrapper around the RMM device buffer like the one we're using in the tests and make the second argument optional so that the `RmmCAIWrapper` could be returned by default, similar to functions I've seen in Scipy. This gives a user the option of doing this:

```python
option1 = pairwise_distance(in1, in2, metric="euclidean")
```

or this

```python
pairwise_distance(in1, in2, out=option2, metric="euclidean")
```

cpp/CMakeLists.txt
cjnolet: Was this comment perhaps made on an older version of the changes in this PR? The Line 153 contains the following, which will add cuco to link library dependencies when it's been found in `find_package`:
`$<$<BOOL:${cuco_ADDED}>:cuco::cuco>`. The goal here is to allow upstream projects to add it if needed but allow it to not be added if `find_package` was never invoked for it.

cpp/include/raft/comms/comms.hpp
```diff
@@ -14,11 +14,18 @@
...
+/**
+ * @warning This file is deprecated and will be removed in release 22.06.
+ * Please use raft_runtime/comms.hpp instead.
+ */
 
-#include <raft/cudart_utils.h>
+#ifndef __RAFT_RT_COMMS_H
+#define __RAFT_RT_COMMS_H
+
+#pragma once
```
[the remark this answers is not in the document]
cjnolet: The pragma itself could be placed outside of the include guards. The include guards are temporary, though, and only needed to eliminate compile errors from projects including old headers. Once the projects are updated and the old files go away, the include guards will be removed.

divyegala (approved):
This PR looks very good to me and after spending some time going over it, I think it is good to go as-is from my side.

DEVELOPER_GUIDE.md
```diff
@@ -4,7 +4,7 @@
...
+The process for working on a CUDA/C++ feature which spans RAFT and one or more consumers can vary slightly depending on whether the consuming project relies on a source build (as outlined in the [BUILD](BUILD.md#install_header_only_cpp) docs). In such a case, the option `CPM_raft_SOURCE=/path/to/raft/source` can be passed to the cmake of the consuming project in order to build the local RAFT from source. The PR with relevant changes to the consuming project can also pin the RAFT version temporarily by explicitly changing the `FORK` and `PINNED_TAG` arguments to the RAFT branch containing their changes when invoking `find_and_configure_raft`.  The pin should be reverted after the changed is merged to the RAFT project and before it is merged to the dependent project(s) downstream.
```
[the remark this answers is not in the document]
cjnolet: I think it doesn't do much to demonstrate the use of `CPM_raft_SOURCE`. Reverting the pinned tag seems straightforward to me and I'm concerned pointing to a PR just for that is going to overcomplicate things.

README.md
```diff
@@ -80,34 +77,56 @@ auto metric = raft::distance::DistanceType::L2SqrtExpanded;
...
+conda install -c rapidsai libraft-headers libraft-nn libraft-distance pyraft pylibraft
 ```
 
-After installing RAFT, `find_package(raft COMPONENTS nn distance)` can be used in your CUDA/C++ build. Note that the `COMPONENTS` are optional and will depend on the packages installed.
+After installing RAFT, `find_package(raft COMPONENTS nn distance)` can be used in your CUDA/C++ build. `COMPONENTS` are optional and will depend on the packages installed.
 
 ### CPM
 
-RAFT uses the [RAPIDS cmake](https://github.com/rapidsai/rapids-cmake) library, which makes it simple to include in downstream cmake projects. RAPIDS cmake provides a convenience layer around the [Cmake Package Manager (CPM)](https://github.com/cpm-cmake/CPM.cmake). 
+RAFT uses the [RAPIDS cmake](https://github.com/rapidsai/rapids-cmake) library, which makes it simple to include in downstream cmake projects. RAPIDS cmake provides a convenience layer around CPM.
```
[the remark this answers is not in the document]
cjnolet: The main goal here is to provide a simplified version of the more comprehensive example in `BUILD.md` as a "quick-start". I'm hesitant to consolidate this simple example w/ the more complicated one, but I'm also hesitant to put the more complex example on the front page. I really don't think this will be too much of a burden to maintain but if it becomes a hassle or ends up diverging more than it should then I don't mind consolidating in the future.

cjnolet:
rerun tests

BUILD.md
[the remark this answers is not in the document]
dantegd: sorry totally missed --name was there at the beginning indeed

cpp/CMakeLists.txt
dantegd: Yeah, when seeing the file github showed me an older version in the branch of the PR that didn't have that change, so now it makes sense

cpp/cmake/versions.json
```diff
@@ -1,9 +1,11 @@
 {
   "packages" : {
-      "Thrust" : {
-        "version" : "1.15.0",
-        "git_url" : "https://github.com/NVIDIA/thrust.git",
-        "git_tag" : "${version}"
+    "Thrust" : {
+      "version" : "1.12.0",
```
dantegd, line 4:
```suggestion
      "version" : "1.15.0",
```

All of RAPIDS has been using 1.15 for releases 22.02 and 22.04, though looking at the CMake files this versions file is not being used to override versions, so it might be good to get rid of it entirely for now?

dantegd (approved):
Just had one minor comment, approving

cjnolet:
@gpucibot merge

cjnolet:
@gpucibot merge

PR #547, opened 2022-03-08
Separating out `raft_frontend` headers

cjnolet:
Separating out "frontend" headers that can be installed to enable RAFT in public APIs of consuming projects without having to install the entire (build time) header-only library. 

Note: This PR changes the name of the deployed conda package from `libraft-headers` to `libraft-frontend`

The main cmake targets and optional components are also changed a bit, which will require a small update to the `get_raft.cmake` in each of the consuming projects to add a `backend` component by default. With the changes in this PR, any consuming projects that just want to build against a project's public API and use the runtime headers can build against raft without the `backend` component. 

Addresses https://github.com/rapidsai/raft/pull/547


ajschmidt8 (approved):

cpp/cmake/modules/raft_export.cmake
```diff
@@ -164,6 +170,10 @@ function(raft_export type project_name)
     set(RAPIDS_PROJECT_FINAL_CODE_BLOCK "${${RAPIDS_FINAL_CODE_BLOCK}}")
   endif()
 
+  if(DEFINED RAPIDS_INSTALL_FILES AND NOT RAPIDS_INSTALL_FILES)
```
robertmaynard, line 173: Suggestion:

```cmake
set(rapids_install_files OFF)
if(DEFINED RAPIDS_INSTALL_FILES)
  set(rapids_install_files ${RAPIDS_INSTALL_FILES})
endif()
```

Afterwards you can use `NOT rapids_install_files` instead of  `NOT DEFINED RAPIDS_INSTALL_FILES`

cpp/cmake/thirdparty/get_libcudacxx.cmake
```diff
@@ -16,7 +16,9 @@
 function(find_and_configure_libcudacxx)
   include(${rapids-cmake-dir}/cpm/libcudacxx.cmake)
 
-  rapids_cpm_libcudacxx(BUILD_EXPORT_SET raft-exports INSTALL_EXPORT_SET raft-exports)
+  rapids_cpm_libcudacxx(BUILD_EXPORT_SET raft-backend-exports
```
robertmaynard, line 19: We should simplify all of these `get_*` changes to the following instead of using `EXCLUDE_FROM_ALL`

```cmake
rapids_cpm_libcudacxx(BUILD_EXPORT_SET raft-backend-exports)
rapids_export_package(INSTALL libcudacxx raft-backend-exports)
```

cjnolet:
I'm closing this PR and shifting focus to https://github.com/rapidsai/raft/pull/540.

PR #548, opened 2022-03-08
[IMP] clang-tidy support: improved clang run scripts with latest changes (see cugraph-ops)

MatthiasKohl:
This PR slightly improves the clang-tidy scripts: for now, they cannot be run successfully on all RAFT source files due to some dependencies not being compatible with clang, this support stays experimental and broken for certain sources.

However, this PR improves the scripts such that they can be used as references for other projects (in particular cudf which is looking into clang-tidy support as well). It makes the GCC tool-chain support much clearer and tries to use whatever compiler was actually used by `cmake` (instead of guessing what that compiler was). It is also compatible with the latest changes in RAPIDS CI docker images (see https://github.com/rapidsai/gpuci-build-environment/pull/246).

For reference of the scripts, see also https://github.com/rapidsai/cugraph-ops/pull/41 (requires RAPIDS organization access), which adds full clang-tidy support in a RAPIDS project.

### DIFF
--- a/cpp/scripts/run-clang-compile.py
+++ b/cpp/scripts/run-clang-compile.py
@@ -20,7 +20,6 @@
 
 from __future__ import print_function
 import argparse
-import glob
 import json
 import multiprocessing as mp
 import os
@@ -29,6 +28,8 @@ import shutil
 import subprocess
 
 
+CMAKE_COMPILER_REGEX = re.compile(
+    r"^\s*CMAKE_CXX_COMPILER:FILEPATH=(.+)\s*$", re.MULTILINE)
 CLANG_COMPILER = "clang++"
 GPU_ARCH_REGEX = re.compile(r"sm_(\d+)")
 SPACES = re.compile(r"\s+")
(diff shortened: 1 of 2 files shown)


cjnolet (approved):
Changes LGTM. Out of curiosity, what dependencies are still blocking clang-tidy support? I’m not opposed to overriding some dependencies in raft if it’s necessary to make this work.

cjnolet:
@gpucibot merge

MatthiasKohl:
@cjnolet I believe it's mainly cuco and libcudacxx.
There is also the CUDA toolkit itself, but that would be fixed by upgrading clang (and tools) to version >= 13.
That update might slightly affect clang-format outputs as well, so we should keep that in mind.

In any case, I can try to see what exactly is causing these issues within cuco and libcudacxx and we might be able to improve things there to be compatible with clang (or maybe add a few intrinsics like I did with the `__ldcg`-kind of functions in CUDA).
Right now, I've created issues on those projects, but I don't think that this is on their roadmaps.

cjnolet:
Darn! I guess we still have a bit before we can support this fully in raft. On due time, I guess. Thanks again for keeping the scripts up to date!

PR #549, opened 2022-03-09
Add benchmarks

achirkin:
Copied the benchmarking setup and one simple benchmark (as an example) from cuml.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -41,6 +41,7 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 # - User Options  ------------------------------------------------------------
 
 option(BUILD_TESTS "Build raft unit-tests" ON)
+option(BUILD_BENCH "Build raft C++ benchmark tests" ON)
 option(CUDA_ENABLE_KERNELINFO "Enable kernel resource usage info" OFF)
 option(CUDA_ENABLE_LINEINFO "Enable the -lineinfo option for nvcc (useful for cuda-memcheck / profiler)" OFF)
 option(CUDA_STATIC_RUNTIME "Statically link the CUDA runtime" OFF)
@@ -58,6 +59,7 @@ include(CMakeDependentOption)
 cmake_dependent_option(RAFT_USE_FAISS_STATIC "Build and statically link the FAISS library for nearest neighbors search on GPU" ON RAFT_COMPILE_LIBRARIES OFF)
 
 message(VERBOSE "RAFT: Build RAFT unit-tests: ${BUILD_TESTS}")
+message(VERBOSE "RAFT: Building raft C++ benchmarks: ${BUILD_BENCH}")
 message(VERBOSE "RAFT: Enable detection of conda environment for dependencies: ${DETECT_CONDA_ENV}")
 message(VERBOSE "RAFT: Disable depreaction warnings " ${DISABLE_DEPRECATION_WARNINGS})
 message(VERBOSE "RAFT: Disable OpenMP: ${DISABLE_OPENMP}")
@@ -120,6 +122,10 @@ if(BUILD_TESTS)
   include(cmake/thirdparty/get_gtest.cmake)
   include(cmake/thirdparty/get_nccl.cmake)
   include(cmake/thirdparty/get_ucx.cmake)
+endif()
+
+if(BUILD_BENCH)
+  include(cmake/thirdparty/get_gbench.cmake)
 endif()
 
 ##############################################################################
@@ -409,6 +415,13 @@ endforeach()
 
 if(BUILD_TESTS)
   include(test/CMakeLists.txt)
+endif()
+
+##############################################################################
+# - build benchmark executable -----------------------------------------------
+
+if(BUILD_BENCH)
+  include(bench/CMakeLists.txt)
 endif()
 
 ##############################################################################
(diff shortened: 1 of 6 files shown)


cpp/bench/linalg/reduce.cu
```diff
@@ -0,0 +1,76 @@
+/*
+ * Copyright (c) 2019-2022, NVIDIA CORPORATION.
```
cjnolet, line 2: New files should only have current year

cpp/CMakeLists.txt
```diff
@@ -41,6 +41,7 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 # - User Options  ------------------------------------------------------------
 
 option(BUILD_TESTS "Build raft unit-tests" ON)
+option(BUILD_BENCH "Build raft C++ benchmark tests" ${BUILD_TESTS})
```
cjnolet, line 44: We also have to get in the habit of updating build docs when we add/change options in raft. I admit I’ve not been great at doing this myself.

cpp/CMakeLists.txt
```diff
@@ -58,6 +59,7 @@ include(CMakeDependentOption)
 cmake_dependent_option(RAFT_USE_FAISS_STATIC "Build and statically link the FAISS library for nearest neighbors search on GPU" ON RAFT_COMPILE_LIBRARIES OFF)
 
 message(VERBOSE "RAFT: Build RAFT unit-tests: ${BUILD_TESTS}")
+message(VERBOSE "RAFT: Building raft C++ benchmarks: ${BUILD_BENCH}")
```
cjnolet, line 62: This option also needs to be added to `build.sh` to match cuml. Also important that we disable the benchmarks by default in `build.sh` so the artifacts doesn’t end up in our conda packages. I can see an argument for maybe doing this eventually but we will want to release a specific package just for the benchmarks.

cjnolet (changes requested):
Thanks so much for getting a start on moving this over! Mostly trivial things like updating docs / build.sh

cpp/bench/common/benchmark.hpp
```diff
@@ -0,0 +1,219 @@
...
+      size = 0;
+    }
+    for (auto _ : state) {
+      CudaEventTimer timer(state, buff, size, stream);
+      benchmarkFunc();
+    }
+  }
+
+  template <typename T>
+  void alloc(T*& ptr, size_t len, bool init = false)
```
cjnolet, line 152: Is there a way we can remove this and use something more stack friendly?

cpp/bench/common/benchmark.hpp
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+ private:
+  ::benchmark::State* state;
+  cudaStream_t stream = 0;
+  cudaEvent_t start;
+  cudaEvent_t stop;
+};  // end struct CudaEventTimer
+
+/** Main fixture to be inherited and used by all other c++ benchmarks in cuml */
```
cjnolet, line 80: Can we remove the reference to cuml here?

cpp/CMakeLists.txt
achirkin: Sorry, forgot about that!
Btw, you can see in the highlighted line, I've set the `BUILD_BENCH` be the same as `BUILD_TESTS` by default. Thus, raft's `--nogtest` flag effectively disables both. Would you like me (A) to keep this behavior and update the `BUILD.md` to reflect that, or (B) to introduce a new `--nobench` flag and add it everywhere in `ci` folder alongside `--nogtest`? I see the benefit of (B) is being more explicit, but (A) is less invasive and simple under the assumption that normally we need to disable/enable them both anyway.

(This also addresses your comment below).

cpp/CMakeLists.txt
achirkin: Hmm, it seems like the `bench` or `prims-bench` parameters in [cuml/build.sh](https://github.com/rapidsai/cuml/blob/4c1d67120cfbd18cb1e9f1b2f9d2a0395b8df8e1/build.sh) are not passed along to `CMakeLists.txt`, and I haven't found anything by `bench`/`BENCH` keywords in the `ci` of cuml.

As for your suggestion, please check my question in the discussion above.

cpp/CMakeLists.txt
cjnolet: It looks like a bug in the cuml build.sh script. I’m assuming at some point support the `bench` option was removed inadvertently. 

Rather than using a `—bench` option, I’d prefer to invoke build.sh like `build.sh pyraft libraft bench`. Since the benchmarks are a top-level cmake project, this also follows the behavior that cuml *should* have (if it worked as documented). 

I would also prefer not to couple the gtests and the benchmarks options in cmake. Also, I would prefer `build.sh` build the benchmarks when no options are specified (just like how it builds the pyraft and libraft artifacts when no other r options are specified)

achirkin: Ok! I've added two targets (`tests` and `bench`) in `build.sh` in place of a single `--nogtest` option and changed the ci/conda scripts accordingly. Let's see if it passes :)

cjnolet (approved):
LGTM

Ethyling (approved):

cjnolet:
@gpucibot merge

PR #551, opened 2022-03-10
KNN select-top-k variants

achirkin:
Integrate two new implementations for knn's `select_k` function.


tfeher (changes requested):
Hi Artem, thanks for this PR! Although it is still a work in progress, I think it makes sense to give you an early feedback. Here are my first batch of comments.

cpp/include/raft/spatial/knn/detail/selection_faiss.cuh
```diff
@@ -127,38 +132,41 @@ inline void select_k_impl(value_t* inK,
...
-    select_k_impl<value_idx, value_t, 512, 8>(
+    select_k_impl<payload_t, key_t, 512, 8>(
       inK, inV, n_rows, n_cols, outK, outV, select_min, k, stream);
-  else if (k <= 1024)
-    select_k_impl<value_idx, value_t, 1024, 8>(
+  else if (k <= 1024 && k <= max_k)
+    select_k_impl<payload_t, key_t, max_k, 8>(
       inK, inV, n_rows, n_cols, outK, outV, select_min, k, stream);
+  else
+    ASSERT(k <= max_k, "Current max k is %d (requested %d)", max_k, k);
```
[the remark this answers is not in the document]
achirkin: Yep

cpp/include/raft/spatial/knn/detail/ivf_flat/bitonic_sort.cuh
```diff
@@ -0,0 +1,168 @@
...
+    const int lane = threadIdx.x % Size;
+    for (int stride = Size / 2; stride > 0; stride /= 2) {
+      bool is_second = lane & stride;
+      KeyT& key      = *keys;
+      KeyT other     = shfl_xor(key, stride, Size);
+
+      bool asc       = Ascending != reverse;
+      bool do_assign = key != other && ((key > other) == (asc != is_second));
+      // Normally, we expect `payloads` to be the array of indices from 0 to len;
+      // in that case, the construct below makes the sorting stable.
```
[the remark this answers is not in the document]
achirkin: I'm actually thinking, maybe I should remove this extra comparison and don't bother with stable sorting? I haven't yet evaluated how this affects the overall performance.

cpp/include/raft/spatial/knn/detail/ivf_flat/warpsort_topk.cuh
```diff
@@ -0,0 +1,862 @@
...
+  }
+
+ private:
+  __device__ void add_first_(const T* in, IdxT start, IdxT end)
+  {
+    IdxT idx = start + lane_;
+    for (int i = 0; i < max_arr_len_; ++i, idx += WarpSize) {
+      if (idx < end) {
+        val_arr_[i] = in[idx];
+        idx_arr_[i] = idx;
```
tfeher, line 378: If we did not have indices in the beginning, then do we really need to generate one?

cpp/include/raft/spatial/knn/detail/ivf_flat/warpsort_topk.cuh
```diff
@@ -0,0 +1,862 @@
...
+  __device__ void done()
+  {
+    if (buf_len_ != 0) {
+      bitonic_sort<capacity, greater>::run(val_buf_, idx_buf_);
+      merge_();
+    }
+  }
+
+ private:
+  __device__ void add_first_(const T* in, IdxT start, IdxT end)
```
tfeher, line 372: Please add a short description, like "Loads values in val_arr_".

cpp/include/raft/spatial/knn/detail/ivf_flat/warpsort_topk.cuh
```diff
@@ -0,0 +1,862 @@
...
+    for (int i = 0; i < max_arr_len_; ++i, idx += WarpSize) {
+      if (idx < end) {
+        val_arr_[i] = in[idx];
+        idx_arr_[i] = in_idx[idx];
+      }
+    }
+    bitonic_sort<capacity, !greater>::run(val_arr_, idx_arr_);
+  }
+
+  __device__ void add_extra_(const T* in, IdxT start, IdxT end)
```
tfeher, line 396: Please add a short description, like "Loads values in val_buf_".

cpp/include/raft/spatial/knn/detail/ivf_flat/warpsort_topk.cuh
```diff
@@ -0,0 +1,862 @@
...
+}
+
+}  // namespace
+
+template <int capacity, bool greater, typename T, typename IdxT>
+class WarpSort {
+ public:
+  __device__ WarpSort(IdxT k, T dummy) : lane_(threadIdx.x % WarpSize), k_(k), dummy_(dummy)
+  {
+    static_assert(capacity >= WarpSize && isPo2(capacity));
```
tfeher, line 145: Is there any relationship between `capacity` and `k` that we should check?

cpp/include/raft/spatial/knn/detail/ivf_flat/warpsort_topk.cuh
```diff
@@ -0,0 +1,862 @@
...
+  IdxT idx_buf_[max_arr_len_];
+  int buf_len_;
+};
+
+template <int capacity, bool greater, typename T, typename IdxT>
+class WarpMerge : public WarpSort<capacity, greater, T, IdxT> {
+ public:
+  __device__ WarpMerge(int k, T dummy) : WarpSort<capacity, greater, T, IdxT>(k, dummy) {}
+
+  __device__ void add(const T* in, const IdxT* in_idx, IdxT start, IdxT end)
```
tfeher, line 444: Is it assumed that the input is already sorted?

tfeher (changes requested):
Hi Artem, here is a second batch of comments.

cpp/include/raft/spatial/knn/knn.hpp
```diff
@@ -42,65 +43,121 @@ namespace knn {
...
+
+  switch (algo) {
+    case SelectKAlgo::FAISS:
+      detail::select_k(
+        in_keys, in_values, n_inputs, input_len, out_keys, out_values, select_min, k, stream);
+      break;
+
+    case SelectKAlgo::RADIX_8_BITS:
+      detail::ivf_flat::radix_topk<value_t, idx_t, 8, 512>(
+        in_keys, in_values, n_inputs, input_len, k, out_keys, out_values, !select_min, stream);
```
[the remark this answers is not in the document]
achirkin: Do you mean to add one more overload in `::detail` with a flipped boolean argument?

cpp/include/raft/spatial/knn/knn.cuh
```diff
@@ -38,65 +43,121 @@ namespace knn {
  *
  * etc...
  *
- * @tparam value_idx
+ * @tparam idx_t
```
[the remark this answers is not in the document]
achirkin: Not sure, this PR mainly updates the `select_k` below. I only update the template parameter names so they look more alike the rest.

cpp/include/raft/spatial/knn/detail/ivf_flat/warpsort_topk.cuh
```diff
@@ -0,0 +1,862 @@
...
+  }
+
+  __device__ void add(T val, IdxT idx)
+  {
+    for (int i = 0; i < max_arr_len_; ++i) {
+      if (i == buf_len_) {
+        val_buf_[i] = val;
+        idx_buf_[i] = idx;
+      }
+    }
```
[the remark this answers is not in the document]
achirkin: Apparently, the loop is unrolled, which ensures the constant indexing of the arrays. When directly indexing using `buf_len_`, the array is placed in local memory instead of the registers, which makes the whole algorithm slower (~2x slower on my machine).

cpp/include/raft/spatial/knn/detail/ivf_flat/warpsort_topk.cuh
achirkin: Yes, for kNN we need to find out distances between which elements are the smallest in the end. Sometimes we don't need to pass indices in the first place, because `0..n-1` is implied (dense inputs).

cpp/include/raft/spatial/knn/detail/ivf_flat/bitonic_sort.cuh
```diff
@@ -0,0 +1,168 @@
...
+struct bitonic_merge {
+  static_assert(isPo2(Size));
+
+  /** How many contiguous elements are processed by one thread. */
+  static constexpr int kArrLen = Size / WarpSize;
+  static constexpr int kStride = kArrLen / 2;
+
+  template <bool Fits, typename Dummy>
+  using when_fits_in_warp =
+    std::enable_if_t<(Fits == (Size <= WarpSize)) && std::is_same_v<Dummy, Dummy>, void>;
```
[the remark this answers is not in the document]
achirkin: That's used to delay the evaluation of this template till the same time the function templates are being compiled, rather than to compile it together with the enclosing struct.

cpp/include/raft/spatial/knn/detail/ivf_flat/radix_topk.cuh
```diff
@@ -0,0 +1,488 @@
...
+      for (int i = threadIdx.x; i < num_buckets; i += blockDim.x) {
+        histogram[i] = 0;
+      }
+      if (threadIdx.x == 0) { counter->filter_cnt = 0; }
+    }
+  }
+}
+
+template <typename T, typename IdxT, int BITS_PER_PASS, int NUM_THREAD>
+void radix_topk(const T* in,
```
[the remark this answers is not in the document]
achirkin: (NB: I removed the other function since the review). The interface now is exactly the same as the public one and the rest implementations in the `detail` namespace. I guess, no need to repeat the same docs multiple times?..

cpp/include/raft/spatial/knn/detail/ivf_flat/radix_topk.cuh
```diff
@@ -0,0 +1,488 @@
...
+  const IdxT stride = blockDim.x * gridDim.x;
+  const int tid     = blockIdx.x * blockDim.x + threadIdx.x;
+  if constexpr (sizeof(T) >= sizeof(WideT)) {
+    for (IdxT i = tid; i < len; i += stride) {
+      f(in[i], i);
+    }
+  } else {
+    static_assert(sizeof(WideT) % sizeof(T) == 0);
+    constexpr int items_per_scalar = sizeof(WideT) / sizeof(T);
+    // TODO: it's UB
```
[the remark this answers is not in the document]
achirkin: I guess, undefined behavior.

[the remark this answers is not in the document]
tfeher: I meant to ask what causes here UB?

cpp/include/raft/spatial/knn/detail/ivf_flat/warpsort_topk.cuh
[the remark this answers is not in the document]
tfeher: Ok, then let us keep the loop how it is.

cpp/include/raft/spatial/knn/detail/ivf_flat/radix_topk.cuh
```diff
@@ -0,0 +1,488 @@
...
+template <typename T, int BITS_PER_PASS>
+__host__ __device__ constexpr int calc_num_passes()
+{
+  return (sizeof(T) * 8 - 1) / BITS_PER_PASS + 1;
+}
+
+// bit 0 is the least significant (rightmost) bit
+// this function works even when pass=-1, which is used in calc_mask()
+template <typename T, int BITS_PER_PASS>
+__device__ constexpr int calc_start_bit(int pass)
```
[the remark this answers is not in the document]
achirkin: I've had a feeling that both versions are common. I guess, here we go from the most to least significant bits, because this way we can save a few passes since, as you said, we don't need to sort within the `k`-sized buckets.

cpp/include/raft/spatial/knn/detail/ivf_flat/warpsort_topk.cuh
achirkin: Yes, it's ensured in the calling code that `k <= capacity`, but we cannot use raft's `ASSERT` in the device code to check the value of `k` at runtime.

tfeher (changes requested):
Hi Artem, I have finished my first review round. Overall it looks good, but as my comments indicate, some more documentation is required at certain places.

cpp/include/raft/spatial/knn/knn.hpp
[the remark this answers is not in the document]
tfeher: No, I meant to have only one variant, that has the same (flipped) boolean arg.

cpp/bench/spatial/selection.cu
```diff
@@ -0,0 +1,150 @@
+/*
+ * Copyright (c) 2019-2022, NVIDIA CORPORATION.
```
cjnolet, line 2: New files should just contain current year

cpp/bench/spatial/selection.cu
```diff
@@ -0,0 +1,150 @@
...
+namespace raft::bench::spatial {
+
+struct params {
+  int n_inputs;
+  int input_len;
+  int k;
+  int select_min;
+};
+
+template <typename KeyT, typename IdxT, raft::spatial::knn::SelectKAlgo Algo>
```
cjnolet, line 35: Since this is a new file, we should use consistent casing/naming of template arguments. I vote for `idx_t` and `key_t` to match other functions in RAFT's nearest neighbors API.

cpp/include/raft/spatial/knn/detail/ivf_flat/bitonic_sort.cuh
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+ private:
+  const int warp_width_;
+  const bool ascending_;
+
+  template <int AnotherSize>
+  friend class bitonic;
+
+  template <typename KeyT, typename... PayloadTs>
```
cjnolet, line 170: same here w/ templates. We should strive to be consistent. Though, if we want to forego this discussion / change to get this PR in quickly for 22.04, can you create an issue for it so we don't lose track?

cpp/include/raft/spatial/knn/knn.cuh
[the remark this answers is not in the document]
cjnolet: @tfeher, yes it's correct because it's performing the same k-select it would perform over a tile of pairwise distances, but it's doing it over potentially multiple k-selected partitions `n_partsxnxk` to reduce them into a single set of size `nxk`.

cpp/include/raft/spatial/knn/knn.cuh
```diff
@@ -38,65 +38,121 @@ namespace knn {
...
- * @param n_rows
- * @param n_cols
- * @param outK
- * @param outV
- * @param select_min
- * @param k
- * @param stream
+ *   the type of the keys (what is being compared).
+ *
+ * @param[in] in_keys
```
cjnolet, line 85: love the expansion in the documentation here.

cpp/include/raft/spatial/knn/knn.cuh
```diff
@@ -38,65 +38,121 @@ namespace knn {
...
- * row3: k0, k1, k2, k0, k1, k2
+ * Select k smallest or largest key/values from each row in the input data.
  *
- * etc...
+ * If you think of the input data `in_keys` as a row-major matrix with input_len columns and
+ * n_inputs rows, then this function selects k smallest/largest values in each row and fills
+ * in the row-major matrix `out_keys` of size (n_inputs, k).
  *
- * @tparam value_idx
+ * Note, depending on the selected algorithm, the values within rows of `out_keys` are not
```
cjnolet, line 77: Can you expand here to note the implementations which would not contain sorted values here? It's often assumed the resulting neighbors will be sorted by distance. For example, HDBSCAN and RBC expect that the kth neighbor is the `neighbor at k`, which is used to compute the radius.

cpp/include/raft/spatial/knn/knn.cuh
```diff
@@ -38,65 +38,121 @@ namespace knn {
...
- * @param n_cols
- * @param outK
- * @param outV
- * @param select_min
- * @param k
- * @param stream
+ *   the type of the keys (what is being compared).
+ *
+ * @param[in] in_keys
+ *   contiguous array of inputs of size (input_len * n_inputs);
```
cjnolet, line 86: Can you add `device array` to all of the inputs/outputs which are on device? We should start doing this across the codebase as we're adding and updating docs.

cjnolet (changes requested):
I've done an initial scan through the changes to provide some feedback while I do a more extensive review. 

If we are to still aim for 22.04 with this PR, we also need a PR to cuML to make sure nothing breaks since it's so close to release time.

cpp/bench/spatial/selection.cu
achirkin: I don't mind your variant, but I changed this cause I thought we're trying now to stick to the style declared in [cpp/.clang-tidy](https://github.com/rapidsai/raft/blob/15eb20a4e90a2c9267dea7046dc9cb9cba3e678e/cpp/.clang-tidy#L211-L212). So I just follow the suggestions of clang-tidy to rename things.

Would you like me to change the template parameter naming here or the rules in .clang-tidy?

cpp/include/raft/spatial/knn/knn.cuh
achirkin: Added to the description for the enum class.

cpp/include/raft/spatial/knn/detail/ivf_flat/radix_topk.cuh
```diff
@@ -0,0 +1,488 @@
...
+
+  BlockScan(temp_storage).InclusiveSum(thread_data, thread_data);
+  __syncthreads();
+  if (index < num_buckets) { histogram[index] = thread_data + current; }
+  __syncthreads();  // This sync is necessary, as the content of histogram needs
+                    // to be read after
+}
+
+template <typename T, typename IdxT, int BITS_PER_PASS, int NUM_THREAD>
+__device__ void choose_bucket(Counter<T, IdxT>* counter, IdxT* histogram, const IdxT k)
```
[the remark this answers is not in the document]
yong-wang: update of the counter:
The k-th value must be in the `index`-th bucket, so we record it in `counter->bucket`.
The number of values in this bucket is `(cur - prev)` since `histogram` now actually contains prefix sum of the histogram. Hence the update of `counter->len`.
counter is used in the next pass, so we keep the len of current pass in `counter->previous_len`.

`prev` means the number of values in the first `index-1` buckets. These values are less than (when `greater`=false) the real k-th value, so must be in the top-k results. In the next pass, we need only to find the `(k - prev)`-th value fall in the `index`-th bucket. Hence the setting of `counter->k`.

cpp/include/raft/spatial/knn/detail/ivf_flat/radix_topk.cuh
```diff
@@ -0,0 +1,488 @@
...
+      unsigned mask           = calc_mask<T, BITS_PER_PASS>(pass);
+
+      if (!out) {  // radix select
+        for (IdxT i = threadIdx.x; i < previous_len; i += blockDim.x) {
+          const T value = out_buf[i];
+          int bucket    = calc_bucket<T, BITS_PER_PASS>(value, start_bit, mask, greater);
+          if (bucket == want_bucket) {
+            // TODO: UB
+            // could use atomicExch, but it's not defined for T=half
+            counter->kth_value = value;
```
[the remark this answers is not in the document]
yong-wang: This `if` branch is for another top-k implementation called "radix-select", which only sets counter->kth_value during the passes without writing top-k values to the result array `out`.
After all passes are done, the counter->kth_value is used as a threshold to filter out the real top-k values from original inputs. So counter->kth_value is just an intermediate value and is never passed back to the caller.

Since this implementation is removed, I think this `if` branch is not used anymore and should be deleted.

cpp/include/raft/spatial/knn/detail/ivf_flat/radix_topk.cuh
```diff
@@ -0,0 +1,488 @@
...
+    if (index < num_buckets) {
+      IdxT prev = (index == 0) ? 0 : histogram[index - 1];
+      IdxT cur  = histogram[index];
+
+      // one and only one thread will satisfy this condition, so only write once
+      if (prev < k && cur >= k) {
+        counter->k            = k - prev;
+        counter->previous_len = counter->len;
+        counter->len          = cur - prev;
+        counter->bucket       = index;
```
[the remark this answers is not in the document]
yong-wang: Yes, it's done by the condition `current_value < k` of the `for` loop.
At the end of the loop, `current_value` is set to the last prefix sum that has been computed. When it >= k, we have found the right bucket, don't need scan further, and we break the loop.

Sorry, the name `current_value` is too vague to reason about its usage, maybe change it to `last_prefix_sum`?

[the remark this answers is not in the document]
yong-wang: It's done by the condition `current_value < k` of the for loop.
At the end of the loop, current_value is set to the last prefix sum that has been computed. When it >= k, we have found the right bucket, don't need scan further, and we break the loop.

Sorry, the name `current_value` is too vague to reason about its usage, maybe change it to `last_prefix_sum`?
(Re-post, I deleted this comment by mistake, sorry)

cjnolet:
rerun tests

cjnolet:
rerun tests

cpp/include/raft/spatial/knn/detail/ivf_flat/bitonic_sort.cuh
achirkin: Closing this with the reference to the new issue https://github.com/rapidsai/raft/issues/591 .

cpp/bench/spatial/selection.cu
achirkin: Closing this with the reference to the new issue https://github.com/rapidsai/raft/issues/591 .

tfeher (changes requested):
Thanks Artem for addressing the issues and improving the documentation! I had  second look at the bitonic sort, here are a few additional comments.

cpp/include/raft/spatial/knn/detail/topk/bitonic_sort.cuh
```diff
@@ -0,0 +1,232 @@
...
+
+/**
+ * Warp-wide bitonic merge and sort.
+ * The data is strided among `warp_width` threads,
+ * e.g. calling `bitonic<4>(ascending=true).sort(arr)` takes a unique 4-element array as input of
+ * each thread in a warp and sorts them, such that for a fixed i, arr[i] are sorted within the
+ * threads in a warp, and for any i < j, arr[j] in any thread is not smaller than arr[i] in any
+ * other thread.
+ *
+ * As an example, assuming `Size = 4`, `warp_width = 16`, and `WarpSize = 32`, the layout is:
```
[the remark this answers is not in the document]
achirkin: Yep, sounds good

cpp/include/raft/spatial/knn/detail/topk/radix_topk.cuh
```diff
@@ -0,0 +1,562 @@
...
+        out_buf[pos] = value;
+        if (out_idx_buf) { out_idx_buf[pos] = in_idx_buf ? in_idx_buf[i] : i; }
+        int bucket = calc_bucket<T, BitsPerPass>(value, start_bit, mask, greater);
+        atomicAdd(histogram_smem + bucket, IdxT(1));
+
+        if (counter_len == 1) {
+          out[k - 1]     = value;
+          out_idx[k - 1] = in_idx_buf ? in_idx_buf[i] : i;
+        }
+      } else if (out && prev_bucket < want_bucket) {
```
tfeher, line 224: Is out always a valid pointer? Three lines above we do not check, but here we do. Please make it consistent.

tfeher (changes requested):
Hi Artem, I had a second look on radix_topk. It looks good overall, just found a few minor things.

cpp/include/raft/spatial/knn/detail/topk/warpsort_topk.cuh
```diff
@@ -0,0 +1,844 @@
...
+    (see the usage of LaunchThreshold<warp_sort_immediate>::len_factor_for_choosing).
+
+    Example:
+      __global__ void kernel() {
+        warp_sort_immediate<...> queue(...);
+        int warp_id = threadIdx.x / WarpSize;
+        int lane_id = threadIdx.x % WarpSize;
+
+        // way 1, [0, len) is same for the whole warp
+        queue.add(in, 0, len);
```
tfeher, line 97:
```suggestion
        queue.load(in, 0, len);
```

tfeher (changes requested):
I have finished a second review. Please find a few more suggestions for improving the docs. Apart from these pending items, the PR looks good to me.

cpp/include/raft/spatial/knn/detail/topk/warpsort_topk.cuh
```diff
@@ -0,0 +1,881 @@
...
+/**
+ * A fixed-size warp-level priority queue.
+ * By feeding the data through this queue, you get the `k <= Capacity`
+ * smallest/greatest values in the data.
+ *
+ * @tparam Capacity
+ *   maximum number of elements in the queue.
+ * @tparam Ascending
+ *   which comparison to use: `true` means `<`, collect the smallest elements,
+ *                                             `false` means `>`, collect the greatest elements.
```
tfeher, line 142: Too much whitespace.

tfeher (approved):
Thanks Artem for addressing the issues! Apart from two more nitpicks, the PR looks good to me. Pre-approving.

cpp/include/raft/spatial/knn/detail/topk/warpsort_topk.cuh
```diff
@@ -0,0 +1,881 @@
...
+
+    After adding is finished, function done() should be called. And finally,
+    store() is used to get the top-k result.
+
+    Example:
+      __global__ void kernel() {
+        block_sort<warp_sort_immediate, ...> queue(...);
+
+        // way 1, [0, len) is same for the whole block
+        queue.add(in, 0, len);
```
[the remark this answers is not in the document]
achirkin: Sorry :)

cpp/bench/spatial/selection.cu
```diff
@@ -0,0 +1,150 @@
...
+  {1000, 10000, 1, true},
+  {1000, 10000, 2, true},
+  {1000, 10000, 4, true},
+  {1000, 10000, 8, true},
+  {1000, 10000, 16, true},
+  {1000, 10000, 32, true},
+  {1000, 10000, 64, true},
+  {1000, 10000, 128, true},
+  {1000, 10000, 256, true},
+  // {1000, 10000, 512, true}, {1000, 10000, 1024, true}, {1000, 10000, 2048, true},
```
cjnolet, line 103: Are these settings for `k` failing? Can we remove the commented out bits in the meantime?

cjnolet (changes requested):
Thanks Artem! Just one little thing from me this time around. I don't see any obvious things in these changes that would break cuml downstream but a PR should be opened and pinned to this branch before merging just to make sure.

cjnolet:
rerun tests

cpp/bench/spatial/selection.cu
achirkin: Yes, since only the radix algorithm supports larger values of `k`.

cjnolet (approved):
LGTM

cjnolet:
@gpucibot merge

PR #553, opened 2022-03-11
Disabling benchmarks building by default.

cjnolet:
This was an oversight because this will break all consumers downstream who are not explicitly turning this off in their `get_raft`. We should not build these by default (and we should eventually turn off compiling gtests by default as well).


seunghwak (approved):

cjnolet:
@gpucibot merge

cjnolet:
rerun tests

PR #564, opened 2022-03-18
Don't generate a gtest-config for build directories

robertmaynard:
Since the install rules are disabled due to not being in an install tree we need to also remove the build directory logic


robertmaynard:
Closing for #565

Issue #568, opened 2022-03-20
[BUG] Fused L2 kNN is broken for larger inputs


zbjornson:
**Describe the bug**
The new fused L2 kNN implementation (ccd5d755719cd6be570405d618de485b62202ca2) doesn't work properly for certain sizes(?) of input data and causes at least t-SNE and UMAP in cuML to fail. I don't know the exact criteria to trigger this, but there's a repro below.

**Steps/Code to reproduce bug**
```cpp
#include <cstdint>
#include <cstddef>
#include <iostream>
#include <vector>
#include <random>
#include <cuml/manifold/tsne.h>
#include <raft/handle.hpp>
#include <rmm/device_uvector.hpp>

void make_blobs(float* data, std::size_t nRows, std::size_t nCols, std::size_t nClust) {
	std::random_device rd{};
	uint64_t randomSeed = rd();

	std::mt19937 gen(randomSeed);
	std::uniform_real_distribution<float> urd{};

	std::vector<std::normal_distribution<float>> generators;
	for (size_t i = 0; i < nClust * nCols; i++) {
		float mean = 100 * urd(gen);
		float stdDev = 10 * urd(gen);
		generators.push_back(std::normal_distribution<float>{mean, stdDev});
	}

	for (size_t j = 0; j < nCols; j++)
		for (size_t i = 0; i < nRows; i++)
			data[j * nRows + i] = generators[(i % nClust) * nCols + j](gen);
}

int main() {
	raft::handle_t handle;
	auto stream = handle.get_stream();

	std::size_t nRows = 131'072; // 🥴 100'000 won't trigger it, not sure what exact criteria is
	std::size_t nCols = 8;

	ML::TSNEParams params{
		.dim=2,
		.n_neighbors=60,
		.perplexity=20,
		.max_iter=1500,
		.min_grad_norm=0.f,
		.algorithm=ML::TSNE_ALGORITHM::FFT
	};

	float* hdata = new float[nRows * nCols];
	make_blobs(hdata, nRows, nCols, 8);
	rmm::device_uvector<float> in(nRows * nCols, stream);
	raft::update_device(in.data(), hdata, nRows * nCols, stream);
	delete[] hdata;

	rmm::device_uvector<float> out(nRows * params.dim, stream);

	float klDiv;

	ML::TSNE_fit(
		handle,
		in.data(),
		out.data(),
		nRows,
		nCols,
		nullptr,
		nullptr,
		params,
		&klDiv
	);

[...]

PR #572, opened 2022-03-22
[REVIEW] Disable fused L2 kNN kernel

zbjornson:
Temporary remedy for #568


zbjornson:
I don't know anything about the ball cover code/tests that are failing so likely won't be able to debug this. I thought the fused L2 kNN kernel was a drop-in optimization of the non-fused ones, so I'm surprised anything would fail by disabling it.

I'll close this since @mdoijade is debugging the issue, unless someone else wants to pick this up.

cjnolet:
@zbjornson no worries, the developer of this primitive is going to look into the issue.

PR #575, opened 2022-03-22
[REVIEW] Add missing thrust includes to transpose.cuh

zbjornson:
Including `raft/linalg/transpose.cuh` appears to work today, but a few weeks ago it didn't because of these missing includes. Either way, these should be here because they're used.

I can't figure out how to get include-what-you-use to process .cuh files, but that would be a nice check for all of the RAPIDS repos.


cjnolet:
rerun tests

cjnolet:
rerun tests

cjnolet:
@gpucibot merge

PR #583, opened 2022-03-22
[FIX] Link to `cuco` and add faiss `EXCLUDE_FROM_ALL` option

trxcllnt:
This PR includes a few fixes to support source-only builds:
1. Defines linkage to `cuco::cuco` if the `RAFT_ENABLE_cuco_DEPENDENCY` variable is set, not if `cuco_ADDED` is true
2. Adds a flag to control the `EXCLUDE_FROM_ALL` for the faiss dependency. This flag can be off for conda builds, but true for C++-only source builds
3. Writes `version_config.hpp` header and fixes a potential GoogleBench issue


cjnolet (approved):
LGTM

cjnolet:
@gpucibot merge

PR #585, opened 2022-03-23
Making cuco, thrust, and mdspan optional dependencies. 

cjnolet:
In addition to the cuco dependency, the following changes are included:
1. ability to turn off thrust and mdspan dependencies (rmm is still required)
2. compiling libraries now defaults to the same setting of `BUILD_TESTS` (tests are still enabled)
3. cuco dependency is disabled by default (unless distance component is enabled)
4. the headers which are safe to expose in public APIs are moved over to `core/` directory.


cpp/CMakeLists.txt
```diff
@@ -160,7 +160,10 @@ target_link_libraries(raft INTERFACE
         CUDA::cudart
         CUDA::cusparse
         rmm::rmm
+<<<<<<< Updated upstream
         $<$<BOOL:${RAFT_ENABLE_cuco_DEPENDENCY}>:cuco::cuco>
+=======
+>>>>>>> Stashed changes
```
dantegd:
```suggestion
        $<$<BOOL:${RAFT_ENABLE_cuco_DEPENDENCY}>:cuco::cuco>
```

Remnants from git merge

dantegd (approved):
Change looks good except for one minor thing so pre-approving from my side

cpp/CMakeLists.txt
cjnolet: Lol, yep CI just informed me

cjnolet:
rerun tests

cjnolet:
rerun tests

cjnolet:
rerun tests

cjnolet:
rerun tests

cjnolet:
rerun tests

cjnolet:
rerun tests

cjnolet:
rerun tests

BUILD.md
```diff
@@ -53,6 +53,11 @@ The following example will download the needed dependencies and install the RAFT
 ./build.sh libraft --install
 ```
 
+The `--no-thrust` flag can be used to install the headers with minimal dependencies:
+```bash
+./build.sh libraft --install --no-thrust
```
robertmaynard, line 58: maybe `--no-deps`? instead of `--no-thrust` if this captures more dependencies than thrust

cpp/CMakeLists.txt
```diff
@@ -59,12 +61,22 @@ option(DISABLE_OPENMP "Disable OpenMP" OFF)
...
 option(RAFT_COMPILE_NN_LIBRARY "Enable building raft nearest neighbors shared library instantiations" OFF)
 option(RAFT_COMPILE_DIST_LIBRARY "Enable building raft distant shared library instantiations" OFF)
 option(RAFT_ENABLE_NN_DEPENDENCIES "Search for raft::nn dependencies like faiss" ${RAFT_COMPILE_LIBRARIES})
 
-option(RAFT_ENABLE_cuco_DEPENDENCY "Enable cuCollections dependency" ON)
+option(RAFT_ENABLE_cuco_DEPENDENCY "Enable cuCollections dependency" OFF)
+option(RAFT_ENABLE_mdspan_DEPENDENCY "Enable mdspan dependency" ON)
+option(RAFT_ENABLE_thrust_DEPENDENCY "Enable Thrust dependency" ON)
+if(distance IN_LIST raft_FIND_COMPONENTS OR RAFT_COMPILE_LIBRARIES OR RAFT_COMPILE_DIST_LIBRARY)
+  set(RAFT_ENABLE_cuco_DEPENDENCY ON)
```
robertmaynard, line 73: If we always force `RAFT_ENABLE_cuco_DEPENDENCY` to be ON if `distance` is an enabled component I don't think we should expose it as an option to the user.

The user will be confused on why the value they set in `RAFT_ENABLE_cuco_DEPENDENCY` is enabled depending on other state,

cpp/cmake/thirdparty/get_cuco.cmake
```diff
@@ -19,8 +19,8 @@ function(find_and_configure_cuco VERSION)
     if(RAFT_ENABLE_cuco_DEPENDENCY)
```
robertmaynard, line 19: We should just not call `find_and_configure_cuco` when `RAFT_ENABLE_cuco_DEPENDENCY` is enabled, instead of having the switch inside the function itself.

cpp/cmake/thirdparty/get_libcudacxx.cmake
```diff
@@ -14,10 +14,13 @@
 
 # This function finds libcudacxx and sets any additional necessary environment variables.
 function(find_and_configure_libcudacxx)
-  include(${rapids-cmake-dir}/cpm/libcudacxx.cmake)
 
-  rapids_cpm_libcudacxx(BUILD_EXPORT_SET raft-exports
-                        INSTALL_EXPORT_SET raft-exports)
+  if(RAFT_ENABLE_cuco_DEPENDENCY)
```
robertmaynard, line 18: Likewise here, we should hoist this out to where we call `find_and_configure_libcudacxx`

robertmaynard (changes requested):

robertmaynard:
To build raft by default I had to disable the following tests since they pull in faiss headers, which isn't enabled by default.

- test/sparse/add.cu
- test/sparse/convert_coo.cu
- test/sparse/convert_csr.cu
- test/sparse/connect_components.cu
- test/sparse/csr_row_slice.cu
- test/sparse/csr_to_dense.cu
- test/sparse/csr_transpose.cu
- test/sparse/degree.cu
- test/sparse/dist_coo_spmv.cu
- test/sparse/distance.cu
- test/sparse/filter.cu
- test/sparse/knn.cu
- test/sparse/knn_graph.cu
- test/sparse/linkage.cu
- test/sparse/norm.cu
- test/sparse/reduce.cu
- test/sparse/row_op.cu
- test/sparse/sort.cu
- test/spatial/knn.cu
- test/spatial/fused_l2_knn.cu
- test/spatial/haversine.cu
- test/spatial/ball_cover.cu
- test/spatial/epsilon_neighborhood.cu
- test/spatial/faiss_mr.cu
- test/spatial/selection.cu
- test/stats/trustworthiness.cu

cjnolet:
> To build raft by default I had to disable the following tests since they pull in faiss headers, which isn't enabled by default.

@robertmaynard most of those files don't include faiss headers, but I do get your point that the tests shouldn't be getting built without FAISS if that's a requirement.

robertmaynard:
> > To build raft by default I had to disable the following tests since they pull in faiss headers, which isn't enabled by default.
> 
> @robertmaynard most of those files don't include faiss headers, but I do get your point that the tests shouldn't be getting built without FAISS if that's a requirement.

You are correct I over removed tests, I don't know how I captured too many. But here is the full breakdown of what includes faiss / cuco headers.

Requires FAISS directly:
- test/spatial/fused_l2_knn.cu
- test/spatial/faiss_mr.cu

Requires FAISS via `knn/detail/knn_brute_force_faiss.cuh`:
- test/sparse/knn_graph.cu
- test/sparse/linkage.cu
- test/sparse/connect_components.cu
- test/sparse/csr_transpose.cu
- test/spatial/knn.cu
- test/spatial/epsilon_neighborhood.cu
- test/stats/selection.cu
- test/stats/trustworthiness.cu

Requires FAISS via `haversine_distance.cuh`:
- test/spatial/haversine.cu
- test/spatial/ball_cover.cu

Requires cuco:
- test/sparse/dist_coo_spmv.cu
- test/sparse/distance.cu
- test/sparse/knn.cu

cjnolet:
rerun tests

cpp/CMakeLists.txt
```diff
@@ -364,7 +385,7 @@ Imported Targets:
 set(code_string
 [=[
 
-if(NOT TARGET raft::Thrust)
+if(RAFT_ENABLE_thrust_DEPENDENCY AND NOT TARGET raft::Thrust)
```
robertmaynard, line 388: The variable `RAFT_ENABLE_thrust_DEPENDENCY` doesn't exist when users of raft call `find_package(raft)`.

Instead you should conditionally append this logic to `code_string` instead

robertmaynard: Something like:
```cmake
set(code_string )
if(RAFT_ENABLE_thrust_DEPENDENCY)
  string(APPEND code_string
  [=[
  if(NOT TARGET raft::Thrust)
    thrust_create_target(raft::Thrust FROM_OPTIONS)
  endif()
  ]=])
endif()

string(APPEND code_string
[=[
...
```

cjnolet:
rerun tests

cpp/include/raft/common/logger.hpp
```diff
@@ -15,284 +15,4 @@
...
-#endif
-
-#if (RAFT_ACTIVE_LEVEL >= RAFT_LEVEL_CRITICAL)
-#define RAFT_LOG_CRITICAL(fmt, ...) \
-  raft::logger::get(RAFT_NAME).log(RAFT_LEVEL_CRITICAL, fmt, ##__VA_ARGS__)
-#else
-#define RAFT_LOG_CRITICAL(fmt, ...) void(0)
-#endif
-/** @} */
+#include <raft/core/logger.hpp>
```
divyegala, line 18: Deprecation message

cpp/include/raft/common/nvtx.hpp
```diff
@@ -16,140 +16,4 @@
...
-  range(const range&) = delete;
-  range(range&&)      = delete;
-  auto operator=(const range&) -> range& = delete;
-  auto operator=(range&&) -> range&                = delete;
-  static auto operator new(std::size_t) -> void*   = delete;
-  static auto operator new[](std::size_t) -> void* = delete;
-};
-
-}  // namespace raft::common::nvtx
+#include <raft/core/nvtx.hpp>
```
divyegala, line 19: Deprecation message

cpp/include/raft/core/nvtx.hpp
```diff
@@ -0,0 +1,155 @@
...
+ * }
+ *
+ * void some_function(int k){
+ *   // This NVTX range appears in the timeline named "my application" in Nsight Systems.
+ *   common::nvtx::range<my_app_domain> fun_scope( r{"some_function_%d", k};
+ *   // some logic inside the loop
+ * }
+ * \endcode
+ */
+namespace raft::common::nvtx {
```
divyegala, line 67: Is this correct?

docs/source/cuda_cpp.rst
```diff
@@ -1,7 +1,7 @@
 CUDA/C++ API
 ============
 
-RAFT is header-only but provides optional shared libraries to speed up compile times for larger projects.
+RAFT is a header-only C++ with optional shared libraries that can speed up compile times for larger projects.
```
divyegala, line 4: The part `header-only C++ with` doesn't read well to me

divyegala (commented):
Some small comments. Should we also have a standard namespace convention for headers in `raft/core`? There's some variation amongst them at this moment

build.sh
```diff
@@ -36,6 +36,8 @@ HELP="$0 [<target> ...] [<flag> ...]
    --compile-libs   - compile shared libraries for all components
    --compile-nn     - compile shared library for nn component
    --compile-dist   - compile shared library for distance component
+   --minimal-deps   - disables dependencies like thrust so they can be overridden.
```
rlratzel, line 39: I didn't see where `--minimal-deps` is used/checked for in this script.

conda/recipes/libraft_headers/build.sh
```diff
@@ -2,3 +2,4 @@
 # Copyright (c) 2022, NVIDIA CORPORATION.
 
 ./build.sh libraft --install -v --allgpuarch
+``
```
rlratzel, line 5: typo?

cpp/cmake/thirdparty/get_mdspan.cmake
```diff
@@ -13,17 +13,19 @@
...
-    GLOBAL_TARGETS std::mdspan
-    BUILD_EXPORT_SET    raft-exports
-    INSTALL_EXPORT_SET  raft-exports
-    CPM_ARGS
-      GIT_REPOSITORY https://github.com/rapidsai/mdspan.git
-      GIT_TAG b3042485358d2ee168ae2b486c98c2c61ec5aec1
-      OPTIONS "MDSPAN_ENABLE_CUDA ON"
-              "MDSPAN_CXX_STANDARD ON"
-  )
+  if(RAFT_ENABLE_mdspan_DEPENDENCY)
```
rlratzel, line 16: I'm curious why this `if` block is here instead of around the call to `find_and_configure_mdspan` below, as the other `get_*.cmake` files did.

cpp/cmake/thirdparty/get_thrust.cmake
```diff
@@ -14,11 +14,13 @@
 
 # Use CPM to find or clone thrust
 function(find_and_configure_thrust)
-    include(${rapids-cmake-dir}/cpm/thrust.cmake)
+    if(RAFT_ENABLE_thrust_DEPENDENCY)
```
rlratzel, line 17: same question: why not put `if` around call to function below?

robertmaynard (approved):

cjnolet:
rerun tests

cjnolet:
@gpucibot merge

msadang (approved):
Lgtm

PR #592, opened 2022-03-25
Update cuco version.

vyasr:
This PR updates the commit hash for cuCollections to include the changes in NVIDIA/cuCollections#138. cudf depends on those changes in 22.04, and some of our CI builds of cudf are finding the version of cuco installed by raft and then failing, so I'm making this change to 22.04 even though we're in code freeze. Happy to work with ops an an alternate solution if there are concerns about the update, though.


cjnolet (approved):

cjnolet:
@gpucibot merge

PR #601, opened 2022-03-30
mdspan/mdarray template functions and utilities

divyegala:


cpp/include/raft/mdarray.hpp
```diff
@@ -647,4 +687,30 @@ auto make_device_vector(raft::handle_t const& handle, size_t n)
 {
   return make_device_vector<ElementType>(n, handle.get_stream());
 }
+
+template <typename host_mdspan_type, typename = is_host_mdspan_t<host_mdspan_type>>
```
divyegala, line 691: Should the second template check with `is_host_mdspan_t` be a `static_assert` to display an error message to the user?

cpp/include/raft/mdarray.hpp
```diff
@@ -642,9 +721,111 @@ auto make_device_vector(size_t n, rmm::cuda_stream_view stream)
...
+}
+
+template <typename mdarray_type, std::enable_if_t<is_mdarray_v<mdarray_type>>* = nullptr>
+auto flatten(const mdarray_type& mda)
+{
+  return flatten(mda.view());
 }
+
+template <typename ElementType, typename LayoutType>
+constexpr auto flatten(host_vector_view<ElementType, LayoutType> h_vv)
```
divyegala, line 757: I only added these overloads for a compile-time no-op when `flatten` is called with a `vector` or a `scalar`. However, since the code is not performance critical, I could simplify this and just keep the generic version.

cpp/include/raft/mdarray.hpp
```diff
@@ -57,6 +106,27 @@ template <typename ElementType,
...
+ */
+template <typename T>
+inline constexpr bool is_host_mdspan_v = __is_host_mdspan<T, is_mdspan_v<T>>::value;
+
+/**
+ * @\brief Boolean to determine if template type T is either raft::host_mdspan/raft::device_mdspan
+ *         or their derived types
+ */
+template <typename T>
+inline constexpr bool is_host_or_device_mdspan_v = is_device_mdspan_v<T> or is_host_mdspan_v<T>;
```
trivialfis, line 128: is there any other type of mdspan?

cpp/include/raft/mdarray.hpp
```diff
@@ -185,6 +254,18 @@ class mdarray {
     return const_view_type(c_.data(), map_, cp_.make_accessor_policy());
   }
 
+  /**
+   * @brief Get an implicitly constructed mdspan that can be passed down to CUDA kernels.
+   */
+  operator view_type() noexcept { return view_type(c_.data(), map_, cp_.make_accessor_policy()); }
```
trivialfis, line 260:
```suggestion
  operator view_type() noexcept { return view(); }
```

cpp/include/raft/mdarray.hpp
```diff
@@ -185,6 +254,18 @@ class mdarray {
...
+  /**
+   * @brief Get an implicitly constructed mdspan that can be passed down to CUDA kernels.
+   */
+  operator const_view_type() const noexcept
+  {
+    return const_view_type(c_.data(), map_, cp_.make_accessor_policy());
```
trivialfis, line 266:
```suggestion
    return view();
```

cpp/include/raft/core/mdarray.hpp
```diff
@@ -57,6 +87,117 @@ template <typename ElementType,
...
+struct __is_array_interface<
+  T,
+  std::void_t<decltype(__takes_an_array_interface_ptr(std::declval<T*>()))>> : std::true_type {
+};
+
+/**
+ * @\brief Boolean to determine if template type T is raft::array_interface or derived type
+ */
+template <typename T>
+inline constexpr bool is_array_interface_v = __is_array_interface<std::remove_const_t<T>>::value;
```
divyegala, line 199: I propose that we give users two options here:
1. They straightforward implement `array_interface`
2. For users with existing owning types, we just ask that they add a `view()` method that returns a host_mdspan or device_mdspan or a derived type

Point 1 is already implemented, and point 2 allows users to update their existing their code minimally.

cpp/include/raft/core/mdarray.hpp
```diff
@@ -642,9 +743,145 @@ auto make_device_vector(size_t n, rmm::cuda_stream_view stream)
...
+ *         ContainerPolicy
+ */
+template <typename mdarray_type, std::enable_if_t<is_mdarray_v<mdarray_type>>* = nullptr>
+auto flatten(const mdarray_type& mda)
+{
+  return flatten(mda.view());
+}
+
+template <typename ElementType, typename LayoutType>
+constexpr auto flatten(host_vector_view<ElementType, LayoutType> h_vv)
```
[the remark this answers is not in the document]
trivialfis: I think you can remove them.  The mdspan is designed to be flexible, having all these overloads doesn't make sense from my perspective.  I would prefer one template function that works for all types.

cpp/include/raft/core/mdarray.hpp
```diff
@@ -642,9 +792,146 @@ auto make_device_vector(size_t n, rmm::cuda_stream_view stream)
...
+ * @tparam Extents raft::extents for dimensions
+ * @param mda raft::array_interace implementing object
+ * @param new_shape Desired new shape of the input
+ * @return raft::host_mdspan or raft::device_mdspan, depending on the underlying
+ *         ContainerPolicy
+ */
+template <typename array_interface_type,
+          size_t... Extents,
+          std::enable_if_t<is_array_interface_v<array_interface_type>>* = nullptr>
+auto reshape(const array_interface_type& mda, extents<Extents...> new_shape)
```
trivialfis, line 932: I haven't looked into the array interface in raft yet.  But if this function accepts (or will potentially accept in the future) an owning type, and might change the size of the underlying memory buffer, one might want to accept a raft handle as a parameter.

cpp/include/raft/core/mdarray.hpp
```diff
@@ -185,6 +263,15 @@ class mdarray {
     return const_view_type(c_.data(), map_, cp_.make_accessor_policy());
   }
 
+  /**
+   * @brief Get an implicitly constructed mdspan that can be passed down to CUDA kernels.
+   */
+  operator view_type() noexcept { return view(); }
```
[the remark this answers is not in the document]
trivialfis: Probably no, if you don't want implicit conversion.

cpp/include/raft/core/mdarray.hpp
```diff
@@ -176,15 +317,24 @@ class mdarray {
   /**
    * @brief Get a mdspan that can be passed down to CUDA kernels.
    */
-  auto view() noexcept { return view_type(c_.data(), map_, cp_.make_accessor_policy()); }
+  view_type view() noexcept { return view_type(c_.data(), map_, cp_.make_accessor_policy()); }
```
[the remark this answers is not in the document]
divyegala: Sounds good, this looks clean.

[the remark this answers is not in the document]
trivialfis: Hmm .. why do you need to make that check? If you want to emit error early, you can use:
``` c++
template <typename ArrayInterface, typename std::enable_if_t<is_mdspan_v<decltype(std::declval<A>().view())>* = nullptr>
auto my_func(ArrayInterface const& arr) {
}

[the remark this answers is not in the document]
Comment (trivialfis) [reply cpp/include/raft/core/mdarray.hpp:320]: Having said that, I'm just trying to avoid c++ virtual table.  Might not be the most rational choice but from my experience template is easier to handle than virtual functions.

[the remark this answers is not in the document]
Comment (trivialfis) [reply cpp/include/raft/core/mdarray.hpp:320]: Alternatively, `std::is_convertible` might help.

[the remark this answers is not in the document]
Comment (divyegala) [reply cpp/include/raft/core/mdarray.hpp:320]: `std:is_convertible` does not help since `mdarray` is a class-template and we do not know ahead of time if a template-type `T` could possibly a class-template of `mdarray`. I like your solution though, we can check if the `view()` returns a type which satisfies `is_mdspan_v`

File: cpp/include/raft/core/mdarray.hpp [hunk 11]
```diff
@@ -27,6 +27,12 @@
 #include <rmm/cuda_stream_view.hpp>
 
 namespace raft {
+/**
+ * @\brief Dimensions extents for raft::host_mdspan or raft::device_mdspan
+ */
+template <size_t... ExtentsPack>
```
wphicks, line 33: Include `stdint.h` for `size_t` or include `cstdint` and use `std::size_t` where possible

wphicks (changes requested):

cjnolet (approved):
We were largely discussing these designs throughout your progress on this PR and I don't see anything particularly concerning. LGTM

divyegala:
rerun tests

cjnolet:
@gpucibot merge

PR #602, opened 2022-03-30
[FEA] Enable building static libs

trxcllnt:
This PR allows building libcugraphops static libs via CMake option [`-DBUILD_SHARED_LIBS=ON|OFF`](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html).

~I was seeing a linker error due to not having `-fPIC` enabled, but I wouldn't have expected this to affect static libs. I'll investigate some more and turn it off if possible, but for now `-fPIC` is enabled.~


robertmaynard (commented):
LGTM to me

trxcllnt:
rerun tests

trxcllnt:
rerun tests

cjnolet:
@gpucibot merge

PR #604, opened 2022-03-31
fix issue in fusedL2knn which happens when rows are multiple of 256

mdoijade:
This PR fixes issue - https://github.com/rapidsai/raft/issues/568 and https://github.com/rapidsai/cuml/issues/4624
-- fix issue in fusedL2knn which happens when rows are multiple of 256.
-- make index value to be size_t to avoid int overflow though this doesn't hamper these issues but it may for higher input sizes. 
-- also add some additional test cases in fusedL2knn test.


mdoijade:
@cjnolet does this PR need to creat cuML verification PR as well? or cuML CI tests are triggered automatically.

cjnolet (approved):
LGTM. I think we should target this for 22.06 since it's somewhat of a minor bug.

cjnolet:
@gpucibot merge

PR #605, opened 2022-03-31
Moving kmeans from cuml to Raft

### LINKED ISSUE #28: [FEA] Move k-means from cuml to RAFT
cuML exposes a Kmeans feature. Another Kmeans implementation has been introduced as part of moving Spectral clustering to RAFT.
We should avoid multiplying implementations for the same feature.

We could move cuML's kmeans backend to RAFT and use it in spectral clustering (and potentially in cuGraph as well later on).

lowener:
This PR replace the current KMeans of Raft with cuml's implementation.
Closes #28.
It is using the new `device_*_view` for the API.


cpp/include/raft/cluster/detail/kmeans.cuh
```diff
@@ -19,967 +19,1056 @@
...
+                                           X,
+                                           centroids,
+                                           minClusterDistance.view(),
+                                           L2NormX.view(),
+                                           L2NormBuf_OR_DistBuf,
+                                           workspace,
+                                           metric,
+                                           stream);
+
+  // RAFT_LOG_INFO(" k-means++ - Sampled %d/%d centroids", n_clusters_picked, n_clusters);
```
cjnolet, line 161: Use `RAFT_LOG_DEBUG` or `RAFT_LOG_TRACE` here if this is something we should keep around for debugging.

cpp/include/raft/cluster/detail/kmeans.cuh
```diff
@@ -19,967 +19,1056 @@
...
+                 stream);
+
+      ++n_clusters_picked;
+      /// <<< End of Step-4 >>>
+    }
 
-    // Move to another row
-    i += blockDim.x * gridDim.x;
-  }
+    // RAFT_LOG_INFO(" k-means++ - Sampled %d/%d centroids", n_clusters_picked, n_clusters);
```
cjnolet, line 257: Same here- use debug or trace if we should keep this.

cpp/include/raft/cluster/detail/kmeans.cuh
```diff
@@ -19,967 +19,1056 @@
...
-    // Get cluster size from global memory
-    clusterSize_private = clusterSizes[gidy];
-
-    // Add vector entries to centroid matrix
-    //   vector entris are determined by global x-index
-    gidx = threadIdx.x + blockIdx.x * blockDim.x;
-    while (gidx < d) {
-      centroids[IDX(gidx, gidy, d)] /= clusterSize_private;
-      gidx += blockDim.x * gridDim.x;
+  /*RAFT_LOG_INFO(
```
cjnolet, line 322: Any commented out log statements should be using `RAFT_LOG_DEBUG` or `RAFT_LOG_TRACE`. The purpose of having different logging levels is to remove the need for commenting them out in the code.

cpp/test/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@
 
 # keep the files in alphabetical order!
 add_executable(test_raft
+    test/cluster/kmeans.cu
```
cjnolet, line 19: Do we also need to move over the gather tests?

cjnolet (changes requested):
Implementation is looking great so far. Most of my comments are related to small API things.

cpp/test/CMakeLists.txt
lowener: True it should be tested as well. I moved the gather tests.

cpp/include/raft/cluster/detail/kmeans.cuh
```diff
@@ -19,967 +19,1056 @@
...
-           index_type_t maxiter,
-           const value_type_t* __restrict__ obs,
-           index_type_t* __restrict__ codes,
-           value_type_t& residual,
-           index_type_t& iters,
-           unsigned long long seed = 123456)
+template <typename DataT, typename IndexT>
+void kmeans_predict(handle_t const& handle,
+                    const KMeansParams& params,
+                    raft::device_matrix_view<const DataT> X,
```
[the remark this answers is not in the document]
lowener: I added const reference specifier to reduce the overhead but it's true, the whole `view` object is not needed.

cpp/include/raft/cluster/detail/kmeans.cuh
```diff
@@ -19,967 +19,1056 @@
...
-                                           index_type_t* __restrict__ clusterSizes)
+template <typename DataT, typename IndexT>
+void kmeans_fit_main(const raft::handle_t& handle,
+                     const KMeansParams& params,
+                     raft::device_matrix_view<const DataT> X,
+                     raft::device_vector_view<const DataT> weight,
+                     rmm::device_uvector<DataT>& centroidsRawData,
+                     DataT& inertia,
+                     IndexT& n_iter,
+                     rmm::device_uvector<char>& workspace)
```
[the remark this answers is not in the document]
lowener: The workspace can't be a device_vector_view because it needs to be resized inside the function.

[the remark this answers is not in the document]
cjnolet: Can you add include guards in core/logger.hpp for now? I think this is happening because both the deprecated and new header files are being included somewhere.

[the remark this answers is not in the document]
lowener: I added the include guards but it didn't solve it. Adding inline to the 3 functions being define multiple times solved it

cpp/test/cluster/kmeans.cu
```diff
@@ -0,0 +1,181 @@
...
+      ss << "Actual: " << raft::arr2Str(d_labels.data(), 25, "d_labels", stream);
+      std::cout << (ss.str().c_str()) << '\n';
+      std::cout << "Score = " << score << '\n';
+    }
+  }
+
+  void SetUp() override { basicTest(); }
+
+ protected:
+  cudaStream_t stream = 0;
```
cjnolet, line 139: This test predates the use of the handle in the testing layer. We should create a handle here and use the stream from that instead of creating an explicit stream. Also less prone to errors.

cpp/include/raft/cluster/detail/kmeans.cuh
```diff
@@ -19,967 +19,1072 @@
...
+                                              params,
+                                              X,
+                                              centroids,
+                                              minClusterAndDistance.view(),
+                                              L2NormX.view(),
+                                              L2NormBuf_OR_DistBuf,
+                                              workspace,
+                                              metric);
+
+  thrust::transform(handle.get_thrust_policy(),
```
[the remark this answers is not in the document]
lowener: Because of the type of `minClusterAndDistance` I can't use binaryOp: the two `InType` are not the same.
I added a TODO comment.

cjnolet:
@lowener this is looking great. Once corresponding cuml and cugraph PRs are ready to merge then I think this is just about ready to go.

cjnolet:
rerun tests

cjnolet:
@gpucibot merge

cjnolet (approved):

Issue #406, opened 2021-12-03
[FEA] RNG device interface


MatthiasKohl:
**Is your feature request related to a problem? Please describe.**
I'd like to add a device interface for RNG. Is this something RAFT can support?

**Describe the solution you'd like**
It would be nice to either have a templated structure (templated on the kind of RNG generator from `random/detail/rng_impl.cuh`) which has methods to generate random numbers.

Alternatively, we could directly expose device functions templated on the kind of RNG generator from `random/detail/rng_impl.cuh`).

In addition to this, I'd like to add methods of generating random values that are not as easy to get right as one might think at first, in particular bounded integers.
See the context in this issue in particular: https://github.com/rapidsai/cugraph/issues/1979

MatthiasKohl:
I'd like to work on this if you agree that this functionality belongs into RAFT

MatthiasKohl:
Tagging @vinaydes since he was interested in fast generation of bounded integers and we have an implementation available.

cjnolet:
I think this functionality sounds reasonable to place in RAFT and it should be fine to wrap the device function through the public API and hide the implementatation details through the `detail` namespace. If this function isn't going to be usable at all on the host, we might also want to consider adding a namespace for it like `raft::random::device::`. 

BTW- a random walk primitive itself might also be useful to have in RAFT eventually. We're planning to consolidate and share more structures between RAFT and cugraph (e.g. csr/dcsr/coo, etc...), which might make this easier to do.

MatthiasKohl:
Thanks for considering it! While working on it, I noticed that what I want is basically almost exactly the `raft::random::detail::Generator` struct, but as a public API.

So I think it makes sense to simply have a sub-class of `raft::random::detail::Generator` in `raft::random::device` namespace (similar to the sub-class of `RngImpl`). Then we can add a few additional distributions and document the number of draws made on the underlying random generator since the caller will have to update their offset accordingly, but that is on the host side.

As for the random walk primitive, I think this makes sense as well, but I guess it will take a bit longer to make a primitive out of it.

vinaydes:
I am currently almost done with reimplementing the RNG class for addressing the exact issues mentioned here. New implementation provides device interface, fixes uniform float and double generation issues and also fixes the bounded integer generation issues. I am also adding a new PCG based generator.
The only issue is bunch of spills which are causing performance regression. I would create PR as soon as I fix those issues. Will update this issue once I do that.

MatthiasKohl:
[...]

PR #609, opened 2022-04-02
[FEA] Rng: expose host-rng-state in host-only API

MatthiasKohl:
Changes:
- Move `RngState` and `GeneratorType` to own header so they can be used in public APIs without a CUDA-compatible compiler (useful for python bindings etc.).
- Move the type of generator into the state object
- Add the `advance` method to the state object as well, allowing libraries to use `RngState` for host-only code
- Remove the `rng.hpp` header: if we want to expose the `Rng` object itself as a host-only object, the implementation/definition would have to be separated from the declaration, and this means it would be hard to keep a header-only library for these objects. Otherwise, there is simply no way that this `rng.hpp` actually fulfills its contract of a host-only header, so we remove it.

This is not the only way of creating a host-only API, but it's most likely the one which needs the least amount of changes in dependent libraries, and these libraries can basically "opt-in" to the host-only API.

@cjnolet Do you agree with this approach? I'd like to get this PR into 22.06 fairly soon, so we can finally go back to sane APIs and compile commands in cugraph-ops and cugraph.


MatthiasKohl:
This PR fixes #406

cpp/include/raft/random/rng_launch.cuh
```diff
@@ -0,0 +1,50 @@
...
+#ifndef __RNG_LAUNCH_H
+#define __RNG_LAUNCH_H
+
+#pragma once
+
+#include "detail/rng_launch.cuh"
+#include "rng_state.hpp"
+
+namespace raft {
+namespace random {
```
cjnolet, line 26: I like the separation between `rng_state` and the actual functions which use it to generate distributions. I think we should continue to use the `rng.cuh` as the main entrypoint for the functions, though. Can we just add the contents of this file and add a deprecation message for the individual pieces which we intend to remove?

cpp/include/raft/random/rng_state.hpp
```diff
@@ -0,0 +1,40 @@
...
+ * This file is deprecated and will be removed in release 22.06.
+ * Please use the cuh version instead.
+ */
+
+#ifndef __RNG_STATE_H
+#define __RNG_STATE_H
+
+#pragma once
+
+#include "detail/rng_state.hpp"
```
cjnolet, line 26: The implementation details of `rng_state` are so trivial that I think we might be okay just exposing them to the public API. I could really go either way with this, though....

cjnolet (changes requested):
Overall I really like the separation of the rng state and the flattened functions because they are more consistent w/ the rest of raft's public APIs. I suggest we still use `rng.cuh` for the flattened API functions if we can.

cpp/include/raft/random/detail/rng_state.hpp
```diff
@@ -0,0 +1,61 @@
...
+  /** curand-based philox generator */
+  GenPhilox = 0,
+  /** Permuted Congruential Generator */
+  GenPC
+};
+
+/**
+ * The RNG state used to keep RNG state around on the host.
+ */
+struct RngState {
```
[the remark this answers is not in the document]
MatthiasKohl: Makes sense, I'll move everything to the main header.

MatthiasKohl:
@cjnolet I believe I addressed your comments. I haven't fully tested all the possible deprecation messages, which are much more tricky since we want to keep the `rng.cuh` file.
But things compile at least and since there is no change to the actual code, I'm confident it works OK :)

cpp/include/raft/random/rng_launch.cuh
MatthiasKohl: I tried to do this, but it does mean it's a little more tricky to properly ensure that we can separate the device-only API from the rest: you'll see I'm using some macro mechanism to make sure that certain parts get deprecated if included through `rng.cuh` rather than `rng_device.cuh` (which is what you should be using in the future for the device-only API).

Please do check this mechanism again because I might have missed something!

cpp/include/raft/random/rng.hpp
```diff
@@ -18,6 +18,20 @@
...
 #pragma once
 
-#include "rng.cuh"
\ No newline at end of file
+#if defined(_MSC_VER)
+#pragma message(                                                               \
+  "raft/random/rng.hpp is deprecated and will be removed in a future relase. " \
+  "Please use raft/random/rng_state.hpp or raft/random/rng.cuh instead")
+#else
+#warning \
```
cjnolet, line 32: I had held off on doing a warning only because I Know this is immediately going to break a few consumers downstream who have enabled "warnings as errors". For this reason, I'm okay removing the warning tag.

cpp/include/raft/random/rng_device.cuh
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright (c) 2018-2022, NVIDIA CORPORATION.
```
cjnolet, line 2: Since this is a new file, we should just use the current year in the copyright.

cpp/include/raft/random/detail/rng_device.cuh
```diff
@@ -0,0 +1,720 @@
+/*
+ * Copyright (c) 2018-2022, NVIDIA CORPORATION.
```
cjnolet, line 2: Use only current year for new files

cpp/include/raft/random/detail/rng_impl.cuh
```diff
@@ -16,906 +16,306 @@
...
+  params.mu    = mu;
+  params.scale = scale;
+  RAFT_CALL_RNG_FUNC(rng_state, call_rng_kernel<1>, rng_state, stream, ptr, len, params);
 }
 
-template <typename GenType, typename OutType, typename LenType>
-DI void custom_next(
-  GenType& gen, OutType* val, SamplingParams<OutType, LenType> params, LenType idx, LenType stride)
+template <typename DataT, typename WeightsT, typename IdxT = int>
+void sampleWithoutReplacement(RngState& rng_state,
```
cjnolet, line 269: One very small nitpick I have here is that the handle should always be the first argument for primitives that accept one.

cpp/include/raft/random/detail/rng_impl_deprecated.cuh
```diff
@@ -0,0 +1,305 @@
+/*
+ * Copyright (c) 2018-2022, NVIDIA CORPORATION.
```
cjnolet, line 2: New file should have current year

cpp/include/raft/random/rng_state.hpp
```diff
@@ -0,0 +1,62 @@
+/*
+ * Copyright (c) 2018-2022, NVIDIA CORPORATION.
```
cjnolet, line 2: Use only current year

cjnolet (changes requested):
I found a few more trivial things in my second pass that I missed in the first pass but overall I think it looks great. Have you tried building cuml/cugraph with these changes yet just to make sure they build successfully? I don't know that we need to open full PRs to test this anymore unless there are direct changes to the public API, but it would be a good idea to check whether things like "warnings as errors" flags cause builds to fail.

MatthiasKohl:
@cjnolet I haven't tried the cuml/cugraph builds yet but I can do that once I've addressed your latest comments

cpp/include/raft/random/rng.hpp
MatthiasKohl: Indeed the builds weren't working, so I removed this warning again

MatthiasKohl: Actually, I'm using the same mechanism as in `raft/spatial/knn.hpp` now

cpp/include/raft/random/detail/rng_impl.cuh
MatthiasKohl: I removed the handle here since it wasn't being used in the implementation at all (the stream is already given as another parameter).
This also keeps the API in line with other RNG APIs

MatthiasKohl:
@cjnolet I've addressed the issues you mentioned, and I've been able to build cuml. I'm currently building cugraph as well, and the files requiring RNG seem to compile fine at least.

cjnolet (approved):
LGTM. Thanks @MatthiasKohl!

cjnolet:
@gpucibot merge

PR #615, opened 2022-04-08
Improve performance of select-top-k RADIX implementation

achirkin:
Improve the launch configuration when the `input_len` (length of a row) is small. This is done by allowing smaller block sizes and more blocks per row. The speedup can be anywhere between ~1.5x and 2x for `input_len < 1000`, or when both `batch_size` and `input_len` are small.


cjnolet:
@achirkin just checking in on the state of this PR for 22.06. Is it ready to be reviewed/merged?

achirkin:
> @achirkin just checking in on the state of this PR for 22.06. Is it ready to be reviewed/merged?

Yep, alongside with https://github.com/rapidsai/raft/pull/606 and https://github.com/rapidsai/raft/pull/618 (the latter should succeed after the former is merged).

tfeher (approved):

PR #618, opened 2022-04-14
Allow nullptr as input-indices argument of select_k

achirkin:
Allow passing `nullptr` as the input-indices-argument of `select_k`. This would imply the input indices are a range from `0` to `input_len - 1` and improve the performance by doing less global memory reads.

Also makes input pointers `const` as they should be.


achirkin:
Note, the tests are expected to fail until https://github.com/rapidsai/raft/pull/606 is merged.

cjnolet:
rerun tests

tfeher (approved):

achirkin:
rerun tests

cjnolet:
@gpucibot merge

PR #619, opened 2022-04-19
Message Flow Graphs Integration Proposal

kaatish:
Exposing cugraph-ops aggregation function via cugraph requires us to move the definitions of the message flow graphs to raft.
This PR is a rough initial draft of the new classes mfg_csr and mfg_csr_rev.
So far the following questions need to be answered :
- which directory/file should this be moved to
- what should be the names. message_flow_graph_view_t/message_flow_graph_transpose_view_t could be used here.

### DIFF
--- a/cpp/include/raft/sparse/csr.hpp
+++ b/cpp/include/raft/sparse/csr.hpp
@@ -178,5 +178,141 @@ void weak_cc(Index_* labels,
     labels, row_ind, row_ind_ptr, nnz, N, 0, N, stream, [](Index_) { return true; });
 }
 
+/**
+ * @brief CSR representation of a MFG.
+ *
+ * @tparam IdxT node index type
+ *
+ * Same as `mfg_ellpack` except the neighbors are in CSR format.
+ *
+ * @note Valid IDs: [0, 2^(sizeof(IdxT) * 8 - 1) - 2] (see `INVALID_ID`)
+ * @note This object does NOT own any of the underlying pointers and thus is
+ *       left to the calling code for their lifetime management.
+ * @note This is always an in-graph (`offsets` over the destination nodes).
+ *       See `fg_csr` for further explanation on this.
+ */
+template <typename IdxT>
+struct mfg_csr {
+  // handle sub-classes correctly
+  mfg_csr()          = default;
+  virtual ~mfg_csr() = default;
+
+  /**
+   * number of output nodes of this MFG. See `mfg_ellpack`.
+   * Note that input/output nodes are defined as in the regular MFG
+   */
+  IdxT n_out_nodes{0};
+  /** sample size: See `mfg_ellpack` */
+  IdxT sample_size{0};
+  /**
+   * number of input nodes of this MFG. See `mfg_ellpack`
+   * Note that input/output nodes are defined as in the regular MFG
+   */
+  IdxT* n_in_nodes{nullptr};
+  /**
+   * total number of edges in this graph (length of `indices` array)
+   * [single element, pinned memory]
+   */
+  IdxT* n_indices{nullptr};
+  /**
+   * Output nodes of this MFG. [len = `n_out_nodes`]
+   * For each output node, contains the index w.r.t. input nodes/features.
+   */
+  IdxT* out_nodes{nullptr};
+  /**
+   * Input nodes of this MFG. [len = `*n_in_nodes`]
+   * For each input node, contains the index w.r.t. input nodes/features of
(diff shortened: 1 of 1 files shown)


cpp/include/raft/sparse/csr.hpp
```diff
@@ -178,5 +178,141 @@ void weak_cc(Index_* labels,
     labels, row_ind, row_ind_ptr, nnz, N, 0, N, stream, [](Index_) { return true; });
 }
 
+/**
+ * @brief CSR representation of a MFG.
+ *
+ * @tparam IdxT node index type
+ *
+ * Same as `mfg_ellpack` except the neighbors are in CSR format.
```
seunghwak, line 186: Where is `mfg_ellpack`?

cpp/include/raft/sparse/csr.hpp
```diff
@@ -178,5 +178,141 @@ void weak_cc(Index_* labels,
...
+ *
+ * @note Valid IDs: [0, 2^(sizeof(IdxT) * 8 - 1) - 2] (see `INVALID_ID`)
```
seunghwak, line 188: Can't we use std::numeirc_limits<IdxT>::max() as INVALID ID? (cugrpah is using this as INVALID ID if vertex_t is unsigned).

https://github.com/rapidsai/cugraph/blob/branch-22.06/cpp/include/cugraph/graph.hpp#L419

We may be able to move this `invalid_vertex_id` machinery to raft and re-use (this will also make transition among raft/cugrpah/cugraph-ops smoother).

cpp/include/raft/sparse/csr.hpp
```diff
@@ -178,5 +178,141 @@ void weak_cc(Index_* labels,
...
+ * @note This object does NOT own any of the underlying pointers and thus is
+ *       left to the calling code for their lifetime management.
+ * @note This is always an in-graph (`offsets` over the destination nodes).
+ *       See `fg_csr` for further explanation on this.
```
seunghwak, line 192: What is `fg_csr`? `mfg_csr`? I think we can combine `mfg_csr` and `mfg_csr_rev` as something like `mfg_adjacency_list` with a template parameter similar to cugraph's graph_view_t.

cpp/include/raft/sparse/csr.hpp
```diff
@@ -178,5 +178,141 @@ void weak_cc(Index_* labels,
...
+   * offset of the neighborhood of that node in the `indices` array. It is of
+   * length `n_out_nodes + 1`.
+   */
+  IdxT* offsets{nullptr};
+  /**
+   * contains neighbor indices of every node belonging to this object. It is of
+   * length `n_indices`. For each node, the number of neighbors is
+   * `<= sample_size`
+   */
+  IdxT* indices{nullptr};
```
seunghwak, line 239: I am thinking about replacing this offsets + indices part with https://github.com/rapidsai/cugraph/blob/branch-22.06/cpp/include/cugraph/edge_partition_device_view.cuh#L36 from cugraph.

I think this class can be implemented with `edge_partition_device_view_base_t` + additional meta data (and cugraph's `edge_partition_device_view_t` for SG & MNMG are also implemented as `edge_partition_device_view_base_t` + additional meta data). We may move `edge_partition_device_view_base_t` to raft (with some renaming to match raft's conventinoo) and use this as the base class for adjacency list.

seunghwak: Also want to mention that we have `edgelist_t` (https://github.com/rapidsai/cugraph/blob/branch-22.06/cpp/include/cugraph/graph.hpp#L33). I think this can also be moved to raft. Then we may have very basic adjacency list/edge list data structures to be extended elsewhere.

kaatish:
@cjnolet Given that mfg classes will only interact with cugraph and cugraph-ops, these are better suited for cugraph rather raft.

cpp/include/raft/sparse/csr.hpp
cjnolet: I agree w/ @seunghwak here, I think these classes could be moved to RAFT (renamed accordingly). We have a COO (owning) class in RAFT currently which is being used by several primitives but I'd rather move the view classes from cugraph and use them throughout RAFT's sparse primitives.

cpp/include/raft/sparse/csr.hpp
cjnolet: This makes sense to me- and perhaps have them extend the csr implementation after it's moved to RAFT. I plan to create a PR for new (d)csr/coo view structures this week.

Issue #348, opened 2021-10-01
[FEA] Consolidate sparse/dense matrix descriptor implementations


cjnolet:
We have a few different matrix descriptor objects available in RAFT currently for wrapping the 3 arrays in sparse formats. Off the top of my head, I can think of the `raft::sparse::COO` object used in many of the sparse primitives, `sparse_matrix_t` used in the spectral clustering code / lanczos solver, and the `Graph_COO` used in the MST prim. 

It would be nice to see an RAII object like `raft::sparse::COO` and `sparse_matrix_t` used but which enables each of the 3 arrays to have a different templated datat type. I also like the inheritance pattern used in the `sparse_matrix_t` implementations because it can make for easier interactions with primitives that can enforce constraints on input. I also like the additional healper functions on the `sparse_matrix_t`, such as `fill()` and it would be nice to see some of the existing sparse primitives, such as those used for conversions, reductions, and linear algebra, to potentially become exposed as methods on the object. This would allow users to manipulate the object in a more object-oriented manner while still keeping the primitives themselves able to be invoked as functions... might need to think about that more but wanted to create this issue as a placeholder. 

@ChuckHastings @BradReesWork @seunghwak I'm wondering if there might be an opportunity to introduce some elements from the core cugraph object here so that they can be shared between cuml and cugraph (as well as other downstream projects).

cjnolet:
Now that `mdspan` and `mdarray` have been merged into 22.04, we can start on a sparse abstraction pending #499

cjnolet:
I've been thinking through a good design for this change and talking with a few folks about things to consider and ideal designs. I want to capture a few notes for things:

1. We should separate the structure itself (connections) from the possibility for weights/values. @mhoemmen has suggested that maybe we capture the structure in `graph_view` while capturing the structure+weights in a `sparse_matrix<graph_view>`. 
2. It's important that we capture expectations in the APIs so our users know what side effects, if any, to expect from different functions- for example, we should differentiate between views that might change weights/values from those that might change structure from those that might require a completely new sparse matrix or graph to be allocated.
3. We need to consider the design of the memory owning containers and the intended user experience for invoking functions whitch might need to change structure. 

@mhoemmen proposed a really nice way to think about this problem, using "solid", "liquid", "gas" nomenclature- 
- solid = object is completely immutable (`sparse_matrix_view<const T, graph_view<const S, const D>>`)
[...]

Issue #312, opened 2021-08-11
[FEA] Sparse primitives should consistently use two separate templates for rows/cols


cjnolet:
While we are at it, we should also consistently use a size_t for the number of nonzeros.

cjnolet:
We should tackle this while implementing the sparse variants of #535 (once #348 is complete)

PR #620, opened 2022-04-20
Standardizing upon CSR and COO view objects from cugraph

cjnolet:
This is an attempt to standardize the API of csr and coo view objects by building upon definitions from cugraph.

addresses #348 
Addresses #312


trivialfis (commented):
I can help create a sparse module based on mdarray and mdspan. Do you think it's something desired for this release?

cjnolet:
@trivialfis, I think the cugraph team would love to start using span/mdspan in the view API and mdarray would be very useful for the memory owning API. This PR is starting from cugraph's base adjacency and edge list objects and abstracting them (which includes the language) so that we can share them. 

I'd like to finish establishing that verbiage and basic operations but if you are interested, it would certainly be very helpful if you could take this from raw pointers into span/mdspan/mdarray.

cpp/include/raft/sparse/sparse_matrix.hpp
```diff
@@ -0,0 +1,269 @@
...
+ * @param [in] nvec total number of vectors for axis==0 (e.g. rows for csr, cols for csc)
+ * @param [in] vdim total number of dimensions (e.g. cols for csr, rows for csc)
+ * @param [in] offsets pointer of offsets into indices array for each vector idx
+ * @param [in] indices pointer of indices from vector offsets (size nnz)
+ * @param [in] values optional pointer of values for each vector offset (size nnz)
+ * @param [in] nnz number of nonzero offsets
+ *
+ */
+template <typename idx_t, typename offset_t, typename value_t, typename ContainerPolicy>
+class comp_sparse_t {
```
mhoemmen, line 216: Would you consider adding a way to get a sparse matrix view directly out of this sparse matrix container, for example by conversion operator, `.to_view()` method, or `to_view` customization point?

This conversion would encourage developers to write all sparse matrix algorithms as functions taking a sparse matrix view.  Otherwise, developers might be tempted to write overloads that take containers.  This would unnecessarily add to build times, and would hinder writing self-documenting code.  (If a function views a sparse matrix, then the function should take the sparse matrix as a view type.  By analogy, the intent of `void f(std::span<float>)` is much more focused and clear than the intent of `void f(std::vector<float>&)`.  Without documentation, I have to assume that the latter could invalidate iterators.  `f(std::span<float>)` _cannot_ invalidate iterators.)

mhoemmen (commented):
Hi all!  I had started working on this a couple days ago and got a bit tied up in other things.  Thanks for asking for feedback! : - )

PR #621, opened 2022-04-22
Enable components installation using CMake

Ethyling:
Enable components installation using CMake. This feature will be used to merge all CPP conda recipes into one, doing a top level build for all the libs, and only installing the right component while packaging.

### DIFF
--- a/cpp/cmake/modules/raft_export.cmake
+++ b/cpp/cmake/modules/raft_export.cmake
@@ -172,7 +172,7 @@ function(raft_export type project_name)
     rapids_cmake_install_lib_dir(install_location)
     set(install_location "${install_location}/cmake/${project_name}")
 
-    set(scratch_dir "${PROJECT_BINARY_DIR}/rapids-cmake/${project_name}/export")
+    set(scratch_dir "${PROJECT_BINARY_DIR}/rapids-cmake/${project_name}/export/raft/")
 
     configure_package_config_file("${CMAKE_CURRENT_FUNCTION_LIST_DIR}/config.cmake.in"
             "${scratch_dir}/${project_name}-config.cmake"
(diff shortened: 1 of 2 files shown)


ajschmidt8 (commented):
This looks okay to me, but I'm no CMake expert. So I'll defer to @robertmaynard for this one.

robertmaynard (changes requested):
The `install` rules in the `raft_export.cmake` need to be updated as well. As it currently stands explicitly specifying the components during install will lead to no CMake config files being installed which stops consumers from using it.

robertmaynard:
The install logic for the CMake raft-config components https://github.com/rapidsai/raft/blob/branch-22.06/cpp/CMakeLists.txt#L445 
 needs to be updated as well.

cpp/cmake/modules/raft_export.cmake
```diff
@@ -202,7 +203,8 @@ function(raft_export type project_name)
     endif()
 
     # Install everything we have generated
-    install(DIRECTORY "${scratch_dir}/" DESTINATION "${install_location}")
+    install(DIRECTORY "${scratch_dir}/" DESTINATION "${install_location}"
```
robertmaynard, line 206: This is wrong as it will capture the dependencies files like `raft-distance-dependencies.cmake` in the `raft` component and not the `raft_distance` component.

I think we want to change the `rapids_export_write_dependencies` in the root CMakeLists.txt to write to a per component folder, and update `raft_export` to map each component name to a install call of the associated component.

robertmaynard (approved):
Works locally as desired

cjnolet:
@Ethyling strangely, there appears to be a cmake error in CI

robertmaynard:
> @Ethyling strangely, there appears to be a cmake error in CI

Ahh we have to make sure the `install(DIRECTORY ` only occurs if the directory exists. This will mean we need to move the `raft_export` commands to be the last functions called in the root CMakeLists.txt

cpp/cmake/modules/raft_export.cmake
```diff
@@ -202,7 +203,13 @@ function(raft_export type project_name)
...
+            COMPONENT raft)
+    foreach(comp nn distance)
+      set(scratch_dir "${PROJECT_BINARY_DIR}/rapids-cmake/${project_name}/export/${comp}/")
+      install(DIRECTORY "${scratch_dir}" DESTINATION "${install_location}"
```
robertmaynard, line 210:
```suggestion
      file(MAKE_DIRECTORY "${scratch_dir}")
      install(DIRECTORY "${scratch_dir}" DESTINATION "${install_location}"
```

cjnolet:
@gpucibot merge

cjnolet:
@gpucibot merge

Issue #501, opened 2022-02-09
[FEA] Reimplement `submdspan`.


trivialfis:
Although the `mdspan` is designed to be flexible, there are limitations on how to integrate customized layouts and accessors.  The `mdspan` in raft is exposed via `host_mdspan` and `device_mdspan` so the compiler can prevent us from making some simple mistakes.  This customization is achieved by having an additional mixin with the default accessor.  Also, in order to implement https://github.com/rapidsai/raft/issues/497 we need customized layouts.  The `submdspan` reference implementation doesn't support this kind of customization.

trivialfis:
I will work on this.

PR #628, opened 2022-04-26
Fix mdspan accessor mixin offset policy.

trivialfis:
* Derive the offset_policy from the mixin instead of the default accessor.

Partially address https://github.com/rapidsai/raft/issues/501 .


cjnolet (approved):

cjnolet:
@gpucibot merge

PR #630, opened 2022-04-26
[BUG] RNG API fixes

MatthiasKohl:
After merging PR #609 , I realized that there were still a few bugs (all minor):
1. The deprecated `Rng` class had the `type` attribute removed. I added it back. This was used in cugraph-ops but not in other projects (cugraph, cuml).
2. The template function to call another function with the device state was actually unusable since you wouldn't be able to instantiate a function to call it with. So I removed that.
3. The macro used to call functions with a device state wasn't using fully namespace-qualified names, which made it unusable outside of RAFT.
4. Removed the deprecation warning of the new `rng_device.cuh` since it has to be included anyway through `rng.cuh`, so those deprecations didn't make sense.

### DIFF
--- a/cpp/include/raft/random/detail/rng_device.cuh
+++ b/cpp/include/raft/random/detail/rng_device.cuh
@@ -29,19 +29,12 @@ namespace raft {
 namespace random {
 namespace detail {
 
-#if defined(__RNG_H_INCLUSION_DEPRECATED)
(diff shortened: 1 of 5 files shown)


MatthiasKohl:
rerun tests

cjnolet (approved):

cjnolet:
@gpucibot merge

Issue #629, opened 2022-04-26
[FEA] Implement unravel for mdspan.


trivialfis:
This is an analog to numpy unravel index: https://numpy.org/doc/stable/reference/generated/numpy.unravel_index.html We often use a single index for parallel launchers like `thrust::for_each`, but need to index into types like matrix that have more than 1 dimension.  unravel index can help us map the computation onto the parallel threads.

PR #631, opened 2022-04-27
Implement unravel_index for row-major array.

trivialfis:
Related: https://github.com/rapidsai/raft/issues/629 Only row-major is currently implemented in this PR.  I can work on column-major support if needed. The function is hidden in the `detail` namespace right now due to the return type (`cuda::std::tuple`) can not be directly used in `mdspan`, which might change in the future. Please see the document and tests for example usage.

The implementation is mostly copied from XGBoost (I'm the author) https://github.com/dmlc/xgboost/blob/fdf533f2b9af9c068cddba50839574c6abb58dc3/include/xgboost/linalg.h#L523 .


trivialfis:
rerun tests

trivialfis:
hmm ? I thought libcu++ is available in raft by default.

cjnolet:
libcu++ is pulled in by cucollections, which is now only enabled when either building tests or the optional "distance" component.

This was done to reduce the dependency burden for projects downstream who aren't using the sparse distances API (which is what  depends on cucollections).

trivialfis:
Ah, thank you for sharing.  I will try to switch to thrust instead.

trivialfis:
I have changed to using thrust tuple. The PR is ready.

cpp/include/raft/detail/mdarray.hpp
```diff
@@ -238,4 +239,128 @@ namespace stdex = std::experimental;
...
+      index[dim] = idx & (s - 1);
+      idx >>= popc(s - 1);
+    }
+  }
+  index[0] = idx;
+  return arr_to_tup(index);
+}
+
+/**
+ * \brief Turns linear index into coordinate.  Similar to numpy unravel_index. This is not
```
divyegala, line 313: I think it should be okay to expose this to public, as we are doing with `flatten` and `reshape` in #601 ?

cpp/include/raft/detail/mdarray.hpp
```diff
@@ -238,4 +239,128 @@ namespace stdex = std::experimental;
...
+ * \param idx    The linear index.
+ * \param shape  The shape of the array to use.
+ * \param layout Must be `layout_right` (row-major) in current implementation.
+ *
+ * \return A thrust::tuple that represents the coordinate.
+ */
+template <typename LayoutPolicy, std::size_t... Exts>
+MDSPAN_INLINE_FUNCTION auto unravel_index(size_t idx,
+                                          detail::stdex::extents<Exts...> shape,
+                                          LayoutPolicy const&)
```
divyegala, line 334: Why is this function arg needed? To stay compatible with NumPy args? I think it's acceptable that we are able to get this information directly from the template type, and thus to remove this arg

cpp/include/raft/detail/mdarray.hpp
```diff
@@ -238,4 +239,128 @@ namespace stdex = std::experimental;
...
+/**
+ * C++ 17 style apply for thrust tuple.
+ *
+ * \param f function to apply
+ * \param t tuple of arguments
+ */
+template <typename Fn,
+          typename Tup,
+          std::size_t kTupSize = thrust::tuple_size<std::remove_reference_t<Tup>>::value>
+MDSPAN_INLINE_FUNCTION auto constexpr apply(Fn&& f, Tup&& t) -> decltype(auto)
```
divyegala, line 361: The trailing return type here is redundant

cpp/include/raft/detail/mdarray.hpp
```diff
@@ -238,4 +239,128 @@ namespace stdex = std::experimental;
...
+
+template <typename Fn, typename Tup, size_t... I>
+MDSPAN_INLINE_FUNCTION auto constexpr apply_impl(Fn&& f, Tup&& t, std::index_sequence<I...>)
+  -> decltype(auto)
+{
+  return f(thrust::get<I>(t)...);
+}
+
+/**
+ * C++ 17 style apply for thrust tuple.
```
divyegala, line 353: Very much like this, and that it just works. I wager this would be very useful in accessing and assigning for COOs

cpp/include/raft/detail/mdarray.hpp
trivialfis: If others think it's public ready then I'm happy to expose it.  For me these are the concerns:
- what if one day mdspan decides to accept std::extents as an index?
- what if we use other implementations of tuple in the future? like `cuda::std::tuple` or just `std::tuple` instead of `thrust::tuple`

cpp/include/raft/detail/mdarray.hpp
trivialfis: Depends on how you like to call the function:

``` c++
unravel_index<stdex::layout_rigth>(idx, shape)
```
or
``` c++
unravel_index(idx, shape, stdex::layout_right{})
```

I chose the second one as it feels more aligned with mdspan design:
``` c++
submdspan(array, std::full_extent_t{}) # use constructor of full_extent_t here
```

cpp/include/raft/detail/mdarray.hpp
trivialfis: it's required to denote that the return type might be a reference.

divyegala: `auto&&` for universal references?

cpp/include/raft/detail/mdarray.hpp
divyegala: @cjnolet had suggested that it's better we rely lesser on thrust as time goes on. I find value in using `cuda::std::tuple` over `thrust::tuple` as I have a feeling the latter is going to be replaced by the former. But maybe Corey feels differently about introducing libcu++ back as a core dependency

cpp/include/raft/detail/mdarray.hpp
divyegala: Actually, I'm not sure if this will work. I'm okay with the former

vyasr:
rerun tests

cpp/include/raft/detail/mdarray.hpp
cjnolet: @trivialfis I prefer to use `std::tuple` here. I'd like to keep both `thrust` and `libcu++` out of public APIs. C++ stdlib is fine.

cjnolet (changes requested):

trivialfis:
@cjnolet but it doesn't always work in CUDA kernel.  Also, it's hidden in detail name space.

cpp/include/raft/detail/mdarray.hpp
cjnolet: > @cjnolet but it doesn't always work in CUDA kernel. Also, it's hidden in detail name space.

The challenge w/ header-only here is that the dependencies for all included headers are required at compile-time by consumers downstream. That includes things pulled in transitively from the `detail` namespace. The goal here is to draw the line and make it so that all the public apis in the `core/` directory can be safely exposed by our consumers through their own public APIs while not imposing any additional dependencies on their users outside of RMM and the CTK libs. Currently, our mdarray header in `detail` pulls in thrust and I'd like to eventually separate that out so that thrust isn't a hard requirement just for including `mdarray`.

If we want to allow this function to be included by files in `core/`, I would propose we find or create some other object to contain the unraveled indices. Though, if this is truly internal code and meant to stay that way, maybe we should consider separating this function out into a different header for now which is documented accordingly (for example, mentioning that it uses thrust, so it shouldn't be included by any headers in `core/`). Maybe something like `detail/mdarray_internal_utils.hpp` just to be very explicit?

trivialfis: I will hide this function as an internal function.  Closing this PR for now and will submit it again when it's actually used.

PR #635, opened 2022-05-03
LinAlg API updates

divyegala:


cjnolet:
Just leaving a note to self that we need to use a pre processor variable to enable/disable the mdspan/mdarray APIs in linalg otherwise we are going to break spark-rapids (they are just closing the raft headers and adding them to the include path)

vyasr:
rerun tests

PR #638, opened 2022-05-04
Require UCX 1.12.1+

jakirkham:
xref: https://github.com/rapidsai/integration/pull/471


conda/recipes/pyraft/conda_build_config.yaml
```diff
@@ -0,0 +1,2 @@
+ucx_version:
+  - "1.12.1"
```
Ethyling, line 2: I think this should be `>=1.12.1`

conda/recipes/pyraft/meta.yaml
```diff
@@ -42,6 +43,7 @@ requirements:
     - libraft-headers {{ version }}
     - nccl>=2.9.9
     - rmm {{ minor_version }}
+    - ucx >={{ ucx_version }}
```
Ethyling, line 46: I think we should remove the `>=` here

conda/recipes/pyraft/conda_build_config.yaml
jakirkham: The intent here is to ensure we have a fixed version at build time so that this sets a minimum version at runtime.

conda/recipes/pyraft/meta.yaml
jakirkham: Please see comment above

conda/recipes/pyraft/conda_build_config.yaml
Ethyling: Oh, I thought we wanted to build using the latest version available. What's the difference between doing this and using `pin_compatible` for the runtime dependency?

conda/recipes/pyraft/meta.yaml
jakirkham: That said, we can probably drop this line as the feedstock has [`run_exports`]( https://github.com/conda-forge/ucx-split-feedstock/pull/92/files#diff-f3725a55bf339595bf865fec73bda8ac99f283b0810c205442021f29c06eea9aR40-R41 ), which covers adding and pinning this.

conda/recipes/pyraft/conda_build_config.yaml
jakirkham: We are using the latest version since 1.12.1 is latest. However we may want more flexibility if a later 1.12.x breaks something and we need to pin an older version in the metapackage. Also we may want to control which version we build against to make it easier to align with other packages that have `ucx` constraints as well.

FWIW am proposing something similar in conda-forge ( https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/2847 ) though maybe we can wait to see what others think about that and update if needed.

raydouglass (approved):

cjnolet:
@gpucibot merge

PR #639, opened 2022-05-04
[BUG] Rng: removed cyclic dependency creating hard-to-debug compiler errors

MatthiasKohl:
This PR removes a cyclic dependency from `raft/random/rng_device.cuh` which can create compiler errors (only in some cases).
Since there are no direct includes of `raft/random/rng_device.cuh` in cuml and cugraph, the impact of the PR is low.
However, it will improve things for cugraph-ops and cuopt which use this header directly.


cjnolet (approved):

cjnolet:
@gpucibot merge

PR #641, opened 2022-05-06
Consolidate C++ conda recipes and add libraft-tests package

Ethyling:
This PR includes the following changes:
- Adds a `libraft-tests` package to the `libraft-split` recipe
  - This is a prerequisite for removing "Project Flash" from our build/CI scripts
   - The `libraft-tests` package was added as an additional output to the `libraft-split` recipe
- Consolidates remaining C++ recipes into `libraft-split` recipe
   - This gets rid of a lot of duplicate code between the recipes and reduces the number of times we have to call conda build in our CI scripts
- Migrate the "from sources" builds done in GPU tests by building packages using conda. This is done for the following reasons:
  - This is required step to improve the Ops CI/CD setup to a more convenient pipeline
  - This is required to start using conda compilers and `mamba` to build RAPIDS packages
  - This prevent us from manually managing and installing the dependencies in GPU job
  - This ensure the packages can be installed (no conda conflict while installing)
  - This ensure the tests are running and working against the package content and not the build results. Currently the Python packages are not tested

The result of consolidating the C++ recipes into a single `libraft-split` recipe is that one top-level build occurs, which builds all of the C++ components (in `conda/recipes/libraft/build.sh`) but does not install them. The installation for each package occurs in the corresponding `conda/recipes/libraft/install_xxx.sh` scripts

There are some implications to these changes. Namely that any top-level host requirement which includes a run_exports value (i.e. shared libraries) must now be manually specified in the run dependencies of the corresponding outputs package. To help keep things DRY, dependency version specs can be specified in `conda/recipes/libraft/conda_build_config.yaml`. The exception here is the version spec used for cudatoolkit since that comes from an environment variable in the CI process.


Ethyling:
rerun tests

conda/recipes/libraft/install_libraft_distance.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 # Copyright (c) 2022, NVIDIA CORPORATION.
 
-./build.sh libraft --install -v --allgpuarch --no-nvtx
+cmake --install cpp/build --component raft_distance
```
cjnolet, line 4: @robertmaynard hoping to have your thoughts here. Should this be enough to install the proper cmake files along w/ the artifact?

conda/recipes/libraft/build.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+# Copyright (c) 2022, NVIDIA CORPORATION.
+
+./build.sh libraft tests bench -v --allgpuarch --compile-dist --compile-nn --no-nvtx
```
cjnolet, line 4: @Ethyling, I need to think about this one a little bit. One of the reasons we were running the compile 3 separate times was because a user should be able to install the headers conda package without the distances or nn packages. They should also be able to install the headers and then install either the distances or the nn packages at a later time and the relevant cmake files should allow for this. 

@robertmaynard do you see any reason compiling them all at once and installing the components separately would create a problem here?

cjnolet:
@Ethyling I had to back out a couple recently merged PRs at the beginning of the week because they had caused some builds to break downstream (cuml specifically). 

IIRC, the changes in those PRs were intended to fix the cmake exports from consolidating and the build steps in this PR, right?

Ethyling:
> @Ethyling I had to back out a couple recently merged PRs at the beginning of the week because they had caused some builds to break downstream (cuml specifically).
> 
> IIRC, the changes in those PRs were intended to fix the cmake exports from consolidating and the build steps in this PR, right?

@cjnolet Yes. This PR requires to have a CMake component installation working. Maybe we can try to work on fixing the issue you were facing with cuml?

cjnolet:
@vyasr CI in this PR seems to be indicating that the cmake for pylibraft cannot find the distance library even though it's being built and installed earlier in the  build script. Have you seen this behavior or know how to handle it?

vyasr:
No, I haven't seen this issue before. Right now what I see in CI is that it's not finding cuco, and you're saying that implicates the distance lib because cuco should be exported as a global target from there, right?

cjnolet (approved):
The resulting packages seem to look correct at a glance. Thanks again @Ethyling!

ajschmidt8 (approved):

PR #643, opened 2022-05-10
Some fixes to pairwise distances for cupy integration

cjnolet:

### DIFF
--- a/python/pylibraft/pylibraft/distance/pairwise_distance.pyx
+++ b/python/pylibraft/pylibraft/distance/pairwise_distance.pyx
@@ -79,7 +79,7 @@ SUPPORTED_DISTANCES = ["euclidean", "l1", "cityblock", "l2", "inner_product",
                        "hamming", "jensenshannon"]
 
 
-def distance(X, Y, dists, metric="euclidean", p=0.0):
+def distance(X, Y, dists, metric="euclidean", p=2.0):
     """
     Compute pairwise distances between X and Y
 
@@ -138,6 +138,8 @@ def distance(X, Y, dists, metric="euclidean", p=0.0):
 
     x_c_contiguous = "strides" not in x_cai or x_cai["strides"] is None
     y_c_contiguous = "strides" not in y_cai or y_cai["strides"] is None
+
+    print(str(x_c_contiguous))
 
     if x_c_contiguous != y_c_contiguous:
         raise ValueError("Inputs must have matching strides")
--- a/python/pylibraft/pylibraft/distance/pairwise_distance.pyx
+++ b/python/pylibraft/pylibraft/distance/pairwise_distance.pyx
@@ -27,10 +27,11 @@ from libcpp cimport bool
 from .distance_type cimport DistanceType
 from pylibraft.common.handle cimport handle_t
 
+
 def is_c_cont(cai, dt):
-  return "strides" not in cai or \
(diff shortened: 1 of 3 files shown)


divyegala (approved):

python/pylibraft/pylibraft/distance/pairwise_distance.pyx
```diff
@@ -68,21 +70,26 @@ DISTANCE_TYPES = {
...
 
-SUPPORTED_DISTANCES = list(DISTANCE_TYPES.keys())
+SUPPORTED_DISTANCES = ["euclidean", "l1", "cityblock", "l2", "inner_product",
+                       "chebyshev", "minkowski", "canberra", "kl_divergence",
+                       "correlation", "russellrao", "hellinger", "lp",
+                       "hamming", "jensenshannon", "cosine", "sqeuclidean"]
 
 
-def distance(X, Y, dists, metric="euclidean"):
+def distance(X, Y, dists, metric="euclidean", p=2.0):
```
dantegd, line 84: New parameter is missing from docstring

python/pylibraft/pylibraft/distance/pairwise_distance.pyx
```diff
@@ -132,6 +142,12 @@ def distance(X, Y, dists, metric="euclidean"):
     y_dt = np.dtype(y_cai["typestr"])
     d_dt = np.dtype(dists_cai["typestr"])
 
+    x_c_contiguous = "strides" not in x_cai or x_cai["strides"] is None
+    y_c_contiguous = "strides" not in y_cai or y_cai["strides"] is None
+
+    if x_c_contiguous != y_c_contiguous:
+        raise ValueError("Inputs must have matching strides")
```
dantegd, line 149: I think this check works 99% of the time because (almost?) all NumPy/CuPy functions that create C contiguous arrays make the strides be `None`, but techically speaking the array interface for a C contiguous array can have the valid strides as opposed to `None` and this check would fail. I think it would be good to be resilient to that case, though it could be a follow up since I think the common case should be covered like this

dantegd (changes requested):
Some minor comments

python/pylibraft/pylibraft/distance/pairwise_distance.pyx
```diff
@@ -68,21 +70,26 @@ DISTANCE_TYPES = {
...
+    "minkowski": DistanceType.LpUnexpanded,
     "russellrao": DistanceType.RusselRaoExpanded,
     "dice": DistanceType.DiceExpanded
 }
 
-SUPPORTED_DISTANCES = list(DISTANCE_TYPES.keys())
+SUPPORTED_DISTANCES = ["euclidean", "l1", "cityblock", "l2", "inner_product",
+                       "chebyshev", "minkowski", "canberra", "kl_divergence",
+                       "correlation", "russellrao", "hellinger", "lp",
+                       "hamming", "jensenshannon", "cosine", "sqeuclidean"]
```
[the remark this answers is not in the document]
cjnolet: The "distance types" is a mapping of all the possible distances whereas only a subset is supported currently. I opted to list our each individual supported distance rather than trying to a weird set/list subtraction.

dantegd (approved):

cjnolet:
@gpucibot merge

PR #646, opened 2022-05-11
Fixing the unit test issue(s) in RAFT

vinaydes:
The call to `uniform()` is getting executed in `handle.get_stream()` and kernels/operations after `uniform()` are executed in separately created `stream`. This causes synchronization hazard. When default RNG was changed from Philox to PCG, the bug got exposed due to the relative difference in generation speed of PCG and Philox (PCG is faster). Adding a stream synchronization call fixes the issue.

### DIFF
--- a/cpp/test/linalg/divide.cu
+++ b/cpp/test/linalg/divide.cu
@@ -57,9 +57,7 @@ class DivideTest : public ::testing::TestWithParam<raft::linalg::UnaryOpInputs<T
   {
     raft::random::RngState r(params.seed);
     int len = params.len;
-    RAFT_CUDA_TRY(cudaStreamCreate(&stream));
     uniform(handle, r, in.data(), len, T(-1.0), T(1.0));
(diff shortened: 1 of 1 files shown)


cpp/test/linalg/divide.cu
```diff
@@ -59,6 +59,7 @@ class DivideTest : public ::testing::TestWithParam<raft::linalg::UnaryOpInputs<T
     int len = params.len;
     RAFT_CUDA_TRY(cudaStreamCreate(&stream));
```
cjnolet, line 60: I've been removing this logic in the other tests as I've been encountering it. I believe this pattern predates the time when the primitives were made aware of the handle. It looks like the handle is already being used in the gtest class. Instead of creating a new stream and having to synchronize it, can we just always use the stream from the handle?

cjnolet (changes requested):

cjnolet (approved):

cjnolet:
@gpucibot merge

PR #648, opened 2022-05-12
Fix components installation

Ethyling:
We need to install the `raft-${comp}-dependencies.cmake` files as part of the `raft` component


vyasr:
Perhaps a naive question, but should we be using `rapids_export` from `rapids-cmake` instead of `raft_export`?

robertmaynard:
> Perhaps a naive question, but should we be using `rapids_export` from `rapids-cmake` instead of `raft_export`?

Not currently, as rapids_export isn't aware of components

cjnolet:
Is this one superseded by #653? If so, can we close this?

Ethyling:
> Is this one superseded by #653? If so, can we close this?

Yes, we can close this

PR #651, opened 2022-05-16
Use sccache or ccache when available

achirkin:
Enable (s)ccache when found in $PATH to speed up builds, preferring sccache over ccache.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -129,6 +129,17 @@ rapids_find_package(CUDAToolkit REQUIRED
 include(cmake/modules/ConfigureCUDA.cmake)
 
 ##############################################################################
+# - Setup cache --------------------------------------------------------------
+
+find_program(CCACHE sccache NAMES sccache ccache)
+if (CCACHE)
+  message(VERBOSE "Using ${CCACHE} for caching.")
+  set(CMAKE_C_COMPILER_LAUNCHER ${CCACHE})
+  set(CMAKE_CXX_COMPILER_LAUNCHER ${CCACHE})
+  set(CMAKE_CUDA_COMPILER_LAUNCHER ${CCACHE})
+endif()
+
+##############################################################################
 # - Requirements -------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


cpp/CMakeLists.txt
```diff
@@ -128,6 +128,17 @@ rapids_find_package(CUDAToolkit REQUIRED
     INSTALL_EXPORT_SET raft-exports)
 include(cmake/modules/ConfigureCUDA.cmake)
 
+##############################################################################
+# - Setup cache --------------------------------------------------------------
+
+find_program(CCACHE sccache NAMES sccache ccache)
```
robertmaynard, line 134: This should be a `build.sh` option and not something that implicit runs on a machine. We shouldn't presume that a user of RAFT that has both sccache and ccache always wants sccache.

As discussed here: https://github.com/rapidsai/cuml/pull/4534#discussion_r798728846 The option to control this should be allow the user to specify which tool they want when using `build.sh`. For users that aren't using `build.sh` they can use the CMake supported enviornment variables to control this if they don't want to specify them on the command line.

robertmaynard (changes requested):

cpp/CMakeLists.txt
achirkin: Thanks for pointing this out for me! I'd be happy to add similar flags in this PR. Is there any fundamental reason to not enable one of the tools by default (if no flags/env vars specified)?

robertmaynard: Yes we shouldn't infer intent from the lack of setting the compiler launcher. For a couple of reasons:

1. Both ccache and sccache also support the use case when they are the actual compiler provided to CMake. This pattern is used by developers and package managers. When this is happening and you have ccache/sccache as the compiler launcher you get errors as both don't support a syntax of `ccache <ccache-symlink>`

2. At times developers need to benchmark the compile time cost / improvement of a change. By automatically injecting a ccache / sccache into the the compile line it will break presumptions of actual performance.

PR #652, opened 2022-05-16
Integrate KNN implementation: ivf-flat

achirkin:
Integrate a new KNN implementation.


cpp/include/raft/spatial/knn/detail/ann_ivf_flat.cuh
```diff
@@ -0,0 +1,1389 @@
...
+                                          neighbors,
+                                          greater_,
+                                          stream);
+#endif
+  }  // end if nprobe=1
+
+  return cuivflStatus_t::CUIVFL_STATUS_SUCCESS;
+}  // end func cuivflHandle::cuivflSearchImpl
+
+void cuivflInit(std::unique_ptr<detail::cuivflHandle>& handle,
```
tfeher, line 1375: Do we really need this init routine? We could call `make_unique` directly.

cpp/include/raft/spatial/knn/detail/ann_ivf_flat.cuh
```diff
@@ -0,0 +1,1389 @@
...
+
+  if ((dim % 4) == 0) {
+    veclen = 4;
+  } else if ((dim % 2) == 0) {
+    veclen = 2;
+  }
+
+  // cuBLAS
+  cublasStatus_t cublasError;
+  cublasError = cublasCreate(&cublas_handle_);
```
tfeher, line 241: Use from raft handle

cpp/include/raft/spatial/knn/detail/ann_ivf_flat.cuh
```diff
@@ -0,0 +1,1389 @@
...
+  // The host memory pointer; Used for list_data_manage_ptr_; size [nlist_]
+  uint32_t* list_prefix_interleaved_host_ptr_;
+  // The host memory pointer; the number of each cluster(list); size [nlist_]
+  uint32_t* list_lengths_host_ptr_;
+  // The host memory pointer; centriod; size [nlist_, dim_]
+  float* centriod_host_ptr_;
+  // The host memory pointer; centriod norm ; size [nlist_, dim_]
+  float* centriod_norm_host_ptr_;
+  // The device memory; used for topk select.
+  void* buf_dev_ptr_;
```
tfeher, line 195: Change to RMM arrays, or if possible `mdarray`.

tfeher (changes requested):
Thanks @mdoijade for preparing the initial version of this PR, and for @achirkin for continuing the work with RAFT integration! Since the PR is large, I am sharing my comments in smaller chunks, so that @achirkin can address them when he gets there.

tfeher (commented):
One more batch of comments.

cpp/bench/common/benchmark.hpp
```diff
@@ -122,6 +122,13 @@ class fixture {
   virtual void run_benchmark(::benchmark::State& state) = 0;
   virtual void generate_metrics(::benchmark::State& state) {}
 
+ protected:
+  /** The helper that writes zeroes to some buffer in GPU memory to flush the L2 cache.  */
+  void flush_L2_cache()
+  {
+    RAFT_CUDA_TRY(cudaMemsetAsync(scratch_buf_.data(), 0, scratch_buf_.size(), stream));
```
teju85, line 129: Just to be sure, it is better to set the size of this buffer to be 3x that of the size of the L2-cache.

achirkin: Sure, why not :)

cpp/include/raft/spatial/knn/ann.cuh
```diff
@@ -64,24 +59,22 @@ inline void approx_knn_build_index(raft::handle_t& handle,
...
  * @param[in] query_array the query to perform a search with
  * @param[in] n number of rows in the query array
  */
-template <typename value_idx = int>
+template <typename T = float, typename value_idx = int>
 inline void approx_knn_search(raft::handle_t& handle,
                               float* distances,
                               int64_t* indices,
                               raft::spatial::knn::knnIndex* index,
+                              knnIndexParam* params,
```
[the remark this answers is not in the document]
achirkin: I guess, for a smoother transition, we should make it optional with sensible defaults?..

cpp/include/raft/spatial/knn/detail/ann_ivf_flat.cuh
```diff
@@ -0,0 +1,789 @@
...
+  if (metric_type_ == raft::distance::DistanceType::L2Expanded ||
+      metric_type_ == raft::distance::DistanceType::L2Unexpanded) {
+    greater_ = false;
+  } else {
+    // Need to set this to true for inner product if need FAISS like behavior for inner product
+    greater_ = false;
+  }
+
+  // Set memory buffer to be reused across searches
+  auto cur_memory_resource = rmm::mr::get_current_device_resource();
```
achirkin, line 606: @tfeher @teju85 @cjnolet, please share your thoughts about this.

I propose to use rmm's pool_memory_resource instead of pre-allocating buffers explicitly, which simplifies the code a lot. I also believe it addresses the concerns on integration with third-party memory resources/allocators.

1. `search_mem_res` is preserved across searches, so we save no repeated allocations, the same as with an explicitly allocated buffer.
2. As I understand, `search_mem_res` acquires memory in somewhat bigger chunks, so it calls the upstream allocator not many times, therefore we can safely start with a small initial size.
3. If a custom memory resource is already set, it's used as an upstream resource; otherwise the behavior is similar to having an explicitly allocated buffer.

I've compared the performance on current (rather limited) set of benchmarks in raft, I don't see any measurable difference (it has actually become slightly better after the change).

cjnolet: @achirkin, we're going to have to think through this one a little bit. The ask here on the FAISS side is that we give them full control over the allocation, which is going to require us providing some type of allocator. This isn't the first time this has come up- as the cunumeric folks are asking for the same thing. At some point very very soon, or even now, we're going to have to implement something like this. We can, for now, back it w/ RMM, but IVF-FLAT would be a good first place to get the ball rolling here.

cjnolet: I'm even okay bringing back the old allocator for now if we need to and then building on that. We're starting on a PoC for the FAISS integration and giving FAISS total control over memory allocations is a hard requirement.

achirkin: Thanks, Corey, for the input!

How much exactly control they want over the allocations? Currently, I [allow to pass the rmm resource to `radix_topk` explicitly](https://github.com/rapidsai/raft/blob/b45b14c620b5bf48ab66e1a494691f3ae8b7adac/cpp/include/raft/spatial/knn/detail/topk/radix_topk.cuh#L558), which means one can pass a pool resource with a fixed initial size == maximum size. I believe it's technically the same as passing a manually allocated ptr... For `cuivflHandle`, the corresponding memory resource is hidden private at this moment, but I can expose that as well. Then, the only missing bit is to allow to calculate the required memory beforehand.

tfeher: @cjnolet other libraries can wrap their allocator in a `rmm::mr::device_memory_resource` object and set that as a default allocator. But the question what "full control" means? 

Right now @achirkin [ties a pooled allocator](https://github.com/achirkin/raft/blob/b45b14c620b5bf48ab66e1a494691f3ae8b7adac/cpp/include/raft/spatial/knn/detail/ann_ivf_flat.cuh#L162) (which uses the default upstream allocator) to the `cuivflHandle` object, and that would keep the pool allocated until the object is alive. This bring up the question of controlling the lifetime of that temporary memory, and whether we want to expose it through `rmm::mr`, or by requiring the user to pass preallocated buffers.

tfeher (changes requested):

cpp/include/raft/spatial/knn/detail/ann_ivf_flat.cuh
```diff
@@ -0,0 +1,787 @@
...
+                                             uint32_t batch_size,
+                                             uint32_t k,
+                                             size_t* neighbors,  // [numQueries, topK]
+                                             float* distances)
+{
+  common::nvtx::range<common::nvtx::domain::raft> fun_scope(
+    "cuivflSearch(%u, %u, %zu)", batch_size, k, neighbors);
+  cuivflSearchImpl<float>(queries, batch_size, k, neighbors, distances);
+  return cuivflStatus_t::CUIVFL_STATUS_SUCCESS;
+}  // end func cuivflSearch
```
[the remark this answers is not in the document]
achirkin: I'm not sure. `cuivflSearchImpl` has this extra template parameter, which, in theory, can be changed to e.g. double. Though I doubt we will ever do that.

cpp/include/raft/spatial/knn/detail/ann_ivf_flat.cuh
tfeher: @cjnolet Below is a draft of how to wrap FAISS allocator as an RMM memory resource. As long as we use the default memory resource, this would ensure that all allocations are done through FAISS. Is this an acceptable solution, or are there other requirements?

```c++
#include <faiss/gpu/StandardGpuResources.h>
#include <rmm/mr/device/callback_memory_resource.hpp>
...
faiss::gpu::StandardGpuResources res;
auto allocate_callback = [](std::size_t bytes, rmm::cuda_stream_view stream, void* arg) {
  auto gpu_res = static_cast<faiss::gpu::GpuResources*>(arg);
  return gpu_res->allocMemory(
    AllocRequest(makeDevAlloc(faiss::gpu::AllocType::Other, stream), bytes));
};
auto deallocate_callback = [](void* ptr, std::size_t, rmm::cuda_stream_view stream, void* arg) {
  auto gpu_res = static_cast<faiss::gpu::GpuResources*>(arg);
  gpu_res->deallocMemory(faiss::gpu::getCurrentDevice(), ptr);
};
rmm::mr::callback_memory_resource faiss_mr(allocate_callback, deallocate_callback, &res, &res);
rmm::mr::set_current_device_resource(&faiss_mr);
```

cpp/include/raft/spatial/knn/ann.cuh
```diff
@@ -64,24 +58,22 @@ inline void approx_knn_build_index(raft::handle_t& handle,
  *                       their query point
  * @param[out] indices indices of the nearest neighbors
  * @param[in] index index to perform a search with
+ * @param[in] params parameters used to build the index
  * @param[in] k the number of nearest neighbors to search for
  * @param[in] query_array the query to perform a search with
```
tfeher, line 63: Please describe the memory layout of the input / output arrays (device vs host, row or column major).

cpp/include/raft/spatial/knn/ann.cuh
```diff
@@ -42,13 +36,13 @@ namespace knn {
  * @param[in] n number of rows in the index array
```
tfeher, line 36: please describe the memory layout of the index array

cpp/include/raft/spatial/knn/detail/ann_ivf_flat.cuh
```diff
@@ -0,0 +1,796 @@
...
+                                                 nullptr,
+                                                 batch_size,
+                                                 nlist_,
+                                                 nprobe,
+                                                 coarse_distances_dev.data(),
+                                                 coarse_indices_dev.data(),
+                                                 !greater_,
+                                                 stream_,
+                                                 &(search_mem_res.value()));
+  }
```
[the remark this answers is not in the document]
tfeher: You are right, does not make sense to use IVF-Flat with such parameters. In that case brute force shall provide better performance. Therefore you can ignore my suggestion above. What might still make sense is to issue a warning if `nprobe` is close to `nlist_`. There we can inform the user about the performance issue.

cpp/include/raft/spatial/knn/detail/ann_ivf_flat.cuh
```diff
@@ -0,0 +1,771 @@
...
+    list_index_host_(0),
+    list_prefix_interleaved_host_(0),
+    list_lengths_host_(0),
+    list_data_host_(0)
+{
+  veclen_ = 16 / sizeof(T);
+  while (dim % veclen_ != 0) {
+    veclen_ = veclen_ >> 1;
+  }
+}
```
tfeher, line 199: Since we copy the dataset to create an interleaved layout, the `dim % veclen == 0` restriction could be removed by padding the dataset. Please add a todo note. Please also create an issue (if we do not have already) to collect such items, and refer that issue here.

cpp/include/raft/spatial/knn/detail/ann_ivf_flat.cuh
```diff
@@ -0,0 +1,771 @@
...
+  T* list_data, const T* dataset, uint32_t dim, size_t index, size_t prefix, uint32_t veclen)
+{
+  size_t group_id = index / WarpSize;
+  size_t in_id    = (index % WarpSize) * veclen;
+  list_data += (prefix + group_id * WarpSize) * dim + in_id;
+
+  for (size_t i = 0; i < dim; i += veclen) {
+    for (size_t j = 0; j < veclen; j++) {
+      list_data[i * WarpSize + j] = dataset[i + j];
+    }
```
tfeher, line 54: Please add a  note: it is assumed that dim % veclen == 0

tfeher (changes requested):

cjnolet:
Providing a brief update here. I've started a subclass of `faiss::gpu::IVFFlat` [here](https://github.com/cjnolet/faiss/blob/2eb94f1d538434edc53745a312f790c7734491ec/faiss/gpu/raft/RaftIndexIVFFlat.h#L74) which is ultimately what will end up invoking the IVF Flat kernels. Of specific importance are the `train_()` and `searchImpl_()` functions. @achirkin we'll need to make sure both the expected inputs and the outputs to these functions match.

cpp/include/raft/spatial/knn/detail/ann_ivf_flat.cuh
```diff
@@ -0,0 +1,709 @@
...
+  for (uint32_t l = 0; l < dim; l += veclen) {
+    for (uint32_t j = 0; j < veclen; j++) {
+      list_data[l * WarpSize + ingroup_id + j] = dataset[l + j];
+    }
+  }
+}
+
+template <typename T>
+void cuivflHandle<T>::cuivflBuildIndex(const T* dataset,
+                                       T* trainset,
```
cjnolet, line 409: Do we need both the dataset (I'm assuming this would be for query?) and the trainset here? Why not just use the trainset?

cpp/include/raft/spatial/knn/detail/ann_ivf_flat.cuh
```diff
@@ -0,0 +1,709 @@
...
+
+  static_assert(std::is_same_v<T, float> || std::is_same_v<T, uint8_t> || std::is_same_v<T, int8_t>,
+                "unsupported data type");
+
+  // Alloc manage memory for centriods, trainset and workspace
+  rmm::device_uvector<uint32_t> labels_buf(n_rows_, stream_, &managed_memory);  // [numDataset]
+  auto labels = labels_buf.data();
+
+  // Predict labels of the whole dataset
+  cuivflBuildOptimizedKmeans(centriod_managed_ptr, dataset, trainset, labels, n_rows, n_rows_train);
```
cjnolet, line 428: Same here- why not just train the clusters on the train set?

cjnolet (changes requested):

cpp/include/raft/spatial/knn/detail/ann_ivf_flat.cuh
achirkin: In the current implementation, `trainset` is a smaller subset of the `dataset` that is used to train k-means. Supposedly, reducing the size of the dataset makes it faster without sacrificing the quality too much.
After that, we use the whole dataset to build the index, which is just grouping the points in clusters (we need all points there, since they are what kNN returns).

cpp/include/raft/spatial/knn/detail/ann_ivf_flat.cuh
achirkin: Hm, we use the `dataset` in the last step of this function, which is to label the dataset and update the cluster centers to be in the centers of whole `dataset` (rather than the smaller `trainset`). In addition, we use the labels later for building the index.

Thanks for the good points, I'll add these answers to the documentation.

achirkin:
NB: added an issue https://github.com/rapidsai/raft/issues/711 to track extra TODOs

cpp/include/raft/spatial/knn/detail/ann_quantized.cuh
```diff
@@ -0,0 +1,305 @@
...
+  int ratio           = 2;  // TODO: take these parameters from API
+  int niter           = 20;
+  const int dim       = D;
+  const size_t ntrain = n / ratio;
+  assert(ntrain > 0);
+
+  rmm::mr::managed_memory_resource managed_memory;
+  rmm::device_uvector<T> trainset(ntrain * dim, stream, &managed_memory);
+
+  RAFT_CUDA_TRY(cudaMemcpy2DAsync(trainset.data(),
```
cjnolet, line 112: Is this assuming that the data has already been properly shuffled? Why not sample here to make sure we're getting a uniform distribution of points?

cpp/include/raft/spatial/knn/detail/ann_ivf_flat.cuh
cjnolet: I see- I think `approx_knn_cuivfl_ivfflat_build_index` is probably what we want to invoke from FAISS. This method is lower level.

cjnolet (changes requested):

cpp/include/raft/spatial/knn/detail/ann_quantized.cuh
achirkin: Yes, I agree! Working on it now. I want to move the allocation of the trainset and the copying of the data inside `cuivflBuildOptimizedKmeans`, which will go into `ann_kmeans_balanced` file. Maybe I'll keep this simple copy for now, but add a TODO proper sampling there.

cjnolet:
rerun tests

cjnolet:
rerunning tests now that sccache is fixed.

cpp/include/raft/spatial/knn/detail/ann_ivf_flat.cuh
```diff
@@ -0,0 +1,465 @@
...
+  void queryIVFFlatGridSize(const uint32_t n_probes, const uint32_t n_queries, const uint32_t k);
+  uint32_t getDim() { return index_.has_value() ? index_->dim() : 0; }
+
+ private:
+  const handle_t& handle_;
+  const rmm::cuda_stream_view stream_;
+  ivf_flat_params params_;
+
+  const raft::distance::DistanceType metric_type_;
+  bool greater_;
```
cjnolet, line 69: Consider the following common usage pattern for FAISS:

1. user trains an ivfflat index
2. user stores index to disk
3. user loads index from disk
4. user performs a search with index

In this case, what would be the indeded lifecycles of cuivflHandle and `ivf_flat_index`? Are we assuming the user would store the handle to disk or the index? I believe we're saying the index, right? So that means the handle is meant to be a temporary object which can be constructed on the fly each time a train or a search is called but doesn't necessarily need to be reused. 

If we removed the methods from `cuivflHandle` and allowed the `greater_` member to be derived from the `metric_type`, the only remaining members which would not be inputs from the functions would be `grid_dim_x_` and `search_mem_res`. It appears they are both reused across multiple calls to `search`. Maybe we could move them to the index itself and re-populated each time the index is instantiated (e.g. by not serializing them to disk).

cjnolet (changes requested):
My apologies for the delay in the review. I'm still reviewing your changes as I'm working through the FAISS integration but I wanted to provide some initial feedback in the meantime.

cpp/include/raft/spatial/knn/detail/ann_ivf_flat_kernel.cuh
```diff
@@ -0,0 +1,1227 @@
...
+   * `
+   *   x[ 0, 0], x[ 0, 1], x[ 1, 0], x[ 1, 1], ... x[14, 0], x[14, 1], x[15, 0], x[15, 1],
+   *   x[16, 0], x[16, 1], x[17, 0], x[17, 1], ... x[30, 0], x[30, 1],    -    ,    -    ,
+   *   x[ 0, 2], x[ 0, 3], x[ 1, 2], x[ 1, 3], ... x[14, 2], x[14, 3], x[15, 2], x[15, 3],
+   *   x[16, 2], x[16, 3], x[17, 2], x[17, 3], ... x[30, 2], x[30, 3],    -    ,    -    ,
+   *   x[ 0, 4], x[ 0, 5], x[ 1, 4], x[ 1, 5], ... x[14, 4], x[14, 5], x[15, 4], x[15, 5],
+   *   x[16, 4], x[16, 5], x[17, 4], x[17, 5], ... x[30, 4], x[30, 5],    -    ,    -    ,
+   * `
+   */
+  device_mdarray<T, extent_2d, row_major> data;
```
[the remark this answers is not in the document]
achirkin: Hmm, I've been thinking to follow our cuml dev guide here, which says to try to keep the structures POD/standard layout (modulo mdarray members). I believe it makes sense in this case for a few reasons: (1) to make it easier for third-parties to write their own serialization/deserialization and (2) to simplify cython wrappers if we choose to write them (like e.g. the linear_svm struct I added in cuml); (3) also it needs less code to maintain :).
At the same time, (4) I keep this structure `const` in the code, so there is no need to do any extra protection using private accessors. On the other hand though, there anyway shouldn't be a need to access the mdarray members except for the two methods we already have plus serialization.

That being said, If you still prefer to make it classy, I'm fine with that; please confirm.

cpp/include/raft/spatial/knn/detail/ann_ivf_flat.cuh
achirkin: `ivf_flat_index` is now supposed to be the persistent part of the things. Without doing changes to the `metric_type`, I think, `metric_type` should be a part of `ivf_flat_index`, because the clusters are built with respect to the chosen metric (hence it does not make sense to have different metrics during building and searching).

I believe, `greater_` parameter could simply be a part of the search arguments. Though, I need to check in the code, how difficult it would be to not use it during `cuivflSetSearchParameters`/`queryIVFFlatGridSize`.

The `grid_dim_x_` member can be queried before search and it remains the same across searches if the batch size (`n_queries`) stays the same. I can check, I'm not sure if I can get rid of this parameter, but without it I'd like the code more.

The last member, `search_mem_res` is the memory resource/pool that can be reused across searches. It's there purely for performance reasons - to make expensive cudaMalloc* functions invoked only during the first search (or to preallocate a large enough buffer in advance).

With this in mind, I suggest the following:
  1. Keep all serialized state in `ivf_flat_index`
  2. Make either of the two:
       - (A)  non-persistent `ivf_flat_handle`, that would contain build/load/store/search methods and keep `ivf_flat_index` as an `optional<const>` member along with `search_mem_res` and possibly `grid_dim_x_`.
       - (B)  non-persistent `ivf_flat_search_state` that would keep  `search_mem_res` and possibly `grid_dim_x_`, and would be passed to the global public `search` function as an argument (build/load/store would be global public functions as well)

Which one would you prefer?

cpp/include/raft/spatial/knn/detail/ann_ivf_flat.cuh
```diff
@@ -0,0 +1,465 @@
...
+
+  void cuivflSearch(
+    const T* queries, uint32_t n_queries, uint32_t k, size_t* neighbors, float* distances);
+
+  void queryIVFFlatGridSize(const uint32_t n_probes, const uint32_t n_queries, const uint32_t k);
+  uint32_t getDim() { return index_.has_value() ? index_->dim() : 0; }
+
+ private:
+  const handle_t& handle_;
+  const rmm::cuda_stream_view stream_;
```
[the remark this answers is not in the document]
achirkin: I think we've had some discussions in the past in cuml, that somebody may want to run one algorithm in several OpenMP threads, using different streams?.. Not sure, how is that applicable to ann search.

cpp/include/raft/spatial/knn/detail/ann_quantized_faiss.cuh
```diff
@@ -82,6 +86,54 @@ void approx_knn_ivfflat_build_index(
   index->index = faiss_index;
 }
 
+template <typename T = float, typename IntType = int>
+void approx_knn_cuivfl_ivfflat_build_index(knnIndex* index,
```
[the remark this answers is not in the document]
achirkin: No longer relevant, as we don't expose any of the managed memory buffers anymore (they are part of the optimized kmeans now).

cpp/include/raft/spatial/knn/detail/ann_ivf_flat.cuh
```diff
@@ -0,0 +1,787 @@
...
+  nprobe_ = nprobe;
+  if (nprobe_ <= 0) { return CUIVFL_STATUS_INVALID_VALUE; }
+  // Set the greater_
+  if (metric_type_ == raft::distance::DistanceType::L2Expanded ||
+      metric_type_ == raft::distance::DistanceType::L2Unexpanded) {
+    greater_ = false;
+  } else {
+    // Need to set this to true for inner product if need FAISS like behavior for inner product
+    greater_ = false;
+  }
```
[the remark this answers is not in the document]
achirkin: Answering to other questions, I stumbled upon this today. Would it make sense, from the theory standpoint, to lift `greater_` as an argument of the search function? i.e. something like `ascending` parameter of the topk, to return closest/furthest points. I'm not sure how the "furthest" neighbors mode would perform in terms of precision/recall (since the clustering is based on the nearest neighbours in either case).

cpp/include/raft/spatial/knn/detail/ann_ivf_flat_kernel.cuh
```diff
@@ -0,0 +1,1227 @@
...
+
+namespace raft::spatial::knn::detail {
+
+constexpr int kThreadsPerBlock = 128;
+
+namespace md = std::experimental;
+
+template <typename T>
+struct ivf_flat_index {
+  using row_major = md::layout_right;
```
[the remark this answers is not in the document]
cjnolet: > Would it be ok if we exposed layout_rightor row_major in the raft namespace as well?

Yes, absolutely I think having both exposed is fine. As the `mdspan`/`mdarray` is a somewhat natural analogue to Numpy's `ndarray`, we preferred to also expose verbiage which is familiar to those users, but that's in addition to and not instead of, the current accepted `mdspan`-specific verbiage.

achirkin:
@cjnolet, in the last commit (2a3eb3385091a815be5701953fb5bf99409453bc), as an experiment, I did rather big api changes by renaming/adding parameter types; I'll revert it back if necessary. What do you think about it? The main idea is to split the parameters into two `knn_index_params` and `ivf_search_params`.

Following the previously established practice, I used dynamic polymorphism to choose between the available ANN implementations. I think, we can do the same with `knnIndex` (extend the base handle with the stuff specific to the implementations). However, I'm fully happy about it and wondering, what if we used a tag system+templates instead? i.e. `approx_knn_build_index` would be parameterized by the implementation tag and would return the index type specific to the implementation. Then, we could either make `search` be a member of that index type, or just make the `approx_knn_search` parameterized the same way as `approx_knn_build_index`.

cpp/include/raft/spatial/knn/detail/ann_ivf_flat_kernel.cuh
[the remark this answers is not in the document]
cjnolet: I wasn't thinking of this as a POD struct from the start, mostly because the bulk of its state is rather complex objects, and data pointers to device memory at that (and `mdarray` itself is not a POD, right?). I'm not completely against sticking w/ using the `check_consistency` method, but my proposal would 1) guarantee the consistency from the time of construction (like `mdarray` does), and 2) make the expectations more immediately obvious to future users of this object.

In terms of serialization, my proposal is more inline w/ the design of the corresponding FAISS indices and serialization frameworks (such as protobuf, as an example) generally provide mechanisms to adapt complex objects to make them serializable. I think we might need to do this anyways, at least for the `mdarray` type?

I think this is a fairly opaque object, since it's constructed by a RAFT function and intended to be used within a RAFT function. I guess the question then becomes, how often do we intend for users outside of RAFT to directly interact w/ its members?

cpp/include/raft/spatial/knn/detail/ann_ivf_flat.cuh
cjnolet: > Without doing changes to the metric_type, I think, metric_type should be a part of ivf_flat_index, because the clusters are built with respect to the chosen metric (hence it does not make sense to have different metrics during building and searching).

Yep, 100% agreed the metric type should be stored in the index. 


> Which one would you prefer?

Does option 2A come w/ the implication that the users would need to construct a handle w/ an index? It's an extra step the users would need to take, unless maybe from the API perspective, they are loading the `handle` from disk instead of the index? Not a show-stopper, but it does seem like a little more awkward flow.

What do you think about maybe a combination of options 1 and 2B? We could create an `ivf_flat_search_state` and add it as a (non-persistent) member of the `ivf_flat_index`?

achirkin: The search state doesn't feel to belong to the index. Since you are more inclined towards option B, I'd prefer to not add an extra structure for the search state at all. I removed the `grid_dim_x_` already, so the only state left is the rmm memory resource. I think, we can just pass it as an optional argument, the same way as I do it now with `radix_topk`.
Afterwards, we can repeat the same discussion with the outermost `knnIndex` class :)

cpp/include/raft/spatial/knn/detail/ann_ivf_flat_kernel.cuh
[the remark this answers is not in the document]
achirkin: Following our other discussion below, I've refactored `build` and `search` into public functions and eliminated the handle type completely. The relevant types are now in [ivf_flat_types.hpp](https://github.com/rapidsai/raft/blob/df17b5b631ddec4e277c2af4386b9aa9f073bdd8/cpp/include/raft/spatial/knn/ivf_flat_types.hpp) and the functions are in [ivf_flat.cuh](https://github.com/rapidsai/raft/blob/df17b5b631ddec4e277c2af4386b9aa9f073bdd8/cpp/include/raft/spatial/knn/ivf_flat.cuh).

Back to our topic. In the current state, the users are not expected to construct the `index<T>` manually anyway, they should call the factory function `build<T>`, which ensures consistency. All public functionality is exposed in `ivf_flat.cuh` anyway, so I don't think there could be any difficulties with the expectations on the use of the type. Later we could add similar functions to serialize/deserialize etc to `ivf_flat.cuh`. At the same time, I can imagine a certain user wants to do access the raw index data for whatever reason; maybe they need a custom serialization (cannot use ours) or do some modifications to the computed clusters. All of the members (mdarrays) are constructible from outside raft, so this is possible now. If we choose to switch to the opaque object, all of these would require us to write more code.

Also, personally I'd prefer to have a more homogenous api: either make everything via object methods or keep everything as public functions. I have a feeling that adding constructors, serialization, and an array of accessor methods would go against that in the presence of our `build` and `search` public functions.

cpp/include/raft/spatial/knn/detail/ann_quantized.cuh
```diff
@@ -0,0 +1,285 @@
...
+    std::unique_ptr<MetricProcessor<float>> query_metric_processor =
+      create_processor<float>(metric, n, D, 0, false, stream);
+
+    if (ivf_ft_pams) {
+      // cuivfl only supports L2/Inner product for now.
+      if (metric == raft::distance::DistanceType::L2SqrtExpanded ||
+          metric == raft::distance::DistanceType::L2SqrtUnexpanded ||
+          metric == raft::distance::DistanceType::L2Unexpanded ||
+          metric == raft::distance::DistanceType::L2Expanded ||
+          metric == raft::distance::DistanceType::InnerProduct) {
```
tfeher, line 172: Currently the [python layer accepts](https://github.com/rapidsai/cuml/blob/6f51c1fd86af337ca1cddb02ce8e03e25bb9df2d/python/cuml/neighbors/__init__.py#L42) cosine similarity and correlation metrics as well. Please add it to the todo list to support them.

cpp/include/raft/spatial/knn/ivf_flat.cuh
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * @tparam T data element type
+ *
+ * @param handle
+ * @param params configure the index building
+ * @param[in] dataset a device pointer to a row-major matrix [n_rows, dim]
+ * @param n_rows the number of samples
+ * @param dim the dimensionality of the data
+ * @param stream
+ *
```
tfeher, line 41: Could you document the supported metric types?

cpp/include/raft/spatial/knn/detail/ann_ivf_flat.cuh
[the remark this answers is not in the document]
cjnolet: I think that's a little tangential to this discussion, and one we're actually starting to have as we begin to discuss the potential for different execution policies within RAFT. I'd like additional streams to be removed from the API wherever possible to alleviate the confusion and potential misuse. This one shouldn't be a problem as it's coming directly from the handle.

cpp/include/raft/spatial/knn/detail/ann_ivf_flat_kernel.cuh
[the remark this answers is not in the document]
cjnolet: > Also, personally I'd prefer to have a more homogenous api: either make everything via object methods or keep everything as public functions. 

RAFT's core public APIs are flat public functions. That said, there are times when we use more complex objects which provide proper constructors and methods to maintain consistency of their own internal state (thinking `handle`, `mdspan`, `mdarray` here). In general, we have already been following the homogenous design that you mention but the `cuivfhandle` breaks this design by exposing public methods on the handle itself. That design isn't really necessary anyways because it's just being invoked from flattened public functions anyways. I don't think we need the extra layer there.

cpp/include/raft/spatial/knn/detail/ann_ivf_flat.cuh
cjnolet: > Afterwards, we can repeat the same discussion with the outermost knnIndex class :)

My take on your modifications to the `knnIndex` class is that you were trying to minimize public API changes by reusing the existing structures. I'm generally in favor of this approach but the goal here is to remove FAISS as a dependency, and so that will alleviate the need for this `knnIndex` class. My suggestion is that we add the new functions using `mdspan` and not worry about reusing the existing functions. The refactoring to cuml will be minimal anyways.

cpp/include/raft/spatial/knn/detail/ann_ivf_flat_kernel.cuh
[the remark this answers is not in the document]
cjnolet: > Following our other discussion below, I've refactored build and search into public functions and eliminated the handle type completely. 

This looks much better.

cpp/include/raft/spatial/knn/detail/ivf_flat_search.cuh
```diff
@@ -0,0 +1,1370 @@
...
+    //@TODO The result with dimension
+    // The start address of the full value of vector for each cluster(list) interleaved
+    auto vecsBase = list_data + size_t(list_prefix_interleave[listId]) * dim;
+    // The start address of index of vector for each cluster(list) interleaved
+    auto indexBase = list_index + list_prefix_interleave[listId];
+    // The number of vectors in each cluster(list); [nlist]
+    const uint32_t numVecs = list_lengths[listId];
+
+    // The number of interleaved group to be processed
+    const uint32_t numBlocks = ceildiv<uint32_t>(numVecs, WarpSize);
```
tfeher, line 814: Since block has multiple meanings, why don't we call this numGroups?

cpp/include/raft/spatial/knn/detail/ivf_flat_search.cuh
```diff
@@ -0,0 +1,1370 @@
...
+      lds(queryRegs[j], &query_shared[d]);
+#pragma unroll
+      for (int k = 0; k < Veclen; ++k) {
+        compute_dist(dist, queryRegs[j][k], encV[j][k]);
+      }
+    }
+  }
+
+  template <typename IdxT>
+  __device__ __forceinline__ void runLoadShflAndCompute(const T*& data,
```
tfeher, line 159: // Computes distance between one query vector (per thread) and one vector from the index. 
// It runs on a set of coordinates defined by baseLoadIndex and kUnroll parameter.

tfeher (changes requested):

cpp/include/raft/spatial/knn/detail/ivf_flat_search.cuh
```diff
@@ -0,0 +1,1370 @@
...
+      // This is where this warp begins reading data
+      const T* data =
+        vecsBase + size_t(block) * wordsPerVectorBlockDim * dim;  // Start position of this block
+
+      if (valid) {
+        /// load query from shared mem
+        for (int dBase = 0; dBase < shLoadDim; dBase += WarpSize) {  //
+          constexpr int kUnroll   = WarpSize / Veclen;
+          constexpr int stride    = kUnroll * Veclen;
+          constexpr int totalIter = WarpSize / stride;
```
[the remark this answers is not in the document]
achirkin: That's a funny one!

cpp/include/raft/spatial/knn/detail/ivf_flat_search.cuh
```diff
@@ -0,0 +1,1297 @@
...
+};
+
+/**
+ * See `ivfflat_interleaved_scan` for parameter docs.
+ *
+ * query_smem_elems must be multiple of WarpSize * Veclen
+ */
+template <int Capacity, int Veclen, bool Ascending, typename T, typename AccT, typename Lambda>
+__global__ void __launch_bounds__(kThreadsPerBlock)
+  interleaved_scan_kernel(Lambda compute_dist,
```
tfeher, line 673: Since the name of this kernel is not too helpful, could you add a short description of what exactly this kernel does? Something like:

```
Scan clusters for nearest neighbors of the query vectors.

The clusters are stored in the interleaved index format described in ivf_flat_types.hpp.
For each query vector, a set of clusters is probed: the distance to each vector in the cluster is calculated, and the top-k nearest neighbors are selected.
```

cpp/include/raft/spatial/knn/detail/ivf_flat_search.cuh
```diff
@@ -0,0 +1,1297 @@
...
+    default: select_min = true;
+  }
+
+  //   // Set memory buffer to be reused across searches
+  //   auto cur_memory_resource = rmm::mr::get_current_device_resource();
+  //   if (!search_mem_res_.has_value() || search_mem_res_->get_upstream() != cur_memory_resource) {
+  //     search_mem_res_.emplace(cur_memory_resource, Pow2<256>::roundUp(n_queries * n_probes * k *
+  //     16));
+  //   }
+  std::optional<rmm::mr::pool_memory_resource<rmm::mr::device_memory_resource>> pool_res;
```
tfeher, line 1286: Remove?

tfeher (changes requested):
Thanks Artem for addressing the issues so far! Overall the internals look good to me, there are just a few additional comments that I have.

cpp/include/raft/spatial/knn/ivf_flat.cuh
achirkin: Not yet, sorry, I wanted first to figure out which metrics we should supported and if we need to involve `MetricProcessor` for this.

cpp/test/spatial/ann_ivf_flat.cu
```diff
@@ -0,0 +1,273 @@
...
+
+ private:
+  raft::handle_t handle_;
+  rmm::cuda_stream_view stream_;
+  AnnIvfFlatInputs ps;
+  rmm::device_uvector<DataT> database;
+  rmm::device_uvector<DataT> search_queries;
+};
+
+const std::vector<AnnIvfFlatInputs> inputs = {
```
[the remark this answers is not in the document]
achirkin: Apparently, they weren't, until now (NB: we have these branches only for selecting the clusters to search i.e. the first phase; in the main part only warp sort is used).

cpp/bench/spatial/knn.cu
```diff
@@ -0,0 +1,377 @@
...
+const std::vector<Scope> kAllScopes{Scope::BUILD, Scope::SEARCH, Scope::BUILD_SEARCH};
+
+#define KNN_REGISTER(ValT, IdxT, ImplT, inputs, strats, scope)                   \
+  namespace BENCHMARK_PRIVATE_NAME(knn)                                          \
+  {                                                                              \
+    using KNN = knn<ValT, IdxT, ImplT<ValT, IdxT>>;                              \
+    RAFT_BENCH_REGISTER(KNN, #ValT "/" #IdxT "/" #ImplT, inputs, strats, scope); \
+  }
+
+KNN_REGISTER(float, int64_t, brute_force_knn, kInputs, kAllStrategies, kScopeFull);
```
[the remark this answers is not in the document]
achirkin: NB: moved this whole file to https://github.com/rapidsai/raft/pull/730

cpp/include/raft/spatial/knn/detail/ann_ivf_flat.cuh
```diff
@@ -0,0 +1,465 @@
...
+ private:
+  const handle_t& handle_;
+  const rmm::cuda_stream_view stream_;
+  ivf_flat_params params_;
+
+  const raft::distance::DistanceType metric_type_;
+  bool greater_;
+  uint32_t grid_dim_x_;  // The number of blocks launched across n_probes.
+  // The built index
+  std::optional<const ivf_flat_index<T>> index_ = std::nullopt;
```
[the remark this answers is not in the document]
achirkin: Not applicable anymore, since `ivf_flat::index` is returned from the build function (if it exists, it's trained).

cpp/include/raft/spatial/knn/ivf_flat.cuh
achirkin: Added this to the documentation here (`ivf_flat::build` method where it provide metric in the arguments).

achirkin:
NB: at the moment, this PR includes changes from #726 and #730

cpp/include/raft/spatial/knn/ivf_flat.cuh
```diff
@@ -0,0 +1,96 @@
...
+ * @return the constructed ivf-flat index
+ */
+template <typename T, typename IdxT = uint32_t>
+inline auto build(const handle_t& handle,
+                  const index_params& params,
+                  const T* dataset,
+                  IdxT n_rows,
+                  uint32_t dim,
+                  rmm::cuda_stream_view stream) -> index<T, IdxT>
```
cjnolet, line 56: We should remove this explicit stream here and it should come from the handle.

cpp/include/raft/spatial/knn/ivf_flat.cuh
```diff
@@ -0,0 +1,96 @@
...
+template <typename T, typename IdxT>
+inline void search(const handle_t& handle,
+                   const search_params& params,
+                   const index<T, IdxT>& index,
+                   const T* queries,
+                   uint32_t n_queries,
+                   uint32_t k,
+                   IdxT* neighbors,
+                   float* distances,
+                   rmm::cuda_stream_view stream,
```
cjnolet, line 89: This should be coming from the handle as well.

cpp/include/raft/spatial/knn/ivf_flat.cuh
```diff
@@ -0,0 +1,96 @@
...
+ * @param[in] dataset a device pointer to a row-major matrix [n_rows, dim]
+ * @param n_rows the number of samples
+ * @param dim the dimensionality of the data
+ * @param stream
+ *
+ * @return the constructed ivf-flat index
+ */
+template <typename T, typename IdxT = uint32_t>
+inline auto build(const handle_t& handle,
+                  const index_params& params,
```
cjnolet, line 52: I notice there's no way to specify the metric here or in the index_params. We should probably add that somewhere, even if we just limit it to euclidean to start, so that we don't have to break the API when add support for more of them.

cpp/include/raft/spatial/knn/ann.cuh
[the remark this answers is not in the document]
achirkin: (I'm reverting the changes to this API and deprecating it)

tfeher (approved):
Thanks Artem for addressing the issues! If there is any additional issue about the API, then I am sure @cjnolet will follow that up. Otherwise, the PR looks good to me.

cpp/include/raft/spatial/knn/detail/ivf_flat_build.cuh
```diff
@@ -0,0 +1,208 @@
...
+  for (uint32_t l = 0; l < dim; l += veclen) {
+    for (uint32_t j = 0; j < veclen; j++) {
+      list_data[l * kIndexGroupSize + ingroup_id + j] = dataset[l + j];
+    }
+  }
+}
+
+/** See raft::spatial::knn::ivf_flat::build docs */
+template <typename T, typename IdxT>
+inline auto build(const handle_t& handle,
```
cjnolet, line 105: As discussed offline, we will need to separate the training of the "quantizer" (kmeans in this case) from the ivf.

cjnolet (changes requested):
Thanks Artem for the changes so far! I made another pass through the implementation now that I've had a chance to go through `RaftIndexIVFFlat`. I think this PR is getting very close.

I'm adding an action item for myself to find out if (and why) support for the allocate/reclaim methods in the FAISS API is absolutely necessary.

cpp/include/raft/spatial/knn/detail/ann_utils.cuh
```diff
@@ -0,0 +1,491 @@
...
+ * @param n_cols
+ * @param[in] src input matrix [..., ld_src]
+ * @param[in] row_ids selection of rows to be copied [n_rows]
+ * @param ld_src number of cols in the input (ld_src >= n_cols)
+ * @param[out] dst output matrix [n_rows, ld_dst]
+ * @param ld_dst number of cols in the output (ld_dst >= n_cols)
+ * @param stream
+ */
+template <typename T, typename S>
+void copy_selected(uint32_t n_rows,
```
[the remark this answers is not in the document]
achirkin: It may have different source and target types. The mapping function is hardcoded, so if we want to move this to the global namespace, we should make it into a lambda.

cpp/include/raft/spatial/knn/detail/ann_utils.cuh
```diff
@@ -0,0 +1,491 @@
...
+ * divisor is zero.
+ *
+ * NB: device-only function
+ *
+ * @param[in] n_rows
+ * @param[in] n_cols
+ * @param[inout] a device pointer to a row-major matrix [n_rows, n_cols]
+ * @param[in] d device pointer to a vector of divisors [n_rows]
+ */
+inline void divide_along_rows(
```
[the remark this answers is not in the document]
achirkin: Same issue as in copyRows

cpp/include/raft/spatial/knn/detail/ivf_flat_build.cuh
achirkin: Done via splitting `build` into `build` and `extend`;  `index_params` has now a parameter to tell if we want to add vectors on `build` (default to true).

cpp/include/raft/spatial/knn/ivf_flat_types.hpp
```diff
@@ -0,0 +1,162 @@
...
+   * `
+   *   x[ 0, 0], x[ 0, 1], x[ 1, 0], x[ 1, 1], ... x[14, 0], x[14, 1], x[15, 0], x[15, 1],
+   *   x[16, 0], x[16, 1], x[17, 0], x[17, 1], ... x[30, 0], x[30, 1],    -    ,    -    ,
+   *   x[ 0, 2], x[ 0, 3], x[ 1, 2], x[ 1, 3], ... x[14, 2], x[14, 3], x[15, 2], x[15, 3],
+   *   x[16, 2], x[16, 3], x[17, 2], x[17, 3], ... x[30, 2], x[30, 3],    -    ,    -    ,
+   *   x[ 0, 4], x[ 0, 5], x[ 1, 4], x[ 1, 5], ... x[14, 4], x[14, 5], x[15, 4], x[15, 5],
+   *   x[16, 4], x[16, 5], x[17, 4], x[17, 5], ... x[30, 4], x[30, 5],    -    ,    -    ,
+   * `
+   */
+  device_mdarray<T, extent_2d, row_major> data;
```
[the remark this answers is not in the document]
achirkin: I'm putting this into a follow-up PR as the first priority change.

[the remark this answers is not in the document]
achirkin: https://github.com/rapidsai/raft/pull/747

cjnolet (approved):
Discussed further changes offline to get this in for 22.08.

cjnolet:
rerun tests

cjnolet:
rerun tests

tfeher:
Thanks Artem for the updates! I had a look at the changes since my last review, and overall it looks good to me. I have some smaller comments (mostly about docs) we can address those in #747.

cjnolet:
rerun tests

cjnolet:
@gpucibot merge

cpp/include/raft/spatial/knn/detail/ann_utils.cuh
```diff
@@ -0,0 +1,491 @@
...
+ * @brief Divide rows by their L2 norm (square root of sum of squares).
+ *
+ * NB: device-only function
+ *
+ * @param[in] n_rows
+ * @param[in] n_cols
+ * @param[inout] a device pointer to a row-major matrix [n_rows, n_cols]
+ * @param stream
+ */
+inline void normalize_rows(uint32_t n_rows, uint32_t n_cols, float* a, rmm::cuda_stream_view stream)
```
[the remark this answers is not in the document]
Nyrio: @achirkin I think it's ok, I don't really see a case where one would compute the norm / normalize a thick matrix, since the rows are typically samples and the cols their dimensionality. In fact, if you look at `raft::linalg::rowNorm`, it is using the `coalescedReductionKernel` which is doing a cub-based reduction with one block per row, quite similarly to `normalize_rows_kernels`.

I'll replace the hardcoded reduction in `normalize_rows_kernel` with a cub-based reduction and move it to the matrix or linalg namespace (and benchmark before/ater perf)

PR #655, opened 2022-05-16
Don't install component dependency files in raft-header only mode

robertmaynard:
Corrects the issues we are seeing in CI with failed `faiss:faiss` dependency


cjnolet (approved):

cjnolet:
@gpucibot merge

PR #659, opened 2022-05-18
Fix SelectionTest: allow different indices when keys are equal.

achirkin:
If the dists (keys) are the same, different corresponding ids may end up in the selection due to non-deterministic nature of some implementations.
This PR updates the comparison functor in the tests to account for this problem.


cjnolet:
PR looks good. We should probably create an issue for the failing test just so it doesn't fall off our radar.

cjnolet (approved):

cjnolet:
@gpucibot merge

PR #662, opened 2022-05-19
Fix minor errors in CMake configuration

vyasr:
This PR fixes a number of small errors that slipped through in #644 and #633. None of them materially affect the build (whether a preexisting libraft exists or not) since 1) benchmarks are currently turned off by default anyway in the C++ build, and 2) pyraft only depends on libraft for headers, not for compiled libraries.


cjnolet (approved):

vyasr:
@cjnolet so it turns out that I did in fact previously test this, and the current build _will_ work even without the changes that I proposed in my last message. The important distinction between pyraft and pylibraft  is that pyraft actually doesn't depend on raft compiled libraries at all, it just requires some of the headers. Those headers are already being inherited by linking in the raft::raft target, so we should be set. I've done a little more cleanup here, but aside from that I think we're good.

cjnolet:
Want to go ahead and merge it?

vyasr:
@gpucibot merge

vyasr:
Done!

PR #664, opened 2022-05-20
gemm: fix [inout] parameter C mistakenly set as const

achirkin:
Fix [inout] parameter C being mistakenly set as const in one of the raft::gemm overloads .


cjnolet (approved):

cjnolet:
@gpucibot merge

PR #653, opened 2022-05-16
Map CMake install components to conda library packages

robertmaynard:
This extends https://github.com/rapidsai/raft/pull/648 and maps the CMake install components to capture what I think is the desire to have them map to conda packages.

Mapping:

- CMake 'raft' componet => conda raft-headers
- CMake 'raft_nn' componet => conda libraft-component
- - CMake 'raft_distance' componet => conda libraft-distance

PR #668, opened 2022-05-23
Re-establishing reverted changes from #653 and #655

cjnolet:


cpp/CMakeLists.txt
```diff
@@ -390,6 +390,44 @@ install(FILES ${CMAKE_CURRENT_BINARY_DIR}/include/raft/version_config.hpp
...
+endif()
+if(TARGET raft_nn_lib)
+  list(APPEND raft_components  nn-lib)
+  list(APPEND raft_install_comp _nn)
+endif()
+
+foreach(comp install_comp IN ZIP_LISTS raft_components raft_install_comp)
+  install(
+          EXPORT raft-${comp}-exports
+          FILE raft-${comp}-targets.cmake
```
vyasr, line 412: You'll end up with double hyphens like `raft--targets.cmake` for the header-only distance/nn components now, which isn't what you want. You probably want to set the filename and the export set in a separate conditional depending on whether the install component is empty or not.

cpp/CMakeLists.txt
```diff
@@ -352,12 +352,12 @@ install(TARGETS raft
 
 install(TARGETS raft_distance
         DESTINATION ${lib_dir}
-        COMPONENT raft_distance
+        COMPONENT raft
```
vyasr, line 355: So the default raft component includes the header-only versions, and requesting specific components is only for pre-built libraries? Making sure that I understand the intention here.

cpp/CMakeLists.txt
vyasr: (This also affects the export commands below)

vyasr (changes requested):

Issue #667, opened 2022-05-23
[BUG] Logger compilation issue


lowener:
**Describe the bug**
The logger of raft is causing compilation issue when the header is included, because of multiple definition of `raft::format`

**Steps/Code to reproduce bug**
Add `#include <raft/common/logger.hpp>` to any file (in this example test/common/handle.cpp).
However it works if I just remove test/common/logger.cpp because the multiple definition is avoided.

**Error message**
```
/bin/ld: CMakeFiles/test_raft.dir/test/handle.cpp.o: in function `raft::format[abi:cxx11](char const*, __va_list_tag (&) [1])':
handle.cpp:(.text+0x340): multiple definition of `raft::format[abi:cxx11](char const*, __va_list_tag (&) [1])'; CMakeFiles/test_raft.dir/test/common/logger.cpp.o:logger.cpp:(.text+0x490): first defined here
/usr/bin/ld: CMakeFiles/test_raft.dir/test/handle.cpp.o: in function `raft::format[abi:cxx11](char const*, ...)':
handle.cpp:(.text+0x4d0): multiple definition of `raft::format[abi:cxx11](char const*, ...)'; CMakeFiles/test_raft.dir/test/common/logger.cpp.o:logger.cpp:(.text+0x620): first defined here
/usr/bin/ld: CMakeFiles/test_raft.dir/test/handle.cpp.o: in function `raft::convert_level_to_spdlog(int)':
handle.cpp:(.text+0x5a0): multiple definition of `raft::convert_level_to_spdlog(int)'; CMakeFiles/test_raft.dir/test/common/logger.cpp.o:logger.cpp:(.text+0x6f0): first defined here
collect2: error: ld returned 1 exit status
ninja: build stopped: subcommand failed.
```

lowener:
Solved by #673

PR #673, opened 2022-05-24
Improve logging

achirkin:
A few fixes/improvements to the logger:

  - Removed one of the two duplicate versions (`common/detail/logger.hpp`)
  - Added `RAFT_LOG_TRACE_VEC` to trace device/host arrays
  - Added `inline` annotations to some of the helper functions (to fix the linker error)
  - Replaced `char[]` with `vector<char>` where applicable for safety and to address `clang-tidy` warnings
  - Changed the default logger level to be controlled by the macro `RAFT_ACTIVE_LEVEL` instead of the fixed `RAFT_LEVEL_INFO`, but set the default `RAFT_ACTIVE_LEVEL` to `RAFT_LEVEL_INFO`.


cpp/include/raft/common/detail/logger.hpp
```diff
@@ -15,139 +15,8 @@
  */
 #pragma once
 
-#include <stdarg.h>
+#pragma message(__FILE__                                                 \
+                " is deprecated and will be removed in future releases." \
+                " Please use the <raft/common/logger.hpp> version instead.")
```
lowener, line 20:
```suggestion
                " Please use the <raft/core/logger.hpp> version instead.")
```

lowener: Or this file could also be deleted since it's in the detail headers. I can't find a file using this header.

lowener (approved):
Nice improvement. I'm also linking #667 that I opened yesterday for the logger compilation issue.

cpp/include/raft/common/detail/logger.hpp
achirkin: I'm not sure. Is an outside user supposed to use `<raft/common/logger.hpp>` or `<raft/core/logger.hpp>`?

lowener: `common/logger.hpp` is deprecated and will be removed in 22.08 according to the file comments.

achirkin: Ah, thanks! I missed that

tfeher (approved):

cjnolet (approved):

cjnolet:
@gpucibot merge

Issue #676, opened 2022-05-24
[BUG] race condition in `raft::spatial::knn::detail::fusedL2kNN`


viclafargue:
**Describe the bug**
Race condition in `raft::spatial::knn::detail::fusedL2kNN`
```
========= WARN: Race reported between Write access at 0x0002f6e0 in _ZN4raft7spatial3knn6detail10fusedL2kNNILb0EffflNS_6linalg12KernelPolicyIfLi1ELi16ELi2ELi8ELi8ELi32EEEZNS2_19fusedL2UnexpKnnImplIffflLi4ELb0ELb1EEEvPKT_SA_T2_SB_SB_SB_SB_SB_bPT1_PSB_SB_P11CUstream_stPvRmEUlRfSJ_SJ_E_ZS7_IffflLi4ELb0ELb1EEvSA_SA_SB_SB_SB_SB_SB_SB_bSD_SE_SB_SG_SH_SI_EUlfiE_Li32ELi2ELb0ELb1EEEvPKT0_SO_SO_SO_T3_SP_SP_SP_SP_SP_T5_T6_bjPViPVSB_PVSP_
=========     and Read access at 0x000302c0 in _ZN4raft7spatial3knn6detail10fusedL2kNNILb0EffflNS_6linalg12KernelPolicyIfLi1ELi16ELi2ELi8ELi8ELi32EEEZNS2_19fusedL2UnexpKnnImplIffflLi4ELb0ELb1EEEvPKT_SA_T2_SB_SB_SB_SB_SB_bPT1_PSB_SB_P11CUstream_stPvRmEUlRfSJ_SJ_E_ZS7_IffflLi4ELb0ELb1EEvSA_SA_SB_SB_SB_SB_SB_SB_bSD_SE_SB_SG_SH_SI_EUlfiE_Li32ELi2ELb0ELb1EEEvPKT0_SO_SO_SO_T3_SP_SP_SP_SP_SP_T5_T6_bjPViPVSB_PVSP_ [120 hazards]
[...]

PR #679, opened 2022-05-25
fix race in fusedL2knn smem read/write by adding a syncwarp

mdoijade:
-- fix race in fusedL2knn smem read/write by adding a syncwarp reported by racecheck tool of compute-sanitizer.
-- this addresses issue - https://github.com/rapidsai/raft/issues/676


cjnolet:
@gpucibot merge

PR #681, opened 2022-05-25
Revert print vector changes because of std::vector<bool>

lowener:
The specialization of `std::vector` when `T=bool` is unfortunately causing compilation issue in cuml because the `data()` function member is not implemented. And the elements may not be stored contiguously.

(Link to the CI failure: https://gpuci.gpuopenanalytics.com/job/rapidsai/job/gpuci/job/cuml/job/prb/job/cuml-cpu-cuda-build-arm64/CUDA=11.5/1364/console)
cc @achirkin

### DIFF
--- a/cpp/include/raft/core/cudart_utils.hpp
+++ b/cpp/include/raft/core/cudart_utils.hpp
@@ -36,7 +36,6 @@
 #include <iomanip>
 #include <iostream>
 #include <mutex>
-#include <vector>
 
 ///@todo: enable once logging has been enabled in raft
 //#include "logger.hpp"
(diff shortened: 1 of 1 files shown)


cjnolet (dismissed):
LGTM

cpp/include/raft/core/cudart_utils.hpp
```diff
@@ -303,10 +302,10 @@ void print_device_vector(const char* variable_name,
                          size_t componentsCount,
                          OutStream& out)
 {
-  std::vector<T> host_mem(componentsCount);
-  CUDA_CHECK(
-    cudaMemcpy(host_mem.data(), devMem, componentsCount * sizeof(T), cudaMemcpyDeviceToHost));
-  print_host_vector(variable_name, host_mem.data(), componentsCount, out);
+  T* host_mem = new T[componentsCount];
```
achirkin, line 305: Then, let it be a `unique_ptr` to avoid a memory leak on exception, e.g.
```
auto host_mem = std::make_unique<T[]>(componentsCount);
```

achirkin (changes requested):
Huh, good point :) still, let us not have manual allocations

cjnolet:
I’m going to go ahead and merge this guy for now to fix the cuml build, though @lowener it would be nice if we could still address @achirkin’s concerns in a follow on PR (maybe for 22.08)

cjnolet:
@gpucibot merge

PR #686, opened 2022-05-25
Restore changes from #653 and #655 and correct cmake component dependencies

robertmaynard:


vyasr:
Was something wrong with #668?

robertmaynard:
> Was something wrong with #668?

I felt a new PR was needed as this isn't just a pure revert, but also changing the behavior of dependency recording.
But if we want me to instead push to #668 I would be happy to.

vyasr:
That's fine, I was just confused to see this.

cjnolet:
@vyasr I opened #668 just out of courtesy since I was the one that did the original revert last week. I certainly have no problem at all with a new PR if it helps us get closer to merging https://github.com/rapidsai/raft/pull/641.

cpp/CMakeLists.txt
```diff
@@ -390,6 +393,44 @@ install(FILES ${CMAKE_CURRENT_BINARY_DIR}/include/raft/version_config.hpp
...
+          EXPORT raft-${comp}-exports
+          FILE ${RAFT_BINARY_DIR}/raft-${comp}-targets.cmake
+          NAMESPACE raft::
+  )
+  rapids_export_write_dependencies(
+          BUILD raft-${comp}-exports "${PROJECT_BINARY_DIR}/raft-${comp}-dependencies.cmake"
+  )
+  rapids_export_write_dependencies(
+          INSTALL raft-${comp}-exports "${PROJECT_BINARY_DIR}/rapids-cmake/raft/export/${install_comp}/raft-${comp}-dependencies.cmake"
+  )
```
vyasr, line 430: For my edification, why do we need both of these?

cpp/CMakeLists.txt
```diff
@@ -390,6 +393,44 @@ install(FILES ${CMAKE_CURRENT_BINARY_DIR}/include/raft/version_config.hpp
...
+include("${rapids-cmake-dir}/export/write_dependencies.cmake")
+
+set(raft_components distance nn)
+set(raft_install_comp "" "")
+if(TARGET raft_distance_lib)
+  list(APPEND raft_components distance-lib)
+  list(APPEND raft_install_comp _distance)
+endif()
+if(TARGET raft_nn_lib)
+  list(APPEND raft_components  nn-lib)
```
vyasr, line 408:
```suggestion
  list(APPEND raft_components nn-lib)
```

cpp/cmake/modules/raft_export.cmake
```diff
@@ -210,6 +210,10 @@ function(raft_export type project_name)
       file(MAKE_DIRECTORY "${scratch_dir}")
       install(DIRECTORY "${scratch_dir}" DESTINATION "${install_location}"
         COMPONENT raft_${comp})
+      if(EXISTS "${scratch_dir}/raft-${comp}-dependencies.cmake")
+        install(FILES "${scratch_dir}/raft-${comp}-dependencies.cmake" DESTINATION "${install_location}"
```
vyasr, line 214: Do we need to do something similar for the `raft-${comp}-targets.cmake` files?

cpp/CMakeLists.txt
robertmaynard: The first one writes the file to the raft build directory, this allow `find_package(raft)` to use the raft build directory.
The second call generates the dependency file that will be installed with RAFT.

We have two separate invocations so that we have finer control over both the build and install directory layouts.

cpp/cmake/modules/raft_export.cmake
robertmaynard: No we don't as we place the `raft-${comp}-targets.cmake` file in a location that is already installed by the `raft` component.  I also we able to remove this install command, as we now also place the `raft-${comp}-dependencies.cmake` file in the correct location.

- raft-*.cmake => installed as part of the `raft` component
- raft-nn-*.cmake => installed as part of the `raft` component
- raft-distance-*.cmake => installed as part of the `raft` component
- raft-nn-lib-*.cmake => installed as part of the `nn` component
- raft-distance-lib-*.cmake => installed as part of the `distance` component

cpp/CMakeLists.txt
```diff
@@ -390,6 +393,44 @@ install(FILES ${CMAKE_CURRENT_BINARY_DIR}/include/raft/version_config.hpp
...
+  list(APPEND raft_install_comp nn)
+endif()
+
+foreach(comp install_comp IN ZIP_LISTS raft_components raft_install_comp)
+  install(
+          EXPORT raft-${comp}-exports
+          FILE raft-${comp}-targets.cmake
+          NAMESPACE raft::
+          DESTINATION "${lib_dir}/cmake/raft"
+          COMPONENT raft${install_comp}
```
vyasr, line 418: With your last change to the `install_comp` names you'll also need to change this line and the second `rapids_export_write_dependencies` on line 429 below.

cpp/CMakeLists.txt
vyasr: Ah I see your other comment below. Now I see why the two were divergent, and how you modified it so that they would work identically.

cpp/CMakeLists.txt
robertmaynard: Updated the install component names to be `nn` and `distance` so that the logic overall is easier to follow

build.sh
```diff
@@ -166,6 +166,7 @@ fi
 if hasArg tests || (( ${NUMARGS} == 0 )); then
     BUILD_TESTS=ON
     ENABLE_NN_DEPENDENCIES=ON
+    COMPILE_NN_LIBRARY=ON
```
vyasr, line 169: Should we also build the DIST library here? Why do tests only require the NN component?

vyasr (approved):
One question, otherwise LGTM.

build.sh
robertmaynard: It is a case where none of the tests need any of the dependencies of the distance component. In the case of NN we need the faiss dependency so requesting `COMPILE_NN_LIBRARY` ensures it will be found and linked.

PR #689, opened 2022-05-26
Fix excessive memory used in selection test

achirkin:
Make sure the gpu vector allocated for the data generation is deallocated before the other vectors are allocated. With this, the test uses under 13 GB of memory.


cjnolet (approved):

cjnolet:
@gpucibot merge

PR #690, opened 2022-05-26
For fixing the cuGraph test failures with PCG

vinaydes:
RAFT's RNG class provides methods that fill a buffer with random numbers belonging to various probability distribution functions. For example`uniform`, `normal` etc. In this methods, multiple instances of random number generator are used in parallel. Each cuda thread gets it own instance of generator. The instance is initialized with three values `seed`, `subsequence` and `offset`. The  value for `seed` is common for all threads and thread id is used for `subsequence`. The `offset`  is kept `0` for all instances. To fill the buffer, the threads work in grid strided fashion  demonstrated by loop below:

```
  for (size_t i = tid; i < buffer_length; i += total_number_of_threads) {
    buffer[i] = rng.next();
  }
```

Due to grid-striding of loops, consecutive elements in the buffer are `ith` element in consecutive subsequences. For example, 10th and 11th elements in the buffer would be 0th element in 10th and 11th subsequences. In the unit tests for cugraph, this scheme seems to introduce a slight bias in certain cases. Easy fix to the issue is to break the lock step increment of individual subsequences. 

This PR sets different offset value for each subsequence by setting `offset = subsequence`. Note that this change has no effect on the period of the random number generator.

This should fix the cuGraph issue https://github.com/rapidsai/cugraph/issues/2266 for now.

### DIFF
--- a/cpp/include/raft/random/detail/rng_device.cuh
+++ b/cpp/include/raft/random/detail/rng_device.cuh
@@ -560,6 +560,7 @@ struct PCGenerator {
     next(discard);
     pcg_state += rng_state.seed;
     next(discard);
+    skipahead(subsequence);
   }
 
   // Based on "Random Number Generation with Arbitrary Strides" F. B. Brown
--- a/cpp/test/linalg/gemm_layout.cu
+++ b/cpp/test/linalg/gemm_layout.cu
@@ -121,7 +121,7 @@ class GemmLayoutTest : public ::testing::TestWithParam<GemmLayoutInputs<T>> {
 };
 
 const std::vector<GemmLayoutInputs<float>> inputsf = {
-  {80, 70, 80, true, true, true, 76433ULL},
+  {80, 70, 80, true, true, true, 76430ULL},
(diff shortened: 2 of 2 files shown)


teju85:
Very interesting find @vinaydes ! Can you also check that this fix doesn't break cuML tests, please?

Also, adding @akifcorduk and @MatthiasKohl to see if this doesn't break anything in cuOpt and cugraph-ops respectively.

teju85 (approved):
Changes LGTM.

vinaydes:
Yes, I'll run cuML tests with this change.

vinaydes:
Created cuml PR here https://github.com/rapidsai/cuml/pull/4760 after making sure it works locally on my test setup.

vinaydes:
Also created cuGraph PR here https://github.com/rapidsai/cugraph/pull/2316

MatthiasKohl:
Interesting, this should not break anything in cugraph-ops. Thanks @vinaydes for the fix!

PR #697, opened 2022-05-31
Change build.sh to find C++ library by default

vyasr:
Discussions with the team indicated a general preference for having build.sh default to finding the C++ library by default (and erroring if one is not found) rather than building a new copy of libcuspatial when building cuspatial Python.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -196,6 +196,11 @@ if [[ ${CMAKE_TARGET} == "" ]]; then
     CMAKE_TARGET="all"
 fi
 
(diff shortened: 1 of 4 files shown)


cjnolet:
> rather than building a new copy of libcuspatial when building cuspatial Python.

Should this say `libraft` and `raft Python`?

vyasr:
> > rather than building a new copy of libcuspatial when building cuspatial Python.
> 
> Should this say `libraft` and `raft Python`?

Nope. Copy-pasting is always correct :)

cjnolet (approved):

cjnolet:
@gpucibot merge

PR #698, opened 2022-05-31
Add masking object to sparse API

cjnolet:
This needs to be used in both cugraph and cugraph-ops and the actual device functions for masking bits and looking them up in unsigned 32-bit integers are already in RAFT so it makes sense to have this object here for now.


ajschmidt8:

PR #706, opened 2022-06-06
Backporting conda compilers

cjnolet:


cjnolet:
So the remaining failure is, to me, an indicator of the main problem we have been facing. The symbol that's failing was introduced into cusparse in cuda toolkit 11.2 and we have an explicit check for the CUDART version when using the symbol, as seen [here](https://github.com/rapidsai/raft/blob/8e1bf946f979aca21674ce4c4fb9cb5872238cba/cpp/include/raft/sparse/detail/cusparse_wrappers.h#L1755). For whatever reason, the cudart version that's being used to compile the tests binary is 11.2+ but and my guess is that this same tests binary ends up being executed on a system with CUDA 11.0.

cc @ajschmidt8

cjnolet:
At least for cuml, it looks like in the gpu build, libcuml (and corresponding greats) as cleaned and built from scratch again during the docs building so the tests will always be built for that cuda version. We have some pre-processor directives in cuml as well which test for cudart versions so I'm wondering if this is why cuml hasn't encountered such a problem.

PR #715, opened 2022-06-16
Introducing raft::mdspan as an alias

divyegala:
- [x] `raft::mdspan` as an alias to `std::experimental::mdspan` to provide seamless integration

- [x] Update to template type checks for `raft::mdspan` with variadics. Ease of use when checking multiple types as `is_mdspan_v<T1, T2, ..., Tn>`

- [x] `raft::device/host_span` are still supported with their own template type checks


cjnolet:
rerun tests

cjnolet (approved):

cjnolet:
@gpucibot merge

PR #723, opened 2022-06-24
[FEA] Implement unravel_index for row-major array.

trivialfis:
Revised version of https://github.com/rapidsai/raft/pull/631 with `thrust::tuple` replaced by `std::tuple` and custom `apply` function replaced by `std::apply`.

Related: https://github.com/rapidsai/raft/issues/629

The implementation is mostly copied from XGBoost  https://github.com/dmlc/xgboost/blob/fdf533f2b9af9c068cddba50839574c6abb58dc3/include/xgboost/linalg.h#L523 .

In the tests, I have used both `__host__ __device__` and `__device__` lambdas to make sure `std::tuple` and `std::apply` are working correctly with CUDA kernel.

### DIFF
--- a/cpp/include/raft/core/mdarray.hpp
+++ b/cpp/include/raft/core/mdarray.hpp
@@ -926,16 +926,17 @@ auto reshape(const array_interface_type& mda, extents<Extents...> new_shape)
  *
  * \param idx    The linear index.
  * \param shape  The shape of the array to use.
- * \param layout Must be `layout_right` (row-major) in current implementation.
+ * \param layout Must be `layout_c_contiguous` (row-major) in current implementation.
  *
  * \return A std::tuple that represents the coordinate.
  */
 template <typename LayoutPolicy, std::size_t... Exts>
 MDSPAN_INLINE_FUNCTION auto unravel_index(size_t idx,
-                                          detail::stdex::extents<Exts...> shape,
-                                          LayoutPolicy const&)
+                                          extents<Exts...> shape,
+                                          LayoutPolicy const& layout)
 {
-  static_assert(std::is_same<LayoutPolicy, layout_c_contiguous>::value,
+  static_assert(std::is_same_v<std::remove_cv_t<std::remove_reference_t<decltype(layout)>>,
+                               layout_c_contiguous>,
                 "Only C layout is supported.");
   if (idx > std::numeric_limits<uint32_t>::max()) {
     return detail::unravel_index_impl<uint64_t, Exts...>(static_cast<uint64_t>(idx), shape);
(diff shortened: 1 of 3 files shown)


trivialfis:
rerun tests

cpp/include/raft/core/mdarray.hpp
```diff
@@ -914,4 +914,34 @@ auto reshape(const array_interface_type& mda, extents<Extents...> new_shape)
...
+ * \param layout Must be `layout_c_contiguous` (row-major) in current implementation.
+ *
+ * \return A std::tuple that represents the coordinate.
+ */
+template <typename LayoutPolicy, std::size_t... Exts>
+MDSPAN_INLINE_FUNCTION auto unravel_index(size_t idx,
+                                          extents<Exts...> shape,
+                                          LayoutPolicy const& layout)
+{
+  static_assert(std::is_same_v<std::remove_cv_t<std::remove_reference_t<decltype(layout)>>,
```
divyegala, line 938: Why `decltype(layout)` when you already have `typename LayoutPolicy`? I imagine you'll not need to remove reference or cv then

cpp/include/raft/core/mdarray.hpp
```diff
@@ -914,4 +914,34 @@ auto reshape(const array_interface_type& mda, extents<Extents...> new_shape)
...
+ * \endcode
+ *
+ * \param idx    The linear index.
+ * \param shape  The shape of the array to use.
+ * \param layout Must be `layout_c_contiguous` (row-major) in current implementation.
+ *
+ * \return A std::tuple that represents the coordinate.
+ */
+template <typename LayoutPolicy, std::size_t... Exts>
+MDSPAN_INLINE_FUNCTION auto unravel_index(size_t idx,
```
divyegala, line 934: `std::size_t idx` as even extents are expressed as that type?

cpp/include/raft/core/mdarray.hpp
trivialfis: Just to find a way to avoid the unused parameter warning while at the same time keep it for doxygen.

cpp/include/raft/core/mdarray.hpp
trivialfis: I haven't tested the updated mdspan and its new extent type yet.

divyegala: Fair. We'll update it when we update mdspan

divyegala: Could this parameter possibly be optional with a default type? For those wishing to use the template param directly?

trivialfis: Done.

divyegala: Am I missing the updates?

trivialfis: I changed the parameter type of `std::size_t idx` into template `Idx idx`

divyegala (approved):

cjnolet (approved):
LGTM. Thanks @trivialfis for the change to use `std::tuple` instead of exposing `thrust`.

Issue #716, opened 2022-06-16
[BUG] Intermittent failures for the KL Divergence GTests


cjnolet:
This has popped up several times now in CI and I've yet to see these locally. The first couple times I saw it, I thought maybe it was just a fluke but now it's happened enough times that it's worth investigating the cause.

cjnolet:
Here's an example of the failure in CI. 

https://gpuci.gpuopenanalytics.com/job/rapidsai/job/gpuci/job/raft/job/prb/job/raft-gpu-test/CUDA=11.0,GPU_LABEL=driver-450,LINUX_VER=centos7,PYTHON=3.8/1556/console

cc @mdoijade

mdoijade:
Thanks for informing, I'm work towards fixing this, from log file seems like naive test kernel has a bug whereas main kernel is generating correct output.

mdoijade:
this issue is fixed, should be good to close this.

PR #724, opened 2022-06-24
fix nans in naive kl divergence kernel introduced by div by 0.

mdoijade:
-- This address issue https://github.com/rapidsai/raft/issues/716
-- remove redundant commented code as well.

### DIFF
--- a/cpp/test/distance/distance_base.cuh
+++ b/cpp/test/distance/distance_base.cuh
@@ -260,9 +260,8 @@ __global__ void naiveKLDivergenceDistanceKernel(
     auto a            = x[xidx];
     auto b            = y[yidx];
     bool b_zero       = (b == 0);
-    const auto m      = (!b_zero) * (a / b);
-    const bool m_zero = (m == 0);
-    acc += (a * (!m_zero) * log(m + m_zero));
+    bool a_zero       = (a == 0);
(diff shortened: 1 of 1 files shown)


cpp/test/distance/distance_base.cuh
```diff
@@ -450,10 +449,6 @@ class DistanceTest : public ::testing::TestWithParam<DistanceInputs<DataType>> {
     }
     naiveDistance(
       dist_ref.data(), x.data(), y.data(), m, n, k, distanceType, isRowMajor, metric_arg, stream);
-    //    size_t worksize = raft::distance::getWorkspaceSize<distanceType, DataType, DataType,
```
cjnolet, line 453: If this is unneeded, can it just be removed?

cjnolet (changes requested):

cpp/test/distance/distance_base.cuh
mdoijade: this is commented code which I've already removed.

cjnolet: Got it! Wow, I totally misread the diff.

cjnolet (approved):

cjnolet:
@gpucibot merge

Issue #497, opened 2022-02-09
[FEA] Padding support for `mdspan` and `mdarray`.


trivialfis:
The integration of `mdspan` introduced in https://github.com/rapidsai/raft/pull/437 allocates the memory based on the function`required_span_size` from layouts.  We can customize the layout classes to change the required size and pad the last dimension of the contiguous array.  This way we can use optimizations like vector load with the new mdspan.

mfoerste4:
I'll start working on this.

mfoerste4:
I added https://github.com/rapidsai/raft/pull/663 as draft for discussion. For now I have chosen *not* to implement a full custom layout as this would require changes to the `submdspan.hpp` implementation which is limited to the three existing layouts (left/right/stride). Instead, i have just provided a set of convenience mappers deriving from `layout_stride::mapper` that assist in configuring the stride configuration to enable padding.
I added a couple of tests to showcase the usage for now. I will add more once we agree upon a design.
@trivialfis, I would be happy to get your feedback on this.

mfoerste4:
I have added a new PR with a full static approach https://github.com/rapidsai/raft/pull/725.

mhoemmen:
Please see also mdspan PR https://github.com/kokkos/mdspan/pull/180 and discussion on PR #725; thanks!

mhoemmen:
To clarify my above comment: `aligned_accessor` (https://github.com/kokkos/mdspan/pull/176) and `layout_padded` (https://github.com/kokkos/mdspan/pull/180), plus a bit of glue code to define an alias for an aligned mdspan, should resolve this issue.

mhoemmen:
mdspan PR https://github.com/kokkos/mdspan/pull/180 (`layout_{left,right}_padded`) is finished.  It reflects a design that works for both padded aligned access, and for BLAS and LAPACK layouts (as in P1673's "BLAS general" layout).

Issue #711, opened 2022-06-15
[ENH] ivf-flat ANN: pad the dataset dim to take advantage of vectorized load/store


achirkin:
We copy the data in ivf_flat while building the index, thus we can pad the data dimensionality `dim` to any vector size. This would improve the search performance compared to the current approach of adapting the vector length `veclen` to the `dim` https://github.com/rapidsai/raft/blob/e6ff267b62bf66bd328ab004a94425cb610cbc89/cpp/include/raft/spatial/knn/detail/ann_ivf_flat.cuh#L196-L198

_Originally posted by @tfeher in https://github.com/rapidsai/raft/pull/652#discussion_r890970098_

Nyrio:
You can mark some items as done:
- (3) #1011 
- (4) Various PRs: #909, #911, #912, #979, #1011

PR #725, opened 2022-06-27
Adding padded layout 'layout_padded_general' 

mfoerste4:
This is a different approach / followup PR of https://github.com/rapidsai/raft/pull/663 for issue https://github.com/rapidsai/raft/issues/497.

I implemented a `layout_padded_general` within raft to statically enforce padding on mdpsan accesses.
* The layout has template parameters for `ValueType`, `StorageOrder `(default `row_major_t`), and `ByteAlignment `(default 128)
* in order to *not* require changes upstream I skipped `submdspan `functionality right now. I have a branch on a mdspan fork where I tested this though (https://github.com/mfoerste4/mdspan/tree/layout_padded).


mfoerste4:
@achirkin, please have a look at this. It is now possible to retrieve the layout / padding width at compile time which makes it possible to target optimized kernel code correctly.

cpp/include/raft/detail/layout_padded_general.hpp
```diff
@@ -0,0 +1,139 @@
...
+  -> std::array<size_t, Extents::rank()>
+{
+  auto alignment                              = std::max<size_t>(ByteAlignment / valueTypeSize, 1);
+  std::array<size_t, Extents::rank()> strides = std::array<size_t, Extents::rank()>{};
+  size_t stride                               = 1;
+  for (size_t r = 0; r + 1 < Extents::rank(); r++) {
+    strides[r] = stride;
+    if (stride == 1) {
+      stride *=
+        std::max<size_t>(alignment, (__exts.extent(r) + alignment - 1) / alignment * alignment);
```
achirkin, line 65: A nitpick: perhaps, we can use `raft::ceildiv` for readability?

cpp/include/raft/matrix/detail/linewise_op.cuh
```diff
@@ -176,6 +186,37 @@ struct Linewise {
...
+                                                      const IdxType blockOffset,
+                                                      const IdxType rowLen,
+                                                      const IdxType rowLenPadded) noexcept
+  {
+    IdxType j = blockOffset + threadIdx.x;
+#pragma unroll VecElems
+    for (int k = threadIdx.x; k < VecElems * BlockSize; k += BlockSize, j += BlockSize) {
+      while (j >= rowLenPadded)
+        j -= rowLenPadded;
+      shm[k] = j < rowLen ? p[j] : Type(1);
```
achirkin, line 211: Out of curiosity: why ones and not zeroes? :)

mfoerste4: I had division ops in mind and did not want to risk any division by zero. I was unsure whether this might cause issues with tools like valgrind?

cpp/include/raft/detail/layout_padded_general.hpp
```diff
@@ -0,0 +1,139 @@
...
+    }
+  }
+  strides[__exts.rank() - 1] = stride;
+  return strides;
+}
+
+// similar to layout_strided, but contiguous with padding in second smallest stride dimension
+template <typename ValueType,
+          StorageOrderType StorageOrder = StorageOrderType::row_major_t,
+          size_t ByteAlignment          = 128>
```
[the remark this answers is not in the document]
achirkin: Sounds good. Maybe you'd also want to constrain it here to always be power-of-two (I've noticed you used Pow2<..> utils somewhere)? That would still cover our hardware-inspired use cases.

cpp/include/raft/matrix/detail/linewise_op.cuh
achirkin: Hmm, good idea, I guess. I don't think it will cost anything anyway.

mfoerste4:
> Looks good! I wonder if we can further simplify/optimize some parts of the `linewiseOp` for padded data?..

The current implementation basically consists of the old main kernel which was running the aligned data portion. I don't think we can simplify it further without loosing performance. Regarding optimizations - we could skip the operation on the padded portion of data, but I guess that would only have very limited effect on larger datasets.

mfoerste4:
@achirkin , what are the next steps here to proceed?

cpp/include/raft/matrix/detail/linewise_op.cuh
```diff
@@ -26,6 +27,15 @@ namespace raft {
 namespace matrix {
 namespace detail {
 
+namespace stdex    = std::experimental;
+using extents_type = stdex::extents<stdex::dynamic_extent, stdex::dynamic_extent>;
+template <typename ValueType>
+using padded_layout_row_major =
```
achirkin, line 33: Perhaps, this, parameterized by the padding size, should go to `raft/core/mdarray.hpp`?..

cpp/include/raft/matrix/detail/linewise_op.cuh
```diff
@@ -26,6 +27,15 @@ namespace raft {
...
+  stdex::layout_padded_general<ValueType, stdex::StorageOrderType::row_major_t, 128>;
+template <typename ElementType>
+using padded_matrix =
```
achirkin, line 36: The same here (so that the api in `matrix.cuh` doesn't show anything in `detail` namespace).

cpp/include/raft/detail/layout_padded_general.hpp
```diff
@@ -0,0 +1,138 @@
...
+struct layout_padded_general {
```
achirkin, line 77: @mhoemmen, sorry to call you out again, could you have a quick look at our attempt to define the padded layout - whether it is properly in the spirit of the mdspan spec?

achirkin (changes requested):
Overall, this looks good to me.

The only thing I'm not sure is whether we'd want to define the padded layout synonyms for public use _with or without the template parameter `padding size`_. On the one hand, the parameterized version seems to be more logical for whatever use case. On the other hand, any application I could imagine, it only matters to have the specific padding of 128 bytes for the coalesced memory access. Or,  maybe, we should have both?

cpp/include/raft/detail/layout_padded_general.hpp
mhoemmen: @achirkin I'm happy to help out! : - )

My main issue with this design is that the layout mapping cannot possibly be responsible for alignment.  The layout can do padding to a particular size, but it can't control whether the pointer is aligned (so that you can do aligned SIMD loads, etc.).  This is because the layout never sees the pointer; it's the accessor that sees the pointer.

Would you instead consider the following design?

* The layout mapping 
    * pads each row (if row-major, or column, if column-major) to a given number of _elements_ (not a number of bytes); and
    * encodes the padding as a `size_type` or `integral_constant<size_type, Value>` template parameter.
* The accessor
    * encodes the alignment as a `size_t` or `integral_constant<size_t, Value>`;
    * defines the `pointer` type with the appropriate attributes to tell the compiler that it is aligned; and
    * has a different `offset_policy` (e.g., `default_accessor`) to express that offset access may not be aligned.

A type alias and possibly also a helper create-an-mdspan function would then use the `value_type` to compute the padding and alignment at compile time.  For example:
```c++
// Specialize this for different value types as needed.
// The value can be size_t to simplify specialization;
// layout_padded_general can convert it to size_type.
template<class ValueType>
struct padding {
    static constexpr size_t value = 128 / sizeof(ValueType);
};

template<class ElementType, class Extents, StorageOrderType order>
using aligned_mdspan = mdspan<ElementType, Extents, layout_padded_general<order, padding_v<remove_cv_t<remove_reference_t<ElementType>>>>, aligned_accessor<ElementType, alignment_v<remove_cv_t<remove_reference_t<ElementType>>>>>

template<class ElementType, class Extents, StorageOrderType order>
aligned_mdspan<ElementType, Extents, order>
make_aligned_mdspan(ElementType* input_pointer, Extents e, StorageOrderType /* order */)
{
    using value_type = remove_cv_t<remove_reference_t<ElementType>>;
    using pointer = typename aligned_accessor<ElementType, alignment_v<value_type>>::pointer;

    // Any run-time check for alignment would happen here.
    // Taking the alignment value as integral_constant permits interesting checks,
    // like wrapping ElementType* in a proxy pointer type
    // that encodes alignment as a non-type template parameter.
    pointer aligned_pointer = 
        assert_aligned(input_pointer, integral_constant<size_t, alignment_v<value_type>>{});
    using mapping = layout_padded_general<order, padding_v<value_type>>::template mapping<Extents>;
    return {aligned_pointer, mapping{e}};
>, 
```

mhoemmen (commented):
Thanks for soliciting my review!  I have a few comments on the design.

cpp/include/raft/core/cudart_utils.hpp
```diff
@@ -43,6 +43,56 @@
...
+{
+  return (a / b) * b;
+}
+
+/**
+ * @brief Check if the input is a power of 2
+ * @tparam IntType data type (checked only for integers)
+ */
+template <typename IntType>
+constexpr bool isPo2(IntType num)
```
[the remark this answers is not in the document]
achirkin: Good idea but NB we cannot remove `isPo2` because it's used in some places in raft and cuml and potentially in other projects. We just move here a few functions from `*.cuh` to `*.hpp` to make them available in the host-only code.

cpp/include/raft/core/mdarray.hpp
```diff
@@ -52,6 +53,54 @@ template <typename ElementType,
...
+  using value_type = std::remove_cv_t<std::remove_reference_t<ElementType>>;
+  using pointer = typename detail::stdex::aligned_accessor<ElementType, alignment::value>::pointer;
+
+  // Any run-time check for alignment would happen here.
+  // Taking the alignment value as integral_constant permits interesting checks,
+  // like wrapping ElementType* in a proxy pointer type
+  // that encodes alignment as a non-type template parameter.
+  assert(input_pointer == alignTo(input_pointer, alignment::value));
+
+  pointer aligned_pointer = input_pointer;  // assert_aligned(input_pointer,
```
mhoemmen, line 96: [INFORMATIVE NOTE]

I've been thinking about how to express this idea.  The issue is that `std::assume_aligned` (C++20) or equivalent compiler built-ins (e.g., GCC's `__builtin_assume_aligned`) don't affect the type.  My current thinking is that `aligned_accessor::access` should do `return (std::assume_aligned(p))[i];`.  Please see my comment below on `aligned_accessor::access`.

cpp/include/raft/detail/layout_padded_general.hpp
```diff
@@ -122,17 +119,44 @@ struct layout_padded_general {
...
+                   element_type (*)[]) &&
+     alignment == OtherAlignment))
+  MDSPAN_INLINE_FUNCTION
+  constexpr aligned_accessor(aligned_accessor<OtherElementType, OtherAlignment>) noexcept {}
+
+  MDSPAN_INLINE_FUNCTION
+  constexpr pointer offset(pointer p, size_t i) const noexcept { return p + i; }
+
+  MDSPAN_FORCE_INLINE_FUNCTION
+  constexpr reference access(pointer p, size_t i) const noexcept { return p[i]; }
```
mhoemmen, line 158: [This is not strictly required for correctness, but I'm guessing you won't get any benefit out of `aligned_accessor` unless you do something here.]

This would be the point at which I make assertions about alignment.  For example:
```c++
constexpr reference access(data_handle_type p, size_t i) const noexcept {
#if defined(__cpp_lib_assume_aligned)
    return (std::assume_aligned<alignment>(p))[i];
#elif defined(__GNUC__)
    return (__builtin_assume_aligned(p, alignment))[i];
#else
    return p[i];
#endif
}
```

cpp/include/raft/detail/layout_padded_general.hpp
```diff
@@ -0,0 +1,138 @@
...
+  class mapping : public layout_stride::mapping<Extents> {
+   public:
+    // This could be a `requires`, but I think it's better and clearer as a `static_assert`.
+    static_assert(detail::__is_extents_v<Extents>,
+                  "std::experimental::layout_padded_general::mapping must be instantiated with a "
+                  "specialization of std::experimental::extents.");
+    // static_assert(Extents::rank() > 1, "std::experimental::layout_padded_general::mapping must be
+    // instantiated with a Extents having rank > 1.");
+
+    using size_type    = typename Extents::size_type;
```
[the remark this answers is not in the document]
mfoerste4: Thanks for the hint - I just merged with the new tip and added the type alias.

cpp/include/raft/detail/layout_padded_general.hpp
mfoerste4: Yes, for now I was just switching to the right accessor class in order to allow for both static and runtime checks later. Thanks for example, I'll work that in.

cpp/include/raft/core/mdarray.hpp
```diff
@@ -74,7 +74,7 @@ using detail::vector_extent;
 template <typename IndexType>
 using extent_1d = vector_extent<IndexType>;
 
-template <typename IndexType>
+template <typename IndexType = std::uint32_t>
```
mhoemmen, line 77: Would you consider getting this default index type from some common source, to avoid code duplication and possible errors between `matrix_extent` and (say) `vector_extent`?

cpp/include/raft/core/mdarray.hpp
```diff
@@ -109,7 +110,8 @@ using padded_layout = detail::stdex::layout_padded_general<
   order>;
 
 template <typename ElementType, storage_order_type order>
-using padded_matrix = mdspan<ElementType, matrix_extent, padded_layout<ElementType, order>>;
+using padded_matrix =
```
mhoemmen, line 113: Given that `uint32_t` is the default, would you consider just relying on the default here, in order to avoid confusion and possible errors?

cpp/include/raft/core/mdarray.hpp
mfoerste4: There were several issues after merging with the introduced IndexType, I did not mean to change that here.

cpp/include/raft/core/mdarray.hpp
mfoerste4: @mhoemmen , thank you very much for your input. I have tried to apply all of your suggestions and also modified the matrix `LinewiseOp `operations to utilize the `aligned_mdspan`.

mfoerste4:
@mhoemmen , thank you very much for your input. I have tried to apply all of your suggestions and also modified the matrix `LinewiseOp `operations to utilize the `aligned_mdspan`.

cpp/include/raft/detail/layout_padded_general.hpp
mhoemmen: @mfoerste4 Some compilers let one express alignment by annotating the pointer; e.g., `double* __attribute__(align_value( alignment ))` -- I haven't experimented with that, but I personally would prefer using the C++ Standard way (`std::assume_aligned`, as above) to the list of `ifdef`s.

cpp/include/raft/detail/layout_padded_general.hpp
```diff
@@ -29,18 +29,16 @@ namespace experimental {
 enum class StorageOrderType { column_major_t, row_major_t };
 
 template <class Extents>
-MDSPAN_INLINE_FUNCTION constexpr auto padded_row_major_strides(size_t ByteAlignment,
-                                                               size_t valueTypeSize,
+MDSPAN_INLINE_FUNCTION constexpr auto padded_row_major_strides(size_t Alignment,
```
achirkin, line 32: Just a neatpick: could you check that raft's clang-tidy does not complain about the naming conventions for the arguments? I believe, the `Alignment` should be lower-case.

(Usually I set up vscode to check this: `clang-tidy -p ${fileWorkspaceFolder}/cpp/build ${file} -extra-arg=-std=c++17`)

achirkin (commented):
Thanks, looks good to me! Just a small suggestion to check if clang-tidy does not complain too much.

BTW, did you have a chance to see how `linewiseOpSpan` compares in terms of speed to `linewiseOp`?

mhoemmen (commented):
I think somehow my review never made it through, yikes!

cpp/include/raft/detail/layout_padded_general.hpp
mfoerste4: I have adjusted the specific parameter you mentioned. Running clang-tidy gives me several more warnings like ` warning: invalid case style for constexpr variable 'storage_order' [readability-identifier-naming]`, but the suggested naming `kStorageOrder ` does seemingly not match our conventions looking at other places in the code. Is my call missing the correct `.clang-tidy` options?

mfoerste4:
> Thanks, looks good to me! Just a small suggestion to check if clang-tidy does not complain too much.
> 
> BTW, did you have a chance to see how `linewiseOpSpan` compares in terms of speed to `linewiseOp`?
I did some benchmarks on a V100. For most matrix shapes the span-implementation is as fast as the existing implementation with aligned data. It suffers significantly if the line-length is comparably small w.r.t. the amount of padding.

cpp/include/raft/detail/layout_padded_general.hpp
achirkin: That's correct `.clang-tidy` and I don't know what's the better option :) I usually choose to blindly follow clang-tidy suggestion, but I wouldn't ask you to do the same

achirkin (approved):
This looks good to me. Looking forward to try this in https://github.com/rapidsai/raft/issues/711 (1)!

mhoemmen:
@mfoerste4 Just to follow up: my mdspan PR https://github.com/kokkos/mdspan/pull/176 adds an `aligned_accessor` example.

cpp/include/raft/core/mdarray.hpp
mhoemmen: For my latest thoughts on `assume_aligned` etc., please see my `aligned_accessor` PR https://github.com/kokkos/mdspan/pull/176 .

mhoemmen:
@mfoerste4 I've been working on a different approach.  The idea is that a padded layout mapping has the same mapping function as a submdspan of a `layout_left` (for column major, or `layout_right` for row-major) mdspan, with suitably padded extents.  This would make `padded_layout::mapping` just a wrapper around `layout_left::mapping` (or `layout_right::mapping`).

This approach has the following advantages.

1. The only nontrivial code is computing the padded extents.  The [`pad_extents` function here here](https://godbolt.org/z/sv7daYGf1) shows how to do that.  Everything else is just a wrapper.  That reduces the testing burden.
2. If the original extents are all known at compile time, the padded extents can also be stored as compile-time values.  This makes storing and passing around padded-layout mdspans cheaper, and can improve compiler optimizations.

mhoemmen:
@mfoerste4 Here is my proposed `layout_padded` with some simple tests: https://godbolt.org/z/odhbzx5eq

mfoerste4:
> @mfoerste4 Here is my proposed `layout_padded` with some simple tests: https://godbolt.org/z/odhbzx5eq
> 
> My extents padding might be more generous than needed -- I think only the first or last extent needs padding, not the middle ones as well -- but that should be easy to fix.

@mhoemmen , I like the idea of compile time 'override' of the extents in your approach. It also feels more natural to base the padded layout on the left/right layout instead of utilizing the more general `strided_layout `with some implicit restrictions regarding the strides. 

Like you already mentioned the padding should only occur on first/last extent w.r.t. the layout. The middle ones don't need extra padding as the strides are implicitly aligned as multiples of the first/last extent respectively.

Regarding next steps - are you planning to merge this idea to the mdspan repo in the near future? Short term I could update the layout class of this PR, but it would be great to have it in the officical repo to avoid future merge issues.

mhoemmen:
@mfoerste4 Thanks for your response and the feedback!

> Like you already mentioned the padding should only occur on first/last extent w.r.t. the layout. 

I'll fix that -- thanks!

> Regarding next steps - are you planning to merge this idea to the mdspan repo in the near future?

Yes -- I'm planning to make a PR in that repo.

Regarding `aligned_accessor`, Christian, Damien, and I have been debating some fine details of my PR https://github.com/kokkos/mdspan/pull/176 .  I think it's close to merging.  (mdspan PRs need approving reviews to merge.)  `layout_padded` doesn't depend on that, but in order to have overaligned mdspan, we will need to combine `aligned_accessor` and `layout_padded` and have examples and tests for their combination.

Btw, I find it a bit awkward that we say "`something_accessor`" but "`layout_something`."  The original convention was to put the specific "something" last: `accessor_basic`.  That got renamed to `default_accessor`, but the renaming didn't touch the layouts; I don't know why.  Thus, I think we should feel free to write "`padded_layout`" if we like, or "`layout_padded`," whichever we prefer.  Which would you prefer?

mhoemmen:
@mfoerste4 Here is my `layout_padded` example's PR: https://github.com/kokkos/mdspan/pull/180

mfoerste4:
> Which would you prefer?
I would prefer `layout_something`, as it is consistent with the other layouts we probably don't want to switch anymore. 

> @mfoerste4 Here is my `layout_padded` example's PR: [kokkos/mdspan#180](https://github.com/kokkos/mdspan/pull/180)
I initially thought you were planning to add the `layer_padded ` as a reusable header part in the [p0009](https://github.com/kokkos/mdspan/tree/stable/include/experimental/__p0009_bits). Having it available as an example is great, but would still require copy/pasting class code for consumers (e.g. raft). Having it generally available would also allow for proper treatment in the `submdspan` going forward. What do you think?

mhoemmen:
@mfoerste4 wrote:
> I initially thought you were planning to add the `layout_padded` as a reusable header part in the [p0009](https://github.com/kokkos/mdspan/tree/stable/include/experimental/__p0009_bits). Having it available as an example is great, but would still require copy/pasting class code for consumers (e.g. raft). Having it generally available would also allow for proper treatment in the submdspan going forward. What do you think?

It's definitely the plan to make `layout_padded` and `aligned_accessor` reusable headers in the reference mdspan implementation.  However, I'm still talking with the other reference mdspan developers about how to do that (e.g., header path, namespace name, testing requirements).  If you have any suggestions, we would welcome them! : - ) 

For now, I just wanted to make the code available in some way, as quickly as possible.  I'll be happy to add `layout_padded` and `aligned_accessor` to RAFT itself for now, if you like.  Once mdspan gets them as reusable headers, we'll be able to remove those implementations.  Do you think that would be a good idea?

mfoerste4:
>  I'll be happy to add layout_padded and aligned_accessor to RAFT itself for now, if you like. Once mdspan gets them as reusable headers, we'll be able to remove those implementations. Do you think that would be a good idea?

I can switch the `layout_padded_general` in this PR to your `layout_padded ` based on your example and modify the test and benchmark code for now. That way we hopefully have not much to change once it is generally available in mdspan. Additionally, we might get some valuable feedback from other people actually using it in the open field.

mfoerste4:
@mhoemmen,  I noticed quite some activity on the layout_padded branch https://github.com/kokkos/mdspan/pull/180. I suppose we can use this implementation for now and move to the mdspan version once it is fully integrated.

mhoemmen:
@mfoerste4 I'm almost done!  I should be finishing up the last few changes today.  It's a different design, based on discussions last week with Christian Trott about what we would like to try to standardize.  It should also be able to replace the "BLAS general" layouts in P1673.

mhoemmen:
@mfoerste4 Please see updated PR here: https://github.com/kokkos/mdspan/pull/180

mfoerste4:
I have updated header versions of `layout_{left_right}_padded` as well as `aligned_accessor` based on the examples in https://github.com/kokkos/mdspan/pull/180 within the thirdparty mdspan include directory (including disclaimer).
A wrapper layout and make utility is located in `cpp/include/raft/core/mdspan.hpp`. 
@cjnolet, could you assign a code-owner for final review?

cpp/include/raft/thirdparty/mdspan/include/experimental/mdspan
```diff
@@ -44,13 +44,15 @@
 #pragma once
 
 #include "__p0009_bits/default_accessor.hpp"
+#include "__p0009_bits/aligned_accessor.hpp"
```
cjnolet, line 47: Is this change already integrated into Kokkos upstream? Just making sure we aren't diverging so when we overwrite these files with upstream changes we aren't finding a bunch of unexpected breaking changes.

cpp/include/raft/core/mdspan.hpp
```diff
@@ -32,6 +32,67 @@ template <typename ElementType,
           typename AccessorPolicy = std::experimental::default_accessor<ElementType>>
 using mdspan = std::experimental::mdspan<ElementType, Extents, LayoutPolicy, AccessorPolicy>;
 
+/**
+ * Some helper templates to handle mdspan with padded layouts / aligned memory
+ */
+enum class StorageOrderType { column_major_t, row_major_t };
+
+namespace detail {
```
cjnolet, line 40: Can we stick this in raft/core/detail/mdspan.hpp just to keep the public API clean?

cpp/test/mdarray.cu
```diff
@@ -436,6 +436,430 @@ TEST(MDArray, FuncArg)
...
+      int alignment_elements = detail::padding<TestElement1, alignment_bytes>::value;
+      auto padded_span       = padded_device_array.view();
+      thrust::for_each_n(
+        rmm::exec_policy(s),
+        thrust::make_counting_iterator(0ul),
+        rows * cols,
+        [rows, cols, padded_span, alignment_elements, p_status] __device__(size_t i) {
+          size_t idx = i / cols;
+          size_t idy = i % cols;
+          if ((&(padded_span(idx, idy)) - &(padded_span(0, idy))) % alignment_elements != 0)
```
cjnolet, line 768: So for a padded mdarray, does this require that we use the accessor directly and can't, for example, just use the underlying data_handle/pointer? Thinking about the case where we have a matrix and want it padded so we call down to something like cublas, which would be expecting a raw pointer.

cjnolet (changes requested):
Thanks for working on this implementation, Malte! I've done a light review after catching myself up on the conversation in the thread. It looks like this PR could be brought up to date with branch-22.12 as well. Looking forward to having this feature in RAFT!

cpp/include/raft/thirdparty/mdspan/include/experimental/mdspan
mhoemmen: @cjnolet I haven't had a chance to look at these changes much.  The reference mdspan implementation (it's not correct to call that "Kokkos"; it just happens to live in that repository but it is a separate project) does not yet have this header file as a reusable component.  `aligned_accessor` does exist there as an example.  However, the project leads are thinking about how to provide "supported extensions" like this as reusable components.

cpp/include/raft/core/mdspan.hpp
mhoemmen: P2642 has moved away from using tag types to express storage order.  It provides separate types `layout_left_padded` and `layout_right_padded`, for consistency with the Standard layouts `layout_left` and `layout_right`.

cpp/test/mdarray.cu
```diff
@@ -436,6 +436,430 @@ TEST(MDArray, FuncArg)
...
+      rows * cols,
+      [d_view, rows, cols] __device__(size_t i) { d_view(i / cols, i % cols) = i; });
+
+    // manually create span with layout
+    {
+      auto data_padded         = padded_device_array.data_handle();
+      using padded_mdspan_type = device_mdspan<float, extents_type, padded_layout_row_major>;
+      auto padded_span         = padded_mdspan_type(data_padded, layout);
+      thrust::for_each_n(rmm::exec_policy(s),
+                         thrust::make_counting_iterator(0ul),
```
mhoemmen, line 559: Does Thrust have a way to do `cartesian_product(iota(0, rows), iota(0, cols))` iteration?  It's painful to have to go back and forth between 1-D and 2-D indices like this.

cpp/test/mdarray.cu
mhoemmen: @cjnolet This might relate to how mdarray's design has changed.  The latest version of the mdarray proposal has a way to get an mdspan viewing the mdarray's elements, with a given accessor.

mdarray always has a container underneath, so you should always be able to use the mdarray's pointer directly.  On the other hand, a much nicer way to access the mdarray's elements on device would be to get an mdspan viewing the elements.

cpp/test/mdarray.cu
mfoerste4: I could not find anything like that. Of course, one could define iterators for each dimension and zip them together, but that would just shift the stride-logic to another piece of manual code.

cpp/include/raft/matrix/matrix.cuh
```diff
@@ -272,6 +272,26 @@ void linewiseOp(m_t* out,
     out, in, lineLen, nLines, alongLines, op, stream, vecs...);
 }
 
+template <typename m_t, typename idx_t = int, typename Lambda, typename... Vecs>
+void linewiseOpSpan(
```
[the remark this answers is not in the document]
mfoerste4: I have aligned the interface as well as the template names with the given host/device pattern.

cpp/include/raft/core/mdspan.hpp
mfoerste4: I have removed the enum and provided two separate layouts.

mfoerste4:
@cjnolet , thanks for reviewing. I re-based the branch to 22.12 and tried to align naming of template classes and API as close to the existing pattern as possible.

cpp/test/mdarray.cu
```diff
@@ -436,6 +436,430 @@ TEST(MDArray, FuncArg)
...
+      !std::is_nothrow_default_constructible<padded_mdarray_type>::value);  // cuda stream
+    static_assert(std::is_nothrow_move_constructible<padded_mdarray_type>::value);
+    static_assert(std::is_nothrow_move_assignable<padded_mdarray_type>::value);
+  }
+}
+
+TEST(MDArray, Padding) { test_mdarray_padding(); }
+
+// Test deactivated as submdspan support requires upstream changes
+/*void test_submdspan_padding()
```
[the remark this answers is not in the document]
mfoerste4: I added https://github.com/rapidsai/raft/issues/922 in order to track.

cjnolet (dismissed):

mhoemmen:
@cjnolet @mfoerste4 I've been completely overwhelmed with my current project and haven't had time to review this.  Please don't feel like you have to wait on me, though I appreciate that you asked me for feedback!

cjnolet (approved):

cjnolet:
@gpucibot merge

PR #729, opened 2022-06-30
[REVIEW] [ENH] Ability to use ccache to speedup local builds

teju85:
I noticed that we have support for `sccache` in `build.sh` for building RAFT, but no `ccache` (especially if one wants to do local builds). This PR adds support for the latter. Also includes changes to install `ccache` in raft dev conda env.

Let me know in case you have any concerns with adding this.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -18,7 +18,7 @@ ARGS=$*
 # script, and that this script resides in the repo dir!
 REPODIR=$(cd $(dirname $0); pwd)
 
-VALIDARGS="clean libraft pyraft pylibraft docs tests bench clean -v -g --install --compile-libs --compile-nn --compile-dist --allgpuarch --sccache --no-nvtx --show_depr_warn -h --buildfaiss --minimal-deps"
+VALIDARGS="clean libraft pyraft pylibraft docs tests bench clean -v -g --install --compile-libs --compile-nn --compile-dist --allgpuarch --sccache --ccache --no-nvtx --show_depr_warn -h --buildfaiss --minimal-deps"
 HELP="$0 [<target> ...] [<flag> ...] [--cmake-args=\"<args>\"]
  where <target> is:
    clean            - remove all existing build artifacts and configuration (start over)
@@ -33,7 +33,8 @@ HELP="$0 [<target> ...] [<flag> ...] [--cmake-args=\"<args>\"]
  and <flag> is:
    -v                          - verbose build mode
    -g                          - build for debug
-   --sccache                    - enable sccache
+   --sccache                   - enable sccache
+   --ccache                    - enable ccache
    --compile-libs              - compile shared libraries for all components
    --compile-nn                - compile shared library for nn component
    --compile-dist              - compile shared library for distance component
@@ -70,6 +71,7 @@ ENABLE_NN_DEPENDENCIES=OFF
 ENABLE_thrust_DEPENDENCY=ON
 
 SCCACHE_ARGS=""
+CCACHE_ARGS=""
 NVTX=ON
 CLEAN=0
 UNINSTALL=0
@@ -134,8 +136,17 @@ if hasArg --install; then
     INSTALL_TARGET="install"
 fi
 
+if hasArg --sccache && hasArg --ccache; then
+    echo "Cannot enable both sccache and ccache at the same time!"
+    exit 1
+fi
+
 if hasArg --sccache; then
   SCCACHE_ARGS="-DCMAKE_CUDA_COMPILER_LAUNCHER=sccache -DCMAKE_C_COMPILER_LAUNCHER=sccache -DCMAKE_CXX_COMPILER_LAUNCHER=sccache"
+fi
+
+if hasArg --ccache; then
+    CCACHE_ARGS="-DCMAKE_CUDA_COMPILER_LAUNCHER=ccache -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache"
 fi
 
 if hasArg --minimal-deps; then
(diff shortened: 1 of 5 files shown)


achirkin:
NB: I meant to follow up on PR https://github.com/rapidsai/raft/pull/651 , but kept postponing it :)

achirkin:
Would it make sense to enable ccache by default if no arguments are passed to `build.sh`?

teju85:
@achirkin I agree with Robert's comment in your PR above. I think this should be an option that should be enabled by the users when they want it. So, that's why this PR makes changes to the `build.sh` only.

achirkin:
Sure, I totally agree. I've just put it here to link the discussion and to not forget to close that PR.

conda/environments/raft_dev_cuda11.5.yml
```diff
@@ -23,6 +23,7 @@ dependencies:
 - doxygen>=1.8.20
 - libfaiss>=1.7.0
 - faiss-proc=*=cuda
+- ccache
```
achirkin, line 26: Shall we do the same changes in cuml? I personally always use `cuml_dev` environment for raft development, and I suspect that's not me only.

build.sh
```diff
@@ -18,7 +18,7 @@ ARGS=$*
 # script, and that this script resides in the repo dir!
 REPODIR=$(cd $(dirname $0); pwd)
 
-VALIDARGS="clean libraft pyraft pylibraft docs tests bench clean -v -g --install --compile-libs --compile-nn --compile-dist --allgpuarch --sccache --no-nvtx --show_depr_warn -h --buildfaiss --minimal-deps"
+VALIDARGS="clean libraft pyraft pylibraft docs tests bench clean -v -g --install --compile-libs --compile-nn --compile-dist --allgpuarch --sccache --ccache --no-nvtx --show_depr_warn -h --buildfaiss --minimal-deps"
```
robertmaynard, line 21: In general I have recommended we switch over to a command line pattern like `--cache-tool=<program>` the reason for this is the following:

- It allows a user to specify either sccache or ccache.
- Since they provide a name or path we don't require the command to be on the PATH
- It removes the need to check if both options are enabled
- It unifies the `hasArg --ccache` and `hasArg --sccache` conditionals into a single branch
- It allows people to also use cache tools like `distcc`

teju85: I like this approach! Thanks Robert. Let me update the PR.

conda/environments/raft_dev_cuda11.5.yml
teju85: Agree with you. If this PR gets accepted by ops team, we could do the same in cuML too.

build.sh
teju85: @robertmaynard can I get a re-review please?

robertmaynard (approved):
Looks amazing! Thank you

build.sh
```diff
@@ -18,8 +18,8 @@ ARGS=$*
 # script, and that this script resides in the repo dir!
 REPODIR=$(cd $(dirname $0); pwd)
 
-VALIDARGS="clean libraft pyraft pylibraft docs tests bench clean -v -g --install --compile-libs --compile-nn --compile-dist --allgpuarch --sccache --no-nvtx --show_depr_warn -h --buildfaiss --minimal-deps"
-HELP="$0 [<target> ...] [<flag> ...] [--cmake-args=\"<args>\"]
+VALIDARGS="clean libraft pyraft pylibraft docs tests bench clean -v -g --install --compile-libs --compile-nn --compile-dist --allgpuarch --no-nvtx --show_depr_warn -h --buildfaiss --minimal-deps"
+HELP="$0 [<target> ...] [<flag> ...] [--cmake-args=\"<args>\"] [--ccache-tool=<tool>]
```
ajschmidt8, line 22:
```suggestion
HELP="$0 [<target> ...] [<flag> ...] [--cmake-args=\"<args>\"] [--cache-tool=<tool>]
```

This is a minor nitpick, but does it make more sense to call this argument `--cache-tool` instead of `--ccache-tool`? If so, it will need to be updated throughout this file.

ajschmidt8 (changes requested):
The rest of the `--ccache-tool` instances in `build.sh` still need to be updated as mentioned in my comment below.


- https://github.com/rapidsai/raft/pull/729#discussion_r914965878

teju85:
ha... I missed your suggestion. Thanks AJ for catching this. The latest commit above should address this now.

ajschmidt8 (approved):

teju85:
@gpucibot merge

ahendriksen:
In the nightly development docker image, the `ccache` integration does not yet work 100% of time time. I have been using 
```
./build.sh  tests bench --cache-tool=ccache
```
today. At the end of the day, compilation still seemed not to hit the cache. These are the statistics that ccache reports: (note the `unsupported source language` statistic)
```
(rapids) root@492c2feb1be4:/workspace/raft/cpp/build# ccache -s
cache directory                     /root/.ccache
primary config                      /root/.ccache/ccache.conf
secondary config      (readonly)    /etc/ccache.conf
stats updated                       Mon Jul 18 15:54:10 2022
cache hit (direct)                    10
cache hit (preprocessed)               2
cache miss                            27
cache hit rate                     30.77 %
unsupported source language          533   <---------------------------
cleanups performed                     0
files in cache                        83
cache size                           5.9 MB
max cache size                       5.0 GB
```

Does anybody know how to resolve this issue? The only thing I have been able to find is this ccache issue https://github.com/ccache/ccache/issues/772.

robertmaynard:
@ahendriksen I expect that the version of ccache you are using doesn't fully support the CUDA language.

In Ubuntu 20.04 the ccache version is 3.7.7 and numerous issues with the CUDA language was backported to the 3 series in 3.7.9. I know locally ( and rapids-compose ) both use ccache 4.5+ for full CUDA support.

ahendriksen:
@robertmaynard Thank you! That solved the issue completely. Cache is now hit 100% of the time and build time is down to 9 seconds.

PR #734, opened 2022-07-07
fix logic bug in include_checker.py utility

grlee77:
Hi, I noticed an issue in the include style checker when I was trying to adapt it for use in another project.

As far as I can tell, the current logic results in no source files actually being checked! Fortunately, the script also passes after the proposed change, so the includes were already in good shape :)

I verified the fix here by changing a <...>-style import to a double-quote one in one of the source files and it was detected.

I don't know if this file originates in RAFT or if it is also in other RAPIDS projects as well? If the later, the fix will need to be propagated to other repositories.

### DIFF
--- a/cpp/scripts/include_checker.py
+++ b/cpp/scripts/include_checker.py
@@ -44,7 +44,7 @@ def list_all_source_file(file_regex, srcdirs):
     for srcdir in srcdirs:
         for root, dirs, files in os.walk(srcdir):
             for f in files:
-                if re.search(file_regex, f):
+                if not re.search(exclusion_regex, f) and re.search(file_regex, f):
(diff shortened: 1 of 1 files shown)


grlee77:
> the includes were already in good shape :)

Looks like I spoke to soon!  The CI found several issues in files under cpp/include. I think I had only run it on cpp/src when testing locally.

grlee77:
Okay, style check is passing now. (The `f` variable is only the base filename, so it is the containing folder name that the `thirdparty`exclusion check should have been applied to)

cjnolet (approved):

cjnolet:
@gpucibot merge

PR #738, opened 2022-07-11
[FEA] [REVIEW] RMAT rectangular graph generator

teju85:
This PR introduces the core RMAT generator primitive into RAFT. It supports the following features:
1. Can specify different `[a, b; c, d]` distribution parameters at each depth.
2. Can generate bipartite graphs (when the depth across rows and columns are different).
3. As a special case, can support the "traditional" RMAT case too.

Cc += @cjnolet, JFYI.
Also Cc += @PiotrBigajNV, JFYI.

_Note to the reviewers_: I know we already have an RMAT generator in cuGraph then why did we have to "duplicate" it in RAFT!? The reason here is that RAFT needs to generate such sparse graphs/adjacency matrices for benchmarking and unit-testing its sparse primitives. Similarly cugraph-ops will also need to do the same in its benchmarks/tests. We cannot introduce cuGraph as a dependency for these things. Hence, this PR. Also, the primitive here is very basic. It _only_ generates the directed edgelist with possible duplicate edges. So, the caller still needs to do this post-processing if/when required. This means, if we are still worried about code duplication, [this](https://github.com/rapidsai/cugraph/blob/branch-22.06/cpp/src/generators/generate_rmat_edgelist.cu#L34) core logic of cuGraph can be replaced by a call to the proposed API in this PR. I also know that the missing thing here is the `clip_and_flip` feature of cuGraph. It will be added soon.


teju85:
@cjnolet did you mean to assign this to someone else? (I'd require a review from somebody as I'm the author of this PR!)

cjnolet:
@teju85 I noticed awhile back that cugraph developers have been assigning the authors of the PRs to the PRs themselves as a way to quickly see the faces of the authors in the PRs. I've been doing that as well because I also find it makes it easier to browse the PR list. We don't have to do it in this case. I started looking at this PR yesterday and I'll try finishing up a review today.

teju85:
Ah I see. Makes sense. Didn't know that. Thanks for letting me know.

JFYI, there's an odd failure of the unit-tests in centos. I still need to debug what's going on there.

cjnolet (approved):

cjnolet:
@gpucibot merge

PR #739, opened 2022-07-12
[FEA] Implement matrix transpose with mdspan.

trivialfis:
* Implement a transpose function that works on both column and row major matrix.

### DIFF
--- a/cpp/include/raft/linalg/detail/transpose.cuh
+++ b/cpp/include/raft/linalg/detail/transpose.cuh
@@ -78,7 +78,6 @@ void transpose(math_t* inout, int n, cudaStream_t stream)
     }
   });
 }
-
 };  // end namespace detail
 };  // end namespace linalg
 };  // end namespace raft
--- a/cpp/include/raft/linalg/transpose.cuh
+++ b/cpp/include/raft/linalg/transpose.cuh
@@ -63,18 +63,18 @@ void transpose(math_t* inout, int n, cudaStream_t stream)
  * @param out Output matirx, storage is pre-allocated by caller and should be contiguous.
  */
 template <typename T, typename LayoutPolicy>
-std::enable_if_t<std::is_floating_point_v<T> &&
-                   (std::is_same_v<LayoutPolicy, layout_c_contiguous> ||
-                    std::is_same_v<LayoutPolicy, layout_f_contiguous>),
-                 void>
-transpose(handle_t const& handle,
-          device_matrix_view<T, LayoutPolicy> in,
-          device_matrix_view<T, LayoutPolicy> out)
+auto transpose(handle_t const& handle,
+               device_matrix_view<T, LayoutPolicy> in,
+               device_matrix_view<T, LayoutPolicy> out)
+  -> std::enable_if_t<std::is_floating_point_v<T> &&
+                        (std::is_same_v<LayoutPolicy, layout_c_contiguous> ||
+                         std::is_same_v<LayoutPolicy, layout_f_contiguous>),
+                      void>
 {
-  ASSERT(out.extent(0) == in.extent(1), "Invalid shape for transpose.");
-  ASSERT(out.extent(1) == in.extent(0), "Invalid shape for transpose.");
-  ASSERT(in.is_contiguous(), "Invalid format for transpose input.");
-  ASSERT(out.is_contiguous(), "Invalid format for transpose output.");
+  RAFT_EXPECTS(out.extent(0) == in.extent(1), "Invalid shape for transpose.");
+  RAFT_EXPECTS(out.extent(1) == in.extent(0), "Invalid shape for transpose.");
+  RAFT_EXPECTS(in.is_contiguous(), "Invalid format for transpose input.");
+  RAFT_EXPECTS(out.is_contiguous(), "Invalid format for transpose output.");
 
   size_t out_n_rows = in.extent(1);
   size_t out_n_cols = in.extent(0);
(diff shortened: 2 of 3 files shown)


cpp/include/raft/linalg/transpose.cuh
```diff
@@ -55,7 +56,95 @@ void transpose(math_t* inout, int n, cudaStream_t stream)
   detail::transpose(inout, n, stream);
 }
 
+/**
+ * @brief Transpose a contiguous matrix. The output has same layout policy as the input.
```
trivialfis, line 60: It doesn't have to be contiguous since we can handle submatrix. Maybe we can relax the constraint by allowing certain types of strides? Suggestions are welcomed.

divyegala (commented):
Looks perfect to me. I just have a question to @cjnolet as we had discussed accepting the standard mdspan without the `host/device` wrappers to our APIs. Is that still the case?

cjnolet:
> Looks perfect to me. I just have a question to @cjnolet as we had discussed accepting the standard mdspan without the host/device wrappers to our APIs. Is that still the case?

That's correct @divyegala. In preparation for our being able to support dispatch to multiple execution environments (e.g. CPU/GPU) in the future, we are coding our public APIs directly to `mdspan`/`mdarray`, rather than specializations.

divyegala:
@trivialfis with Corey's confirmation on this, can we use the vanilla mdspan API? Also we should leave out the overload that returns our implementation of `mdarray` for now, and add it when we adopt the proposal's `mdarray` implementation

trivialfis:
Thank you for the review, I hid the functions into tests and changed the input parameters to vanilla mdspan.

cpp/include/raft/linalg/transpose.cuh
```diff
@@ -55,7 +56,44 @@ void transpose(math_t* inout, int n, cudaStream_t stream)
...
+ * @tparam AccessorPolicy Accessor for the input and output, must be valid accessor on
+ *                        device.
+ *
+ * @param[in]  handle raft handle for managing expensive cuda resources.
+ * @param[in]  in     Input matrix.
+ * @param[out] out    Output matirx, storage is pre-allocated by caller.
+ */
+template <typename T, typename LayoutPolicy, typename AccessorPolicy>
+auto transpose(handle_t const& handle,
+               mdspan<T, ::raft::detail::matrix_extent, LayoutPolicy, AccessorPolicy> in,
```
divyegala, line 74: These have been brought into the `raft` namespace https://github.com/rapidsai/raft/blob/af4f35cc0608730f090122dc793fd8cc8cccc9da/cpp/include/raft/core/mdarray.hpp#L64

divyegala: I'd also like to see `mdspan` being used as `raft::mdspan`

cpp/include/raft/linalg/detail/transpose.cuh
```diff
@@ -79,6 +80,55 @@ void transpose(math_t* inout, int n, cudaStream_t stream)
   });
 }
 
+template <typename T, typename LayoutPolicy>
+void transpose_row_major_impl(handle_t const& handle,
+                              device_matrix_view<T, LayoutPolicy> in,
```
divyegala, line 85: These should also be vanilla mdspan? I am surprised we are able to construct a `device_matrix_view` with mdspan

trivialfis: I'm not sure if this should be Henry span. It's a safe guard that prevents users from passing the wrong pointer.

cpp/include/raft/linalg/transpose.cuh
trivialfis: Fixed, thank you for the suggestions.

cpp/include/raft/linalg/detail/transpose.cuh
trivialfis: It wouldn't be a breaking change when we support host mdspan since we will be just relaxing the constraint.

cjnolet: I think we might be okay on the `impl` side. Because this isn't directly exposed to users, we have the freedom to change this as needed.

divyegala: `device_matrix_view` doesn't do any check on the pointer's underlying memory though, right? We can construct it even with a host mdspan

cjnolet: Yeah I agree we should probably do a check there and raise exception for now if it's a host mdspan.

divyegala: I think that's why we should just accept vanilla mdspan all the way through

trivialfis: It's a compile time guard. Given a host mdspan it won't compile.

divyegala: @trivialfis can we accept vanilla mdspan on our public API, and either use pointers or vanilla mdspan all the way after that? In the future, we will discuss the design of an "execution environment" that will allow us to handle memory guards better (compile time/run time). Until then, it will be easier to use the standard and then refactor if needed

trivialfis: Got it, will make the changes.  Could you please loop me in on the idea of "execution environment"? I think that might benefit the development for XGBoost.

trivialfis: Changed to using vanilla mdspan.

trivialfis:
rerun tests

cjnolet (approved):

cjnolet:
@gpucibot merge

PR #746, opened 2022-07-18
Replace csr_adj_graph functions with faster equivalent

ahendriksen:
The `csr_adj_graph` functions are a performance bottleneck in the DBSCAN implementation in cuML. They are not used anywhere  else.

This PR replaces the `csr_adj_graph` functions with the faster `dense_bool_to_unsorted_csr` function. It has the same functionality, *but*

1. It requires the input adjacency matrix to be in row-major order (rather than column-major).
2. The output column indices are not guaranteed to be in ascending order (hence unsorted).


tfeher (changes requested):
Hi Allard, thanks for the PR. In principle it looks good, but what if we go a step further?

Currently, the function that you are adding needs to receive the `row_ind` values. It would be more self contained if it only takes the boolean matrix, and calculates `row_ind` internally, and returns it together with the `row_ind_ptr`

Earlier these steps were separated, because the vertex degree was calculated in the epsilon neighborhood kernel. But I see that in https://github.com/rapidsai/cuml/pull/4803 you have added an explicit vertex degree calculation. If that is necessary, then it would make sense to move all the steps of building the csr matrix here (vertex degree, scan, adj_to_csr). What is your opinion?

cjnolet:
rerun tests

ahendriksen:
@cjnolet Is this PR good to go now? If it's merged into 22.08 that would make finishing up the cuML follow up easier (https://github.com/rapidsai/cuml/pull/4803).

cjnolet:
@gpucibot merge

PR #747, opened 2022-07-21
ivf_flat::index: hide implementation details

achirkin:
Hide the mutable `mdarray` members of `ivf_flat::index` behind immutable `mdspan` views.


achirkin:
NB: This is a follow-up to #652 and addresses https://github.com/rapidsai/raft/pull/652#discussion_r915035439

cjnolet:
rerun tests

cpp/include/raft/spatial/knn/detail/ann_utils.cuh
```diff
@@ -0,0 +1,560 @@
...
+  uint64_t gid = threadIdx.x + blockDim.x * blockIdx.x;
+  uint64_t i   = gid / n_cols;
+  if (i >= n_rows) return;
+  float& x = a[gid];
+  x        = map(x, d[i]);
+}
+
+/**
+ * @brief Divide matrix values along rows by an integer value, skipping rows if the corresponding
+ * divisor is zero.
```
tfeher, line 372:
```suggestion
 * @brief Change matrix values using elementwise operation. 
 * 
 *  On exit a = op(a, d) where vector d is broadcast along each row.
```

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
```diff
@@ -0,0 +1,706 @@
...
+  minibatch_size = uint32_t(std::min<size_t>(minibatch_size, n_rows));
+  return minibatch_size;
+}
+
+/**
+ * @brief Given the data and labels, calculate cluster centers and sizes in one sweep.
+ *
+ * Let S_i = {x_k | x_k \in dataset & labels[k] == i} be the vectors in the dataset with label i.
+ *   On exit centers_i = normalize(\sum_{x \in S_i} x), where `normalize` depends on the distance
+ * type.
```
tfeher, line 136: Shall be updated since normalization is moved out, and it would help if the weighted average is explained:

```suggestion
 * Let S_i = {x_k | x_k \in dataset & labels[k] == i} be the vectors in the dataset with label i.
 *   On exit centers_i = (\sum_{x \in S_i} x + w_i * center_i) / (|S_i| + w_i), where  w_i = reset_counters ?  0 : cluster_size[i]. In other words, the updated cluster centers are a weighted average of the existing cluster center, and the coordinates of the points labeled with i.
```

tfeher (changes requested):
Thanks @achirkin for these updates, I have a few small comments, otherwise in general it looks good to me.

cpp/include/raft/spatial/knn/ivf_flat_types.hpp
```diff
@@ -0,0 +1,212 @@
...
+   *     x[16, 4], x[16, 5], x[17, 4], x[17, 5], ... x[30, 4], x[30, 5],    -    ,    -    ,
+   *
+   */
+  [[nodiscard]] inline auto data() const noexcept -> device_mdspan<const T, extent_2d, row_major>
+  {
+    return data_.view();
+  }
+
+  /** Inverted list indices: ids of items in the source data [size] */
+  [[nodiscard]] inline auto indices() const noexcept
```
cjnolet, line 83: This looks really nice and clean and it designates the `ivf_flat::index` as the owning object without having to deal w/ the complexities and maintenance of transferring that ownership. I like this a lot.

achirkin:
This PR is ready for review, but should only be merged after https://github.com/rapidsai/raft/pull/764 due to a bug in mdarrays

tfeher (approved):
Thanks Artem for addressing the issues! The PR looks good to me.

cjnolet (approved):

cjnolet:
rerun tests

cjnolet:
@achirkin I notice this PR is breaking but from the looks of the changes, it's only breaking for the new IVF-flat index API, right? If that's the case, are there any active consumers of this API that you know of which would break if we merged this?

achirkin:
Yes, I marked it breaking because of the changes in the bew ivf-flat api, and I am not aware of anybody using it yet.

cjnolet:
@gpucibot merge

PR #749, opened 2022-07-22
Update rapids-cmake version for pyraft in update-version.sh

vyasr:
Previously only pylibraft's copy was being updated, leading to an out-of-date branch of rapids-cmake being pulled for pyraft.

### DIFF
--- a/python/pylibraft/CMakeLists.txt
+++ b/python/pylibraft/CMakeLists.txt
@@ -16,7 +16,7 @@ cmake_minimum_required(VERSION 3.20.1 FATAL_ERROR)
 
 set(pylibraft_version 22.08.00)
 
-include(../fetch_rapids.cmake)
+include(../../fetch_rapids.cmake)
 
 project(
   pylibraft
--- a/python/pylibraft/CMakeLists.txt
+++ b/python/pylibraft/CMakeLists.txt
@@ -16,9 +16,7 @@ cmake_minimum_required(VERSION 3.20.1 FATAL_ERROR)
 
 set(pylibraft_version 22.08.00)
 
-file(DOWNLOAD https://raw.githubusercontent.com/rapidsai/rapids-cmake/branch-22.08/RAPIDS.cmake
-        ${CMAKE_BINARY_DIR}/RAPIDS.cmake)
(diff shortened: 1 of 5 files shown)


sevagh (approved):

robertmaynard (changes requested):
As we ae now up to 3 places using rapids-cmake we should switch over to the approach cudf is using where we have a single `fetch_rapids` that each project includes. 

https://github.com/rapidsai/cudf/blob/branch-22.08/fetch_rapids.cmake

python/pylibraft/CMakeLists.txt
```diff
@@ -16,9 +16,7 @@ cmake_minimum_required(VERSION 3.20.1 FATAL_ERROR)
 
 set(pylibraft_version 22.08.00)
 
-file(DOWNLOAD https://raw.githubusercontent.com/rapidsai/rapids-cmake/branch-22.08/RAPIDS.cmake
-        ${CMAKE_BINARY_DIR}/RAPIDS.cmake)
-include(${CMAKE_BINARY_DIR}/RAPIDS.cmake)
+include(../fetch_rapids.cmake)
```
robertmaynard, line 19: up two directories for `fetch_rapids.cmake`

python/raft/CMakeLists.txt
```diff
@@ -16,9 +16,7 @@ cmake_minimum_required(VERSION 3.20.1 FATAL_ERROR)
 
 set(pyraft_version 22.08.00)
 
-file(DOWNLOAD https://raw.githubusercontent.com/rapidsai/rapids-cmake/branch-22.06/RAPIDS.cmake
-        ${CMAKE_BINARY_DIR}/RAPIDS.cmake)
-include(${CMAKE_BINARY_DIR}/RAPIDS.cmake)
+include(../fetch_rapids.cmake)
```
robertmaynard, line 19: up two directories for `fetch_rapids.cmake`

robertmaynard (changes requested):

vyasr:
Whoops sorry about that.

robertmaynard (approved):

cjnolet:
@gpucibot merge

PR #750, opened 2022-07-22
Add inline to nccl_sync_stream()

seunghwak:
Including `cpp/include/raft/comms/detail/util.hpp` more than once in separate files end up a link error:  "more than one definition ... ".

Add inline to fix this.


ChuckHastings (approved):

cjnolet (approved):

cjnolet:
@gpucibot merge

PR #751, opened 2022-07-22
Update `mdspan` to account for changes to `extents`

divyegala:
`extents` now allows for an `IndexType` to index dimensions. This PR reflects those updates and tags to https://github.com/kokkos/mdspan/commit/3309ce9059e04b42f8d5223f423a78854e3d0b58

The update also follows three API changes:

1. `raft::mdarray/mdspan::data()` is renamed to `raft::mdarray/mdspan::data_handle()`
2. `raft::mdarray/mdspan::is_contiguous()` is renamed to `raft::mdarray/mdspan::is_exhaustive()`
3. `raft::mdarray/mdspan::is_always_contiguous()` is renamed to `raft::mdarray/mdspan::is_always_exhaustive()`


cpp/include/raft/core/mdarray.hpp
```diff
@@ -437,11 +450,11 @@ class mdarray
   }
 
   // basic_mdarray observers of the domain multidimensional index space (also in basic_mdspan)
-  [[nodiscard]] MDSPAN_INLINE_FUNCTION static constexpr auto rank() noexcept -> index_type
+  [[nodiscard]] MDSPAN_INLINE_FUNCTION static constexpr auto rank() noexcept -> std::size_t
   {
     return extents_type::rank();
   }
-  [[nodiscard]] MDSPAN_INLINE_FUNCTION static constexpr auto rank_dynamic() noexcept -> index_type
+  [[nodiscard]] MDSPAN_INLINE_FUNCTION static constexpr auto rank_dynamic() noexcept -> std::size_t
```
cjnolet, line 457: Is the mdarray itself not able to return `index_type`?

cjnolet (approved):

cpp/include/raft/distance/distance.cuh
```diff
@@ -415,9 +415,9 @@ void pairwise_distance(const raft::handle_t& handle,
  */
 template <typename Type, typename Index_ = int, typename layout = layout_c_contiguous>
 void pairwise_distance(raft::handle_t const& handle,
-                       device_matrix_view<Type, layout> const x,
-                       device_matrix_view<Type, layout> const y,
-                       device_matrix_view<Type, layout> dist,
+                       device_matrix_view<Type, Index_, layout> const x,
```
cjnolet, line 418: If we add the `Index_` as a third type, we can default it to `std::size_t` and then we won't break any APIs downstream.

divyegala: Will do, although it will be non-breaking if default is `int` as that is what it was before

cpp/include/raft/core/mdarray.hpp
divyegala: Done

cjnolet:
@gpucibot merge

PR #758, opened 2022-07-27
Fix for KMeans raw pointers API

lowener:


cjnolet:
@lowener was this breaking in cuml but not in RAFT?

cjnolet (approved):

cjnolet:
@gpucibot merge

PR #760, opened 2022-07-27
Fea 2208 kmeans use specializations

cjnolet:


cjnolet:
@Nyrio just FYI, I did realize after some investigation that this doesn't fix the horrid build time of k-means because the standard `distance` instantiations seem to be getting ignored- telling me k-means must be using a template combination that we're not instantiating. I haven't investigated exactly which ones it's using yet, though.

Nyrio:
@cjnolet I've compiled the kmeans test with your changes and then run `nm` and this is what I found:

- 28/32 of the `raft:distance::detail::distance` are marked as "U" i.e undefined in this file (correctly identified as extern)
- The 4 remaining correspond to a metric that hasn't been added to the specializations, as I pointed in a comment on my PR
- 20 instances of `fusedL2NNkernel` are recompiled every time.

I think your PR is working. Horrid compilation times are due to `fusedL2NN`, which is why I made a specialization for it in #795

cjnolet:
Perfect and thanks for investigating this! If you'd like, we can get this PR merged ASAP and then add the specializations for the missing metrics in your PR. Or I can add them here, whichever you prefer.

Nyrio:
I think we can merge this one, and I'll add the rest to my PR.

divyegala (approved):

cjnolet:
@gpucibot merge

PR #761, opened 2022-07-27
Adding old kmeans impl back in (as kmeans_deprecated)

cjnolet:
The new k-means implementation increased the build time significantly, which has a direct impact on our users who are also having to build against it. For now, I'm adding the old kmeans back in (albeit deprecating it) so that we unblock our users.


divyegala (approved):

cjnolet:
@gpucibot merge

Issue #740, opened 2022-07-13
[BUG] Memory leak when mdarray is used in a file that is used by cython/python interface


rg20:
**Describe the bug**
When using mdarray in C++ code alone, it works as expected. However, when using the python interface, we get memory issues such as: 

- Caught signal 11 (Segmentation fault: address not mapped to object at address (nil))
- munmap_chunk(): invalid pointer, Aborted (core dumped)
- corrupted double-linked list. Aborted (core dumped)

**Steps/Code to reproduce bug**
A PR in cuOpt that reproduces the bug is here: https://github.com/rapidsai/cuopt/pull/605. In this PR, mdarray is declared as a member variable in a class and not at all used (null op) 

**Expected behavior**
The C++ version is working just fine.  The `mdarray` (in this case) is not part of the cython or python class, so it should work fine. 

**Environment details (please complete the following information):**
 cuOpt needs to be installed to reproduce the bug. There may be other projects (like cuGraph) that can be used to reproduce the bug. 
 
**Additional context**

afender:
We are getting close to release and impacted authors will go on PTO soon. Unless this is a quick fix that can happen this week, cuOpt should fall back on rmm for 22.08.

tfeher:
Note that cuML had the same issue: segfault during the destruction of objects, if we were wrapping a class with an mdarray class member (https://github.com/rapidsai/cuml/pull/4823).

The current workround is to replace the `device_mdarray`s with `device_uvector`s https://github.com/achirkin/raft/commit/8b267506f4f48936a6609d8d7d1866eca754bd56

tfeher:
This looks like a similar issue: https://github.com/rapidsai/rmm/pull/931

PR #764, opened 2022-07-28
Restrict the use of `[[no_unique_address]]` to CXX 20 only.

trivialfis:
Close https://github.com/rapidsai/raft/issues/740 .

Found by @achirkin , the size of `mapping` is non-deterministic. This is caused by the macro definition around whether `no_unique_address` is supported with nvcc. This PR uses the standard cpp version to define the macro regardless of the compiler being used.

### DIFF
--- a/cpp/include/raft/thirdparty/mdspan/include/experimental/__p0009_bits/config.hpp
+++ b/cpp/include/raft/thirdparty/mdspan/include/experimental/__p0009_bits/config.hpp
@@ -121,10 +121,8 @@ static_assert(_MDSPAN_CPLUSPLUS >= MDSPAN_CXX_STD_14, "mdspan requires C++14 or 
 #  define _MDSPAN_PRESERVE_STANDARD_LAYOUT 1
 #endif
 
-// no unique address starts working in NVCC 11.6
-#if !defined(_MDSPAN_USE_ATTRIBUTE_NO_UNIQUE_ADDRESS) && \
-   (!defined(__NVCC__) || ((__CUDACC_VER_MAJOR__ > 11) && (__CUDACC_VER_MINOR__ > 5)))
-#  if (__has_cpp_attribute(no_unique_address) >= 201803L) && !defined(_MDSPAN_COMPILER_MSVC)
+#if !defined(_MDSPAN_USE_ATTRIBUTE_NO_UNIQUE_ADDRESS)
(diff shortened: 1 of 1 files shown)


cpp/include/raft/thirdparty/mdspan/include/experimental/__p0009_bits/config.hpp
```diff
@@ -121,10 +121,8 @@ static_assert(_MDSPAN_CPLUSPLUS >= MDSPAN_CXX_STD_14, "mdspan requires C++14 or
 #  define _MDSPAN_PRESERVE_STANDARD_LAYOUT 1
 #endif
 
-// no unique address starts working in NVCC 11.6
-#if !defined(_MDSPAN_USE_ATTRIBUTE_NO_UNIQUE_ADDRESS) && \
-   (!defined(__NVCC__) || ((__CUDACC_VER_MAJOR__ > 11) && (__CUDACC_VER_MINOR__ > 5)))
-#  if (__has_cpp_attribute(no_unique_address) >= 201803L) && !defined(_MDSPAN_COMPILER_MSVC)
+#if !defined(_MDSPAN_USE_ATTRIBUTE_NO_UNIQUE_ADDRESS)
+#  if MDSPAN_HAS_CXX_20 && (__has_cpp_attribute(no_unique_address) >= 201803L)
```
mhoemmen, line 125: Thanks for the proposed fix!  This looks reasonable.  Users can always turn off the feature if their compiler doesn't support it.  Are you able to test with `MDSPAN_HAS_CXX_20`?

mhoemmen (approved):

cjnolet (approved):

cjnolet:
@gpucibot merge

trivialfis:
I will open a PR upstream tomorrow.

cjnolet:
Thankis again for the fix @trivialfis (and @achirkin)!

cpp/include/raft/thirdparty/mdspan/include/experimental/__p0009_bits/config.hpp
crtrott: doesn't that mean you always turn it off for NVCC since nvcc does not support C++20?

mhoemmen:
Thanks @trivialfis !

PR #766, opened 2022-07-29
Relax ivf-flat test recall thresholds

achirkin:
When `n_probes = n_lists`, ivf-flat should perform equal to the brute-force kNN (`recall == 1` in tests). However, it looks like in some rare cases in CI the recall is slightly smaller. This could be due to some numeric errors or due to the radix_topk not being stable.
This PR reduces the recall threshold in the tests for now to avoid occasional failures in the main branch while the issue is investigated.


cjnolet (approved):

cjnolet:
@gpucibot merge

Issue #752, opened 2022-07-25
[FEA] IVF-flat serialization


tfeher:
**Is your feature request related to a problem? Please describe.**
cuML will need to serialize/deserialize the `ivf_flat::index` (see https://github.com/rapidsai/cuml/issues/4743).  Some other downstream libraries might want to do the same thing. 

It would be easier if RAFT provides methods to save / restore the state from simple byte strings.

**Describe the solution you'd like**

- Provide serialize  method that return host byte string with the index state.
- Provide a deserialize method that creates an index from a byte string.

**Describe alternatives you've considered**
The new IVF flat code #747 provides accessors for the index state, so it is possible for each library to roll out their preferred serialization method. This requires downstream library developers to understand (part of the) implementation details of the index. It would be easier if RAFT provides the serialization method.

teju85:
@cjnolet and I had a long discussion around this topic. Thought it would if a short summary of this discussion was provided here:

Pros of RAFT providing serialization APIs:
1. Code reusability.
2. Users don't have to worry about serialization

Cons:
1. Users are restricted to the way we serialize (although they could technically use the index structures we expose and write their own serialization methods)

cjnolet:
@tfeher I think this issue has been addressed. Can it be closed?

PR #770, opened 2022-08-02
Serialization for commonly used types

achirkin:
This draft pull request adds a simple serialization to commonly used types in raft.

__Serialized layout__
_Simple_ here refers to the way the data is stored: at the moment, no safety/type/versioning information is preserved; only a bare minimum of information is stored to a byte array and read back.

__Target type__
In the most basic form, the target of serialization is a raw pointer. Currently, I also provide overloads to store data to `vector<uint8_t>`.

__ContextArgs... args__
Plain structures and arithmetic types do not require any context to serialize/deserialize (zero args). The data on gpu devices needs, at least, a cuda stream (one arg). mdarrays / rmm vectors may need a memory resource to tell where to keep the deserialized live object (two args). To allow all of these in a single template, I chose to use a parameter pack in this iteration.

Closes: https://github.com/rapidsai/raft/issues/752

### DIFF
--- a/cpp/include/raft/core/serialization.hpp
+++ b/cpp/include/raft/core/serialization.hpp
@@ -80,7 +80,7 @@ auto deserialize(T* p, const void* in, ContextArgs&&... args) -> size_t
  * @tparam T type of the serializable object
  * @tparam ContextArgs types of context required for serialization
  *
- * @param[out] p an unitialized host pointer to a location where the object should be created.
+ * @param[out] p an uninitialized host pointer to a location where the object should be created.
  * @param in a vector where the state should be read from.
  * @param args context required for serialization
  * @return the number of bytes read from the vector
--- a/cpp/include/raft/core/serialization.hpp
+++ b/cpp/include/raft/core/serialization.hpp
@@ -1,0 +1,117 @@
+/*
+ * Copyright (c) 2022, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <raft/core/handle.hpp>
+#include <raft/detail/serialization.hpp>
+
+#include <vector>
+
+namespace raft {
+
+/**
+ * @brief Write a serializable state of an object to memory.
+ *
+ * @tparam Type of the serializable object
+ *
+ * @param handle provides a GPU context for objects, whos state depends on it
+ * @param obj the object to be serialized
+ * @param[out] out a host pointer to the location where to store the state;
+ *   when `nullptr`, the actual data is not written (only the size-to-written is calculated).
+ * @return the number of bytes (to be) written by the pointer
+ */
+template <typename T>
+auto serialize(const handle_t& handle, const T& obj, void* out) -> size_t
+{
+  return detail::call_serialize<T>(handle, obj, reinterpret_cast<uint8_t*>(out));
+}
+
+/**
+ * @brief Write a serializable state of an object to a host vector.
+ *
+ * @tparam Type of the serializable object
+ *
+ * @param handle provides a GPU context for objects, whos state depends on it
+ * @param obj the object to be serialized
+ * @return the serialized state
+ */
+template <typename T>
+auto serialize(const handle_t& handle, const T& obj) -> std::vector<uint8_t>
+{
+  return detail::call_serialize<T>(handle, obj);
+}
+
+/**
+ * @brief Read a serializable state of an object from memory.
+ *
+ * @tparam Type of the serializable object
+ *
+ * @param handle provides a GPU context for objects, whos state depends on it.
+ * @param[out] p an unitialized host pointer to a location where the object should be created.
+ * @param[in] in a host pointer to the location where the state should be read from.
+ */
+template <typename T>
+void deserialize(const handle_t& handle, T* p, const void* in)
+{
+  return detail::call_deserialize<T>(handle, p, reinterpret_cast<const uint8_t*>(in));
+}
+
+/**
+ * @brief Read a serializable state of an object from a vector.
+ *
+ * @tparam Type of the serializable object
+ *
+ * @param handle provides a GPU context for objects, whos state depends on it.
+ * @param[out] p an unitialized host pointer to a location where the object should be created.
+ * @param in a vector where the state should be read from.
+ */
+template <typename T>
+void deserialize(const handle_t& handle, T* p, const std::vector<uint8_t>& in)
+{
+  return detail::call_deserialize<T>(handle, p, in);
+}
+
+/**
+ * @brief Read a serializable state of an object.
+ *
+ * @tparam Type of the serializable object
+ *
+ * @param handle provides a GPU context for objects, whos state depends on it.
+ * @param[in] in a host pointer to the location where the state should be read from.
+ * @return the deserialized object;
+ */
+template <typename T>
+auto deserialize(const handle_t& handle, const void* in) -> T
+{
+  return detail::call_deserialize<T>(handle, reinterpret_cast<const uint8_t*>(in));
+}
+
+/**
+ * @brief Read a serializable state of an object.
+ *
+ * @tparam Type of the serializable object
+ *
+ * @param handle provides a GPU context for objects, whos state depends on it.
+ * @param in a vector where the state should be read from.
+ * @return the deserialized object;
+ */
+template <typename T>
+auto deserialize(const handle_t& handle, const std::vector<uint8_t>& in) -> T
+{
+  return detail::call_deserialize<T>(handle, in);
+}
+
+}  // namespace raft
--- a/cpp/include/raft/core/serialization.hpp
+++ b/cpp/include/raft/core/serialization.hpp
@@ -25,93 +25,103 @@ namespace raft {
 /**
  * @brief Write a serializable state of an object to memory.
  *
- * @tparam Type of the serializable object
+ * @tparam T type of the serializable object
+ * @tparam ContextArgs types of context required for serialization
  *
- * @param handle provides a GPU context for objects, whos state depends on it
- * @param obj the object to be serialized
  * @param[out] out a host pointer to the location where to store the state;
  *   when `nullptr`, the actual data is not written (only the size-to-written is calculated).
+ * @param obj the object to be serialized
+ * @param args context required for serialization
  * @return the number of bytes (to be) written by the pointer
  */
-template <typename T>
-auto serialize(const handle_t& handle, const T& obj, void* out) -> size_t
+template <typename T, typename... ContextArgs>
+auto serialize(uint8_t* out, const T& obj, ContextArgs&&... args) -> size_t
 {
-  return detail::call_serialize<T>(handle, obj, reinterpret_cast<uint8_t*>(out));
+  return detail::call_serialize<T, ContextArgs...>(out, obj, std::forward<ContextArgs>(args)...);
 }
 
 /**
  * @brief Write a serializable state of an object to a host vector.
  *
- * @tparam Type of the serializable object
+ * @tparam T type of the serializable object
+ * @tparam ContextArgs types of context required for serialization
  *
- * @param handle provides a GPU context for objects, whos state depends on it
  * @param obj the object to be serialized
+ * @param args context required for serialization
  * @return the serialized state
  */
-template <typename T>
-auto serialize(const handle_t& handle, const T& obj) -> std::vector<uint8_t>
+template <typename T, typename... ContextArgs>
+auto serialize(const T& obj, ContextArgs&&... args) -> std::vector<uint8_t>
 {
-  return detail::call_serialize<T>(handle, obj);
+  return detail::call_serialize<T, ContextArgs...>(obj, std::forward<ContextArgs>(args)...);
 }
 
 /**
  * @brief Read a serializable state of an object from memory.
  *
- * @tparam Type of the serializable object
+ * @tparam T type of the serializable object
+ * @tparam ContextArgs types of context required for serialization
  *
- * @param handle provides a GPU context for objects, whos state depends on it.
  * @param[out] p an unitialized host pointer to a location where the object should be created.
  * @param[in] in a host pointer to the location where the state should be read from.
+ * @param args context required for serialization
+ * @return the number of bytes read by the pointer
  */
-template <typename T>
-void deserialize(const handle_t& handle, T* p, const void* in)
+template <typename T, typename... ContextArgs>
+auto deserialize(T* p, const void* in, ContextArgs&&... args) -> size_t
 {
-  return detail::call_deserialize<T>(handle, p, reinterpret_cast<const uint8_t*>(in));
+  return detail::call_deserialize<T, ContextArgs...>(
+    p, reinterpret_cast<const uint8_t*>(in), std::forward<ContextArgs>(args)...);
 }
 
 /**
  * @brief Read a serializable state of an object from a vector.
  *
- * @tparam Type of the serializable object
+ * @tparam T type of the serializable object
+ * @tparam ContextArgs types of context required for serialization
  *
- * @param handle provides a GPU context for objects, whos state depends on it.
  * @param[out] p an unitialized host pointer to a location where the object should be created.
  * @param in a vector where the state should be read from.
+ * @param args context required for serialization
+ * @return the number of bytes read from the vector
  */
-template <typename T>
-void deserialize(const handle_t& handle, T* p, const std::vector<uint8_t>& in)
+template <typename T, typename... ContextArgs>
+auto deserialize(T* p, const std::vector<uint8_t>& in, ContextArgs&&... args) -> size_t
 {
-  return detail::call_deserialize<T>(handle, p, in);
+  return detail::call_deserialize<T, ContextArgs...>(p, in, std::forward<ContextArgs>(args)...);
 }
 
 /**
  * @brief Read a serializable state of an object.
  *
- * @tparam Type of the serializable object
+ * @tparam T type of the serializable object
+ * @tparam ContextArgs types of context required for serialization
  *
- * @param handle provides a GPU context for objects, whos state depends on it.
  * @param[in] in a host pointer to the location where the state should be read from.
+ * @param args context required for serialization
  * @return the deserialized object;
  */
-template <typename T>
-auto deserialize(const handle_t& handle, const void* in) -> T
+template <typename T, typename... ContextArgs>
+auto deserialize(const void* in, ContextArgs&&... args) -> T
 {
-  return detail::call_deserialize<T>(handle, reinterpret_cast<const uint8_t*>(in));
+  return detail::call_deserialize<T, ContextArgs...>(reinterpret_cast<const uint8_t*>(in),
+                                                     std::forward<ContextArgs>(args)...);
 }
 
 /**
  * @brief Read a serializable state of an object.
  *
- * @tparam Type of the serializable object
+ * @tparam T type of the serializable object
+ * @tparam ContextArgs types of context required for serialization
  *
- * @param handle provides a GPU context for objects, whos state depends on it.
  * @param in a vector where the state should be read from.
+ * @param args context required for serialization
  * @return the deserialized object;
  */
-template <typename T>
-auto deserialize(const handle_t& handle, const std::vector<uint8_t>& in) -> T
+template <typename T, typename... ContextArgs>
+auto deserialize(const std::vector<uint8_t>& in, ContextArgs&&... args) -> T
 {
-  return detail::call_deserialize<T>(handle, in);
+  return detail::call_deserialize<T, ContextArgs...>(in, std::forward<ContextArgs>(args)...);
 }
 
 }  // namespace raft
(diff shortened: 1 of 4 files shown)


trivialfis:
Hi, have you considered using a schema like onnx that can be reused by others and provide version info.  I think versioning is quite important if we want to keep some stability over serialized object.

cpp/include/raft/core/serialization.hpp
```diff
@@ -0,0 +1,127 @@
...
+ * @tparam T type of the serializable object
+ * @tparam ContextArgs types of context required for serialization
+ *
+ * @param[out] p an unitialized host pointer to a location where the object should be created.
+ * @param[in] in a host pointer to the location where the state should be read from.
+ * @param args context required for serialization
+ * @return the number of bytes read by the pointer
+ */
+template <typename T, typename... ContextArgs>
+auto deserialize(T* p, const void* in, ContextArgs&&... args) -> size_t
```
mhoemmen, line 71: 1. What happens if deserialization fails (e.g., because the representation is not correct)?  How would you report the error?
2. It sounds like the intent is for `deserialize` to do placement new in `p`.  Is that correct?
3. Would you instead consider an interface that writes to an existing T via `T& out`?  The issue with a raw pointer interface is that users may forget the invariant "an *uninitialized* host pointer" and pass in a pointer to an existing object.  In that case, your interface can totally break any invariants of `T`, e.g., possibly leaking any heap allocations that `T` might have inside.

cpp/include/raft/core/serialization.hpp
```diff
@@ -0,0 +1,127 @@
...
+ * @tparam T type of the serializable object
+ * @tparam ContextArgs types of context required for serialization
+ *
+ * @param[out] p an unitialized host pointer to a location where the object should be created.
+ * @param in a vector where the state should be read from.
+ * @param args context required for serialization
+ * @return the number of bytes read from the vector
+ */
+template <typename T, typename... ContextArgs>
+auto deserialize(T* p, const std::vector<uint8_t>& in, ContextArgs&&... args) -> size_t
```
mhoemmen, line 89: 1. Would you consider using `std::span<const uint8_t>` instead of `const std::vector&` to pass in the buffer?
2. How would this interface report errors?

cpp/include/raft/core/serialization.hpp
```diff
@@ -0,0 +1,127 @@
...
+  return detail::call_serialize<T, ContextArgs...>(obj, std::forward<ContextArgs>(args)...);
+}
+
+/**
+ * @brief Read a serializable state of an object from memory.
+ *
+ * @tparam T type of the serializable object
+ * @tparam ContextArgs types of context required for serialization
+ *
+ * @param[out] p an unitialized host pointer to a location where the object should be created.
```
mhoemmen, line 65:
```suggestion
 * @param[out] p an uninitialized host pointer to a location where the object should be created.
```

cpp/include/raft/core/serialization.hpp
```diff
@@ -0,0 +1,127 @@
...
+    p, reinterpret_cast<const uint8_t*>(in), std::forward<ContextArgs>(args)...);
+}
+
+/**
+ * @brief Read a serializable state of an object from a vector.
+ *
+ * @tparam T type of the serializable object
+ * @tparam ContextArgs types of context required for serialization
+ *
+ * @param[out] p an unitialized host pointer to a location where the object should be created.
```
mhoemmen, line 83:
```suggestion
 * @param[out] p an uninitialized host pointer to a location where the object should be created.
```

cpp/include/raft/core/serialization.hpp
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * @tparam T type of the serializable object
+ * @tparam ContextArgs types of context required for serialization
+ *
+ * @param[in] in a host pointer to the location where the state should be read from.
+ * @param args context required for serialization
+ * @return the deserialized object;
+ */
+template <typename T, typename... ContextArgs>
+auto deserialize(const void* in, ContextArgs&&... args) -> T
```
mhoemmen, line 105: Instead of taking the input buffer as `const void*`, would you consider taking it as `std::span<uint8_t>`?  That would better express both the length of the buffer, and its desired element type.  Otherwise, users have no idea what input representation you intend.

cpp/include/raft/core/serialization.hpp
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * @tparam T type of the serializable object
+ * @tparam ContextArgs types of context required for serialization
+ *
+ * @param in a vector where the state should be read from.
+ * @param args context required for serialization
+ * @return the deserialized object;
+ */
+template <typename T, typename... ContextArgs>
+auto deserialize(const std::vector<uint8_t>& in, ContextArgs&&... args) -> T
```
mhoemmen, line 122: Would you consider taking the input buffer as `std::span<const uint8_t>` instead of `const vector&`?

cpp/include/raft/detail/serialization.hpp
```diff
@@ -0,0 +1,205 @@
...
+template <typename T, typename... ContextArgs>
+auto call_deserialize(T* p, const std::vector<uint8_t>& in, ContextArgs&&... args) -> size_t
+{
+  return call_deserialize<T, ContextArgs...>(p, in.data(), std::forward<ContextArgs>(args)...);
+}
+
+template <typename T, typename... ContextArgs>
+auto call_deserialize(const uint8_t* in, ContextArgs&&... args) -> T
+{
+  union res {
```
mhoemmen, line 61: A more idiomatic way to avoid initialization of T would be to have `call_serialize` return `optional<T>` or `expected<T, some_error_code>`.

cpp/include/raft/detail/serialization.hpp
```diff
@@ -0,0 +1,205 @@
...
+      call_deserialize<T, ContextArgs...>(&value, in, std::forward<ContextArgs>(args)...);
+    }
+    ~res() { value.~T(); }  // NOLINT
+  };
+  // using a union to avoid initialization of T and force copy elision.
+  return res(in, std::forward<ContextArgs>(args)...).value;
+}
+
+template <typename T, typename... ContextArgs>
+auto call_deserialize(const std::vector<uint8_t>& in, ContextArgs&&... args) -> T
```
mhoemmen, line 74: Please see above notes on using `span<const uint8_t>` instead of `const vector<uint8_t>&` -- thanks!

cpp/include/raft/detail/serialization.hpp
```diff
@@ -0,0 +1,205 @@
...
+}
+
+template <typename T>
+struct serial {
+  // Default implementation for all arithmetic types: just write the value by the pointer.
+  template <typename S = T>
+  static auto to_bytes(uint8_t* out, const S& obj)
+    -> std::enable_if_t<std::is_arithmetic_v<S>, size_t>
+  {
+    if (out) { *reinterpret_cast<T*>(out) = obj; }
```
mhoemmen, line 86: This smells of UB, even if it isn't.  More importantly, it assumes that the lifetime of `out` has begun.  I don't think you can dereference `out` as `T` unless you know that it points to some `T`, even if it points to valid memory and is correctly aligned (which you haven't checked here).  Would you instead consider `memcpy`?
```c++
memcpy(&obj, out, sizeof(S));
```

cpp/include/raft/detail/serialization.hpp
```diff
@@ -0,0 +1,205 @@
...
+                        detail::device_accessor<detail::device_uvector_policy<ElementType>>>;
+
+  static auto to_bytes(uint8_t* out, const obj_t& obj, const handle_t& handle) -> size_t
+  {
+    auto extents_size = call_serialize<Extents>(out, obj.extents());
+    auto total_size   = obj.size() * sizeof(ElementType) + extents_size;
+    if (out) {
+      out += extents_size;
+      raft::copy(
+        reinterpret_cast<ElementType*>(out), obj.data_handle(), obj.size(), handle.get_stream());
```
mhoemmen, line 152: [INCORRECT]

This is only correct if the layout is exhaustive (contiguous in memory, not strided or otherwise skipping parts of memory).  Otherwise, it will both write memory that are not part of the `mdarray`'s elements, and it will miss elements of the `mdarray`.

The right thing to do would be to define an `mdspan` copy operation, as in P1673.  Then you can copy from the input `mdspan` to a temporary output `mdspan` that views `out` with the desired output layout.

cpp/include/raft/detail/serialization.hpp
```diff
@@ -0,0 +1,205 @@
...
+
+  static auto from_bytes(obj_t* p, const uint8_t* in) -> size_t
+  {
+    new (p) obj_t{*reinterpret_cast<const obj_t*>(in)};
+    return sizeof(obj_t);
+  }
+};
+
+template <typename ElementType, typename Extents, typename LayoutPolicy>
+struct serial<mdarray<ElementType,
```
mhoemmen, line 136: Serializing `mdarray` instead of `mdspan` implies code duplication.  Would you consider instead only defining serialization for `mdspan`?

cpp/include/raft/detail/serialization.hpp
```diff
@@ -0,0 +1,205 @@
...
+    return total_size;
+  }
+
+  static auto from_bytes(obj_t* p,
+                         const uint8_t* in,
+                         const handle_t& handle,
+                         rmm::mr::device_memory_resource* mr = nullptr) -> size_t
+  {
+    Extents exts;
+    auto extents_size = call_deserialize<Extents>(&exts, in);
```
mhoemmen, line 163: What is your representation of `extents` if all the extents are static?  Are you able to distinguish this case from simply not having any data at all?

cpp/include/raft/detail/serialization.hpp
```diff
@@ -0,0 +1,205 @@
...
+    in += extents_size;
+    typename obj_t::mapping_type layout{exts};
+    typename obj_t::container_policy_type policy{handle.get_stream(), mr};
+    new (p) obj_t{layout, policy};
```
mhoemmen, line 167: This approach first value-initializes all the elements of the mdarray, then deserializes into the mdarray.  This iterates over the mdarray twice.  If your mdarray has `std::vector` as its container, you could replace the `raft::copy` call with a range over the input, do `ranges::to` initialization of a `std::vector`, and then move the vector into the mdarray.

cpp/include/raft/detail/serialization.hpp
```diff
@@ -0,0 +1,205 @@
...
+    raft::copy(
```
mhoemmen, line 168: [INCORRECT]

This is only correct if the layout is exhaustive (contiguous in memory, not strided or otherwise skipping parts of memory).  Please see above note.

mhoemmen (changes requested):
Thanks for asking me to review!  I've marked the code that's definitely incorrect with `[INCORRECT]`.  It will be wrong for `layout_stride` if `! is_exhaustive()`.

cpp/include/raft/core/serialization.hpp
```diff
@@ -0,0 +1,127 @@
...
+ * @tparam ContextArgs types of context required for serialization
+ *
+ * @param[out] out a host pointer to the location where to store the state;
+ *   when `nullptr`, the actual data is not written (only the size-to-written is calculated).
+ * @param obj the object to be serialized
+ * @param args context required for serialization
+ * @return the number of bytes (to be) written by the pointer
+ */
+template <typename T, typename... ContextArgs>
+auto serialize(uint8_t* out, const T& obj, ContextArgs&&... args) -> size_t
```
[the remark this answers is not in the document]
achirkin: I'd be happy to add overloads to have span as the output, but raft is currently set to `C++17`. It's not available till C++20, right?

cpp/include/raft/core/serialization.hpp
achirkin: Let's keep the related discussion here
https://github.com/rapidsai/raft/pull/770#discussion_r935679007

cpp/include/raft/core/serialization.hpp
achirkin: Let's keep the related discussion here
https://github.com/rapidsai/raft/pull/770#discussion_r935679007

cpp/include/raft/core/serialization.hpp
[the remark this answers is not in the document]
cjnolet: What if we returned a raft::span?

cpp/include/raft/detail/serialization.hpp
achirkin: Thanks for the suggestion! Is there an official way to populate the optional using a callback accepting a pointer to the data (i.e. `call_deserialize(T*, ...)`?

I see I haven't explained why I've gone through such a hussle with these many overloads. My goal here is to provide an easy way for other developers to add serialization to new structures. That is, to add serialization to a new structure, one would need to write just two methods:
```
template <T>
struct serial {
  static auto to_bytes(uint8_t* out, const T& obj) -> size_t { ... }
  static auto from_bytes(T* p, const uint8_t* in) -> size_t { ... }
};
```
Although I love the return-value approach, I'm afraid I cannot use it here, because some of the classes in raft and cuml do not have some of the copy, move, or nullary constructors. Hence I've come up with this idea that it should be the responsibility of the serialization implementer to initialize the object in the `from_bytes` method.

The rest, all `call_serialize`/`call_deserialize` are implemented in terms of these two. At this moment, we don't seem to have a consensus over which target type to use for serialization - so I added a few of them here.

Assuming that `from_bytes` is available as-is, the union approach is the only way I could come up to return the value in this overload. I'd be happy to use `optional` for this, if it's possible!

cpp/include/raft/detail/serialization.hpp
achirkin: Let's keep the related discussion here
https://github.com/rapidsai/raft/pull/770#discussion_r935679007

cpp/include/raft/core/serialization.hpp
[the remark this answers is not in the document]
achirkin: Ah, great, I can add overloads that accept `raft::span`!

cpp/include/raft/detail/serialization.hpp
achirkin: Hmm, at this moment we don't have any sorts of RTTI at all. That is, in this version the caller should know it has serialized the static extents, by possibly not writing anything at all; then `from_bytes` should read the same thing - possibly nothing.

cpp/include/raft/core/serialization.hpp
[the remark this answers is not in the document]
achirkin: > What if we returned a raft::span?

It does not seem to have python bindings at the moment, right? I guess, we'd need at least one version of serialize that could be easily wrapped by cython for https://github.com/rapidsai/raft/issues/752

cpp/include/raft/core/serialization.hpp
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * @tparam T type of the serializable object
+ * @tparam ContextArgs types of context required for serialization
+ *
+ * @param obj the object to be serialized
+ * @param args context required for serialization
+ * @return the serialized state
+ */
+template <typename T, typename... ContextArgs>
+auto serialize(const T& obj, ContextArgs&&... args) -> std::vector<uint8_t>
```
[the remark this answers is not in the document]
achirkin: The base implementation (`to_bytes`) at the moment writes to existing storage, but I don't have any checks whatsoever to see if there is enough space allocated. In the current state, the user is supposed to call `serialize` with `nullptr` to find out how much space it needs.

cpp/include/raft/core/serialization.hpp
achirkin: 1. At the moment, there are no safety checks at all. That being said, I think, we could add them in `to_bytes`/`from_bytes`, i.e. make it the responsibility of a person who adds serialization to a particular class - just throw one of the raft errors. What do you think? CC @cjnolet
2. Yes
3. I'd be fine with that, but I'm not sure how to do that better, while keeping the `serial::from_bytes` the same. Shall I add an overload that accepts `T& out`, manually calls its destructor and then passes the `&out` pointer further?.. sounds hackish.

In general, the reason why I'm hesitant to change `serial::from_bytes` to take a reference to an initialized object is that some of the classes in raft/cuml don't seem to have constructors that _do nothing_. On the one hand, the pre-initialization would likely mean some extra overhead and a minor nuisance of passing dummy arguments if there is no nullary constructor available. On the other hand, that would make impossible the value-returning overload of deserialize (`deserialize(const std::vector<uint8_t>& in, ContextArgs&&... args) -> T`) for the types that don't have the nullary constructor.

cpp/include/raft/core/serialization.hpp
achirkin: Thanks for pointing this out, let's use the following threads for the discussion:
 1. https://github.com/rapidsai/raft/pull/770#discussion_r935679007
 2. https://github.com/rapidsai/raft/pull/770#discussion_r935712803

cpp/include/raft/core/serialization.hpp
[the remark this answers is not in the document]
mhoemmen: @achirkin Would you consider a design that names the space query as a separate function from serialization?  Compare to cuSPARSE, which separates `*_bufferSize` functions from `*_solve` functions.

cpp/include/raft/core/serialization.hpp
mhoemmen: > Shall I add an overload that accepts `T& out`, manually calls its destructor and then passes the `&out` pointer further?.. sounds hackish.

Please do not : - )  

> On the other hand, that would make impossible the value-returning overload ... for the types that don't have the nullary constructor.

It's possible if you have assignment, but ugly -- https://godbolt.org/z/774Ebfvh5 -- so I understand if you don't want to do it.

cpp/include/raft/detail/serialization.hpp
mhoemmen: @achirkin Thanks for explaining!  I see the difficulty now in using `optional`.  Another way to express this besides a `union` would be to allocate an array of bytes with the same size and alignment as `T`, and then placement new into it.  Here is an example.
```c++
class NotDefaultConstructible {
public:
  NotDefaultConstructible() = delete;
  NotDefaultConstructible(int i) : i_(i) {}

  int value() const { return i_; }

private:
  int i_;
};

void deserialize_raw(NotDefaultConstructible* out, std::string_view s)
{
    int value{};  
    auto result = std::from_chars(s.data(), s.data()+s.size(), value);
    if (result.ec == std::errc::invalid_argument) {
        throw std::invalid_argument{"String does not represent an integer"};
    }
    else if (result.ec == std::errc::result_out_of_range) {
        throw std::out_of_range{"String represents an integer, but it does not fit in int"};
    }
    // WARNING: assumes out does NOT point to a valid object.
    new(out) NotDefaultConstructible(value);
}

int main()
{
    constexpr size_t num_bytes = sizeof(NotDefaultConstructible);
    alignas(num_bytes) std::byte ptr[num_bytes];

    NotDefaultConstructible* out = reinterpret_cast<NotDefaultConstructible*>(ptr);
    deserialize_raw(out, s);

    std::cout << "Value: " << out->value();
    out->~NotDefaultConstructible(); // oof, we have to do this by hand

    return 0;
}
```

cpp/include/raft/core/serialization.hpp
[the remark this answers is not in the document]
cjnolet: It should be straightforward to expose `raft::span` in a `pxd` file. We can also rename the functions as needed (for e.g. `to_span`/`from_span` and `to_bytes`/`from_bytes`).

cpp/include/raft/core/serialization.hpp
achirkin: I meant to say, it would be impossible if I changed `serial::from_bytes` to accept the reference instead of the pointer (`deserialize_raw` in your example). Also, `deserialize_using_raw` in your example does not call the destructor, does it? :) (nb my `union` trick to address this and copy elision)

cpp/include/raft/detail/serialization.hpp
achirkin: Ouch, I don't want to call the destructor by hand in the presence of exceptions in `deserialize_raw` :) I guess I'd better go with the `union` approach if you don't see anything criminal in that.

cpp/include/raft/core/serialization.hpp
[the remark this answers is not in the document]
achirkin: I'd be totally fine separating them, though for some reason I've had an impression that in raft/cuml we follow the CUB approach of using the same function. @cjnolet , which one do you prefer?

cpp/include/raft/core/serialization.hpp
mhoemmen: @achirkin Right, I definitely forgot the important detail of manually invoking the destructor on a placement new'd thing.

cpp/include/raft/core/serialization.hpp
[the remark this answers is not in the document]
cjnolet: This one is hard. @mhoemmen, we've been trying to keep the APIs in RAFT a little more C++ friendly, so generally trying to avoid the two-stage `compute buffer size` -> `perform computation` where at all possible. I'm usually not particularly in love w/ returning objects from functions either, unless we are explicitly using something like a factory for a complex object, because then the user cannot control the memory allocation. 

In this particular case, if we are only ever expecting the serialized buffer to be in host memory, I could see an argument for having serialize return a vector, or even accepting a vector and populating it. I think if I was to choose between the various options presented, I most like the idea of returning/populating some sort of streamable/iterable type if we can. I find that to be the most flexible design because a large stream can go right to disk and have to buffer only a small portion into memory at any given time. Further, the stream is also nice because you can collect it into a byte buffer at any point if needed.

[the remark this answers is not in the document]
mhoemmen: @cjnolet wrote:
> I think if I was to choose between the various options presented, I most like the idea of returning/populating some sort of streamable/iterable type if we can.

I do like the idea of writing to some kind of output range.  If it's a random access range, it would be nice to test if it's long enough and report an error if not, rather than just clobbering it.  This is always detectable, and is generally recoverable (allocate more memory) or at least worth reporting ("insufficient disk space").

Please do consider not allocating the return buffer each time.  This would make it impossible to write sequential records to contiguous storage.  It can also be surprisingly expensive in a tight loop.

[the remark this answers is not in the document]
cjnolet: > Please do consider not allocating the return buffer each time. This would make it impossible to write sequential records to contiguous storage. It can also be surprisingly expensive in a tight loop.

I do wholeheartedly agree with this.

achirkin:
Closing in favor of https://github.com/rapidsai/raft/pull/1173

PR #771, opened 2022-08-02
Various fixes for build.sh

vyasr:
This PR includes the following fixes for build.sh:
- The parallel argument is for the build tool (make/ninja) and not CMake or setup.py, so it needs to be passed after a second `--`.
- pylibraft requires the distances component of libraft to be compiled. If pylibraft is passed to build.sh, we need to force building libraft with the distance component
- build.sh doesn't install by default, so the `CMAKE_PREFIX_PATH` that the Python build points to needs to also include the build directory in addition to the install directory. I think giving the build directory precedence makes more sense, but that could be changed if necessary.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -250,7 +250,7 @@ fi
 
 ################################################################################
 # Configure for building all C++ targets
-if (( ${NUMARGS} == 0 )) || hasArg libraft || hasArg docs || hasArg tests || hasArg bench; then
+if (( ${NUMARGS} == 0 )) || hasArg libraft || hasArg pylibraft || hasArg docs || hasArg tests || hasArg bench; then
     if (( ${BUILD_ALL_GPU_ARCH} == 0 )); then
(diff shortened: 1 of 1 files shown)


teju85:
I confirm that this PR fixes the issues I was facing: (Thanks @vyasr !)
1. parallel build not working
2. "distances" sources getting compiled twice (due to me having to set `FIND_RAFT_CPP=OFF`)

After this PR gets merged, if one wants reuse the compiled sources from cpp/src, then one needs to make sure NOT to pass the above `FIND_RAFT_CPP=OFF` option!

That said, there's still a minor hiccup: the `python setup.py ...` commands in build.sh somehow doesn't seem to honor setting up ccache through options like `-DCMAKE_CUDA_COMPILER_LAUNCHER=ccache`. For now, this is not an issue for me as I'll not be choosing the `FIND_RAFT_CPP=OFF` path.

teju85 (approved):

teju85:
@gpucibot merge

PR #778, opened 2022-08-03
[FEA] Exposing python wrapper for the RMAT generator logic

teju85:
This PR adds a thin python wrapper for the RMAT generator logic (as per the request from @PiotrBigajNV)

@cjnolet I was not sure what's the right way to wrap other such RAFT prims to python layer. Hence, I just followed the same path as that of `raft_distance` and created a `raft_random` build target instead. Let me know if you have a better approach for this one.


python/pylibraft/pylibraft/test/test_random.py
```diff
@@ -0,0 +1,75 @@
...
+
+import pytest
+import numpy as np
+
+import rmm
+
+from pylibraft.random import rmat
+
+
+class TestDeviceBuffer:
```
cjnolet, line 24: Since we're using this in two places now, we should probably centralize it in its own file so we can update it in one place when needed.

cjnolet (changes requested):
Super excited to start exposing more of our primitives through Python! Other than a discussion on the shared libs / components, I only have minor feedback

build.sh
```diff
@@ -187,9 +189,15 @@ if hasArg --compile-dist || hasArg --compile-libs || hasArg pylibraft || (( ${NU
...
 
+if hasArg --compile-rand || hasArg --compile-libs || hasArg pylibraft || (( ${NUMARGS} == 0 )); then
+    COMPILE_RAND_LIBRARY=ON
+    CMAKE_TARGET="${CMAKE_TARGET};raft_random_lib"
+fi
+
 if hasArg tests || (( ${NUMARGS} == 0 )); then
     BUILD_TESTS=ON
     COMPILE_DIST_LIBRARY=ON
+    COMPILE_RAND_LIBRARY=ON
```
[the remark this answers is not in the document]
teju85: This was the exact thing I wasn't sure of! I'm glad to know that you'd like to keep this generic. Let me propose a change.

python/pylibraft/pylibraft/test/test_random.py
teju85: Done

python/pylibraft/pylibraft/test/test_random.py
```diff
@@ -0,0 +1,75 @@
...
+                             like=self.ndarray_)\
+            .astype(self.ndarray_.dtype)\
+            .reshape(self.ndarray_.shape)
+
+
+@pytest.mark.parametrize("n_edges", [10000, 20000])
+@pytest.mark.parametrize("r_scale", [16, 18])
+@pytest.mark.parametrize("c_scale", [16, 18])
+@pytest.mark.parametrize("dtype", [np.int32, np.int64])
+def test_rmat(n_edges, r_scale, c_scale, dtype):
```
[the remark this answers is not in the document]
teju85: Good point. Done.

python/pylibraft/pylibraft/random/rmat_rectangular_generator.pyx
```diff
@@ -0,0 +1,169 @@
...
+def rmat(out, theta, r_scale, c_scale, seed=12345):
+    """
+    Generate RMAT adjacency list based on the input distribution.
+
+    Parameters
+    ----------
+
+    out: CUDA array interface compliant matrix shape (n_edges, 2). This will
+         contain the src/dst node ids stored consecutively like a pair. If this
+         output is not needed, pass None.
```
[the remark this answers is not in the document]
teju85: Oops.. that was an oversight. I was doing this when I was trying to support generating both AoS and SoA output formats with a single call (c++ API supports this). But then decided to settle to AoS format for the python layer, to begin with. Thanks for catching this. Fixed the docstring.

build.sh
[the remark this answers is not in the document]
teju85: Can you check now @cjnolet ?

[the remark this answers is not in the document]
cjnolet: I like it a lot- so users have the option of header-only device or runtime APIs or shared lib w/ host accessable API

cjnolet:
Not something we need to do for this PR, but we might also consider having a separate section in the raft docs for the `header-only` APIs and the `host accessable APIs`.

cjnolet (approved):

teju85:
@cjnolet and/or @vyasr can one of you help me dealing with the CI errors please? Unable to figure out what I'm doing wrong here. :(

cpp/include/raft_compiled/random/rmat_rectangular_generator.hpp
```diff
@@ -0,0 +1,42 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <cstdint>
+
+#include <raft/handle.hpp>
```
lowener, line 19:
```suggestion
#include <raft/core/handle.hpp>
```

teju85:
@cjnolet and @vyasr This PR doesn't touch the "distance-lib" but the following errors can be seen in the CI and I'm not sure why this is happening. Any suggestions/ideas?
```
CMake Error at /workspace/.conda-bld/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/lib/cmake/raft/raft-distance-lib-targets.cmake:61 (set_target_properties):
  The link interface of target "raft::raft_distance_lib" contains:

    cuco::cuco

    * An ALIAS target is missing.
```

cjnolet:
@vyasr do you notice anything obvious or stange in the build? What's weird is that it seems to be looking for that target and I'd assume that would also have caused it to fail when running the `find_package` for cuco, but it doesn't seem to find that.

What's even more strange is that it only seems to be happening in CI. 

```
if(RAFT_COMPILE_DIST_LIBRARY OR distance IN_LIST raft_FIND_COMPONENTS)
  set(RAFT_ENABLE_cuco_DEPENDENCY ON)
endif()
```

teju85:
@vyasr and/or @robertmaynard when you get time, can one of you guys help us here? I'm not sure why CI is failing here!

cjnolet:
@teju85 I'm thinking maybe for now we should just have `libraft-distance.so` compile the `random` source files as well and then add a note/todo in the cmakelists that we're eventually going to be renaming `libraft-distance.so` to `libraft.so` (and likely merging in the nearest neighbors source files as well once we remove faiss).

cpp/CMakeLists.txt
```diff
@@ -407,7 +475,7 @@ install(FILES ${CMAKE_CURRENT_BINARY_DIR}/include/raft/version_config.hpp
 
 include("${rapids-cmake-dir}/export/write_dependencies.cmake")
 
-set(raft_components distance nn)
+set(raft_components distance nn compiled)
 set(raft_install_comp raft raft)
```
robertmaynard, line 479:
```suggestion
set(raft_install_comp raft raft raft)
```

The `raft_components` and `raft_install_comp` need to be equal lengths

robertmaynard: I expect this is the bug as it would cause the final sets to look like:

```
-- raft-distance-targets.cmake => install component raft
-- raft-nn-targets.cmake => install component raft
-- raft-compiled-targets.cmake => install component distance
-- raft-distance-lib-targets.cmake => install component nn
-- raft-nn-lib-targets.cmake => install component compiled
-- raft-compiled-lib-targets.cmake => install component 
```

robertmaynard (changes requested):

cpp/CMakeLists.txt
teju85: Duh... such a silly mistake. Sorry about that. Thanks a lot @robertmaynard ! This has fixed the build issue for me. Now somehow the pytest isn't able to find the right modules. I'll figure out that.

cjnolet:
@teju85 Looks like we got it finally! I think I traced the issue to `__init__.py` files that shouldn't have been created (one in the root of `pylibraft` and another in the `tests` directory. I think we're good to go at this point, pending @robertmaynard's blessing, of course.

teju85:
woohoo! Thanks @cjnolet and @robertmaynard . We need a final approval from @robertmaynard before this can be merged.

robertmaynard (approved):

cjnolet:
@gpucibot merge

cjnolet (approved):

Ethyling (approved):

PR #781, opened 2022-08-05
Remove type punning from TxN_t

wphicks:
Avoid undefined behavior related to accessing inactive union members in TxN_t


cpp/include/raft/vectorized.cuh
```diff
@@ -310,22 +310,22 @@ struct TxN_t {
   template <typename idx_t = int>
   DI void load(const math_t* ptr, idx_t idx)
   {
-    const io_t* bptr = reinterpret_cast<const io_t*>(&ptr[idx]);
-    val.internal     = __ldg(bptr);
+    const io_t* bptr   = reinterpret_cast<const io_t*>(&ptr[idx]);
```
mhoemmen, line 313: I don't think this change actually solves the problem of accessing an inactive union member.  When default-initializing `TxN_t`, no members of the union are active until you assign to one of them.  Calling `vectorized_data()` at that point accesses `val.data`, an uninitialized union member.

I can think of a few ways to fix this, but all of them would involve value-initializing `data`.

wphicks: Agreed. Moved it to a direct member array.

cjnolet (approved):

mhoemmen (approved):

wphicks:
@gpucibot merge

PR #785, opened 2022-08-09
Fix adj_to_csr_kernel 

ahendriksen:
This PR adds launch bounds to adj_to_csr kernel, making it forward compatible with future architectures.

### DIFF
--- a/cpp/include/raft/sparse/convert/detail/adj_to_csr.cuh
+++ b/cpp/include/raft/sparse/convert/detail/adj_to_csr.cuh
@@ -29,6 +29,9 @@ namespace sparse {
 namespace convert {
 namespace detail {
 
+// Threads per block in adj_to_csr_kernel.
+static const constexpr int adj_to_csr_tpb = 512;
+
 /**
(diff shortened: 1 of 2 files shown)


cpp/include/raft/sparse/convert/detail/adj_to_csr.cuh
```diff
@@ -29,6 +29,9 @@ namespace sparse {
 namespace convert {
 namespace detail {
 
+// Threads per block in adj_to_csr_kernel.
+const int adj_to_csr_tpb = 512;
```
wphicks, line 33:
```suggestion
static const constexpr int adj_to_csr_tpb = 512;
```

ahendriksen: Thanks! Changed.

wphicks (approved):

tfeher (approved):
Thanks @ahendriksen for the PR, it looks good to me!

wphicks:
@gpucibot merge

PR #786, opened 2022-08-09
Using std::vector instead of arrays in gtests

cjnolet:
This fixes some strange behavior we're encountering in the `stats/SumTest` and `AtomicWarpInc` test. For some reason, when using a fixed-sized array we get a segfault. We're not sure yet if it's a `g++ 9.4.0` specific problem or related to something else in the system configuration but we aren't able to reproduce this in `g++ 9.3.0`.

### DIFF
--- a/cpp/test/device_atomics.cu
+++ b/cpp/test/device_atomics.cu
@@ -15,7 +15,7 @@
  */
 
 #include <algorithm>
-#include <array>
+#include <vector>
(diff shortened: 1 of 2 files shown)


viclafargue (approved):
LGTM

cjnolet:
Closing this PR as we realized the issue is related to incorrect limits flags when launching the docker container.

PR #793, opened 2022-08-25
Complete the deprecation of duplicated hpp headers

ahendriksen:
Replace all .hpp headers that have a .cuh header in the same directory with the
same name by a simple include of the cuh header and a pragma warning of
deprecation.

This change hopefully prevents future head scratching when changes in a
file are seemingly not picked up by the compiler..

Care has been taken to copy the right start year for the copyright line.
Copyright lines have been updated to 2022 when necessary.

The following template has been used for the .hpp header replacement
text:

```
/*
 * %%COPYRIGHT_LINE%%
 *
 * [... snip license .. ]
 */
/**
 * This file is deprecated and will be removed in release 22.06.
 * Please use the cuh version instead.
 */

/**
 * DISCLAIMER: this file is deprecated: use %%CUH_FILE%% instead
 */
#pragma once

#pragma message(__FILE__                                                               \
                " is deprecated and will be removed in a future release." \
                " Please use the cuh version instead.")
```


ahendriksen:
This is a follow up to https://github.com/rapidsai/raft/pull/524

I am note sure that I understood the rationale for duplicating some files and not duplicating others. In any case, I think it would be wise to make sure that there is only one header that contains the actual code. The risk of the .cuh and .hpp headers diverging is quite real. I for one did not know we had full code duplication in some (i.e. 83) headers.

cjnolet:
@ahendriksen this is definitely a welcome change. Thank you for doing this! 

When the original headers were duplicated, we weren't expecting for them to stick around as long as they have. @lowener has graciously started reconciling the raft includes in cuml and related libs so that we can remove the additional headers as a next step.

achirkin:
NB: [cpp/include/raft/spatial/knn/detail/processing.hpp](https://github.com/rapidsai/raft/pull/793/files#diff-a77b7e8bf97c78dbb8776f55bbc070629a84ad0d38719c8f95493d46202b6123) is one exception where we currently need both `.hpp` and `.cuh` files. There may be more, but I'm not aware of others.

ahendriksen:
@achirkin : I will roll back the removal of `processing.hpp`. It looks like this file is only included from `ann_common.h` (which is deprecated), and `processing.cuh`. Is it part of the external interface of RAFT? In that case, should it be moved out of the detail directory?

achirkin:
No, it's not really a part of the external interface, but it's a member of a struct which is currently public but should be deprecated and removed at some point (https://github.com/rapidsai/raft/blob/branch-22.10/cpp/include/raft/spatial/knn/ann_common.h#L40)

ahendriksen:
I found the following files to have non-trivial diffs: 

```
include/raft/linalg/transpose.hpp   (mdspan based transpose only in .cuh header)
raft_linalg_add.hpp                 (more doxygen comments in .cuh header)
spatial/knn/detail/processing.hpp   (as mentioned in discussion above)
```

Repushing.

ahendriksen:
@cjnolet Tamas and Artem have taken a look and I have taken their feedback into account. There are three files where this PR is non-trivial: 

```
include/raft/linalg/transpose.hpp   (mdspan based transpose only in .cuh header)
raft_linalg_add.hpp                 (more doxygen comments in .cuh header)
spatial/knn/detail/processing.hpp   (as mentioned in discussion above)
```

Could you have a look? (no rush, just an update)

cjnolet (approved):
Looks great. Thanks Allard!

cjnolet:
@gpucibot merge

PR #794, opened 2022-08-26
Optimize fusedL2NN when data is skinny

ahendriksen:
The fusedL2NN kernel uses tiling to maximize performance. The current implementation assumes that the input matrices are at least 32 elements wide. When this is not the case, it performs redundant computations. 

This PR adds a policy to apply when the matrix is skinny (less than 32 elements wide). This results in a 1.5 - 2x performance improvement across GPU architectures.


ahendriksen:
rerun tests

cpp/include/raft/distance/detail/fused_l2_nn.cuh
```diff
@@ -92,14 +92,14 @@ DI void updateReducedVal(
...
+  // Update each output row in order within a warp. This will resolve hang
+  // issues with pre-Volta architectures
 #pragma unroll
   for (int j = 0; j < (raft::WarpSize / P::AccThCols); j++) {
-    if (lid == 0) {
+    if (lid == j * P::AccThCols) {
 #pragma unroll
       for (int i = 0; i < P::AccRowsPerTh; ++i) {
-        auto rid = gridStrideY + accrowid + j + i * P::AccThRows;
+        auto rid = gridStrideY + accrowid + i * P::AccThRows;
```
tfeher, line 102: Shouldn't we keep `j` here?

cpp/include/raft/distance/detail/fused_l2_nn.cuh
```diff
@@ -210,8 +202,8 @@ __global__ __launch_bounds__(P::Nthreads, 2) void fusedL2NNkernel(OutT* min,
       for (int i = 0; i < P::AccRowsPerTh; ++i) {
 #pragma unroll
         for (int j = P::AccThCols / 2; j > 0; j >>= 1) {
-          auto tmpkey   = raft::shfl(val[i].key, lid + j);
-          auto tmpvalue = raft::shfl(val[i].value, lid + j);
+          auto tmpkey   = raft::shfl(val[i].key, lid + j, P::AccThCols);
```
tfeher, line 205: Please add a comment why this works:
```
// Actually the srcLane (lid +j) should be (lid +j) % P:AccThCols, but the shlf op applies the module internally.
```

cpp/include/raft/distance/detail/fused_l2_nn.cuh
ahendriksen: No, but it took me a while to figure out why! 

The role of `j` is taken over by `accrowid`, which changes from thread to thread: 
```
const auto accrowid = threadIdx.x / P::AccThCols;
```
Since we are no longer doing all the work in thread 0, we can remove `j`.

tfeher (approved):
Thanks @ahendriksen for addressing the issues. LGTM.

cjnolet:
@gpucibot merge

PR #796, opened 2022-09-01
Minor follow-up fixes for ivf-flat

achirkin:
Small fixes to ivf-flat and its dependency warp_sort:

- fix the template parameter for the call of `calc_smem_size_for_block_wide` (may reduce the use of shared memory);
- force initialize warp_sort internal value buffers to avoid uninitialized output in case of very small input data size;
- small readability fixes.


tfeher (approved):
Thanks Artem for the PR, LGTM.

cjnolet:
@gpucibot merge

PR #797, opened 2022-09-01
[FEA] Statically link all CUDA toolkit libraries

trxcllnt:
This PR ensures RAFT statically links all the CUDA toolkit libraries (not just `cudart`) if a user enables `CUDA_STATIC_RUNTIME`.

### DIFF
--- a/cpp/cmake/thirdparty/get_faiss.cmake
+++ b/cpp/cmake/thirdparty/get_faiss.cmake
@@ -15,7 +15,7 @@
 #=============================================================================
 
 function(find_and_configure_faiss)
-    set(oneValueArgs VERSION PINNED_TAG BUILD_STATIC_LIBS EXCLUDE_FROM_ALL)
+    set(oneValueArgs VERSION REPOSITORY PINNED_TAG BUILD_STATIC_LIBS EXCLUDE_FROM_ALL)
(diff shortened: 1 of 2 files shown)


cpp/cmake/thirdparty/get_faiss.cmake
```diff
@@ -66,7 +68,22 @@ function(find_and_configure_faiss)
     rapids_export_find_package_root(BUILD faiss [=[${CMAKE_CURRENT_LIST_DIR}]=] raft-nn-lib-exports)
 endfunction()
 
+if(NOT RAFT_FAISS_GIT_TAG)
+  # TODO: Remove this once faiss supports FAISS_USE_CUDA_TOOLKIT_STATIC
```
cjnolet, line 72: Is the plan to wait for this before we merge this PR? Or should we create a RAFT issue just to make sure it stays on our radar?

trxcllnt: @cjnolet it's up to you -- I based this branch from the commit sha here. If pointing at my fork is fine with you, then great. Otherwise, it might be worth forking faiss into rapidsai and cherry-picking my commit.

cjnolet (approved):

cjnolet:
@gpucibot merge

PR #801, opened 2022-09-02
Renaming `pyraft` -> `raft-dask`

cjnolet:
`dask-raft` is a more appropriate name. If needed, I will also move the contents of `raft.common` from `dask-raft` over to `pylibraft` and make `pylibraft` a dependency of `dask-raft`. I will do that after seeing who's still using those classes, though.


cjnolet:
These change should not be breaking to downstream projects, however, any projects dependent upon `pyraft` will need to be updated to use `dask-raft` before the 22.10 release.

vyasr:
@cjnolet let me know if you think this PR needs a general review (from a non-raft person) at some point.

cjnolet:
@vyasr sure, if you have some time you are welcome to review this PR.

ajschmidt8 (approved):

ci/gpu/build.sh
```diff
@@ -86,12 +90,12 @@ set +x
...
 nvidia-smi
 
-gpuci_logger "GoogleTest for raft"
+gpuci_logger "GoogleTest for libraft"
 GTEST_OUTPUT="xml:${WORKSPACE}/test-results/raft_cpp/" $CONDA_PREFIX/bin/libraft/gtests/test_raft
 
-gpuci_logger "Python pytest for pyraft"
-cd "$WORKSPACE/python/raft/raft/test"
-pytest --cache-clear --junitxml="$WORKSPACE/junit-pyraft.xml" -v -s
+gpuci_logger "Python pytest for raft-dask"
```
dantegd, line 96: Question a bit unrelated to the PR, wouldn't it be better to run the `pylibraft` pytests first, so that tests run from lower to higher level?

dantegd (approved):
PR and name change look great to me

ci/gpu/build.sh
cjnolet: Ah! That's a good point. Assuming CI Passes, let me merge this guy over and I'll create a follow-on to fix that.

python/raft-dask/CMakeLists.txt
```diff
@@ -43,11 +43,11 @@ if(NOT raft_FOUND)
   # pull in the required languages for the C++ project even if this project
   # does not require those languges.
   include(rapids-cuda)
-  rapids_cuda_init_architectures(pyraft)
+  rapids_cuda_init_architectures(raft-dask)
   enable_language(CUDA)
   # Since pyraft only enables CUDA optionally we need to manually include the file that
```
vyasr, line 48:
```suggestion
  # Since raft-dask only enables CUDA optionally we need to manually include the file that
```

python/raft-dask/CMakeLists.txt
```diff
@@ -43,11 +43,11 @@ if(NOT raft_FOUND)
...
   # rapids_cuda_init_architectures relies on `project` including.
-  include("${CMAKE_PROJECT_pyraft_INCLUDE}")
+  include("${CMAKE_PROJECT_raft_dask_INCLUDE}")
 
   # pyraft doesn't actually use raft libraries, it just needs the headers, so
```
vyasr, line 52: pyraft->raft-dask

python/pylibraft/pyproject.toml
```diff
@@ -20,5 +20,5 @@ requires = [
     "cython>=0.29,<0.30",
     "scikit-build>=0.13.1",
     "cmake>=3.20.1,!=3.23.0",
-    "ninja",
+    "ninja"
```
vyasr, line 23: Did a linter complain here?

ci/cpu/upload.sh
```diff
@@ -38,12 +38,12 @@ if [[ "$BUILD_LIBRAFT" == "1" && "$UPLOAD_LIBRAFT" == "1" ]]; then
 fi
 
 if [[ "$BUILD_RAFT" == "1" && "$UPLOAD_RAFT" == "1" ]]; then
-  PYRAFT_FILE=$(conda build --no-build-id --croot ${CONDA_BLD_DIR} -c ${CONDA_LOCAL_CHANNEL} conda/recipes/pyraft --python=$PYTHON --output)
+  DASKRAFT_FILE=$(conda build --no-build-id --croot ${CONDA_BLD_DIR} -c ${CONDA_LOCAL_CHANNEL} conda/recipes/raft-dask --python=$PYTHON --output)
```
vyasr, line 41:
```suggestion
  DASK_RAFT_FILE=$(conda build --no-build-id --croot ${CONDA_BLD_DIR} -c ${CONDA_LOCAL_CHANNEL} conda/recipes/raft-dask --python=$PYTHON --output)
```

ci/cpu/upload.sh
```diff
@@ -38,12 +38,12 @@ if [[ "$BUILD_LIBRAFT" == "1" && "$UPLOAD_LIBRAFT" == "1" ]]; then
...
   PYLIBRAFT_FILE=$(conda build --no-build-id --croot ${CONDA_BLD_DIR} -c ${CONDA_LOCAL_CHANNEL} conda/recipes/pylibraft --python=$PYTHON --output)
-  test -e ${PYRAFT_FILE}
-  echo "Upload pyraft"
-  echo ${PYRAFT_FILE}
-  gpuci_retry anaconda -t ${MY_UPLOAD_KEY} upload -u ${CONDA_USERNAME:-rapidsai} ${LABEL_OPTION} --skip-existing ${PYRAFT_FILE} --no-progress
+  test -e ${DASKRAFT_FILE}
```
vyasr, line 43:
```suggestion
  test -e ${DASK_RAFT_FILE}
```

ci/cpu/upload.sh
```diff
@@ -38,12 +38,12 @@ if [[ "$BUILD_LIBRAFT" == "1" && "$UPLOAD_LIBRAFT" == "1" ]]; then
...
+  echo "Upload raft-dask"
+  echo ${DASKRAFT_FILE}
```
vyasr, line 45:
```suggestion
  echo ${DASK_RAFT_FILE}
```

ci/cpu/upload.sh
```diff
@@ -38,12 +38,12 @@ if [[ "$BUILD_LIBRAFT" == "1" && "$UPLOAD_LIBRAFT" == "1" ]]; then
...
+  gpuci_retry anaconda -t ${MY_UPLOAD_KEY} upload -u ${CONDA_USERNAME:-rapidsai} ${LABEL_OPTION} --skip-existing ${DASKRAFT_FILE} --no-progress
```
vyasr, line 46:
```suggestion
  gpuci_retry anaconda -t ${MY_UPLOAD_KEY} upload -u ${CONDA_USERNAME:-rapidsai} ${LABEL_OPTION} --skip-existing ${DASK_RAFT_FILE} --no-progress
```

ci/gpu/build.sh
```diff
@@ -63,9 +63,13 @@ conda list --show-channel-urls
 gpuci_logger "Build and install Python targets"
 CONDA_BLD_DIR="$WORKSPACE/.conda-bld"
 gpuci_mamba_retry install boa
-gpuci_conda_retry mambabuild --no-build-id --croot "${CONDA_BLD_DIR}" conda/recipes/pyraft -c "${CONDA_ARTIFACT_PATH}" --python="${PYTHON}"
+
+# Install pylibraft first since it's a dependency of raft-dask
 gpuci_conda_retry mambabuild --no-build-id --croot "${CONDA_BLD_DIR}" conda/recipes/pylibraft -c "${CONDA_ARTIFACT_PATH}" --python="${PYTHON}"
-gpuci_mamba_retry install -y -c "${CONDA_BLD_DIR}" -c "${CONDA_ARTIFACT_PATH}" pyraft pylibraft
+gpuci_mamba_retry install -y -c "${CONDA_BLD_DIR}" -c "${CONDA_ARTIFACT_PATH}" pylibraft
```
vyasr, line 69: This should just be installing the conda package that was already built on the previous (including before this PR), correct?

python/pylibraft/pylibraft/common/interruptible.pxd
```diff
@@ -22,11 +22,11 @@
 from libcpp.memory cimport shared_ptr
 from rmm._lib.cuda_stream_view cimport cuda_stream_view
 
-cdef extern from "raft/interruptible.hpp" namespace "raft" nogil:
+cdef extern from "raft/core/interruptible.hpp" namespace "raft" nogil:
```
vyasr, line 25: Were these just errors before? It looks like one file just includes the other, what's the difference between the core and non-core file?

python/pylibraft/pylibraft/common/handle.pxd
```diff
@@ -33,3 +34,8 @@ cdef extern from "raft/core/handle.hpp" namespace "raft" nogil:
                  shared_ptr[cuda_stream_pool] stream_pool) except +
         cuda_stream_view get_stream() except +
         void sync_stream() except +
+
+cdef class Handle:
```
vyasr, line 38: It looks like pylibraft already had this in the pyx file. Were these just untyped Python objects before, then? I'm sort of surprised that worked, am I missing something?

vyasr (changes requested):
A few minor comments to address. You should also update the `update-version.sh` script.

ci/gpu/build.sh
cjnolet: Unfortunately, there was a long-time issue w/ using the pylibraft that was built in the previous step. I can't recall exactly what that issue was but it does predate this PR (and I consider it out of scope to fix here).

python/pylibraft/pylibraft/common/interruptible.pxd
cjnolet: We've been moving things around to establish a formal public API so the non-core version of this file is deprecated. It just makes sense to fix the path in this PR. In general, files in `core/` are safe to use in headers of downstream APIs as they won't bring in additional dependencies outside of the cuda toolkit. This isn't quite true for 1 or 2 of the files but we're working on it.

python/pylibraft/pylibraft/common/handle.pxd
cjnolet: We just weren't using the `Handle` before, so we didn't explicitly need to build it.

vyasr (approved):
LGTM

cjnolet:
@gpucibot merge

Issue #535, opened 2022-03-01
[FEA] Implement new API functions using `mdspan` 


cjnolet:
Now that we have `mdspan` and `mdarray` in RAFT, we are able to start implementing our new public API contract facades with them. 

To start, we can add new functions that assume the underlying layouts are contiguous, deprecate the old functions, and then as we iterate on the `mdspan` and `mdarray` and use them in more in device functions, we can eventually support non-contiguous layouts. 

Ideally we would be consistent about the ordering of arguments to the new functions. We should start using `std::optional` for any optional arguments (instead of null pointers) and perhaps follow Thrust's format of accepting inputs before outputs (raft handle should always be the first argument). 

Namespaces to add the new functions:
- [x] raft::cluster (Mickael)
- [x] raft::distance (Corey)
- [x] raft::random (Mark)
- [x] raft::linalg (Divye)
- [X] raft::matrix (Corey) 
- [x] raft::stats (Mickael)
- [x] raft::spatial::knn (Corey) 

It might also make sense to break these up so different people can take on a couple namespaces. This could help 1) more developers getting quickly spun up on mdarray/mdspan, and 2) getting these knocked out more quickly while folks are able to work on different parts concurrently. 

This issue will also close #360

PR #802, opened 2022-09-05
Implement `raft::stats` API with mdspan

lowener:
Part of #535.
Implementation of the raft::stats API with mdspan, with the C++ tests
14/22 Files implemented. The remaining files will come in a following PR.


cpp/include/raft/stats/accuracy.cuh
```diff
@@ -39,6 +40,29 @@ float accuracy(const math_t* predictions, const math_t* ref_predictions, int n,
...
+ *                        device.
+ * @param[in] handle: the raft handle.
+ * @param[in] predictions: array of predictions (GPU pointer).
+ * @param[in] ref_predictions: array of reference (ground-truth) predictions (GPU pointer).
+ * @return: Accuracy score in [0, 1]; higher is better.
+ */
+template <typename math_t, typename IdxType, typename LayoutPolicy, typename AccessorPolicy>
+float accuracy(
+  const raft::handle_t& handle,
+  raft::mdspan<math_t, raft::vector_extent<IdxType>, LayoutPolicy, AccessorPolicy> predictions,
```
mhoemmen, line 58: It looks like the mdspan are read-only, so would you consider making them views of const?  For example:

```c++
raft::mdspan<const math_t, raft::vector_extent<IdxType>, LayoutPolicy, AccessorPolicy> predictions,
```

(This suggestion generally applies to all the functions in this PR.)

cpp/include/raft/stats/accuracy.cuh
```diff
@@ -39,6 +40,29 @@ float accuracy(const math_t* predictions, const math_t* ref_predictions, int n,
...
+ * @tparam IdxType Index type of matrix extent.
+ * @tparam LayoutPolicy Layout type of the input data.
+ * @tparam AccessorPolicy Accessor for the input and output, must be valid accessor on
+ *                        device.
+ * @param[in] handle: the raft handle.
+ * @param[in] predictions: array of predictions (GPU pointer).
+ * @param[in] ref_predictions: array of reference (ground-truth) predictions (GPU pointer).
+ * @return: Accuracy score in [0, 1]; higher is better.
+ */
+template <typename math_t, typename IdxType, typename LayoutPolicy, typename AccessorPolicy>
```
mhoemmen, line 55: The two `mdspan` could have different layouts and/or accessors, so would you consider separate layout and accessor template parameters for each?  (This suggestion generally applies to all the functions in this PR.)

cpp/include/raft/stats/accuracy.cuh
```diff
@@ -39,6 +40,29 @@ float accuracy(const math_t* predictions, const math_t* ref_predictions, int n,
...
+float accuracy(
+  const raft::handle_t& handle,
+  raft::mdspan<math_t, raft::vector_extent<IdxType>, LayoutPolicy, AccessorPolicy> predictions,
+  raft::mdspan<math_t, raft::vector_extent<IdxType>, LayoutPolicy, AccessorPolicy> ref_predictions)
+{
+  return detail::accuracy_score(predictions.data_handle(),
```
mhoemmen, line 61: Would you consider first `static_assert`ing that the layouts are always exhaustive, strided, and unique, before handing them off to a function that assumes these properties?

(This suggestion generally applies to all the functions in this PR.)

cpp/include/raft/stats/contingency_matrix.cuh
```diff
@@ -40,6 +41,29 @@ void getInputClassCardinality(
...
+ * @param groundTruth: device 1-d array for ground truth (num of rows)
+ * @param nSamples: number of elements in input array
+ * @param minLabel: [out] calculated min value in input array
+ * @param maxLabel: [out] calculated max value in input array
+ */
+template <typename T, typename IdxType, typename LayoutPolicy, typename AccessorPolicy>
+void getInputClassCardinality(
+  const raft::handle_t& handle,
+  raft::mdspan<T, raft::vector_extent<IdxType>, LayoutPolicy, AccessorPolicy> groundTruth,
+  const raft::host_scalar_view<T>& minLabel,
```
mhoemmen, line 57: Given that `host_scalar_view` is an mdspan, would you consider passing it by value instead of const reference, since mdspan is a view and views are idiomatically passed by value?

(This suggestion generally applies to all the functions in this PR.)

cpp/include/raft/stats/contingency_matrix.cuh
```diff
@@ -61,6 +85,30 @@ size_t getContingencyMatrixWorkspaceSize(int nSamples,
...
+template <typename T,
+          typename OutT = int,
+          typename IdxType,
+          typename LayoutPolicy,
+          typename AccessorPolicy>
+size_t getContingencyMatrixWorkspaceSize(
+  const raft::handle_t& handle,
+  raft::mdspan<T, raft::vector_extent<IdxType>, LayoutPolicy, AccessorPolicy> groundTruth,
+  T minLabel = std::numeric_limits<T>::max(),
+  T maxLabel = std::numeric_limits<T>::max())
```
mhoemmen, line 106: This doesn't seem quite right -- should the `maxLabel`'s default value be `numeric_limits<T>::min()`, so that everything else is greater than it?

cpp/include/raft/stats/contingency_matrix.cuh
```diff
@@ -100,6 +148,52 @@ void contingencyMatrix(const T* groundTruth,
...
+ * @param groundTruth: device 1-d array for ground truth (num of rows)
+ * @param predictedLabel: device 1-d array for prediction (num of columns)
+ * @param outMat: output buffer for contingecy matrix
+ * @param workspace: Optional, workspace memory allocation
+ * @param workspaceSize: Optional, size of workspace memory
+ * @param minLabel: Optional, min value in input ground truth array
+ * @param maxLabel: Optional, max value in input ground truth array
+ */
+template <typename T,
+          typename OutT = int,
```
mhoemmen, line 172: I don't think you need or even want a default value for `OutT`, given that the function can deduce it from `outMat`.

cpp/include/raft/stats/contingency_matrix.cuh
```diff
@@ -100,6 +148,52 @@ void contingencyMatrix(const T* groundTruth,
...
+          typename IdxType,
+          typename LayoutPolicy,
+          typename AccessorPolicy>
+void contingencyMatrix(
+  const raft::handle_t& handle,
+  raft::mdspan<T, raft::vector_extent<IdxType>, LayoutPolicy, AccessorPolicy> groundTruth,
+  raft::mdspan<T, raft::vector_extent<IdxType>, LayoutPolicy, AccessorPolicy> predictedLabel,
+  raft::mdspan<OutT, raft::matrix_extent<IdxType>, LayoutPolicy, AccessorPolicy> outMat,
+  void* workspace      = nullptr,
```
mhoemmen, line 181: If the `workspace` will always be treated as an array of `T` (or `OutT`), then would you consider giving it that appropriate type (at least `T*`, preferably an `mdspan` of `T`)?

Otherwise, if the `workspace` is just a bunch of contiguous bytes, then would you consider making it an mdspan or span of `char` instead of a `void*`?

Using mdspan or span here would let you get rid of the `workspaceSize` parameter.

cpp/include/raft/stats/cov.cuh
```diff
@@ -57,6 +58,47 @@ void cov(const raft::handle_t& handle,
...
+         bool sample,
+         bool stable)
+{
+  detail::cov(handle,
+              covar.data_handle(),
+              data.data_handle(),
+              mu.data_handle(),
+              data.extent(1),
+              data.extent(0),
+              std::is_same_v<LayoutPolicy, raft::row_major>,
```
mhoemmen, line 97: Given that it looks like this implementation only supports `row_major` or `col_major`, would you consider `static_assert`ing that `LayoutPolicy` is one of these?

cpp/include/raft/stats/dispersion.cuh
```diff
@@ -51,10 +53,53 @@ DataT dispersion(const DataT* centroids,
...
+ */
+template <typename DataT,
+          typename IdxType = int,
+          typename LayoutPolicy,
+          typename AccessorPolicy,
+          int TPB = 256>
+DataT dispersion(
+  const raft::handle_t& handle,
+  raft::mdspan<DataT, raft::matrix_extent<IdxType>, LayoutPolicy, AccessorPolicy> centroids,
+  raft::mdspan<IdxType, raft::vector_extent<IdxType>> clusterSizes,
```
mhoemmen, line 87: Is it intentional that this uses the default accessor policy?

cpp/include/raft/stats/histogram.cuh
```diff
@@ -42,7 +43,7 @@ namespace stats {
  * @param nbins number of bins
  * @param data input data (length = ncols * nrows)
  * @param nrows data array length in each column (or batch)
- * @param ncols number of columsn (or batch size)
+ * @param ncols number of columns (or batch size)
```
mhoemmen, line 46: Nice catch! : - )

cpp/include/raft/stats/histogram.cuh
```diff
@@ -61,6 +62,44 @@ void histogram(HistType type,
...
+ * @note signature of BinnerOp is `int func(DataT, IdxT);`
+ */
+template <typename DataT,
+          typename IdxType  = int,
+          typename BinnerOp = IdentityBinner<DataT, IdxType>,
+          typename LayoutPolicy,
+          typename AccessorPolicy>
+void histogram(const raft::handle_t& handle,
+               HistType type,
+               raft::mdspan<int, raft::matrix_extent<IdxType>, LayoutPolicy, AccessorPolicy> bins,
```
mhoemmen, line 90: Given that `bins` is an mdspan of `int`, it looks like the implementation currently only knows how to deal with `int` indices.  Thus, would you consider either requiring that `IdxType` is `int`, or checking (possibly at run time) whether the arrays' extents fit in `int`?

cpp/include/raft/stats/mean.cuh
```diff
@@ -50,6 +50,40 @@ void mean(
...
+ * @param mu: the output mean vector
+ * @param data: the input matrix
+ * @param sample: whether to evaluate sample mean or not. In other words, whether
+ *   to normalize the output using N-1 or N, for true or false, respectively
+ */
+template <typename Type,
+          typename IdxType      = int,
+          typename LayoutPolicy = raft::row_major,
+          typename AccessorPolicy>
+void mean(const raft::handle_t& handle,
```
mhoemmen, line 73: Please see my remarks on `histogram` about the `IdxType`; thanks!

cpp/test/stats/dispersion.cu
```diff
@@ -16,6 +16,7 @@
 
 #include "../test_utils.h"
 #include <gtest/gtest.h>
+#include <optional>
```
mhoemmen, line 19: `optional` requires C++17 -- which is fine with me : - ) -- but just want to make sure it's OK to use.

cpp/include/raft/stats/minmax.cuh
```diff
@@ -68,6 +70,73 @@ void minmax(const T* data,
...
+ * @note This method makes the following assumptions:
+ * 1. input and output matrices are assumed to be col-major
+ * 2. ncols is small enough to fit the whole of min/max values across all cols
+ *    in shared memory
+ */
+template <typename T,
+          typename IdxType,
+          typename LayoutPolicy,
+          typename AccessorPolicy,
+          int TPB = 512>
```
mhoemmen, line 103: Would you consider letting users pass in the value of `TPB` as a `std::integral_constant<int, TPB>` parameter?  That way, users wouldn't have to call this function as `minmax<TPB>(...)`, and they could pass down the integral constant through a chain of nested function calls.

```
template <typename T,
          typename IdxType,
          typename LayoutPolicy,
          typename AccessorPolicy,
          int TPB = 512>
Styling with Markdown is supported
void minmax(const raft::handle_t& handle,
  /* the other parameters */,
  std::integral_constant<int, TPB> /* threads_per_block */);
```

cpp/include/raft/stats/accuracy.cuh
mhoemmen: Thanks!  I should perhaps have also written that if we're calling into some existing raw array code, then we're probably baking in the assumption of layout left or layout right.

It's OK to use `.size()` here as it's the product of the one extent, but normally I would spell this `.extent(0)` instead of `.size()`.

cpp/include/raft/stats/accuracy.cuh
```diff
@@ -42,25 +42,25 @@ float accuracy(const math_t* predictions, const math_t* ref_predictions, int n,
...
+               raft::device_vector_view<const DataT, IdxType> predictions,
+               raft::device_vector_view<const DataT, IdxType> ref_predictions)
 {
+  RAFT_EXPECTS(predictions.size() == ref_predictions.size(), "Size mismatch");
+  RAFT_EXPECTS(predictions.is_exhaustive(), "predictions must be contiguous");
+  RAFT_EXPECTS(ref_predictions.is_exhaustive(), "ref_predictions must be contiguous");
+
   return detail::accuracy_score(predictions.data_handle(),
                                 ref_predictions.data_handle(),
-                                predictions.extent(0),
```
mhoemmen: `.extent(0)` is really the right way to spell this, though `.size()` is not wrong here (as it's the product of the extents, and there's only one extent here).

cpp/include/raft/stats/contingency_matrix.cuh
```diff
@@ -153,45 +146,53 @@ void contingencyMatrix(const T* groundTruth,
...
+ * @tparam OutType output matrix type
  * @tparam IdxType Index type of matrix extent.
  * @tparam LayoutPolicy Layout type of the input data.
- * @tparam AccessorPolicy Accessor for the input and output, must be valid accessor on
- *                        device.
  * @param handle: the raft handle.
  * @param groundTruth: device 1-d array for ground truth (num of rows)
  * @param predictedLabel: device 1-d array for prediction (num of columns)
- * @param outMat: output buffer for contingecy matrix
+ * @param outMat: output buffer for contingency matrix
```
mhoemmen, line 156: Nice catch! : - )

cpp/include/raft/stats/contingency_matrix.cuh
```diff
@@ -153,45 +146,53 @@ void contingencyMatrix(const T* groundTruth,
...
-void contingencyMatrix(
-  const raft::handle_t& handle,
-  raft::mdspan<T, raft::vector_extent<IdxType>, LayoutPolicy, AccessorPolicy> groundTruth,
-  raft::mdspan<T, raft::vector_extent<IdxType>, LayoutPolicy, AccessorPolicy> predictedLabel,
-  raft::mdspan<OutT, raft::matrix_extent<IdxType>, LayoutPolicy, AccessorPolicy> outMat,
-  void* workspace      = nullptr,
-  size_t workspaceSize = 0,
-  T minLabel           = std::numeric_limits<T>::max(),
-  T maxLabel           = std::numeric_limits<T>::max())
+          typename WorkspaceType,
```
mhoemmen, line 165: Given that the code assumes `WorkspaceType` is some kind of mdspan, would you consider just spelling that out in its type?  `workspace` would just be `std::optional<raft::device_vector_view<DataT, IdxType>>`.  That way, you wouldn't even need a template parameter.

If you want to say that `WorkspaceType` could be any device-accessible mdspan, you could use `mdspan<DataT, extents<WorkspaceIndexType, WorkspaceExtent>, layout_right, Accessor>`.  For rank-1 mdspan, `layout_left` can be converted to `layout_right` and vice versa.  Then, just constrain the Accessor to be device accessible, using RAFT's traits.

cpp/include/raft/stats/contingency_matrix.cuh
lowener: It's used as a special value in the detail function: if one of those label is `numeric_limits<T>::max()`, then both `minLabels` and `maxLabels` are recomputed.
Instead of using this behaviour on the public API I put this parameter as optional now.

cpp/include/raft/stats/contingency_matrix.cuh
```diff
@@ -95,11 +101,18 @@ size_t getContingencyMatrixWorkspaceSize(int nSamples,
 template <typename DataT, typename IdxType>
 size_t getContingencyMatrixWorkspaceSize(const raft::handle_t& handle,
                                          raft::device_vector_view<const DataT, IdxType> groundTruth,
-                                         DataT minLabel = std::numeric_limits<DataT>::max(),
-                                         DataT maxLabel = std::numeric_limits<DataT>::max())
+                                         std::optional<DataT> minLabel = std::nullopt,
```
mhoemmen, line 104: Note that if users provide `nullopt` explicitly as an argument, sometimes the compiler has trouble deducing types. It can be worth adding a test to make sure that compiles. If not, there are a couple different ways to fix it. My PR #830 takes one approach that can handle multiple optional arguments.

cpp/include/raft/stats/contingency_matrix.cuh
```diff
@@ -61,6 +89,32 @@ size_t getContingencyMatrixWorkspaceSize(int nSamples,
...
+ * @brief Calculate workspace size for running contingency matrix calculations
+ * @tparam DataT label type
+ * @tparam IdxType Index type of matrix extent.
+ * @param handle: the raft handle.
+ * @param groundTruth: device 1-d array for ground truth (num of rows)
+ * @param minLabel: Optional, min value in input array
+ * @param maxLabel: Optional, max value in input array
+ */
+template <typename DataT, typename IdxType>
+size_t getContingencyMatrixWorkspaceSize(const raft::handle_t& handle,
```
cjnolet, line 102: We should not expose workspaces to the public API. The original functions were written before we had RMM (and could just use a pool allocator).

cpp/include/raft/stats/contingency_matrix.cuh
```diff
@@ -100,6 +154,63 @@ void contingencyMatrix(const T* groundTruth,
...
+ * @param workspace: Optional, workspace memory allocation
+ * @param minLabel: Optional, min value in input ground truth array
+ * @param maxLabel: Optional, max value in input ground truth array
+ */
+template <typename DataT,
+          typename OutType,
+          typename IdxType,
+          typename LayoutPolicy,
+          typename WorkspaceDataType>
+void contingencyMatrix(
```
cjnolet, line 179: Use snake case everywhere for consistency (eg `contingency_matrix`). This applies to function names, arguments. Also templates if there's time but we can fix those later if needed.

cpp/include/raft/stats/cov.cuh
```diff
@@ -89,7 +90,8 @@ void cov(const raft::handle_t& handle,
   static_assert(
     std::is_same_v<LayoutPolicy, raft::row_major> || std::is_same_v<LayoutPolicy, raft::col_major>,
     "Data layout not supported");
-  RAFT_EXPECTS(data.size() == covar.size(), "Size mismatch");
+  RAFT_EXPECTS(data.extent(1) == covar.extent(0) && data.extent(1) == covar.extent(1),
```
mhoemmen, line 93: `data.extents() == covar.extents()` works too!

cpp/include/raft/stats/mean_center.cuh
```diff
@@ -51,6 +52,44 @@ void meanCenter(Type* out,
...
+ * @tparam LayoutPolicy Layout type of the input matrix.
+ * @tparam TPB threads per block of the cuda kernel launched
+ * @param handle the raft handle
+ * @param data input matrix
+ * @param mu the mean vector
+ * @param out the output mean-centered matrix
+ * @param bcastAlongRows whether to broadcast vector along rows or columns
+ */
+template <typename DataT, typename IdxType, typename LayoutPolicy, int TPB = 256>
+void meanCenter(const raft::handle_t& handle,
```
cjnolet, line 68: Use snake case everywhere

cpp/include/raft/stats/mean_center.cuh
```diff
@@ -51,6 +52,44 @@ void meanCenter(Type* out,
...
+ * @tparam IdxType index type
+ * @tparam LayoutPolicy Layout type of the input matrix.
+ * @tparam TPB threads per block of the cuda kernel launched
+ * @param handle the raft handle
+ * @param data input matrix
+ * @param mu the mean vector
+ * @param out the output mean-centered matrix
+ * @param bcastAlongRows whether to broadcast vector along rows or columns
+ */
+template <typename DataT, typename IdxType, typename LayoutPolicy, int TPB = 256>
```
cjnolet, line 67: We shouldn't be exposing any cuda-specific templates so we can remove TPB from the public API

cjnolet (changes requested):
PR is looking good so far. Aside from Divye and Mark's reviews, mine are mostly about consistency with the other APIs

cpp/include/raft/stats/cov.cuh
lowener: Not for this one because covar is `ncols*ncols` while data is `nrows*ncols`

cpp/include/raft/stats/contingency_matrix.cuh
```diff
@@ -40,6 +43,31 @@ void getInputClassCardinality(
...
+ * size of matrix = (maxLabel - minLabel + 1)^2 * sizeof(int)
+ * @tparam value_t label type
+ * @tparam idx_t Index type of matrix extent.
+ * @param handle: the raft handle.
+ * @param groundTruth: device 1-d array for ground truth (num of rows)
+ * @param minLabel: [out] calculated min value in input array
+ * @param maxLabel: [out] calculated max value in input array
+ */
+template <typename value_t, typename idx_t>
+void getInputClassCardinality(const raft::handle_t& handle,
```
cjnolet, line 57: All new functions should be snake cased.

cpp/include/raft/stats/contingency_matrix.cuh
cjnolet: This one still needs to be removed from the public API. Rather than having `contingency_matrix` accept a workspace, please have it compute and allocate the workspace internally. Eventually we will have an optional rmm device memory resources attached to the raft handle that can be used to override the allocation strategy.

cpp/include/raft/stats/r2_score.cuh
```diff
@@ -45,6 +46,33 @@ math_t r2_score(math_t* y, math_t* y_hat, int n, cudaStream_t stream)
...
+ * normalizing the sum of squared errors by the total sum of squares.
+ *
+ * This score indicates the proportionate amount of variation in an
+ * expected response variable is explained by the independent variables
+ * in a linear regression model. The larger the R-squared value, the
+ * more variability is explained by the linear regression model.
+ *
+ * @tparam value_t the data type
+ * @tparam idx_t index type
+ * @param handle the raft handle
```
cjnolet, line 60: Please add `[in]`, `[inout]` and `[out]` to the docs for all params (applies for all changes in this PR).

cpp/include/raft/stats/trustworthiness_score.cuh
```diff
@@ -48,6 +50,37 @@ double trustworthiness_score(const raft::handle_t& h,
...
+ * @param[in] X: Data in original dimension
+ * @param[in] X_embedded: Data in target dimension (embedding)
+ * @param[in] n_neighbors Number of neighbors considered by trustworthiness score
+ * @param[in] batch_size Batch size
+ * @return[out] Trustworthiness score
+ */
+template <raft::distance::DistanceType distance_type, typename value_t, typename idx_t>
+double trustworthiness_score(const raft::handle_t& handle,
+                             raft::device_matrix_view<const value_t, idx_t, raft::row_major> X,
+                             raft::device_matrix_view<value_t, idx_t, raft::row_major> X_embedded,
```
cjnolet, line 68: `X_embedded` is an input as well, which I don't think should be modified in this function. We shouldmark that `const value_t` as well.

cpp/include/raft/stats/contingency_matrix.cuh
```diff
@@ -40,6 +43,31 @@ void getInputClassCardinality(
   detail::getInputClassCardinality(groundTruth, nSamples, stream, minLabel, maxLabel);
 }
 
+/**
+ * @brief use this to allocate output matrix size
+ * size of matrix = (maxLabel - minLabel + 1)^2 * sizeof(int)
+ * @tparam value_t label type
+ * @tparam idx_t Index type of matrix extent.
+ * @param handle: the raft handle.
+ * @param groundTruth: device 1-d array for ground truth (num of rows)
```
cjnolet, line 52: Add `[in]` to `groundTruth` and handle.

cjnolet (changes requested):
This is getting closer!

cpp/include/raft/stats/cov.cuh
mhoemmen: Oh oof, sorry!  Yes, for next time : - )

cpp/include/raft/stats/trustworthiness_score.cuh
lowener: `X_embedded` is used as input to run a kNN, which is not taking const inputs as well, so it would require some changes on the `raft::spatial::knn`

cjnolet: For now, you can accept the input as `const` and then do a `const_cast` inside the public API function (with a TODO and a note). The goal here is to make sure these arguments are sending the right message to the user, even if the data ends up being modified slightly in place (and then reverted back).

cjnolet:
rerun tests

cpp/include/raft/stats/contingency_matrix.cuh
```diff
@@ -19,6 +19,10 @@
 
 #pragma once
 
+#include <raft/core/device_mdarray.hpp>
```
cjnolet, line 22: I don't think we need the `device_mdarray` includes in any  of these files unless they are explicitly calling `make_device_*` (without `_view`).

cpp/include/raft/stats/cov.cuh
```diff
@@ -57,6 +58,55 @@ void cov(const raft::handle_t& handle,
...
+ * @param[in]  stable whether to run the slower-but-numerically-stable version or not
+ * @note if stable=true, then the input data will be mean centered after this
+ * function returns!
+ */
+template <typename value_t, typename idx_t, typename layout_t>
+void cov(const raft::handle_t& handle,
+         raft::device_matrix_view<value_t, idx_t, layout_t> data,
+         raft::device_vector_view<const value_t, idx_t> mu,
+         raft::device_matrix_view<value_t, idx_t, layout_t> covar,
+         bool sample,
```
cjnolet, line 87: Having these two booleans one right after the other is a bit confusing. I guess I can live with this , but I wonder if maybe we should add another overload (maybe `cov` and `cov_stable`?)

cpp/include/raft/stats/dispersion.cuh
```diff
@@ -51,10 +53,51 @@ DataT dispersion(const DataT* centroids,
...
+ *   and of dimension (n_clusters x dim)
+ * @param[in]  cluster_sizes number of points in the dataset which belong to each
+ *   cluster. This is of length n_clusters
+ * @param[out] global_centroid compute the global weighted centroid of all cluster
+ *   centroids. This is of length dim. Use std::nullopt to not return it.
+ * @param[in]  n_points number of points in the dataset
+ * @return the cluster dispersion value
+ */
+template <typename value_t, typename idx_t>
+value_t dispersion(const raft::handle_t& handle,
```
cjnolet, line 76: Now might be a good time to rename this to cluster_dispersion (there are many types of dispersion measures, and also many types of cluster dispersion measures). We probably also want to spell out the math in the description so a user knows when to use this (the use of the term "centroids" already reduces the usefulness of this with an algorithm that assume gaussian distributed data- such as most density-based methods).

cpp/include/raft/stats/histogram.cuh
```diff
@@ -61,6 +62,44 @@ void histogram(HistType type,
...
+ * @param[in]  handle the raft handle
+ * @param[in]  type histogram implementation type to choose
+ * @param[in]  data input data col-major (length = nrows * ncols)
+ * @param[out] bins the output bins col-major (length = nbins * ncols)
+ * @param[in]  binner the operation that computes the bin index of the input data
+ *
+ * @note signature of binner_op is `int func(value_t, IdxT);`
+ */
+template <typename value_t,
+          typename idx_t     = int,
```
cjnolet, line 81: We should consider removing the default value_t and idx_t in all of these functions.

cpp/include/raft/stats/r2_score.cuh
```diff
@@ -45,6 +46,38 @@ math_t r2_score(math_t* y, math_t* y_hat, int n, cudaStream_t stream)
...
+ * more variability is explained by the linear regression model.
+ *
+ * @tparam value_t the data type
+ * @tparam idx_t index type
+ * @param[in] handle the raft handle
+ * @param[in] y: Array of ground-truth response variables
+ * @param[in] y_hat: Array of predicted response variables
+ * @return: The R-squared value.
+ * @note The constness of y and y_hat is currently casted away.
+ * TODO: Change the underlying implementation to remove the need to const_cast
```
cjnolet, line 65: Let's add the todo inside the function's implementation so user's don't mistakenly see it in the docs.

cpp/include/raft/stats/trustworthiness_score.cuh
```diff
@@ -48,6 +50,42 @@ double trustworthiness_score(const raft::handle_t& h,
...
+ * @tparam idx_t Integer type used to for addressing
+ * @param[in] handle the raft handle
+ * @param[in] X: Data in original dimension
+ * @param[in] X_embedded: Data in target dimension (embedding)
+ * @param[in] n_neighbors Number of neighbors considered by trustworthiness score
+ * @param[in] batch_size Batch size
+ * @return Trustworthiness score
+ * @note The constness of the data in X_embedded is currently casted away and the data is slightly
+ * modified.
+ * TODO: Change the underlying implementation to remove the need to const_cast X_embedded.
```
cjnolet, line 66: Same as above, move the todo inside the implementation. We want the docs to be strictly for users

cpp/include/raft/stats/weighted_mean.cuh
```diff
@@ -91,6 +92,89 @@ void colWeightedMean(
...
+ *
+ * @tparam value_t the data type
+ * @tparam idx_t Integer type used to for addressing
+ * @param[in]  handle the raft handle
+ * @param[in]  data the input matrix of size nrows * ncols
+ * @param[in]  weights per-col weight
+ * @param[out] mu the output mean vector of size ncols
+ */
+template <typename value_t, typename idx_t = int>
+void rowWeightedMean(const raft::handle_t& handle,
```
cjnolet, line 151: These should be `row_weighted_mean` and `col_weighted_mean` for consistency.

cjnolet (changes requested):
Looking good! This should be my last set of requested changes, I believe.

cpp/include/raft/stats/contingency_matrix.cuh
lowener: `make_device_vector` is used here in this file to allocate a workspace.

cpp/include/raft/stats/mean_center.cuh
```diff
@@ -78,6 +115,42 @@ void meanAdd(Type* out,
...
+              raft::device_vector_view<const value_t, idx_t> mu,
+              raft::device_matrix_view<value_t, idx_t, layout_t> out,
+              bool bcast_along_rows)
+{
+  static_assert(
+    std::is_same_v<layout_t, raft::row_major> || std::is_same_v<layout_t, raft::col_major>,
+    "Data layout not supported");
+  auto mean_vec_size = bcast_along_rows ? data.extent(1) : data.extent(0);
+  RAFT_EXPECTS(out.size() == data.size(), "Size mismatch");
+  RAFT_EXPECTS(mean_vec_size == mu.extent(0), "Size mismatch betwen data and mu");
```
cjnolet, line 142: I've been taking `bcast_along_rows` to mean that the vector must be the number of columns if `bcast_along_rows==false` (e.g bcast_across_rows) and the vector must be the number of rows if `bcast_along_rows==true`.

cpp/include/raft/stats/weighted_mean.cuh
```diff
@@ -91,6 +92,89 @@ void colWeightedMean(
...
+  constexpr bool is_row_major = std::is_same_v<layout_t, raft::row_major>;
+  constexpr bool is_col_major = std::is_same_v<layout_t, raft::col_major>;
+  static_assert(is_row_major || is_col_major,
+                "weighted_mean: Layout must be either "
+                "raft::row_major or raft::col_major (or one of their aliases)");
+  auto mean_vec_size = along_rows ? data.extent(1) : data.extent(0);
+
+  RAFT_EXPECTS(weights.extent(0) == mean_vec_size,
+               "Size mismatch betwen weights and mean_vec_size");
+  RAFT_EXPECTS(mu.extent(0) == mean_vec_size, "Size mismatch betwen mu and mean_vec_size");
```
cjnolet, line 125: Same here, shouldn't `mu` be the number of columns if `along_rows==true` and the number of rows if `along_rows==false`?

cpp/test/stats/v_measure.cu
```diff
@@ -93,19 +93,17 @@ class vMeasureTest : public ::testing::TestWithParam<vMeasureParam> {
...
+      lowerLabelRange,
+      upperLabelRange,
+      params.beta);
   }
 
-  // the destructor
-  void TearDown() override { RAFT_CUDA_TRY(cudaStreamDestroy(stream)); }
-
   // declaring the data values
+  raft::handle_t handle;
```
cjnolet, line 106: I notice we are missing gtests for a bunch of these. Can you please create an issue and list out the functions which aren't yet being tested? Also, since some of these aren't being tested, can you create a cuml PR and pin it to this branch so we can test that nothing breaks in cuml? We didn't do this for the linalg and matrix APIs and found that some changes actually did break cuml inadvertently.

cpp/include/raft/stats/cov.cuh
lowener: I think we should leave it like that.

cpp/test/stats/mean_center.cu
```diff
@@ -50,7 +50,7 @@ class MeanCenterTest : public ::testing::TestWithParam<MeanCenterInputs<T, IdxTy
       out(rows * cols, stream),
       out_ref(rows * cols, stream),
       data(rows * cols, stream),
-      meanVec(params.bcastAlongRows ? cols : rows, stream)
```
cjnolet, line 53: I think this logic is correct here- when `bcastAlongRows == true`, the vector should be the columns, since `along` rows would be across columns. Your original implementation was correct and I think I missed this line initially when reviewing this pr:

```
auto mean_vec_size = along_rows ? data.extent(1) : data.extent(0);
```

cpp/include/raft/stats/weighted_mean.cuh
```diff
@@ -102,7 +102,7 @@ void colWeightedMean(
  * @tparam layout_t Layout type of the input matrix.
  * @param[in]  handle the raft handle
  * @param[in]  data the input matrix of size nrows * ncols
- * @param[in]  weights weight of size ncols if along_row is true, else of size nrows
+ * @param[in]  weights weight of size nrows if along_row is true, else of size ncols
```
cjnolet, line 105: This should be `weights weight of size ncols if along_row is true, else of size nrows`

cpp/include/raft/stats/weighted_mean.cuh
```diff
@@ -139,13 +139,13 @@ void weighted_mean(const raft::handle_t& handle,
...
  * @brief Compute the row-wise weighted mean of the input matrix with a
- * vector of column weights
+ * vector of row weights
  *
  * @tparam value_t the data type
  * @tparam idx_t Integer type used to for addressing
  * @param[in]  handle the raft handle
  * @param[in]  data the input matrix of size nrows * ncols
- * @param[in]  weights per-col weight
+ * @param[in]  weights weight of size nrows
```
cjnolet, line 148: `per-col` weight was correct here.

cpp/include/raft/stats/weighted_mean.cuh
```diff
@@ -159,13 +159,13 @@ void row_weighted_mean(const raft::handle_t& handle,
 
 /**
  * @brief Compute the column-wise weighted mean of the input matrix with a
- * vector of row weights
+ * vector of col weights
```
cjnolet, line 162: col weights should be correct, though you might want to add it's only correct when `along_rows==true`, otherwise it's row-wise weights.

cpp/include/raft/stats/weighted_mean.cuh
```diff
@@ -159,13 +159,13 @@ void row_weighted_mean(const raft::handle_t& handle,
...
  *
  * @tparam value_t the data type
  * @tparam idx_t Integer type used to for addressing
  * @param[in]  handle the raft handle
  * @param[in]  data the input matrix of size nrows * ncols
- * @param[in]  weights per-row weight
```
cjnolet: Same as above

cpp/test/stats/weighted_mean.cu
```diff
@@ -73,18 +94,18 @@ class RowWeightedMeanTest : public ::testing::TestWithParam<WeightedMeanInputs<T
     auto stream = handle.get_stream();
     // device-side data
     din.resize(len);
-    dweights.resize(cols);
-    dexp.resize(rows);
-    dact.resize(rows);
+    dweights.resize(rows);
```
cjnolet, line 97: I would take `RowWeightedMean` to be that we want to do the mean along rows, right? That would mean we want the weights to be sized to the cols so we do the weighted average of each row, right? So the final array produced would be the number of rows when along rows == true and the number of columns when along rows == false.

cjnolet:
@lowener since it's getting so close to burndown, I reverted the `weighted_mean` changes. Let's aim to get those in early in 22.12.

cjnolet (approved):

cjnolet:
rerun tests

cjnolet (approved):

cjnolet:
rerun tests

cjnolet:
@gpucibot merge

PR #808, opened 2022-09-06
[WIP] Fix malloc/delete[] mismatch

### LINKED ISSUE #807: [BUG] delete[] being used for malloc'd allocations
**Describe the bug**

`./build.sh libraft tests --compile-libs` with CUDA Toolkit 11.7 and GCC 11.2 as the host compiler gives build errors such as the following.

```
.../raft/cpp/test/sparse/sort.cu:84:11: error: ‘void operator delete [](void*)’ called on pointer returned from a mismatched allocation function [-Werror=mismatched-new-delete]
   84 |   delete[] in_rows_h;
      |           ^~~~~~~~~
.../raft/cpp/test/sparse/sort.cu:63:31: note: returned from ‘void* malloc(size_t)’
   63 |   int* in_rows_h = (int*)malloc(params.nnz * sizeof(int));
      |                         ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

1. Check out current HEAD (8c639d994801cb00b7af6c4756cf75aedcf693b9) of branch-22.10. 
2. `./build.sh libraft tests --compile-libs` with CUDA Toolkit 11.7 and GCC 11.2 as the host compiler (the default GCC on Ubuntu 22.04 LTS).
3. Observe build errors.

**Expected behavior**

* The default build should not fail.
* Allocations performed using `malloc` should be freed using `free`, not `delete[]`.  Better yet, C++ code should not be doing manual memory management.  `std::make_unique<int[]>(params.nnz)` is the right way to perform these allocations.

 - Environment location: Bare-metal
 - Method of RAFT install: from source

**Additional context**

N/A

mhoemmen:
Fix `malloc` allocations that use `delete[]` for deallocation, by using `std::make_unique<int[]>` to manage memory instead.

Fixes #807.


cjnolet (approved):
Looks like a style updated is needed, but this LGTM pending CO.

Thanks for fixing this Mark!

cpp/test/sparse/sort.cu
```diff
@@ -60,30 +61,26 @@ TEST_P(COOSort, Result)
 
   uniform(h, r, in_vals.data(), params.nnz, float(-1.0), float(1.0));
 
-  int* in_rows_h = (int*)malloc(params.nnz * sizeof(int));
-  int* in_cols_h = (int*)malloc(params.nnz * sizeof(int));
-  int* verify_h  = (int*)malloc(params.nnz * sizeof(int));
+  auto in_rows_h = std::make_unique<int[]>(params.nnz);
```
cjnolet, line 64: Maybe we could just use a std::vector here?

mhoemmen: @cjnolet A good point, though I would prefer not to change the test's existing behavior by zero-initializing all the elements of the arrays.  (That would also hinder a sanitizer's ability to track use of uninitialized values.)  It's up to you, though!

mhoemmen:
Thanks @cjnolet !  Your formatting script is fantastic, btw!  It told me exactly what version of clang-format to use.  Different clang-format versions change the formatting a lot, so it's really helpful that the script reports the required version.

cpp/test/sparse/sort.cu
cjnolet: In general I prefer the use of vectors over smart pointers for the host arrays, especially as we are just performing a loop over the array to directly populate it in the next step.

It's not a huge deal in this test, though, as it's pretty trivial.

cjnolet:
@gpucibot merge

PR #809, opened 2022-09-07
Updating `raft::linalg` APIs to use `mdspan`

divyegala:


cpp/include/raft/linalg/eig.cuh
```diff
@@ -112,6 +114,117 @@ void eigJacobi(const raft::handle_t& handle,
...
+ * @tparam IntegerType Integer used for addressing
+ * @param handle raft::handle_t
+ * @param in input raft::device_matrix_view (symmetric matrix that has real eig values and
+ * vectors)
+ * @param n_eig_vals: number of eigenvectors to be generated
+ * @param eig_vectors: eigenvectors output of type raft::device_matrix_view
+ * @param eig_vals: eigen values output of type raft::device_vector_view
+ */
+template <typename ElementType, typename IndexType = std::uint32_t>
+void eig_dc_select(const raft::handle_t& handle,
```
cjnolet, line 163: Can we call this `eig_dc_selective` maybe?

cpp/include/raft/linalg/eig.cuh
```diff
@@ -112,6 +114,117 @@ void eigJacobi(const raft::handle_t& handle,
...
+                   const raft::device_matrix_view<ElementType, IndexType, raft::col_major> in,
+                   std::size_t n_eig_vals,
+                   raft::device_matrix_view<ElementType, IndexType, raft::col_major> eig_vectors,
+                   raft::device_vector_view<ElementType, IndexType> eig_vals,
+                   EigVecMemUsage memUsage)
+{
+  RAFT_EXPECTS(in.is_exhaustive(), "Input must be contiguous");
```
cjnolet, line 170: Same wrt exhaustive vs row-/col-major. This applies to the rest of the PR so I won't bother pointing out each and every instance.

cpp/include/raft/linalg/map.cuh
```diff
@@ -37,17 +39,63 @@ namespace linalg {
...
 template <typename InType,
           typename MapOp,
-          int TPB = 256,
+          int TPB          = 256,
+          typename OutType = InType,
           typename... Args,
-          typename OutType = InType>
-void map(OutType* out, size_t len, MapOp map, cudaStream_t stream, const InType* in, Args... args)
+          typename = enable_if_device_mdspan<InType, OutType>>
+void map(const raft::handle_t& handle, OutType out, MapOp map, const InType in, Args... args)
```
cjnolet, line 79: Is this missing the `mdspan` overloads?

cpp/include/raft/linalg/map_then_reduce.cuh
```diff
@@ -82,9 +86,65 @@ void mapThenReduce(OutType* out,
...
+ */
+template <typename InElementType,
+          typename MapOp,
+          typename ReduceLambda,
+          typename IndexType      = std::uint32_t,
+          typename OutElementType = InElementType,
+          int TPB                 = 256,
+          typename... Args>
+void map_reduce(const raft::handle_t& handle,
+                raft::device_scalar_view<OutElementType> out,
```
cjnolet, line 124: We need to keep the order of arguments consistent across the new API (which means not following the order of the arguments from the existing API functions in many cases). I captured the expected orderings [here](https://github.com/rapidsai/raft/issues/535). Following a more thrust-like ordering, we should put the handle first, then inputs, then outputs, then additional params/arguments.

cpp/include/raft/linalg/qr.cuh
```diff
@@ -70,6 +69,51 @@ void qrGetQR(const raft::handle_t& handle,
...
   detail::qrGetQR(handle, M, Q, R, n_rows, n_cols, stream);
 }
+
+/**
+ * @brief compute QR decomp and return only Q matrix
+ * @param handle raft::handle_t
+ * @param M Input raft::device_matrix_view
+ * @param Q Output raft::device_matrix_view
+ */
+template <typename ElementType, typename IndexType = std::uint32_t>
```
cjnolet, line 79: This file following the snake-cased format that we've been using everywhere else: (e.g. `qr_get_q`)

cpp/include/raft/linalg/qr.cuh
```diff
@@ -70,6 +69,51 @@ void qrGetQR(const raft::handle_t& handle,
 {
   detail::qrGetQR(handle, M, Q, R, n_rows, n_cols, stream);
 }
+
+/**
+ * @brief compute QR decomp and return only Q matrix
```
cjnolet, line 74: Can we spell out the decumentation more clearly? I'd prefer for clarity over potential redundancy. Something like `Compute the QR decomposition of matrix M and return only the Q matrix.`

cpp/include/raft/linalg/qr.cuh
```diff
@@ -70,6 +69,51 @@ void qrGetQR(const raft::handle_t& handle,
...
+{
+  RAFT_EXPECTS(Q.is_exhaustive(), "Output must be contiguous");
+  RAFT_EXPECTS(M.is_exhaustive(), "Input must be contiguous");
+  RAFT_EXPECTS(Q.size() == M.size(), "Size mismatch between Output and Input");
+
+  qrGetQ(handle, M.data_handle(), Q.data_handle(), M.extent(0), M.extent(1), handle.get_stream());
+}
+
+/**
+ * @brief compute QR decomp and return Q and R matrices
```
cjnolet, line 92: Same here wrt clarity

cpp/include/raft/linalg/reduce.cuh
```diff
@@ -75,6 +78,98 @@ void reduce(OutType* dots,
     dots, data, D, N, init, rowMajor, alongRows, stream, inplace, main_op, reduce_op, final_op);
 }
 
+/**
+ * @defgroup reduction Reduction Along Requested Dimension
+ * @{
+ */
+
+/**
+ * @brief Compute reduction of the input matrix along the requested dimension
```
cjnolet, line 87: The descriptions for the new functions could be made more verbose and explicit. The way I think about it is that a user should be able to look at the docs for this function and see right in the description 1) what exactly it's going to do to their data. The parameter/argument document should be expand upon that. 

Also, now that we have a nice clean API, usage examples are super helpful as well (like CUB and Thrust), it's really useful to have a copy/paste example that just works, vs having to figure out all the details before making the call.

cpp/include/raft/linalg/reduce.cuh
```diff
@@ -75,6 +78,98 @@ void reduce(OutType* dots,
...
+ * <pre>OutType (*ReduceLambda)(OutType);</pre>
+ * @tparam FinalLambda the final lambda applied before STG (eg: Sqrt for L2 norm)
+ * It must be a 'callable' supporting the following input and output:
+ * <pre>OutType (*FinalLambda)(OutType);</pre>
+ * @param handle raft::handle_t
+ * @param dots Output of type raft::device_matrix_view
+ * @param data Input of type raft::device_matrix_view
+ * @param init initial value to use for the reduction
+ * @param apply whether to reduce along rows or along columns (using raft::linalg::Apply)
+ * @param main_op elementwise operation to apply before reduction
```
cjnolet, line 107: when we accept lambdas like this, can we make use of the term `fused` somewhere in these descriptions so that we can let the user know the benefits of this element-wise operation up front? We have an element-wise primitive so why are they better off using this one? (the answer: because it fuses the element-wise op into the kernel before performing the reduction). 

If a lambda op is not fused and needs to be performed as a separate operation, we'd be better off removing it as an argument because it would save the additional template parameter which is non-trivial to specialize generically.

cpp/include/raft/linalg/sqrt.cuh
```diff
@@ -34,14 +35,54 @@ namespace linalg {
...
 }
 /** @} */
 
+/**
+ * @defgroup power Sqrt Arithmetic
+ * @{
+ */
+
+/**
+ * @brief Elementwise sqrt operation on the input buffer
```
cjnolet, line 52: maybe input buffer isn't the best word to use here now that it's working on mdspans?

cpp/include/raft/linalg/reduce_cols_by_key.cuh
```diff
@@ -52,6 +54,54 @@ void reduce_cols_by_key(const T* data,
...
   detail::reduce_cols_by_key(data, keys, out, nrows, ncols, nkeys, stream);
 }
+
+/**
+ * @defgroup reduce_cols_by_key Reduce Across Columns by Key
+ * @{
+ */
+
+/**
+ * @brief Computes the sum-reduction of matrix columns for each given key
```
cjnolet, line 64: I guess I haven't noticed how confusing this was before. If i was a user looking at a function called `reduce_cols_by_key`, I'd be expecting that I'd pass my own reduction function in. However, if I were invoking such a function without a reduction function (where another overload were provided that accepts a lambda), I'd expect it to perform a sum by default. It doesn't look like we provide such an option, though. What makes things even more confusing to me is that the `reduce_rows_by_key` function performs a weighted sum. 

I'm not saying we need to rename or remove these functions, but it would be useful if we had a github issue (and a todo here in this file that references it) to allow generic reduction functions. That would make this much more useful and also allow it to be extended for generalized row-norm computations).

cpp/include/raft/linalg/reduce_rows_by_key.cuh
```diff
@@ -108,6 +110,78 @@ void reduce_rows_by_key(const DataIteratorT d_A,
                      stream);
 }
 
+/**
+ * @defgroup reduce_rows_by_key Reduce Across Rows by Key
+ * @{
+ */
+
+/**
+ * @brief Computes the weighted reduction of matrix rows for each given key
```
cjnolet, line 119: Same as above- can you put a todo in this file and reference a github issue to add an overload w/ generic reduction functions? I believe the weighted reduction here should be using a fused multiply-add intrinsic but a generalized reduction would make this primitive more widely applicable to different problems.

cpp/include/raft/linalg/strided_reduction.cuh
```diff
@@ -71,6 +74,84 @@ void stridedReduction(OutType* dots,
   detail::stridedReduction(dots, data, D, N, init, stream, inplace, main_op, reduce_op, final_op);
 }
 
+/**
+ * @defgroup strided_reduction Strided Memory Access Reductions
```
cjnolet, line 78: The capitalization here seems a little weird. Is this a title?

cpp/include/raft/linalg/strided_reduction.cuh
```diff
@@ -71,6 +74,84 @@ void stridedReduction(OutType* dots,
...
+ * For reducing along rows for row-major and along columns for column-major
+ * @{
+ */
+
+/**
+ * @brief Compute reduction of the input matrix along the strided dimension
```
cjnolet, line 84: Can you expand upon this description a little bit? These primitives have been infamous for not being the most descriptive and if we can provide more information, I think they will reach users who may have avoided using them due to lack of clarity in the docs.

cjnolet: Also we've been trying to mark `[in]` and `[out]` in the docs and it's super useful to keep that consistent especially for our new public API functions.

cjnolet (changes requested):
I've done an initial pass at what you have so far. I haven't looked things through exhaustive- such as making sure each new function has an associated test, or even that all the functions have been `mdspan`-ified. I'll do a couple more passes but wanted to provide some immediate thoughts/feedback in the meantime to help accelerate the process.

cpp/include/raft/linalg/map.cuh
divyegala: No, why? The types have `enable_if_mdspan` on them

cpp/include/raft/linalg/strided_reduction.cuh
divyegala: Yes it's the title for the group

cpp/include/raft/linalg/map.cuh
cjnolet: Ah, weird, I must have missed that when going through the file diff.

cpp/include/raft/linalg/strided_reduction.cuh
divyegala: Yes, it's the title of the doc group

cpp/include/raft/linalg/coalesced_reduction.cuh
```diff
@@ -70,6 +73,85 @@ void coalescedReduction(OutType* dots,
...
+ * @param init initial value to use for the reduction
+ * @param main_op elementwise operation to apply before reduction
+ * @param reduce_op binary reduction operation
+ * @param final_op elementwise operation to apply before storing results
+ * @param inplace reduction result added inplace or overwrites old values?
+ */
+template <typename InElementType,
+          typename LayoutPolicy,
+          typename OutElementType = InElementType,
+          typename IndexType      = std::uint32_t,
```
mhoemmen, line 110: Would you consider removing the default values of `OutElementType` and `IndexType`, since these template parameters can always be deduced from the function's arguments?

cpp/include/raft/linalg/coalesced_reduction.cuh
```diff
@@ -70,6 +73,85 @@ void coalescedReduction(OutType* dots,
...
+          typename MainLambda     = raft::Nop<InElementType>,
+          typename ReduceLambda   = raft::Sum<OutElementType>,
+          typename FinalLambda    = raft::Nop<OutElementType>>
+void coalesced_reduction(
+  const raft::handle_t& handle,
+  raft::device_matrix_view<OutElementType, IndexType, LayoutPolicy> dots,
+  const raft::device_matrix_view<InElementType, IndexType, LayoutPolicy> data,
```
mhoemmen, line 117:
```suggestion
  raft::device_matrix_view<const InElementType, IndexType, LayoutPolicy> data,
```
Would you also consider renaming `InElementType` to `InValueType`, since it is meant not to be cv-qualified?  Compare `mdspan::value_type` (not cv-qualified) to `mdspan::element_type` (the user's template argument, with any cv qualifications).

cpp/include/raft/linalg/coalesced_reduction.cuh
```diff
@@ -70,6 +73,85 @@ void coalescedReduction(OutType* dots,
...
+  OutElementType init,
+  bool inplace           = false,
```
mhoemmen, line 119: Would you consider making the "in-place" version a different function or overload, rather than specifying this by `bool` parameter?  This is the approach taken by stdblas (P1673) for distinguishing in-place and out-of-place operations.

cpp/include/raft/linalg/detail/map_then_reduce.cuh
```diff
@@ -86,8 +88,8 @@ void mapThenReduceImpl(OutType* out,
                        Args... args)
 {
   raft::update_device(out, &neutral, 1, stream);
-  const int nblks = raft::ceildiv(len, (size_t)TPB);
-  mapThenReduceKernel<InType, OutType, MapOp, ReduceLambda, TPB, Args...>
+  const int nblks = raft::ceildiv(len, (IdxType)TPB);
```
mhoemmen, line 91: Would you consider for new code using C++ casts (`IdxType(TPB)` or `static_cast<IdxType>(TPB)`) instead of C-style casts?

cpp/include/raft/linalg/divide.cuh
```diff
@@ -36,13 +39,62 @@ using detail::divides_scalar;
...
+ * @brief Elementwise division of input by host scalar
+ * @tparam InType    Input Type raft::device_mdspan
+ * @tparam OutType   Output Type raft::device_mdspan
+ * @tparam ScalarIdxType Index Type of scalar
+ * @param handle raft::handle_t
+ * @param out    Output
+ * @param in    Input
+ * @param scalar    raft::host_scalar_view
+ */
+template <typename InType,
```
mhoemmen, line 64: Please see above notes; thanks!

cpp/include/raft/linalg/eig.cuh
```diff
@@ -112,6 +114,117 @@ void eigJacobi(const raft::handle_t& handle,
...
+ * @tparam IntegerType Integer used for addressing
+ * @param handle raft::handle_t
+ * @param in input raft::device_matrix_view (symmetric matrix that has real eig values and
+ * vectors)
+ * @param eig_vectors: eigenvectors output of type raft::device_matrix_view
+ * @param eig_vals: eigen values output of type raft::device_vector_view
+ */
+template <typename ElementType, typename IndexType = std::uint32_t>
+void eig_dc(const raft::handle_t& handle,
+            const raft::device_matrix_view<ElementType, IndexType, raft::col_major> in,
```
mhoemmen, line 131:
```suggestion
            raft::device_matrix_view<const ElementType, IndexType, raft::col_major> in,
```
Would you consider applying this change generally to the entire PR?

Also, would you consider renaming `ElementType` to `ValueType` here?  (Please see above notes.)

cpp/include/raft/linalg/eig.cuh
```diff
@@ -112,6 +114,117 @@ void eigJacobi(const raft::handle_t& handle,
...
+ * error is below tol
+ * @param sweeps: number of sweeps in the Jacobi algorithm. The more the better
+ * accuracy.
+ */
+template <typename ElementType, typename IndexType = std::uint32_t>
+void eig_jacobi(const raft::handle_t& handle,
+                const raft::device_matrix_view<ElementType, IndexType, raft::col_major> in,
+                raft::device_matrix_view<ElementType, IndexType, raft::col_major> eig_vectors,
+                raft::device_vector_view<ElementType, IndexType> eig_vals,
+                ElementType tol = 1.e-7,
```
mhoemmen, line 208: 1. Would you consider passing multiple algorithm parameters as a struct?  Users could then use designated initializers to name the input values, if they want to specify any nondefault values.

2. Would you consider making the default value of `tol` depend on `ElementType`, by using `numeric_limits::epsilon()`?

```c++
template<class ValueType>
struct eig_jacobi_params {
  using magnitude_type = decltype(std::abs(std::declval<ValueType>()));

  magnitude_type tol = std::numeric_limits<magnitude_type>::epsilon();
  int sweeps = 15;
};
```

cpp/include/raft/linalg/eig.cuh
```diff
@@ -112,6 +114,117 @@ void eigJacobi(const raft::handle_t& handle,
...
+ * @brief overloaded function for eig decomp with Jacobi method for the
+ * column-major symmetric matrices (in parameter)
+ * @tparam ElementType the data-type of input and output
+ * @tparam IntegerType Integer used for addressing
+ * @param handle raft::handle_t
+ * @param in input raft::device_matrix_view (symmetric matrix that has real eig values and
+ * vectors)
+ * @param eig_vectors: eigenvectors output of type raft::device_matrix_view
+ * @param eig_vals: eigen values output of type raft::device_vector_view
+ * @param tol: error tolerance for the jacobi method. Algorithm stops when the
```
mhoemmen, line 198: Would you consider specifying whether this is a relative or absolute tolerance, and what norm is used?

cpp/include/raft/linalg/lstsq.cuh
```diff
@@ -115,6 +115,143 @@ void lstsqQR(const raft::handle_t& handle,
...
 }
 
+/**
+ * @defgroup lstsq Least Squares Methods
+ * @{
+ */
+
+/**
+ * @brief Solves the linear ordinary least squares problem `Aw = b`
+ * Via SVD decomposition of `A = U S Vt` using default cuSOLVER routine.
```
mhoemmen, line 125: Given comments above, it sounds like whether cuSOLVER is used is an implementation detail that we don't have to reveal to users.

cpp/include/raft/linalg/map.cuh
mhoemmen, line 79:
```suggestion
void map(const raft::handle_t& handle, OutType out, MapOp map, InType in, Args... args)
```
Please see notes above about using `mdspan<const T, ...>` instead of `const mdspan<T, ...>`; thanks!

cpp/include/raft/linalg/map_then_reduce.cuh
```diff
@@ -82,9 +86,65 @@ void mapThenReduce(OutType* out,
...
+                OutElementType neutral,
+                MapOp map,
+                ReduceLambda op,
+                const raft::device_vector_view<InElementType, IndexType> in,
```
mhoemmen, line 128:
```suggestion
                raft::device_vector_view<const InElementType, IndexType> in,
```

cpp/include/raft/linalg/matrix_vector_op.cuh
```diff
@@ -99,6 +102,142 @@ void matrixVectorOp(Type* out,
...
+ */
+template <typename InElementType,
+          typename LayoutPolicy,
+          typename Lambda,
+          typename OutElementType = InElementType,
+          typename IndexType      = std::uint32_t,
+          int TPB                 = 256>
+void matrix_vector_op(const raft::handle_t& handle,
+                      raft::device_matrix_view<OutElementType, IndexType, LayoutPolicy> out,
+                      const raft::device_matrix_view<InElementType, IndexType, LayoutPolicy> matrix,
```
mhoemmen, line 140:
```suggestion
                      raft::device_matrix_view<const InElementType, IndexType, LayoutPolicy> matrix,
```

cpp/include/raft/linalg/matrix_vector_op.cuh
```diff
@@ -99,6 +102,142 @@ void matrixVectorOp(Type* out,
...
+                      const raft::device_vector_view<InElementType, IndexType> vec,
```
mhoemmen, line 141:
```suggestion
                      raft::device_vector_view<const InElementType, IndexType> vec,
```

cpp/include/raft/linalg/power.cuh
```diff
@@ -55,14 +56,100 @@ void powerScalar(math_t* out, const math_t* in, math_t scalar, IdxType len, cuda
...
+ * @tparam OutType   Output Type raft::device_mdspan
+ * @param handle raft::handle_t
+ * @param out    Output
+ * @param in1    First Input
+ * @param in2    Second Input
+ */
+template <typename InType,
+          typename OutType = InType,
+          typename         = raft::enable_if_device_mdspan<OutType, InType>>
+void power(const raft::handle_t& handle, OutType out, const InType in1, const InType in2)
```
mhoemmen, line 84:
```suggestion
void power(const raft::handle_t& handle, OutType out, InType in1, InType in2)
```

cpp/include/raft/linalg/power.cuh
```diff
@@ -55,14 +56,100 @@ void powerScalar(math_t* out, const math_t* in, math_t scalar, IdxType len, cuda
...
+ * @param scalar    raft::host_scalar_view
+ */
+template <typename InType,
+          typename OutType       = InType,
+          typename ScalarIdxType = std::uint32_t,
+          typename               = raft::enable_if_device_mdspan<OutType, InType>>
+void power_scalar(const raft::handle_t& handle,
+                  OutType out,
+                  const InType in,
+                  const raft::host_scalar_view<typename InType::element_type, ScalarIdxType> scalar)
```
mhoemmen, line 127:
```suggestion
                  raft::host_scalar_view<const typename InType::value_type, ScalarIdxType> scalar)
```

cpp/include/raft/linalg/power.cuh
```diff
@@ -55,14 +56,100 @@ void powerScalar(math_t* out, const math_t* in, math_t scalar, IdxType len, cuda
...
+ * @param in    Input
+ * @param scalar    raft::host_scalar_view
+ */
+template <typename InType,
+          typename OutType       = InType,
+          typename ScalarIdxType = std::uint32_t,
+          typename               = raft::enable_if_device_mdspan<OutType, InType>>
+void power_scalar(const raft::handle_t& handle,
+                  OutType out,
+                  const InType in,
```
mhoemmen, line 126:
```suggestion
                  InType in,
```

mhoemmen (changes requested):
Thanks for letting me review! : - )  Some general comments:

1. A read-only mdspan is `mdspan<const T, ...>`, not `const mdspan<T, ...>`.  I can't think of a reason why one would want to use `const mdspan` as a function parameter type.
2. `mdspan::element_type` is the template argument that the user gave to mdspan.  `mdspan::value_type` is that with any cv-qualifiers removed.
3. Would you consider separating workspace queries from actual computations by using different function names?
4. Would you consider making tolerances a function of the value type?

cpp/include/raft/linalg/cholesky_r1_update.cuh
```diff
@@ -132,6 +132,137 @@ void choleskyRank1Update(const raft::handle_t& handle,
...
+ *    matrix (CUBLAS_FILL_MODE_UPPER or CUBLAS_FILL_MODE_LOWER)
+ * @param eps numerical parameter that can act as a regularizer for ill
+ *    conditioned systems. Negative values mean no regularizaton.
+ */
+template <typename ElementType, typename IndexType = std::uint32_t>
+void cholesky_rank1_update(const raft::handle_t& handle,
+                           raft::device_matrix_view<ElementType, IndexType, raft::col_major> L,
+                           int n,
+                           int ld,
+                           std::optional<raft::device_vector_view<char>> workspace,
```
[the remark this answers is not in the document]
cjnolet: I'd actually prefer for us to not expose the workspace at all. This is something we can add back if needed, but we spent a lot of time supporting it in the early APIs and it turns out it would just get wrapped and no longer used directly.

cpp/include/raft/linalg/add.cuh
```diff
@@ -72,24 +74,150 @@ void add(OutT* out, const InT* in1, const InT* in2, IdxType len, cudaStream_t st
...
+  using in_element_t  = typename InType::element_type;
+  using out_element_t = typename OutType::element_type;
+
+  RAFT_EXPECTS(out.is_exhaustive(), "Output must be contiguous");
+  RAFT_EXPECTS(in1.is_exhaustive(), "Input 1 must be contiguous");
+  RAFT_EXPECTS(in2.is_exhaustive(), "Input 2 must be contiguous");
+  RAFT_EXPECTS(out.size() == in1.size() && in1.size() == in2.size(),
+               "Size mismatch between Output and Inputs");
+
+  if (out.size() <= std::numeric_limits<std::uint32_t>::max()) {
```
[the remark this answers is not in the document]
divyegala: In point 2, what happens in extreme cases? Consider `index_type=uint32_t` with extents `{2^32, 2}`. In this case, will `required_span_size()` by representable by `index_type` or will it cause an overflow?

cpp/include/raft/linalg/cholesky_r1_update.cuh
```diff
@@ -132,6 +132,137 @@ void choleskyRank1Update(const raft::handle_t& handle,
...
+ * raft::linalg::choleskyRank1Update(handle, L, n_rows, ld_L, nullptr,
+ *                                       &n_bytes, CUBLAS_FILL_MODE_LOWER,
+ *                                       stream);
+ * rmm::device_uvector<char> workspace(n_bytes, stream);
+ *
+ * for (n=1; n<=n_rows; rank++) {
+ *   // Calculate a new row/column of matrix A into A_new
+ *   // ...
+ *   // Copy new row to L[rank-1,:]
+ *   RAFT_CUBLAS_TRY(cublasCopy(handle.get_cublas_handle(), n - 1, A_new, 1,
```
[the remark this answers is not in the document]
divyegala: We could consider adding it in the next release, if we want to account for `submdspan` as well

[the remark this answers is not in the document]
cjnolet: If this is just copying a contiguous memory location into a new memory location, I would consider just using `raft::copy` here (which is actually what's being done in the LARS solver in cuML).

cpp/include/raft/linalg/add.cuh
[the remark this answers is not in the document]
mhoemmen: @divyegala `required_span_size()` is _not_ representable by `index_type` in this case.  For `layout_left` and `layout_right`, `required_span_size()` and `size()` are the same mathematically.  The only difference is the return type (`index_type` resp. `size_t`).  For `layout_stride`, though, `required_span_size()` can be greater than the `size()`.  For other layouts (e.g., the "matrix of a single value" layout that maps all multidimensional indices to the offset zero), `required_span_size()` can be less than `size()`. 

Note that while it's UB for users to violate preconditions, implementations aren't required to _check_ preconditions.  The reference implementation of `layout_left` does _not_ currently check preconditions, as [you can see here](https://github.com/kokkos/mdspan/blob/d58ddaaa1869aa5ad8469c3b02abe025efab52a4/include/experimental/__p0009_bits/layout_left.hpp#L117), for instance.  This means two things.

1. If someone gives you a `layout_{left,right,stride}::mapping` instance (e.g., in an `mdspan`), then you can assume that the precondition is satisfied.

2. If you are _constructing_ a `layout_{left,right,stride}::mapping` instance (e.g., by constructing an `mdspan` with a pointer and extents), then you are responsible for ensuring that the precondition is satisfied.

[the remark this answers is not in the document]
mhoemmen: @divyegala wrote:
> In this case, will `required_span_size()` by representable by `index_type` or will it cause an overflow?

Those are two separate questions, actually! : - ) 

1. `required_span_size()` is _not_ representable by `index_type` in this case.
2. Giving this `extents` object to `layout_{left,right,stride}::mapping`'s constructor violates the constructor's precondition.  It could overflow, or it could open a portal to the Awesome Dimension and let loose a swarm of nasal demons who search out precondition violators and boop them gently on the nose.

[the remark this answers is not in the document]
divyegala: @mhoemmen thanks for the explanations! How do we really represent such edge cases and safely obtain the product of the extents? Sounds like `size()` is the safe way to obtain the product without violating any pre-conditions since it's representable by `size_t`?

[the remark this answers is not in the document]
mhoemmen: @divyegala Gladly! : - )

> How do we really represent such edge cases and safely obtain the product of the extents? 

By the time the user has created a layout mapping, it's already too late.  What I mean by that is that if `required_span_size()` doesn't fit `index_type`, then the user will likely get the wrong answer when they try to index into the mdspan.

In what follows in my comment, I'll distinguish between "the Preconditions in the spec" and "what the reference implementation does."  The reference implementation currently does _not_ check this precondition in the layout mapping.  This means that it's possible for users to construct extents for which the mapping's `required_span_size()` can overflow.

We can prevent this by wrapping mdspan creation to check the extents object for potential overflow, before it goes into a layout mapping's constructor.  It's not UB to construct, e.g., `dextents<uint16_t, 2>(` 2^{15} `,` 2^{15} `)`.  We just need to intercept that naughty `extents` value before it goes into a layout mapping's constructor.  Otherwise, the layout mapping has the freedom to do whatever it likes, including calling `abort()`.

Our mdarray implementation's conversion to mdspan can also check, but again, we're probably better off making the wrapper explicit and not part of the mdarray proposal.  WG21 likes Preconditions and wants violating them to be UB.  If we want some specified behavior (e.g., throwing a particular exception, or calling `terminate()` after printing a helpful error message), then we'll have to implement that ourselves.

cpp/include/raft/linalg/cholesky_r1_update.cuh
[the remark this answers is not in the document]
divyegala: How would we not expose the workspace without exposing an interface to accept an allocator?

cpp/include/raft/linalg/coalesced_reduction.cuh
divyegala: Looking at the code implementation, I'm not sure if `inplace` is doing exactly what the name suggests it should be doing

cpp/include/raft/linalg/cholesky_r1_update.cuh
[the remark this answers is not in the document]
mhoemmen: Perhaps we could hide a scratch space allocation in the RAFT handle.  That would ensure coupling of the allocation to a CUDA stream, permitting e.g., asynchronous allocations.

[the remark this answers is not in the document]
cjnolet: Take a similar case in the `raft::distance` API as an example: https://github.com/rapidsai/raft/blob/0d71ceb613432a4cd438b78a9180666e411f64c4/cpp/include/raft/distance/distance.cuh#L215.

cpp/include/raft/linalg/rsvd.cuh
```diff
@@ -137,6 +139,131 @@ void rsvdPerc(const raft::handle_t& handle,
...
+ * @param[out] V optional right singular values of raft::device_matrix_view with layout
+ * raft::col_major
+ */
+template <typename ValueType, typename IndexType>
+void rsvd_fixed_rank(
+  const raft::handle_t& handle,
+  raft::device_matrix_view<ValueType, IndexType, raft::col_major> M,
+  raft::device_vector_view<ValueType, IndexType> S_vec,
+  IndexType p,
+  bool use_bbt,
```
mhoemmen, line 170: Would you consider putting algorithm configuration parameters (not input, output, or options like `p`, which are algorithm-independent) in a struct, rather than passing them in one by one?  That would reduce the number of arguments, make it easier for users to get default values, and would avoid common errors like mixing up the order of arguments.

cpp/include/raft/linalg/strided_reduction.cuh
```diff
@@ -71,6 +74,93 @@ void stridedReduction(OutType* dots,
...
+ * It must be a 'callable' supporting the following input and output:
+ * <pre>OutType (*FinalLambda)(OutType);</pre>
+ * @param[in] handle raft::handle_t
+ * @param[in] data Input of type raft::device_matrix_view
+ * @param[out] dots Output of type raft::device_matrix_view
+ * @param[in] init initial value to use for the reduction
+ * @param[in] main_op fused elementwise operation to apply before reduction
+ * @param[in] reduce_op fused binary reduction operation
+ * @param[in] final_op fused elementwise operation to apply before storing results
+ * @param[in] inplace reduction result added inplace or overwrites old values?
```
cjnolet, line 110: It's unclear to me from reading this doc 1) whether or not the `dots` output is needed when `inplace==true` and 2) if `data == dots` when `inplace==true`. I felt the same while looking over the API for `reduce`.  This might just require some additional documentation in these functions, though I wonder if it's an indication that an update to the API might be needed (I'm just really not sure what that would be as of yet).

cpp/include/raft/linalg/rsvd.cuh
cjnolet: Looks like this function has been simplified quite a bit, which is great! An alternative to a struct might also be to break this apart into separate functions with names that might imply a specific configuration. What's not clear from the documentation at all is that `use_bbt` determines whether a rectangular or symmetric eigen decomposition is performed and `use_jacobi` determines whether Jacobi approx is used in place of QR (bbt=false) or divide-and-conquer (bbt=true). 

Maybe we could keep the `use_jacobi`, remove the `use_bbt` option, and set it to false for this particular one. Then can add a new function: `rsvd_fixed_rank_symmetric`. I think that makes the intention more clear and self explanatory. To make things even more clear, we could create `rsvd_fixed_rank_jacobi` and `rsvd_fixed_rank_symmetric_jacobi`. This would follow the convention used in `svd.cuh` as well for consistency.

cpp/include/raft/linalg/rsvd.cuh
```diff
@@ -137,6 +139,131 @@ void rsvdPerc(const raft::handle_t& handle,
...
+ * @param[in] tol tolerance for Jacobi-based solvers
+ * @param[in] max_sweeps maximum number of sweeps for Jacobi-based solvers
+ * @param[out] U optional left singular values of raft::device_matrix_view with layout
+ * raft::col_major
+ * @param[out] V optional right singular values of raft::device_matrix_view with layout
+ * raft::col_major
+ */
+template <typename ValueType, typename IndexType>
+void rsvd_perc(
+  const raft::handle_t& handle,
```
cjnolet, line 226: Could do the same as above for this one too in order to remove some params and make the usage more clear. Both definitely need a doc update though

cpp/include/raft/linalg/unary_op.cuh
```diff
@@ -71,6 +73,79 @@ void writeOnlyUnaryOp(OutType* out, IdxType len, Lambda op, cudaStream_t stream)
...
+ * @param[in] in Input
+ * @param[out] out Output
+ * @param[in] op the device-lambda
+ * @note Lambda must be a functor with the following signature:
+ *       `InType func(const InType& val);`
+ */
+template <typename InType,
+          typename Lambda,
+          typename OutType,
+          int TPB  = 256,
```
cjnolet, line 97: This is exposing underlying implementation details and we should remove it from public APIs everywhere. We should also do a quick grep through cuml and cugraph to make sure this isn't being set to anything other than 256.

cpp/include/raft/linalg/unary_op.cuh
```diff
@@ -71,6 +73,79 @@ void writeOnlyUnaryOp(OutType* out, IdxType len, Lambda op, cudaStream_t stream)
...
+ * @brief perform element-wise binary operation on the input arrays
+ * @tparam InType Input Type raft::device_mdspan
+ * @tparam Lambda the device-lambda performing the actual operation
+ * @tparam TPB threads-per-block in the final kernel launched
+ * @param[in] handle raft::handle_t
+ * @param[inout] in Input/Output
+ * @param[in] op the device-lambda
+ * @note Lambda must be a functor with the following signature:
+ *       `InType func(const InType& val);`
+ */
```
cjnolet, line 127: The docs for this function look identical to the one above but this has a different name and it's not immediately obvious to me what it means.

cjnolet (changes requested):
Things are coming together great! I just did another deeper review. I'll try and do another before EOD.

cpp/include/raft/linalg/apply.hpp
```diff
@@ -0,0 +1,28 @@
...
+#pragma once
+
+namespace raft::linalg {
+
+/**
+ * @brief Enum for reduction/broadcast where an operation is to be performed along
+ *        a matrix's rows or columns
+ *
+ */
+enum class Apply { ALONG_ROWS, ALONG_COLUMNS };
```
cjnolet, line 26: Maybe also put this in `linalg_types.hpp`?

cpp/include/raft/linalg/axpy.cuh
```diff
@@ -50,6 +50,79 @@ void axpy(const raft::handle_t& handle,
...
+ * @param [in] alpha raft::device_scalar_view
+ * @param [in] x Input vector
+ * @param [in] incx stride between consecutive elements of x
+ * @param [inout] y Output vector
+ * @param [in] incy stride between consecutive elements of y
+ */
+template <typename MdspanType, typename = raft::enable_if_device_mdspan<MdspanType>>
+void axpy(const raft::handle_t& handle,
+          raft::device_scalar_view<typename MdspanType::element_type, ScalarIdxType> alpha,
+          const MdspanType x,
```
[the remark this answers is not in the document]
mhoemmen: `const mdspan<T>` might not even compile if you need to make it from, say, a `const std::vector<T>&`.

cpp/include/raft/linalg/apply.hpp
cjnolet: I've also been spending time on the docs for 22.10 and consolidating these types into a single file that follows a consistent naming convention makes it a lot easier to list these files out in the documentation explicitly

cpp/include/raft/linalg/map_then_reduce.cuh
```diff
@@ -82,9 +86,64 @@ void mapThenReduce(OutType* out,
...
+ * @param[in] args additional input arrays
+ */
+template <typename InValueType,
+          typename MapOp,
+          typename ReduceLambda,
+          typename IndexType,
+          typename OutValueType,
+          int TPB = 256,
+          typename... Args>
+void map_reduce(const raft::handle_t& handle,
```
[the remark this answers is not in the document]
divyegala: Let's change the name of the file in that case and mark it a breaking change?

cpp/include/raft/linalg/reduce.cuh
```diff
@@ -75,6 +78,88 @@ void reduce(OutType* dots,
...
+ * <pre>OutType (*FinalLambda)(OutType);</pre>
+ * @param[in] handle raft::handle_t
+ * @param[in] data Input of type raft::device_matrix_view
+ * @param[out] dots Output of type raft::device_matrix_view
+ * @param[in] init initial value to use for the reduction
+ * @param[in] apply whether to reduce along rows or along columns (using raft::linalg::Apply)
+ * @param[in] main_op fused elementwise operation to apply before reduction
+ * @param[in] reduce_op fused binary reduction operation
+ * @param[in] final_op fused elementwise operation to apply before storing results
+ * @param[in] inplace reduction result added inplace or overwrites old values?
```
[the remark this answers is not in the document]
divyegala: That's my fault after my refactor of following input->output->args. I will be doing a thorough pass of docs again

cpp/include/raft/linalg/strided_reduction.cuh
divyegala: It's unclear to me as well, and I mentioned it to another of @mhoemmen 's comments. Reading the impl did not make it any clearer for me.

cpp/include/raft/linalg/map_then_reduce.cuh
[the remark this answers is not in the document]
cjnolet: Maybe instead of changing the name of the file and making it breaking, we could add a new file with the new name, just add the mdspan overload (not the one that accepts pointers) and deprecate this one?

cpp/include/raft/linalg/cholesky_r1_update.cuh
```diff
@@ -132,6 +135,139 @@ void choleskyRank1Update(const raft::handle_t& handle,
...
+ *    of size [n_bytes].
+ * @param[in] n_bytes size of workspace is returned here if workspace==std::nullopt.
+ * @param[in] uplo indicates whether L is stored as an upper or lower triangular
+ *    matrix (raft::linalg::FillMode::UPPER or raft::linalg::FillMode::LOWER)
+ * @param[in] eps numerical parameter that can act as a regularizer for ill
+ *    conditioned systems. Negative values mean no regularizaton.
+ */
+template <typename ElementType, typename IndexType = std::uint32_t>
+void cholesky_rank1_update(const raft::handle_t& handle,
+                           raft::device_matrix_view<ElementType, IndexType, raft::col_major> L,
```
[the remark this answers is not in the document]
divyegala: It's already that, isn't it? Only one input/output in this function, and they are both the same parameter

cpp/include/raft/linalg/cholesky_r1_update.cuh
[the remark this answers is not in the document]
divyegala: I don't know if this is possible to use, because this is an iterative method that is called from the host multiple times by the user. The workspace is used repeatedly for every iteration. So is it okay if I keep the current design?

cpp/include/raft/linalg/eig.cuh
divyegala: I'm quite unaware to-be-honest, maybe @cjnolet can help here?

cpp/include/raft/linalg/cholesky_r1_update.cuh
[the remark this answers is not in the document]
cjnolet: I would prefer we not keep workspaces on the public API. As discussed in the past, a pool allocator will remove the overhead from intermediate allocations if that becomes a bottleneck here. Something tells me it's not the bottleneck, though. Another option we have, which I was recently discussing w/ @mhoemmen (and I've discussed w/ others in the past) is to add an option in the handle to attach an additional allocator just for intermediate / temporary allocations in between calls to primitives (this allocator would be beneficial as a pool allocator).

cpp/include/raft/linalg/cholesky_r1_update.cuh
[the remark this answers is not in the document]
cjnolet: (this would mean removing the `workspace` and `n_bytes` arguments also).

cpp/include/raft/linalg/cholesky_r1_update.cuh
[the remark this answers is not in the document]
divyegala: That would be a future discussion, would it not? Allocating and de-allocating a workspace every iteration will severely slow this function down though if a synchronous allocator is used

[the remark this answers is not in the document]
cjnolet: The more I think about this one, the more I'm realizing this might be an indicator of a bad code smell. This doesn't really fit the patterns used in the other primitives and really what it means is that the LARS solver itself should be in RAFT (which is the the plan) and this primitive should really be an implementation detail. 

Maybe we should skip mdspanifying this function altogether and leave a note in the original code that it's not moving to the new API.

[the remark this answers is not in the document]
cjnolet: >  Allocating and de-allocating a workspace every iteration will severely slow this function down though if a synchronous allocator is used

The idea behind attaching an optional allocator to the handle is that a pool allocator could be used for temporary workspace allocations in the primitives while a non-pool allocator (could be device synchronous, asynchronous, or stream synchronous, really) could be used everywhere else globally. That type of design would keep the knowledge of workspaces out of the public APIs while still giving the control over the allocations to the user.

cpp/include/raft/linalg/eig.cuh
cjnolet: You can maybe say `error tolerance for the jacobi method. Algorithm stops when the Frobenius norm of the absolute error is below tol`

cpp/include/raft/linalg/rsvd.cuh
```diff
@@ -255,15 +412,188 @@ void rsvd_perc(
...
+template <typename ValueType, typename IndexType>
+void rsvd_perc_symmetric(
+  const raft::handle_t& handle,
+  raft::device_matrix_view<ValueType, IndexType, raft::col_major> M,
+  raft::device_vector_view<ValueType, IndexType> S_vec,
+  ValueType PC_perc,
+  ValueType UpS_perc,
+  ValueType tol,
+  int max_sweeps,
+  std::optional<raft::device_matrix_view<ValueType, IndexType, raft::col_major>> U = std::nullopt,
```
mhoemmen, line 449: Note that if users provide `nullopt` explicitly as an argument, sometimes the compiler has trouble deducing types.  It can be worth adding a test to make sure that compiles.  If not, there are a couple different ways to fix it.  My PR #830 takes one approach that can handle multiple `optional` arguments.

divyegala: Doesn't seem like we are deducing the type here though, are we? I have explicitly specified which type is expected

cpp/include/raft/linalg/cholesky_r1_update.cuh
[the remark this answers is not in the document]
divyegala: Sounds good, I can remove the API then

cpp/include/raft/linalg/rsvd.cuh
mhoemmen: @divyegala Surprisingly, this doesn't always work.  It could be worth writing a little test that passes in `std::nullopt`.  If it compiles, you're fine.

divyegala: Understood! Yes, it does fail the compilation. On that note, I figured out that your solution is a little bit more complicated than necessary. The aliases and type-traits are not required because `std::nullopt` arguments will always go in the overloaded function. From there on, you can directly construct the optional and call the first overload.

mhoemmen: @divyegala wrote:
> On that note, I figured out that your solution is a little bit more complicated than necessary. The aliases and type-traits are not required because std::nullopt arguments will always go in the overloaded function. 

The benefit of putting the optional arguments at the end and defaulting them to `nullopt` : - )

cpp/include/raft/linalg/map.cuh
```diff
@@ -37,17 +40,63 @@ namespace linalg {
  * @param in the input array
  * @param args additional input arrays
  */
+template <typename InType,
```
cjnolet, line 43: I think this eventually needs to go into `raft::matrix` as well.

cpp/include/raft/linalg/map_then_reduce.cuh
```diff
@@ -66,25 +67,27 @@ void mapThenSumReduce(
...
           typename... Args>
-void mapThenReduce(OutType* out,
-                   size_t len,
-                   OutType neutral,
-                   MapOp map,
-                   ReduceLambda op,
-                   cudaStream_t stream,
-                   const InType* in,
-                   Args... args)
+[[deprecated("Use function `mapReduce` from `raft/linalg/map_reduce.cuh")]] void mapThenReduce(
```
cjnolet, line 77: Do you know if `breathe` / `doxygen` support this?

cpp/include/raft/linalg/reduce_cols_by_key.cuh
cjnolet: Still waiting on todo / github issue here

cpp/include/raft/linalg/reduce_rows_by_key.cuh
```diff
@@ -108,6 +111,69 @@ void reduce_rows_by_key(const DataIteratorT d_A,
...
+ *
+ * @tparam ElementType data-type of input and output
+ * @tparam KeyType data-type of keys
+ * @tparam WeightType data-type of weights
+ * @tparam IndexType index type
+ * @param[in]  handle      raft::handle_t
+ * @param[in]  d_A         Input raft::device_mdspan (ncols * nrows)
+ * @param[in]  d_keys      Keys for each row raft::device_vector_view (1 x nrows)
+ * @param[out] d_sums      Row sums by key raft::device_matrix_view (ncols x d_keys)
+ * @param[in]  nkeys       Number of unique keys in d_keys
```
cjnolet, line 130: Maybe we should rename this to `n_unique_keys`

cpp/include/raft/linalg/rsvd.cuh
```diff
@@ -137,6 +139,667 @@ void rsvdPerc(const raft::handle_t& handle,
...
+ * @param[out] V optional right singular values of raft::device_matrix_view with layout
+ * raft::col_major
+ */
+template <typename ValueType, typename IndexType>
+void rsvd_fixed_rank_symmetric(
+  const raft::handle_t& handle,
+  raft::device_matrix_view<ValueType, IndexType, raft::col_major> M,
+  raft::device_vector_view<ValueType, IndexType> S_vec,
+  IndexType p,
+  ValueType tol,
```
cjnolet, line 248: Do we still need this argument for the non-jacobi versions?

cpp/include/raft/linalg/rsvd.cuh
```diff
@@ -137,6 +139,667 @@ void rsvdPerc(const raft::handle_t& handle,
...
+ * raft::col_major
+ */
+template <typename ValueType, typename IndexType>
+void rsvd_perc_symmetric(
+  const raft::handle_t& handle,
+  raft::device_matrix_view<ValueType, IndexType, raft::col_major> M,
+  raft::device_vector_view<ValueType, IndexType> S_vec,
+  ValueType PC_perc,
+  ValueType UpS_perc,
+  ValueType tol,
```
cjnolet, line 574: Same here- do we still need the tolerance for non-jacobi overloads?

cpp/include/raft/linalg/strided_reduction.cuh
cjnolet: Maybe we should do a separate overload for the inplace version like we do in `raft::matrix`. That makes it much more straightforward to the user that it's in place (and which array(s) end up being overwriten).

cpp/include/raft/linalg/strided_reduction.cuh
```diff
@@ -71,6 +74,90 @@ void stridedReduction(OutType* dots,
...
+ */
+template <typename InElementType,
+          typename LayoutPolicy,
+          typename OutElementType = InElementType,
+          typename IndexType      = std::uint32_t,
+          typename MainLambda     = raft::Nop<InElementType>,
+          typename ReduceLambda   = raft::Sum<OutElementType>,
+          typename FinalLambda    = raft::Nop<OutElementType>>
+void strided_reduction(const raft::handle_t& handle,
+                       const raft::device_matrix_view<InElementType, IndexType, LayoutPolicy> data,
```
cjnolet, line 120: I think this should be `const InElementType` right?

cpp/include/raft/linalg/subtract.cuh
```diff
@@ -84,6 +88,137 @@ void subtractDevScalar(math_t* outDev,
...
+ * @tparam OutType   Output Type raft::device_mdspan
+ * @param handle raft::handle_t
+ * @param out    Output
+ * @param in1    First Input
+ * @param in2    Second Input
+ */
+template <typename InType,
+          typename OutType,
+          typename = raft::enable_if_device_mdspan<OutType, InType>>
+void subtract(const raft::handle_t& handle, InType in1, InType in2, OutType out)
```
cjnolet, line 108: I notice the const missing here. Can we still declare the inputs `mdspan<const value_t>` somehow?

cpp/include/raft/linalg/subtract.cuh
```diff
@@ -84,6 +88,137 @@ void subtractDevScalar(math_t* outDev,
...
+ * @defgroup sub Subtraction Arithmetic
+ * @{
+ */
+
+/**
+ * @brief Elementwise subtraction operation on the input buffers
+ * @tparam InType    Input Type raft::device_mdspan
+ * @tparam OutType   Output Type raft::device_mdspan
+ * @param handle raft::handle_t
+ * @param out    Output
```
cjnolet, line 101: Docs need `[in]`, [out]` across the PR.

cpp/include/raft/linalg/svd.cuh
```diff
@@ -182,6 +181,218 @@ bool evaluateSVDByL2Norm(const raft::handle_t& handle,
...
+ * @param[out] sing_vals singular values raft::device_vector_view of shape (K)
+ * @param[out] left_sing_vecs optional left singular values of raft::device_matrix_view with layout
+ * raft::col_major and dimensions (m, n)
+ * @param[out] right_sing_vecs optional right singular values of raft::device_matrix_view with
+ * layout raft::col_major and dimensions (n, n)
+ */
+template <typename ValueType, typename IndexType>
+void svd_qr(
+  const raft::handle_t& handle,
+  raft::device_matrix_view<ValueType, IndexType, raft::col_major> in,
```
cjnolet, line 203: Use `const ValueType`. This applies throughout the PR.

cjnolet (changes requested):
Looks like this is coming across the final bend. Really minor things at this point.

cpp/include/raft/linalg/subtract.cuh
mhoemmen: @cjnolet Two issues here:

1. how to make sure that the inputs are `mdspan<const T>` for some `T`; and
2. the function should, but does not currently, constrain the mdspan all to have the same rank; and
3. the function should, but does not currently, constrain the mdspan to have all the same layouts.

Regarding (1), there are a few different ways to do this.  For example, we could constrain the parameters to be `mdspan` whose `element_type` is `const`.

```c++
template<class T>
constexpr bool is_const_mdspan_v = false;

constexpr bool is_const_mdspan_v<mdspan<ElementType, Extents, Layout, Accessor>> =
  std::is_const_v<ElementType>;

template <typename InType,
  typename OutType,
  typename std::enable_if<
    is_device_mdspan<InType> &&
      is_device_mdspan<OutType> &&
      is_const_mdspan_v<InType>,
    bool>::type = true>
void subtract(const raft::handle_t& handle, InType in1, InType in2, OutType out);
```

This style is concise, and good choice of traits names can help document the interface.  However, use of SFINAE for overload resolution may also increase compilation time.  Another approach would be to overload with `mdspan` whose accessors are device accessible.  It's more verbose, but self-documenting, and it can avoid issues with traits classes needing to do `std::remove_const_t<std::remove_cv_t<T>>`.

```c++
template<class ElementType, class Extents, class Layout, class NestedAccessor, bool host_accessible>
using device_accessible_mdspan =
  mdspan<ElementType, Extents, Layout, host_device_accessor<NestedAccessor, host_accessible, true>>;

template<
  class InValueType1, class InExtents1, class InLayout1, class InNestedAccessor1, bool input1_host_accessible,
  class InValueType2, class InExtents2, class InLayout2, class InNestedAccessor2, bool input2_host_accessible,
  class OutValueType, class OutExtents, class OutLayout, class OutNestedAccessor, bool output_host_accessible>
void subtract(const raft::handle_t& handle,
  device_accessible_mdspan<const InValueType1, InExtents1, InLayout1, InNestedAccessor1, input1_host_accessible> in1,
  device_accessible_mdspan<const InValueType2, InExtents2, InLayout2, InNestedAccessor2, input2_host_accessible> in2,
  device_accessible_mdspan<OutValueType, OutExtents, OutLayout, OutNestedAccessor, output_host_accessible> out);
```

You might like to document that users should not provide template arguments explicitly.  That would give you implementation freedom to change the template parameters later, for example if you want to redefine `host_device_accessor` in the future.

cjnolet: Thanks @mhoemmen! I agree that I don't see a reason these functions couldn't constrain the ins and outs to having the same layouts and indexing types, but your second example is exactly what we were trying to avoid with all of this (having to accept all the template parameters individually for each input).  I think your first example might not be a bad trade-off between forcing the inputs to be const qualified while still being agnostic of the layouts and indexing types and not bloating the number of template params. It also looks really clean, as you point out. 

Just out of curiosity, how much would you expect the SFINAE in your first example to slow down the compile time in comparison to the current approach (which just does the `enable_if_mdspan`?

cpp/include/raft/linalg/map_then_reduce.cuh
divyegala: Not sure but even Mark has used it in his `sampleWithReplacement` PR

cpp/include/raft/linalg/coalesced_reduction.cuh
```diff
@@ -70,6 +73,90 @@ void coalescedReduction(OutType* dots,
...
+          typename OutValueType,
+          typename IndexType,
+          typename MainLambda   = raft::Nop<InValueType>,
+          typename ReduceLambda = raft::Sum<OutValueType>,
+          typename FinalLambda  = raft::Nop<OutValueType>>
+void coalesced_reduction(const raft::handle_t& handle,
+                         raft::device_matrix_view<const InValueType, IndexType, LayoutPolicy> data,
+                         raft::device_vector_view<OutValueType, IndexType> dots,
+                         OutValueType init,
+                         bool inplace           = false,
```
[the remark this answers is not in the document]
divyegala: Should I do the separate overload in this PR or wait for the move to `raft::matrix`?

cpp/include/raft/linalg/subtract.cuh
divyegala: Why do we need to force the inputs to be const-qualified though? The interface they eventually call to translate to const pointers

divyegala: @mhoemmen @cjnolet instead of doing the device mdspan and const mdspan checks separately, I can just chain them together so we don't do extra work by having `enable_if_input_device_mdspan` and `enable_if_output_device_mdspan`. The first one enforces const-ness and the second one enforces the inverse

cpp/include/raft/linalg/svd.cuh
divyegala: Many of these svd/eig functions that use cusolver don't have const-qualifers even on the original pointers-based API. I *think* it is because cusolver doesn't accept const-qualified pointers. What to do?

cpp/include/raft/linalg/coalesced_reduction.cuh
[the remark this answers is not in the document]
divyegala: I also really don't think `inplace` means what we think it means here. Looking at the kernel, it seems like all `inplace=true` does is run the `reduction_op` one more time over the output pointer.

cpp/include/raft/linalg/svd.cuh
cjnolet: Cast the constness away underneath but use it on the outermost bits. This future-proofs the API as well.

cjnolet (approved):

cjnolet:
@gpucibot merge

PR #810, opened 2022-09-07
Separating mdspan/mdarray infra into host_* and device_* variants

### LINKED ISSUE #806: [FEA] Separate device_mdspan and host_mdspan from mdarray.hpp into device_mdspan.hpp and host_mdspan.hpp
We are doing this for the purpose of 1) making the includes more specific to mdspan (rather than putting everything in mdarray.hpp as they are currently, and 2) allowing the host_mdspan to be included and used without ever having to know anything about CUDA (meaning it can be compiled with gcc and doesn't require any runtime libs or headers). 

It's also important the corresponding accessors be moved into these files.

cjnolet:
This is a breaking change as it provides users with a more granular set of headers to import `host` separately from `device` and `managed` versions. It also separates the headers for `mdspan` and `mdarray`. 

As an example, the following public headers can now be imported individually:
```c++
raft/core/host_mdspan.hpp
raft/core/device_mdspan.hpp

raft/core/host_mdarray.hpp
raft/core/device_mdarray.hpp
```

cc @rg20 @afender @akifcorduk for awareness. 

Closes #806


divyegala (approved):
Just a question about `.cuh` vs `.hpp` file extensions, and whether `device_` files should be `.cuh`?

cjnolet:
@divyegala I'm on the fence about whether the `device_md{span|array}` files should be in `cuh` files. On one hand, the `default_accessor` for the reference `mdspan` has a `__device__` function but it's still in an `hpp` file. I think so long as the `__device__` functions are kept trivial (like just for data loading), i'm inclined to keep the files as `hpp` for consistency. We should also be using a macro that can properly disable the `__device__` in the event the compiler is not `NVCC`. 

But, for example, if we start supplying more complex `__device__` primitives that actually perform complex computations on device, or that directly use device intrinsics, I'd say we should opt to put those in files with the `cuh` extension (such as in `mdspan_util.cuh`.

cpp/include/raft/core/device_mdspan.hpp
```diff
@@ -0,0 +1,194 @@
...
+#include <raft/core/detail/accessor_mixin.hpp>
+#include <raft/core/mdspan.hpp>
+
+namespace raft {
+
+template <typename AccessorPolicy>
+using device_accessor = detail::accessor_mixin<AccessorPolicy, false, true>;
+
+template <typename AccessorPolicy>
+using managed_accessor = detail::accessor_mixin<AccessorPolicy, true, true>;
```
mhoemmen, line 28: `accessor_mixin<AccessorPolicy, true, true>` just implies that the memory is accessible from either host or device.  It could be either pinned memory or managed memory.  Should we thus consider a different name for this alias, such as `host_device_accessor`?

cpp/include/raft/core/device_mdspan.hpp
```diff
@@ -0,0 +1,194 @@
...
+ * @param[in] ptr on device to wrap
+ * @param[in] n number of elements in pointer
+ * @return raft::device_vector_view
+ */
+template <typename ElementType,
+          typename IndexType    = std::uint32_t,
+          typename LayoutPolicy = layout_c_contiguous>
+auto make_device_vector_view(ElementType* ptr, IndexType n)
+{
+  vector_extent<IndexType> extents{n};
```
mhoemmen, line 190: You can also construct the mdspan from `n` directly; you don't need to create an intermediate `extents` object.  Would you then consider just removing `make_device_vector_view`?

cpp/include/raft/core/device_span.hpp
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+#pragma once
+
+#include <raft/core/span.hpp>
+
+namespace raft {
+
+/**
+ * @brief A span class for device pointer.
```
mhoemmen, line 24: Given that `std::span` is _just_ a pointer and a length, with no accessor or other information, does it still make sense for RAFT to have a `span` class separate from `mdspan`?

cpp/include/raft/core/host_mdarray.hpp
```diff
@@ -0,0 +1,146 @@
...
+ * @tparam LayoutPolicy policy for strides and layout ordering
+ * @param[in] n number of elements in vector
+ * @return raft::host_vector
+ */
+template <typename ElementType,
+          typename IndexType    = std::uint32_t,
+          typename LayoutPolicy = layout_c_contiguous>
+auto make_host_vector(IndexType n)
+{
+  return make_host_mdarray<ElementType, IndexType, LayoutPolicy>(make_extents<IndexType>(n));
```
mhoemmen, line 143: Please see above note: you should just be able to pass in `n` directly here.

cpp/include/raft/core/host_mdspan.hpp
```diff
@@ -0,0 +1,143 @@
...
+          typename AccessorPolicy = std::experimental::default_accessor<ElementType>>
+using host_mdspan = mdspan<ElementType, Extents, LayoutPolicy, host_accessor<AccessorPolicy>>;
+
+namespace detail {
+
+template <typename T, bool B>
+struct is_host_mdspan : std::false_type {
+};
+template <typename T>
+struct is_host_mdspan<T, true> : std::bool_constant<T::accessor_type::is_host_accessible> {
```
mhoemmen, line 43: `is_host_mdspan<managed_mdspan<...>>::value` is `true` as well.  Is that OK?  If so, would you consider perhaps calling this trait something else, like `is_host_accessible_mdspan`?  It could be confusing that `is_host_mdspan<T>::value` could be `true` for things that aren't `host_mdspan`.

cpp/include/raft/core/mdspan.hpp
```diff
@@ -22,4 +15,212 @@
...
+ */
+
+template <typename ElementType, typename Extents, typename LayoutPolicy, typename AccessorPolicy>
+void __takes_an_mdspan_ptr(mdspan<ElementType, Extents, LayoutPolicy, AccessorPolicy>*);
+
+template <typename T, typename = void>
+struct is_mdspan : std::false_type {
+};
+template <typename T>
+struct is_mdspan<T, std::void_t<decltype(__takes_an_mdspan_ptr(std::declval<T*>()))>>
```
mhoemmen, line 55: Would you consider a simpler design that just takes a single template parameter T and specializes on it?

```c++
template<typename T>
struct is_mdspan : std::false_type {};

template<typename ElementType, typename Extents, typename Layout, typename Accessor>
struct is_mdspan<mdspan<ElementType, Extents, Layout, Accessor>> : std::true_type {};
```

cpp/include/raft/core/mdspan.hpp
```diff
@@ -22,4 +15,212 @@
...
+/**
+ * @brief Create raft::extents to specify dimensionality
+ *
+ * @tparam IndexType The type of each dimension of the extents
+ * @tparam Extents Dimensions (a series of integers)
+ * @param exts The desired dimensions
+ * @return raft::extents
+ */
+template <typename IndexType, typename... Extents, typename = ensure_integral_extents<Extents...>>
+auto make_extents(Extents... exts)
```
mhoemmen, line 136: It looks like this does the same thing as `dextents`.  If so, would you consider just using that?

mhoemmen (approved):

cpp/include/raft/core/device_mdspan.hpp
cjnolet: Ah, that's a nice simplification. I'll remove the need to create the extents here but I still like the explicit `vector_view` (and corresponding factory function) for consistency.

cpp/include/raft/core/device_span.hpp
cjnolet: I've brought up in the past, actually, that a `span` and `mdspan` are similar enough that we could probably just specialize the mdspan into a span. I think for now, while we have users downstream still using span, we should keep it in the meantime while we continue the discussion and consolidate in a future follow-up.

cpp/include/raft/core/mdspan.hpp
```diff
@@ -22,4 +15,212 @@
...
+ *
+ * @tparam mdspan_type Expected type raft::host_mdspan or raft::device_mdspan
+ * @param mds raft::host_mdspan or raft::device_mdspan object
+ * @return raft::host_mdspan or raft::device_mdspan with vector_extent
+ *         depending on AccessoryPolicy
+ */
+template <typename mdspan_type, typename = enable_if_mdspan<mdspan_type>>
+auto flatten(mdspan_type mds)
+{
+  RAFT_EXPECTS(mds.is_exhaustive(), "Input must be contiguous.");
```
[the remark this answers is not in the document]
cjnolet: Can you provide an example of how we might force this to be row or col major?

cjnolet:
@mhoemmen, thank you for this thorough review! I've implemented most of your suggestions. I wasn't able to get [this](https://github.com/rapidsai/raft/pull/810#discussion_r966335111) suggestion or [this](https://github.com/rapidsai/raft/pull/810#discussion_r966336529) suggestion building, though most of the changes in this PR are just moving stuff so I think we can continue to make things more clean and correct in follow-on PRs.  Would you be interested in opening a PR w/ those changes?

cjnolet:
rerun tests

cpp/include/raft/core/device_span.hpp
cjnolet: Created issue to continue the discussion more formally w/ different stakeholders: https://github.com/rapidsai/raft/issues/817

cpp/include/raft/core/mdspan.hpp
cjnolet: Created for this: https://github.com/rapidsai/raft/issues/818

cpp/include/raft/core/mdspan.hpp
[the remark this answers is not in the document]
mhoemmen: Here's a way to check if an mdspan is row or column major.
```c++
constexpr bool is_row_or_column_major(mdspan<ElementType, Extents, Layout, Accessor> /* m */ )
{
  return false;
}

template<class ElementType, class Extents, class Accessor>
constexpr bool is_row_or_column_major(mdspan<ElementType, Extents, layout_left, Accessor> /* m */ )
{
  return true;
}

template<class ElementType, class Extents, class Accessor>
constexpr bool is_row_or_column_major(mdspan<ElementType, Extents, layout_right, Accessor> /* m */ )
{
  return true;
}

template<class ElementType, class Extents, class Accessor>
constexpr bool is_row_or_column_major(mdspan<ElementType, Extents, layout_stride, Accessor> m)
{
  return m.is_exhaustive();
}
```

(Relevant issue that you posted; thank you! https://github.com/rapidsai/raft/issues/819 )

cpp/include/raft/core/detail/macros.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+#ifndef _RAFT_HAS_CUDA
+#if defined(__CUDACC__)
+#define _RAFT_HAS_CUDA __CUDACC__
+#endif
+#endif
+
+#ifndef _RAFT_HOST_DEVICE
+#if defined(_RAFT_HAS_CUDA)
+#define _RAFT_HOST_DEVICE __host__ __device__
```
wphicks, line 27: Would we consider separating this into `_RAFT_HOST` and `_RAFT_DEVICE`?

cpp/include/raft/core/detail/macros.hpp
```diff
@@ -0,0 +1,35 @@
...
+#else
+#define _RAFT_HOST_DEVICE
+#endif
+#endif
+
+#ifndef RAFT_INLINE_FUNCTION
+#define RAFT_INLINE_FUNCTION inline _RAFT_HOST_DEVICE
```
wphicks, line 34: Was there any particular use-case in mind for this macro where it provides benefit beyond invoking `inline _RAFT_HOST_DEVICE` directly? I could see an argument for introducing a `RAFT_FORCE_INLINE` so that we conditionally make use of the CUDA-specific `__force_inline__`, but it's less obvious to me what the benefit is in tacking on an ordinary `inline` here.

wphicks (approved):
Love this! Had two questions on details, but this is a huge step forward for making use of mdspan in the FIL backend. I was not able to _absolutely_ confirm that there were no gotchas with the CPU-only compilation, but everything looks good as best I can tell without finishing a version of the FIL code that makes use of this.

cpp/include/raft/core/detail/macros.hpp
cjnolet: This is more for general convenience. It was also being used in our `mdarray` implementation (copied over from `mdspan`, actually) so I figured I'd include it in this refactor (but redefine it as well since we're already redefining the conditional pieces).

cpp/include/raft/core/detail/macros.hpp
cjnolet: I can definitely split these out but I'd prefer to do it in a follow-on only because I'm trying to orchestrate this merge w/ cugraph and cuopt (the changes break them downstream) and I'm hoping not to have to re-run CI if I can avoid it.

cjnolet:
@gpucibot merge

PR #820, opened 2022-09-09
Separating test executables

cjnolet:
Other RAPIDS projects have been doing this for awhile. It's time RAFT did this as well.


robertmaynard (approved):

raydouglass (approved):

cpp/test/CMakeLists.txt
```diff
@@ -14,172 +14,308 @@
...
+endfunction()
+
+
+function(ConfigureDistanceTest CMAKE_TEST_NAME)
+    add_executable(${CMAKE_TEST_NAME} ${ARGN})
+
+    target_link_libraries(${CMAKE_TEST_NAME}
+            PRIVATE
+            raft::raft
+            raft::distance
```
dantegd, line 68: I think you could avoid the definitions of `ConfigureDistanceTest` and `ConfigureNearestNeighborsTest` by adding boolean parameters to `ConfigureTest`, which could save a lot of lines of code in this file.

dantegd (changes requested):

cpp/test/CMakeLists.txt
dantegd: We do that in cuML for nccl, cumlprims amd others in https://github.com/rapidsai/cuml/blob/6f42ced4d6f84cc04ea27f8940449d57fa08ab51/cpp/test/CMakeLists.txt#L53 And https://github.com/rapidsai/cuml/blob/6f42ced4d6f84cc04ea27f8940449d57fa08ab51/cpp/test/CMakeLists.txt#L19

cjnolet: Ah, yes that is much cleaner indeed. I've adjusted the cmake accordingly.

build.sh
```diff
@@ -194,7 +194,7 @@ if hasArg tests || (( ${NUMARGS} == 0 )); then
     COMPILE_DIST_LIBRARY=ON
     ENABLE_NN_DEPENDENCIES=ON
     COMPILE_NN_LIBRARY=ON
-    CMAKE_TARGET="${CMAKE_TARGET};test_raft"
+    CMAKE_TARGET="${CMAKE_TARGET};CLUSTER_TEST;CORE_TEST;DISTANCE_TEST;LABEL_TEST;LINALG_TEST;MATRIX_TEST;RANDOM_TEST;SOLVERS_TEST;SPARSE_TEST;SPARSE_DIST_TEST;SPARSE_NN_TEST;SPATIAL_TEST;STATS_TEST;UTILS_TEST"
```
cjnolet, line 197: @robertmaynard I'm hoping to avoid having to do this. The problem is, `build.sh` is explicitly specifying the artifact that it's expecting to create by manually specifying the target(s) which are built based on the given options. When the tests are split into separate executables, I now need to specify those targets here- so a single `test_raft` target got split into several smaller targets.

Eventually, I'd like to add an option to `build.sh` to manually specify which tests get built, but in the meantime this is pretty verbose and it's a chicken and egg problem- the `build.sh` script doesn't know the targets at the time it invokes cmake so they need to specified manually, but the targets are defined within cmake itself. Do you have any suggestions on how I could avoid such a pattern here? Is there a way I could just use something similar to `--component testing` to specify all the targets marked testing? (`--component testing` seems to be defined only in the `install()` function but is there a similar pattern I could use to reference the full list of executable targets?)

cjnolet:
@gpucibot merge

PR #821, opened 2022-09-12
Use fusedL2NN in ANN kmeans

Nyrio:
The bottleneck of `ivf_flat::build` is the label prediction, currently done with GEMM + argmin. This PR gives a 25-30% speedup by replacing it with the fused primitive.


cpp/include/raft/spatial/knn/detail/ann_utils.cuh
```diff
@@ -151,17 +151,18 @@ inline void memzero(T* ptr, size_t n_elems, rmm::cuda_stream_view stream)
...
+template <typename LabelT>
 __global__ void argmin_along_rows_kernel(uint32_t n_rows,
                                          uint32_t n_cols,
                                          const float* a,
-                                         uint32_t* out)
+                                         LabelT* out)
 {
-  __shared__ uint32_t shm_ids[1024];  // NOLINT
-  __shared__ float shm_vals[1024];    // NOLINT
+  __shared__ LabelT shm_ids[1024];  // NOLINT
```
achirkin, line 160: This probably will break as soon as we more than one instantiation of the template. Consider changing the type to `uint8_t` and then `reinterpret_cast` it to `LabelT`.

cpp/include/raft/spatial/knn/detail/ann_utils.cuh
```diff
@@ -253,19 +257,19 @@ inline void dots_along_rows(
...
                                                 const T* input,
-                                                const uint32_t* row_ids)
+                                                const LabelT* row_ids)
 {
   uint64_t gid = threadIdx.x + (blockDim.x * blockIdx.x);
   uint64_t j   = gid % n_cols;
   uint64_t i   = gid / n_cols;
   if (i >= n_rows) return;
-  uint64_t l = row_ids[i];
+  uint64_t l = (uint64_t)row_ids[i];
```
achirkin, line 272: Consider not using C-style conversion (clang-tidy in our config should have generated a warning for this).

cpp/include/raft/spatial/knn/detail/ann_utils.cuh
```diff
@@ -305,7 +310,7 @@ void accumulate_into_selected(size_t n_rows,
     case pointer_residency::host_only: {
       stream.synchronize();
       for (size_t i = 0; i < n_rows; i++) {
-        uint32_t l = row_ids[i];
+        uint32_t l = (uint32_t)row_ids[i];
```
achirkin, line 313: Consider not using C-style conversion (clang-tidy in our config should have generated a warning for this).

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
```diff
@@ -791,27 +905,28 @@ void build_hierarchical(const handle_t& handle,
...
+                                                              mesocluster_size_max,
+                                                              fine_clusters_nums_max,
+                                                              cluster_centers,
+                                                              metric,
+                                                              &managed_memory,
+                                                              device_memory,
+                                                              stream);
   RAFT_EXPECTS(n_clusters_done == n_clusters, "Didn't process all clusters.");
 
   rmm::device_uvector<uint32_t> cluster_sizes(n_clusters, stream, device_memory);
```
achirkin, line 928: `cluster_sizes` and, in general, all arrays containing sizes/offsets into the clusters should use the same type as labels, i.e. `LabelT`. Originally, the premise for this module was that the maximum cluster size fits in `uint32_t`; now we assume it can be anything that `LabelT` can represent, right?

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
```diff
@@ -45,67 +49,105 @@ constexpr static inline const float kAdjustCentersWeight = 7.0f;
...
+    // todo(lsugy): this is temporary!
+    // Copy keys to output labels
+    thrust::transform(handle.get_thrust_policy(),
+                      minClusterAndDistance.data_handle(),
+                      minClusterAndDistance.data_handle() + n_rows,
+                      labels,
+                      [=] __device__(cub::KeyValuePair<LabelT, float> kvp) { return kvp.key; });
+
+    // todo(lsugy): return distances!
+  } else if (metric == raft::distance::DistanceType::InnerProduct) {
```
achirkin, line 125: Would you consider using `switch` statement instead of multiple `if` when matching against enums? It feels more explicit.

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
```diff
@@ -45,67 +49,105 @@ constexpr static inline const float kAdjustCentersWeight = 7.0f;
...
  * @param mr (optional) memory resource to use for temporary allocations
  */
+template <typename IdxT, typename LabelT>
 inline void predict_float_core(const handle_t& handle,
                                const float* centers,
                                uint32_t n_clusters,
                                uint32_t dim,
                                const float* dataset,
+                               const float* dataset_norm,
                                size_t n_rows,
```
achirkin, line 73: Should `n_rows` and all variables/arrays referring to the dataset indices/sizes be the same type as indices, i.e. `IdxT`?

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
```diff
@@ -522,45 +574,47 @@ void balancing_em_iters(const handle_t& handle,
 }
 
 /** Randomly initialize cluster centers and then call `balancing_em_iters`. */
-template <typename T>
+template <typename T, typename IdxT, typename LabelT>
 void build_clusters(const handle_t& handle,
```
achirkin, line 578: Please provide the compatible overload/wrapper for `predict` and `build_clusters` with the previous api (with no `dataset_norm` argument). I have a number of usages of these two in #789 scattered across the build module.

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
```diff
@@ -730,7 +832,7 @@ auto build_fine_clusters(const handle_t& handle,
  * @param metric the distance type
  * @param stream
  */
-template <typename T>
+template <typename T, typename IdxT, typename LabelT = int>
```
achirkin, line 835: Would you consider removing the default value for `LabelT` and move it to the first parameter? Since this is the only one that cannot be inferred. Gladly, it could only silently break the code in my draft PR at the moment, but still :)

achirkin (changes requested):
That's very nice speedup! Two main points:

  1.  Would you consider to templatize the `fusedL2NN` to allow unsigned labels instead? It feels like more internals of ann_kmeans there would still need to be changed to properly switch to `LabelT` if we go this way.
  2. Out of curiosity, did you check if the `argmin_along_rows` could be optimized to achieve the same speedup not fused? I remember it looked awfully slow in some settings in the nsys timeline when I tested it...

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
Nyrio: So, in those cases `predict` is only invoked once with the same dataset, hence it doesn't make any difference to compute the norm in `predict` as opposed to pre-computing?

Nyrio:
> Would you consider to templatize the fusedL2NN to allow unsigned labels instead? It feels like more internals of ann_kmeans there would still need to be changed to properly switch to LabelT if we go this way.

Which internals of ann kmeans still need to be changed? The IVF-Flat tests are working at least. I'll look into `fusedL2NN` to see if I can simply add support to unsigned label types, but it doesn't hurt to have the flexibility in the ann kmeans code, even if we end up only instantiating one type (be it `uint32_t` or `int`).

> Out of curiosity, did you check if the argmin_along_rows could be optimized to achieve the same speedup not fused? I remember it looked awfully slow in some settings in the nsys timeline when I tested it...

I'll look into this, but by my quick calculation, I would have to speed up `argmin_along_rows` by a factor of at least 10 to achieve similar performance.

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
Nyrio: I don't see why it should be `LabelT`. It would make more sense to use `IdxT` since the cluster size is proportional to the number of samples and inversely proportional to the number of clusters.

But it's also fair to assume that the cluster size will not exceed the range of `uint32_t` and that using a larger type would be inefficient. At the moment, cluster sizes are order of magnitudes below that limit in the worst case.

Nyrio: It would almost make sense to have another index type for this, but it might get confusing if we have tons of different index types.

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
Nyrio: Yes, I could do that.
But my thinking was that it's actually better if the caller doesn't change that, at least for now. I could also have a `using` statement rather than a template parameter.

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
achirkin: Ah, you're right, I mixed these up again. This is the number of clusters what should have the `LabelT` type, not the cluster sizes.

achirkin:
> Which internals of ann kmeans still need to be changed?

It's the things like [`fine_clusters_csum`](https://github.com/rapidsai/raft/blob/4b497ff4cb04060573001e9c97a4d7e402df32d4/cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh#L627) variable, which are not obvious and do not raise a compile-time errors. It's not always clear which element type should they have, what makes me hesitant to introduce more index types in the code. I've had already one fix for something there in the merged follow-up pr :)

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
achirkin: Yep. Depending on the mode of operation, there is a loop over either clusters or subspaces of the original/transformed dataset.

Nyrio:
> It's the things like [fine_clusters_csum](https://github.com/rapidsai/raft/blob/4b497ff4cb04060573001e9c97a4d7e402df32d4/cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh#L627) variable, which are not obvious and do not raise a compile-time errors. It's not always clear which element type should they have, what makes me hesitant to introduce more index types in the code.

Yes, it's not always clear which index types should be used, and I tried not to change all the types in this PR but only those related the labels. `fine_clusters_csum` is not impacted by the current change so I didn't touch it.

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
Nyrio: That would make sense, but it's quite dangerous to let the type inference depend on a scalar parameter. It's very easy to overlook those types when calling the function and end up with 64-bit labels.

By the way, regarding unsigned label types, it seems that I only have to replace `-1` with `0` and it should work, in which case I can switch `LabelT` back to `uint32_t`. I'll try this out when I'm done debugging some overflows

Nyrio: It's worth a try. Probably not a substantial difference for `fusedL2NN` if `dim` is large, but let's see.

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
```diff
@@ -45,67 +49,109 @@ constexpr static inline const float kAdjustCentersWeight = 7.0f;
...
-      utils::outer_add(
-        sqsum_data.data(), n_rows, sqsum_centers.data(), n_clusters, distances.data(), stream);
-      alpha = -2.0;
-      beta  = 1.0;
-    } break;
-    // NB: update the description of `knn::ivf_flat::build` when adding here a new metric.
-    default: RAFT_FAIL("The chosen distance metric is not supported (%d)", int(metric));
+    case raft::distance::DistanceType::L2SqrtExpanded: {
+      // todo(lsugy): pass
+      auto workspace = raft::make_device_vector<char, IdxT>(handle, (sizeof(int)) * n_rows);
```
tfeher, line 83: We could make use of the memory_resoure argument (alternatively, one could expose an mr arg for make_device_vector).
```suggestion
      auto workspace = raft::make_device_mdarray<char, IdxT>(handle, mr, make_extents<IdxT>((sizeof(int)) * n_rows));
```

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
```diff
@@ -45,67 +49,109 @@ constexpr static inline const float kAdjustCentersWeight = 7.0f;
...
+                   minClusterAndDistance.data_handle(),
+                   minClusterAndDistance.data_handle() + minClusterAndDistance.size(),
+                   initial_value);
+
+      auto centroidsNorm = raft::make_device_vector<float, IdxT>(handle, n_clusters);
+      raft::linalg::rowNorm<float, IdxT>(
+        centroidsNorm.data_handle(), centers, dim, n_clusters, raft::linalg::L2Norm, true, stream);
+
+      // todo(lsugy): should fusedL2NN take IdxT!=LabelT?
+      raft::distance::fusedL2NN<float, cub::KeyValuePair<LabelT, float>, LabelT>(
```
tfeher, line 101: Probably a comment for #795: shouldn't we have a specialization for fusedLNN that selects only the label? That would simplify things here

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
tfeher: +1 for Artem's request, since it simplifies the code on the call site.

tfeher (changes requested):
Thanks @Nyrio for the PR! It is nice to see this improvement in the balanced kmeans imlementation. I have a few comments, and I see that some items for @achirkin's review are still pending, please have a look.

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
Nyrio: I'm not sure if it's that easy to implement, since you can't really define a reduction operation between two labels without a distance.

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
Nyrio: For now, I left `dim` and `n_clusters` in `uint32_t`, replaced `uint64_t` with `IdxT`, and solved a few overflows (I've tested with very large datasets to debug overflows).

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
```diff
@@ -45,67 +49,111 @@ constexpr static inline const float kAdjustCentersWeight = 7.0f;
...
-    default: RAFT_FAIL("The chosen distance metric is not supported (%d)", int(metric));
+    case raft::distance::DistanceType::L2SqrtExpanded: {
+      auto workspace = raft::make_device_mdarray<char, IdxT>(
+        handle, mr, make_extents<IdxT>((sizeof(int)) * n_rows));
+
+      raft::distance::MinAndDistanceReduceOp<IdxT, float> redOp;
+      raft::distance::KVPMinReduce<IdxT, float> pairRedOp;
+
+      auto minClusterAndDistance =
+        raft::make_device_vector<cub::KeyValuePair<IdxT, float>, IdxT>(handle, n_rows);
```
tfeher, line 90: Please switch to `make_device_mdarray` and pass the memory resource arguments.

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
```diff
@@ -45,67 +49,111 @@ constexpr static inline const float kAdjustCentersWeight = 7.0f;
...
+      cub::KeyValuePair<IdxT, float> initial_value(0, std::numeric_limits<float>::max());
+      thrust::fill(handle.get_thrust_policy(),
+                   minClusterAndDistance.data_handle(),
+                   minClusterAndDistance.data_handle() + minClusterAndDistance.size(),
+                   initial_value);
+
+      auto centroidsNorm = raft::make_device_vector<float, IdxT>(handle, n_clusters);
```
tfeher, line 97: Please switch to `make_device_mdarray` and pass the memory resource arguments.

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
```diff
@@ -522,51 +643,54 @@ void balancing_em_iters(const handle_t& handle,
 }
 
 /** Randomly initialize cluster centers and then call `balancing_em_iters`. */
-template <typename T>
+template <typename T, typename IdxT, typename LabelT>
 void build_clusters(const handle_t& handle,
```
tfeher, line 647: Since this is called by external algorithms (ivf-pq), could you add an assert here as well, to check that `n_rows * dim` can be represented with `IdxT`?

tfeher (changes requested):
Thanks Louis, for addressing the issues so far. I have a few more nitpicks, otherwise it looks good.

tfeher (approved):
Thanks Louis for addressing the issues, the PR looks good to me!

@achirkin could you also have another look at the PR?

cjnolet (commented):
I think this PR looks good so far. I'm just reviewing as comments for now since I have a few questions but I don't want to hold up progress.

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
```diff
@@ -689,21 +821,40 @@ auto build_fine_clusters(const handle_t& handle,
                    "Number of fine clusters must be non-zero for a non-empty mesocluster");
     }
 
-    utils::copy_selected(
-      mesocluster_sizes[i], dim, dataset_mptr, mc_trainset_ids, dim, mc_trainset, dim, stream);
+    utils::copy_selected((IdxT)mesocluster_sizes[i],
```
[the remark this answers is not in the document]
achirkin: This one is used all over the place in both new ivf-* implementations and it uses the `mapping` type to transform between input and output along the way (e.g. case uint8_t to float and then divide by 256).

cpp/include/raft/spatial/knn/detail/ann_utils.cuh
```diff
@@ -192,14 +191,18 @@ __global__ void argmin_along_rows_kernel(uint32_t n_rows,
...
+ * @tparam LabelT label type
+ *
  * @param n_rows
  * @param n_cols
  * @param[in] a device pointer to the row-major matrix [n_rows, n_cols]
  * @param[out] out device pointer to the vector of selected indices [n_rows]
  * @param stream
  */
+template <typename IdxT, typename LabelT>
 inline void argmin_along_rows(
```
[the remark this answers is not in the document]
achirkin: I've noted there in the docs that it would be nice to merge this into select-top-k as a special case https://github.com/rapidsai/raft/blob/2e98138c7d022e754e2b2f3bbab377d9d47150c3/cpp/include/raft/spatial/knn/detail/ann_utils.cuh#L193

I suggest I will do this as soon as ivf-pq implementation is merged. Here's the linked issue https://github.com/rapidsai/raft/issues/853

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
```diff
@@ -201,76 +268,125 @@ void calc_centers_and_sizes(float* centers,
...
  * @param n_rows number samples in the `dataset`
  * @param[out] labels output predictions [n_rows]
  * @param metric
  * @param stream
  * @param mr (optional) memory resource to use for temporary allocations
  */
-
-template <typename T>
+template <typename T, typename IdxT, typename LabelT>
 void predict(const handle_t& handle,
```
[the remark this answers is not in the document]
Nyrio: Moving as a separate API sounds feasible. Merging the two APIs requires first using similar primitives, for which I have some ongoing work. I'm not sure which path will be preferred, but if we expose that as a separate API now, it's an API we're also going to have to maintain for a while once we merge it with the existing kmeans.

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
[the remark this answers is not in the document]
Nyrio: > However, in case of ann_kmeans_balanced, I think we'd better use raft::matrix::copy_rows, because we copy indices here rather than values, and logically it has nothing to do with the ann mapping of the value types.

What do you mean "we copy indices"? The two calls of `copy_selected` in `ann_kmeans_balanced` copy parts of the dataset and the norm respectively. While the norm is `float` and can easily make use of `matrix::copyRows`, the dataset can be `int8`/`uint8` and definitely needs the mapping.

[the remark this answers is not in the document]
Nyrio: But maybe we could adapt `copyRows` to accept an iterator, and then branch inside `matrix::detail::copyRows` to select the most efficient implementation based on whether the iterator is a simple pointer or not.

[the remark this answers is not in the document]
achirkin: > What do you mean "we copy indices"?

Oh, sorry, I looked the wrong way. Indeed you're right, we do copy values here and do make use of the mapping `T`->`float`. Which, strictly speaking, is a part of our implementation of ANN methods.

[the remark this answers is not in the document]
achirkin: Therefore, I think, if we want to move ann_kmeans_balanced into the cluster namespace, we should make the mapping public (at least non-dependent on ANN) as well.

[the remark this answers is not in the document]
Nyrio: I think a first step is to change `copyRows` to take an iterator and then do the mapping with `cub::TransformInputIterator`. I've applied similar techniques to use raft prims in a WIP branch.

For the copy of the norm I'll replace it right now with `thrust::gather` though.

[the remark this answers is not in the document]
achirkin: You can go this way, sure, but there still is the logical issue of how we expect a dataset should be transformed within k-means algorithm (which right now is hardcoded in the mapping type).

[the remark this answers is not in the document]
Nyrio: Yes, but in that regard, it matters little whether that hardcoded mapping type is given through an iterator or directly in the kernel such as in `copy_selected_kernel`, I'd even argue I prefer the former.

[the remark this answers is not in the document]
achirkin: Ah, yes, I totally agree (just make sure we don't lose much in terms of performance). I meant to explain issues with the other topic discussed above - moving ann_kmeans_balanced into the cluster namespace, whereas it currently contains ANN-implementation-specific logic for transforming the data.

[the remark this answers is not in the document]
cjnolet: @Nyrio trying to use raft primitives where at all possible, would raft::matrix::gather work in place of thrust::gather? 

Also, just FYI, we are in the process of migrating the RAFT APIs to accept mdspan everywhere to make them consistent across the board, cleaner, easier to use. Maybe we could also use raft::span (which can construct an interator)?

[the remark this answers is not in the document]
Nyrio: @cjnolet `matrix::gather` will be very inefficient for `D=1`:

```cuda
for (int i = threadIdx.x; i < D; i += TPB) {
  out[outRowStart + i] = in[inRowStart + i];
}
```
We could have an array/vector gather but it would essentially be a simple wrapper around thrust.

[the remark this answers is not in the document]
Nyrio: Note that I used `thrust::gather` only for the norm, we could however use `matrix::gather` to select rows in the dataset (in another PR). Its support for iterators is convenient, I could use that to do the int-float mapping.

Don't we have some redundancy between `copyRows` and `gather`? The latter seems to be a generalization of the former, am I right?

achirkin (approved):
Just to confirm, all of the k-means changes seem to be compatible with my ivf-pq PR and hence LGTM :)

cpp/include/raft/spatial/knn/detail/ann_utils.cuh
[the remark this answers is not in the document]
cjnolet: Argmin and argmax themselves are useful primitives. We do have an argmax primitive that uses cub underneath currently. We could create an argmin from that, or use this one. Either way, i think it's generally useful and could be moved to `raft::matrix::argmin`

cjnolet:
@Nyrio, it looks like there's a couple conflicts but otherwise good to go. We can merge after the conflicts are resolved.

cjnolet (approved):

Nyrio:
@cjnolet Yes, I am just running some tests on local changes (conflicts + using the new compiled wrapper around `fusedL2NN`) and after I push it should be ready to merge

cjnolet:
@gpucibot merge

cjnolet:
@Nyrio it looks like this guy needs some conflicts some conflicts resolved before it can be merged

PR #827, opened 2022-09-14
Mdspan-ifying `raft::spatial`

cjnolet:


cpp/include/raft/spatial/knn/knn.cuh
```diff
@@ -65,6 +66,60 @@ inline void knn_merge_parts(value_t* in_keys,
     in_keys, in_values, out_keys, out_values, n_samples, n_parts, k, stream, translations);
 }
 
+/**
+ * Performs a k-select across row partitioned index/distance
+ * matrices formatted like the following:
+ * row1: k0, k1, k2
```
mhoemmen, line 72: Does this mean that the matrices always have 3 columns?  If so, then would you consider expressing that at compile time in the `extents` types of the matrices?

cpp/include/raft/spatial/knn/knn.cuh
```diff
@@ -168,6 +223,72 @@ inline void select_k(const value_t* in_keys,
...
+ *   the implementation of the algorithm
+ */
+template <typename idx_t = int, typename value_t = float>
+inline void select_k(const raft::handle_t& handle,
+                     raft::device_matrix_view<const value_t, idx_t, row_major> in_keys,
+                     raft::device_matrix_view<const idx_t, idx_t, row_major> in_values,
+                     raft::device_matrix_view<value_t, idx_t, row_major> out_keys,
+                     raft::device_matrix_view<idx_t, idx_t, row_major> out_values,
+                     int k,
+                     bool select_min  = true,
```
mhoemmen, line 271: If `select_min` and `algo` are algorithm parameters that have reasonable defaults, then would you consider gathering them into a struct, in order to avoid potential mix-ups with the order of `k` and `select_min`?  Alternately (or additionally), would you consider making `select_min` an `enum class` instead of a `bool`, so it can't be mixed up with `k`?

cpp/include/raft/spatial/knn/knn.cuh
```diff
@@ -224,4 +345,84 @@ void brute_force_knn(raft::handle_t const& handle,
...
+ *
+ *
+ *
+ * @param[in] handle the cuml handle to use
+ * @param[in] index vector of device matrices (each size m_i*d) to be used as the knn index
+ * @param[in] search matrix (size n*d) to be used for searching the index
+ * @param[out] indices matrix (size n*k) to store output knn indices
+ * @param[out] distances matrix (size n*k) to store the output knn distance
+ * @param[in] k the number of nearest neighbors to return
+ * @param[in] metric_arg the value of `p` for Minkowski (l-p) distances. This
```
mhoemmen, line 366: It looks like `metric_arg` comes after `metric` in the parameter list, but before in the documentation.

cpp/include/raft/spatial/knn/knn.cuh
```diff
@@ -224,4 +345,84 @@ void brute_force_knn(raft::handle_t const& handle,
...
+ * 					 is ignored if the metric_type is not Minkowski.
```
mhoemmen, line 367: Would you consider combining `metric` and `metric_arg` into a single type, where it's impossible to construct an instance with a `p` value for a non-Minkowski metric?

cpp/test/spatial/ann_ivf_flat.cu
```diff
@@ -206,25 +207,30 @@ class AnnIVFFlatTest : public ::testing::TestWithParam<AnnIvfFlatInputs> {
...
                          thrust::device_pointer_cast(vector_indices.data()),
                          thrust::device_pointer_cast(vector_indices.data() + ps.num_db_vecs));
         handle_.sync_stream(stream_);
 
         int64_t half_of_data = ps.num_db_vecs / 2;
 
-        auto index_2 =
-          ivf_flat::extend<DataT, int64_t>(handle_, index, database.data(), nullptr, half_of_data);
+        auto half_of_data_view = raft::make_device_matrix_view<const DataT>(
+          (const DataT*)database.data(), static_cast<int>(half_of_data), ps.dim);
```
mhoemmen, line 225: Would you consider using `const_cast` here instead of a C-style cast?

cpp/test/spatial/epsilon_neighborhood.cu
```diff
@@ -65,14 +72,13 @@ class EpsNeighTest : public ::testing::TestWithParam<EpsInputs<T, IdxT>> {
...
 
-  void TearDown() override { RAFT_CUDA_TRY(cudaStreamDestroy(stream)); }
-
   EpsInputs<T, IdxT> param;
   cudaStream_t stream = 0;
   rmm::device_uvector<T> data;
   rmm::device_uvector<bool> adj;
   rmm::device_uvector<IdxT> labels, vd;
   IdxT batchSize;
+  const raft::handle_t handle;
```
mhoemmen, line 81: I noticed that here, the handle is captured by (const) value, whereas in another similar class above, the handle is captured by const reference.  If capturing by value produces correct results, would you consider just capturing the handle by (nonconst) value?  In general, capturing nonconst values instead of const values or references can help make the class `semiregular`, which makes it easier to reason about the class' behavior.  This is also consistent with `mdspan` behavior (which is pointer-like, therefore `semiregular`).

mhoemmen (approved):
Lots of work!!!  I made a few suggestions, none of which are required.

cpp/test/spatial/epsilon_neighborhood.cu
cjnolet: Which similar class above are you referring to? This is a test class, which both creates and owns the handle so it's not using a reference. It's also being used in  the setup in a test function above- it could be made const there, but doesn't need to be a reference.

mhoemmen: Sorry, I had forgotten that this was a test (so you have to create a handle anyway) and that a default-constructed handle is meaningful : - ) .

cpp/include/raft/spatial/knn/ball_cover.cuh
```diff
@@ -111,16 +126,71 @@ void rbc_all_knn_query(const raft::handle_t& handle,
...
+          typename value_t,
+          typename int_t        = std::uint32_t,
+          typename matrix_idx_t = std::uint32_t>
+void rbc_all_knn_query(const raft::handle_t& handle,
+                       BallCoverIndex<idx_t, value_t, int_t, matrix_idx_t>& index,
+                       raft::device_matrix_view<idx_t, matrix_idx_t, row_major> inds,
+                       raft::device_matrix_view<value_t, matrix_idx_t, row_major> dists,
+                       int_t k                     = 5,
+                       bool perform_post_filtering = true,
+                       float weight                = 1.0)
```
divyegala, line 168: Can this be an `std::optional`?

cpp/include/raft/spatial/knn/ball_cover.cuh
```diff
@@ -111,16 +126,71 @@ void rbc_all_knn_query(const raft::handle_t& handle,
...
+ */
+template <typename idx_t = std::int64_t,
+          typename value_t,
+          typename int_t        = std::uint32_t,
+          typename matrix_idx_t = std::uint32_t>
+void rbc_all_knn_query(const raft::handle_t& handle,
+                       BallCoverIndex<idx_t, value_t, int_t, matrix_idx_t>& index,
+                       raft::device_matrix_view<idx_t, matrix_idx_t, row_major> inds,
+                       raft::device_matrix_view<value_t, matrix_idx_t, row_major> dists,
+                       int_t k                     = 5,
```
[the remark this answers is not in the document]
divyegala: I quite liked @cjnolet 's idea in my `raft::linalg` PR where we add another API for such booleans. Maybe that will be the cleanest here?

cpp/include/raft/spatial/knn/ball_cover.cuh
```diff
@@ -172,14 +242,79 @@ void rbc_knn_query(const raft::handle_t& handle,
...
+template <typename idx_t = std::int64_t,
+          typename value_t,
+          typename int_t        = std::uint32_t,
+          typename matrix_idx_t = std::uint32_t>
+void rbc_knn_query(const raft::handle_t& handle,
+                   BallCoverIndex<idx_t, value_t, int_t, matrix_idx_t>& index,
+                   raft::device_matrix_view<const value_t, matrix_idx_t, row_major> query,
+                   raft::device_matrix_view<idx_t, matrix_idx_t, row_major> inds,
+                   raft::device_matrix_view<value_t, matrix_idx_t, row_major> dists,
+                   int_t k                     = 5,
```
divyegala, line 291: `std::optional` for `weight` parameter, and also same as the conclusion of review discussions on the previous function

cpp/include/raft/spatial/knn/ivf_flat.cuh
```diff
@@ -191,4 +322,79 @@ inline void search(const handle_t& handle,
...
+ * The exact size of the temporary buffer depends on multiple factors and is an implementation
+ * detail. However, you can safely specify a small initial size for the memory pool, so that only a
+ * few allocations happen to grow it during the first invocations of the `search`.
+ *
+ * @tparam value_t data element type
+ * @tparam idx_t type of the indices
+ * @tparam int_t precision / type of integral arguments
+ * @tparam matrix_idx_t matrix indexing type
+ *
+ * @param handle
```
divyegala, line 357: Consistency of `[in/out]` across all arguments?

cpp/include/raft/spatial/knn/knn.cuh
```diff
@@ -65,6 +66,60 @@ inline void knn_merge_parts(value_t* in_keys,
...
+ * row1: k0, k1, k2
+ * row2: k0, k1, k2
+ * row3: k0, k1, k2
+ * row1: k0, k1, k2
+ * row2: k0, k1, k2
+ * row3: k0, k1, k2
+ *
+ * etc...
+ *
+ * @tparam idx_t
```
divyegala, line 81: Is this an internal function?

cpp/include/raft/spatial/knn/knn.cuh
divyegala: Again proposing separating out into TWO APIs, `select_k_max` and `select_k_min`

cpp/test/spatial/ann_ivf_flat.cu
divyegala: Is the cast even needed? I think it's clean enough to just express in the templates `raft::make_device_matrix_view<const DataT, int>(database.data(), half_of_data, ps.dim);`

cpp/include/raft/spatial/knn/ball_cover.cuh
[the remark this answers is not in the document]
cjnolet: @divyegala, the `linalg` case was a tad different because there were several options in addition to the booleans and the results of the booleans were each completely different algorithms. In this case, the boolean really is just a simple option. I think once we support an official "approximate" rbc, it would make sense to have 2 different functions.

cpp/include/raft/spatial/knn/knn.cuh
cjnolet: I think I might actually remove the `select_k` for now because we're in the process of pulling the individual algorithms out of `detail` and moving them to `raft::matrix`. When that happens, I don't think `select_k` belongs in `raft::spatial`.

cpp/include/raft/spatial/knn/knn.cuh
cjnolet: No, not internal. The mnmg knn uses this. I'd like to eventually move it to `raft::matrix`, but probably not until the top-k is moved there.

cpp/test/spatial/ann_ivf_flat.cu
cjnolet: It's test code so I honestly wasn't particularly worried about it. I'll update it, though.

cpp/include/raft/spatial/knn/knn.cuh
cjnolet: We've adopted this method of specifying a metric w/ an optional metric arg from Scipy and Scikit-learn. This was done intentionally to provide a familiar face for the APIs. There are also a few other places where this design is used, such as pairwise distances and sparse pw dists/knn. I don't mind having a larger discussion on this topic (maybe on a Github issue?) but I'm wondering if we should hold that change for a post 22.10 release.

cpp/include/raft/spatial/knn/brute_force.cuh
```diff
@@ -0,0 +1,154 @@
...
+ * @param[in] handle
+ * @param[in] in_keys matrix of input keys (size n_samples * n_parts * k)
+ * @param[in] in_values matrix of input values (size n_samples * n_parts * k)
+ * @param[out] out_keys matrix of output keys (size n_samples * k)
+ * @param[out] out_values matrix of output values (size n_samples * k)
+ * @param[in] n_samples number of rows in each part
+ * @param[in] k number of neighbors for each part
+ * @param[in] translations optional vector of starting index mappings for each partition
+ */
+template <typename idx_t = int64_t, typename value_t = float>
```
divyegala, line 48: Per @mhoemmen I also removed defaults for mdspan-related templates that are inferred from my `raft::linalg` PR. I think we can apply this across the whole PR

divyegala (approved):

cjnolet:
@gpucibot merge

PR #831, opened 2022-09-20
Mdspanifying remaining `raft::matrix` functions

cjnolet:


cjnolet:
Giving a heads up that I realized after separating out the `matrix.cuh` and `math.cuh` that many of these functions are not tested. 

I'd really prefer not to make it worse and add new public APIs for functions which are not being explicitly tested so I think I'm going to break this into two different PRs- one for the `raft::matrix` functions that have corresponding tests, which we can expect to release with 22.10, and another for the remaining functions, which I will write the tests for and hopefully merge during 22.12.

PR #834, opened 2022-09-21
[REVIEW] Mdspanify permute

mhoemmen:
I added an overload of `raft::random::permute` that takes device mdspan instead of raw arrays, and `std::optional<mdspan<...>>` for optional output arrays.  I also added two overloads that let users pass in `std::nullopt`.

I've added a unit test that imitates the existing unit test for the raw-array overloads; it builds and passes.  The test ensures that the two `std::nullopt` overloads also compile.

I also opportunistically fixed some unrelated existing small build errors that were blocking forward progress.  That commit is included in PRs #830 and #833 as well, so if it merges, the change should rebase correctly.

### DIFF
--- a/cpp/include/raft/random/permute.cuh
+++ b/cpp/include/raft/random/permute.cuh
@@ -21,30 +21,162 @@
 
 #include "detail/permute.cuh"
 
+#include <optional>
+#include <raft/mdarray.hpp>
+#include <type_traits>
+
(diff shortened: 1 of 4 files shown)


cjnolet (approved):
LGTM

cjnolet:
@gpucibot merge

PR #835, opened 2022-09-21
Fix integer overflow in ANN kmeans

Nyrio:
CUDA block and thread indices are uint32, so the following operation results in an overflow before being cast to uint64:

```cuda
uint64_t gid = threadIdx.x + (blockDim.x * blockIdx.x);
```

This PR fixes the error that @tfeher and myself have encountered while benchmarking large datasets:

```
Incorrect mesocluster size at 0. 625618 vs 625858
```

cc @achirkin


tfeher (approved):
Thanks for the fix Louis, LGTM!

cjnolet:
@gpucibot merge

PR #837, opened 2022-09-22
Move contractions tiling logic outside of Contractions_NT

ahendriksen:
The main functionality of Contractions_NT involves loading tiles of data into shared memory to enable fast GEMM-like kernels. In practice, this requires keeping track of tiles of data (2D submatrices of a bigger matrix) and distributing the data in the tiles over shared memory and registers of thread in a thread block.

Currently, Contractions_NT performs indexing logic for both: 
1. The distribution of data in a tile over registers and shared memory; 
2. Looping over tiles of data in a 2D matrix.

In this PR, we move functionality 2 out of Contractions_NT. Moving over the tiles of data and keeping track of the grid stride loop is now the responsibility of the calling code. 

Splitting these responsibilities is helpful when non-trivial tiling logic is required, as in the upcoming sparseL2NN functionality.

**Note**: This PR also cleans up one unfortunate wart in the current implementation. Depending on which of the two overloaded constructors was called, the tiling logic was transposed leading to extremely difficult to track down bugs.


ajschmidt8:
Please use GitHub's _Draft PR_ feature instead of `WIP` tags in the future. Draft PRs have the benefit of preventing notifications to `codeowners` until PRs are marked _Ready for Review_, which helps cut down on excessive notifications for PRs that are still being worked on. CI will still run on Draft PRs.

Some useful information about Draft PRs:

- https://github.blog/2019-02-14-introducing-draft-pull-requests/
- https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests#draft-pull-requests
- https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request

ahendriksen:
@tfeher @cjnolet : I have rebased the branch onto the latest 22.03 branch. It would be nice to get this PR merged. Both for sparseL2NN and hopper-based pairwise distance kernels.

cjnolet (commented):
These changes LGTM but I'd like to have @tfeher's review on this as well.

cpp/include/raft/distance/detail/pairwise_distance_base.cuh
```diff
@@ -116,96 +121,81 @@ struct PairwiseDistances : public BaseClass {
...
-         gridStrideY += P::Mblk * gridDim.y) {
-      for (auto gridStrideX = blockIdx.x * P::Nblk; gridStrideX < this->n;
-           gridStrideX += P::Nblk * gridDim.x) {
-        prolog(gridStrideX, gridStrideY);
-        loop();
-        epilog(gridStrideX, gridStrideY);
+    for (auto tile_idx_m = grid_offset_m; tile_idx_m < this->m; tile_idx_m += grid_stride_m) {
+      this->ldgXY(tile_idx_m, grid_offset_n, 0);
+      for (auto tile_idx_n = grid_offset_n; tile_idx_n < this->n; tile_idx_n += grid_stride_n) {
+        reset_accumulator();
```
tfeher, line 137: The previous implementation was simply calling `prolog`, `loop`, `epilog` inside the grid offset loops. I see that you have restructured this. Is this motivated because there is an overlap between different operation? Could you still add simple comments as `//main loop`,` //epiloge` to guide the eye?

tfeher (changes requested):
Thanks Allard for this PR! In general it looks good, I just have a few small comments.

cpp/include/raft/distance/detail/pairwise_distance_base.cuh
```diff
@@ -116,96 +121,81 @@ struct PairwiseDistances : public BaseClass {
...
+          // accumulate in registers.
+          accumulate();
+          this->stsXY();
+          __syncthreads();
+          this->switch_write_buffer();
+          this->switch_read_buffer();
+        }
+        accumulate();  // last iteration
+        // This is needed for making sure next grid stride of
+        // non-norm based metrics uses previously accumulated buffer so
```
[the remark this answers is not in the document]
ahendriksen: Non-norm based metrics are metrics where `useNorms == false`, i.e., the unexpanded forms of L2 and Cosine, as well as L1, Chebyshev, etc.. 

I copied this comment over from the `loop()` method in the previous implementation, but I am not sure that it makes sense. There is no real difference between "norm-based" and "non norm-based" metrics in this respect. 

I have changed the comment to: 

        // The pre-condition for the loop over tile_idx_n is that write_buffer
        // and read_buffer point to the same buffer. This flips read_buffer back
        // so that it satisfies the pre-condition of this loop.

cpp/include/raft/distance/detail/pairwise_distance_base.cuh
ahendriksen: Done.

tfeher (approved):
Thanks Allard for addressing the issues, the PR looks good to me!

cjnolet (changes requested):
Holding off on merging right away so we can investigate the CI timeouts.

cjnolet (approved):

PR #838, opened 2022-09-22
Add MaskedL2NN

ahendriksen:
This PR adds the sparseL2NN functionality. 

This enables faster computing pairwise distances by making use of sparsity in the problem: the computation of distances between point pairs can be skipped.

The sparsity between arrays of points X and Y is expressed as follows: 
- X is split into rows (points)
- Y is split into contiguous groups of points (i.e. all points in a group are adjacent)
- A boolean adjacency matrix indicates for each row of X and each group in Y whether to compute the distance.

To speed up computation, the adjacency matrix is compressed into a bitfield.
To ensure competitive speeds, the caller must make sure that consecutive rows in X are adjacent to the same groups in Y (as much as possible) to enable efficient skipping in the kernel.

Some work is still TODO:
- Flesh out documentation
- Discuss / remove allocation of intermediate array
- Optimize for skinny matrices by using a different KernelPolicy.


cjnolet:
@ahendriksen mind if we bump this forward to 23.02? Do you think you might have something working by then?

ahendriksen:
Yes that is okay.

cjnolet:
Hey @ahendriksen, we are still really excited about this feature. I was looking through your changes a bit mostly to see where we are with it. I have a few review items but I've held off from submitting the review until you feel this is in a good place to do so. 

> To ensure competitive speeds, the caller must make sure that consecutive rows in X are adjacent to the same groups in Y (as much as possible) to enable efficient skipping in the kernel.

This won't be easy to do in an algorithm like HDBSCAN without modifying the user's input data, though I understand the importance of memory co-location like coalescing the reads. The pattern that this PR solves pops up quite a bit, though. Do you know the expected perf hit from not co-locating the data points w/ the groups? Is it pretty much all from coalescing?

ahendriksen:
Hi @cjnolet, 

Feel free to add your comments!

> > To ensure competitive speeds, the caller must make sure that consecutive rows in X are adjacent to the same groups in Y (as much as possible) to enable efficient skipping in the kernel.
> 
> This won't be easy to do in an algorithm like HDBSCAN without modifying the user's input data, though I understand the importance of memory co-location like coalescing the reads. The pattern that this PR solves pops up quite a bit, though. Do you know the expected perf hit from not co-locating the data points w/ the groups? Is it pretty much all from coalescing?

Coalescing is absolutely necessary.
I ran a representative benchmark where roughly 99% of distance calculations could be skipped. In the theoretical best case, sparseL2NN should have taken 1% of the time of fusedL2NN. In practice, the relative performance was
- ~98% without coalescing (sorting input points);
- ~10% with coalescing.

There will be a cut off point before which it does not make sense to sort the inputs. However, since the number of distance calculations scales quadratically with the number of input points, sorting will be worth it especially for medium to large data sizes.

As you suggest, the sorting may even be done in place and reverted after the sparseL2NN call (rather than allocating a separate buffer with sorted inputs).

ajschmidt8:
Please use GitHub's _Draft PR_ feature instead of `WIP` tags in the future. Draft PRs have the benefit of preventing notifications to `codeowners` until PRs are marked _Ready for Review_, which helps cut down on excessive notifications for PRs that are still being worked on. CI will still run on Draft PRs.

Some useful information about Draft PRs:

- https://github.blog/2019-02-14-introducing-draft-pull-requests/
- https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests#draft-pull-requests
- https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request

cjnolet:
@ahendriksen I'm working my way through a more formal review. One thing I'd like to discuss is the naming of this primitive. I'd prefer that we rename it to something more descriptive of its purpose because `sparseL2NN` seems to indicate that's actually accepting sparse inputs for both sides. I'm leaning towards something like `Masked L2NN` since masked matrix multiplication is, itself, a known and easily identifiable routine. What do you think?

ahendriksen:
MaskedL2NN sounds good. I can imagine it's more descriptive than sparseL2NN. Looking forward to your comments.

cpp/bench/CMakeLists.txt
```diff
@@ -81,6 +81,7 @@ if(BUILD_BENCH)
     bench/distance/distance_l1.cu
     bench/distance/distance_unexp_l2.cu
     bench/distance/fused_l2_nn.cu
+    bench/spatial/sparse_l2_nn.cu
```
cjnolet, line 84: We've deprecated the `spatial` namespace in favor of `raft/neighbors`, though I also see the file below is in `bench/distance`. For example, we have the `fusedl2nn` primitive in `distance` so maybe the `masked-nn` belongs there as well. Once we have the `masked-knn` we can add that to neighbors (like the `fusedl2knn`).

cpp/include/raft/distance/detail/compress_to_bits.cuh
```diff
@@ -0,0 +1,49 @@
...
+
+#include <raft/util/cuda_utils.cuh>
+#include <raft/util/device_atomics.cuh>
+
+namespace raft {
+namespace distance {
+namespace detail {
+
+template <typename T = uint64_t, typename = std::enable_if_t<std::is_integral<T>::value>>
+__global__ void compress_to_bits_naive(const bool* in, int in_rows, int in_cols, T* out)
```
cjnolet, line 26: So this accepts a boolean adjacency matrix and compresses the bytes to a `uint64_t` bitset? This seems super useful. Any reason we shouldn't add this to `raft/util/mask.cuh` or even `raft/sparse/mask.cuh`?

Either way, I would suggest putting a non-`__global__` helper function in this file that can wrap this kernel to provide the needed geometry and kernel settings.

cpp/include/raft/distance/detail/sparse_distance_base.cuh
```diff
@@ -0,0 +1,362 @@
...
+          typename OutT,
+          typename IdxT,
+          typename Policy,
+          typename CoreLambda,
+          typename EpilogueLambda,
+          typename FinalLambda,
+          typename rowEpilogueLambda,
+          bool isRowMajor    = true,
+          typename BaseClass = raft::linalg::Contractions_NT<DataT, IdxT, Policy, isRowMajor>>
+struct SparseDistances : public BaseClass {
```
cjnolet, line 73: Maybe `MaskedDistances`?

cpp/include/raft/distance/detail/sparse_distance_base.cuh
```diff
@@ -0,0 +1,362 @@
...
+          typename OutT,
+          typename IdxT,
+          typename Policy,
+          typename CoreLambda,
+          typename EpilogueLambda,
+          typename FinalLambda,
+          bool isRowMajor = true>
+__global__ __launch_bounds__(Policy::Nthreads, 2)
+
+  void sparseDistanceMatKernel(const DataT* x,
```
cjnolet, line 327: maskedDistanceMatKernel?

cpp/include/raft/distance/detail/sparse_l2_nn.cuh
```diff
@@ -0,0 +1,303 @@
+/*
```
cjnolet, line 1: Maybe rename file to `maked_l2_nn.cuh` (or even just `masked_nn.cuh` since it's new. I'd like to eventually deprecate `fused_l2_nn.cuh` in favor of `fused_nn.cuh`).

cpp/include/raft/distance/sparse_l2_nn.cuh
```diff
@@ -0,0 +1,114 @@
...
+ * @param[in]  redOp         reduction operator in the epilogue
+ * @param[in] pairRedOp reduction operation on key value pairs
+ * @param[in]  sqrt          Whether the output `minDist` should contain L2-sqrt
+ * @param[in]  initOutBuffer whether to initialize the output buffer before the
+ *                           main kernel launch
+ * @param[in]  stream        cuda stream
+ */
+template <typename DataT, typename OutT, typename IdxT, typename ReduceOpT, typename KVPReduceOpT>
+void sparseL2NN(OutT* min,
+                const DataT* x,
```
cjnolet, line 71: All new functions should be accepting `raft::mdspan` instead of raw pointers.

cpp/include/raft/distance/sparse_l2_nn.cuh
```diff
@@ -0,0 +1,114 @@
...
+                const DataT* y,
+                const DataT* xn,
+                const DataT* yn,
+                const bool* adj,
+                const IdxT* group_idxs,
+                IdxT num_groups,
+                IdxT m,
+                IdxT n,
+                IdxT k,
+                void* workspace,
```
cjnolet, line 81: We shouldn't be exposing workspaces to public APIs, but should be using `handle.get_temp_workspace_resource()` to get an allocator (which could be a pool allocator) for temp memory.

cpp/test/distance/sparse_l2_nn.cu
```diff
@@ -0,0 +1,494 @@
+/*
```
cjnolet, line 1: T[his is a new file so it should only have the current year.

cpp/test/distance/sparse_l2_nn.cu
```diff
@@ -0,0 +1,494 @@
...
+
+#include <iostream>
+#include <stdio.h>
+
+namespace raft {
+namespace distance {
+namespace sparse_l2_nn {
+
+template <typename LabelT, typename DataT>
+struct CubKVPMinReduce {
```
cjnolet, line 34: Are you able to use the `raft::KeyValuePair` here? I understand some of the cub functions are only overloaded internally for the cub version but it would be nice to use the raft variant wherever it's possible.

cpp/test/distance/sparse_l2_nn.cu
```diff
@@ -0,0 +1,494 @@
...
+      return ::testing::AssertionFailure()
+             << "actual=" << act.key << "," << act.value << " != expected=" << exp.key << ","
+             << exp.value << " @" << i;
+    }
+  }
+  return ::testing::AssertionSuccess();
+}
+
+const std::vector<Inputs<float>> inputsf = {
+  {0.001f, 32, 32, 32, 2, 1234ULL, AdjacencyPattern::all_true},
```
cjnolet, line 358: Not a huge deal but I'm wondering if it might make sense to also include a skewed degree distribution that might follow small-world or power-law characteristics (and it could be a good opportunity to use the rmat generator).

cpp/bench/distance/sparse_l2_nn.cu
```diff
@@ -0,0 +1,194 @@
...
+#include <raft/handle.hpp>
+#include <raft/linalg/norm.cuh>
+#include <raft/random/rng.cuh>
+#include <raft/util/cudart_utils.hpp>
+
+#if defined RAFT_NN_COMPILED
+#include <raft/spatial/knn/specializations.hpp>
+#endif
+
+namespace raft::bench::spatial::sparse {
```
cjnolet, line 35: I think rather than the `spatial::sparse` namespace, we should use `distance::masked_nn`

cpp/include/raft/distance/detail/sparse_l2_nn.cuh
```diff
@@ -0,0 +1,303 @@
...
+  rmm::device_uvector<uint64_t> adj64(raft::ceildiv(m, IdxT(64)) * num_groups, stream);
+  RAFT_CUDA_TRY(cudaMemsetAsync(adj64.data(), 0, adj64.size() * sizeof(uint64_t), stream));
+  dim3 compress_grid(raft::ceildiv(m, 32), raft::ceildiv(num_groups, 32));
+  compress_to_bits_naive<<<compress_grid, dim3(32, 32), 0, stream>>>(
+    adj, num_groups, m, adj64.data());
+
+  dim3 blk(P::Nthreads);
+  auto nblks            = raft::ceildiv<int>(m, P::Nthreads);
+  constexpr auto maxVal = std::numeric_limits<DataT>::max();
+  typedef cub::KeyValuePair<IdxT, DataT> KVPair;
```
cjnolet, line 223: Are we able to use `raft::KeyValuePair` here? It's nice when we can use our own vocabulary types

cpp/test/distance/sparse_l2_nn.cu
```diff
@@ -0,0 +1,494 @@
...
+    dim3 block(32, 32);
+    dim3 grid(10, 10);
+    init_adj<<<grid, block, 0, stream>>>(
+      m, n, num_groups, params.pattern, adj.data(), group_idxs.data());
+    RAFT_CUDA_TRY(cudaGetLastError());
+
+    generateGoldenResult();
+    raft::linalg::rowNorm(xn.data(), x.data(), k, m, raft::linalg::L2Norm, true, stream);
+    raft::linalg::rowNorm(yn.data(), y.data(), k, n, raft::linalg::L2Norm, true, stream);
+    RAFT_CUDA_TRY(cudaStreamSynchronize(stream));
```
cjnolet, line 239: Nitpick: `handle.sync_stream()` is a little safer and less error-prone, though I know we set a stream explicitly in most of the other googletests.

cpp/test/distance/sparse_l2_nn.cu
```diff
@@ -0,0 +1,494 @@
...
+                       group_idxs.data(),
+                       m,
+                       n,
+                       k,
+                       num_groups,
+                       (int*)workspace.data(),
+                       stream);
+  }
+
+  void runTest(cub::KeyValuePair<int, DataT>* out)
```
cjnolet, line 276: Can we use `raft::KeyValuePair` here instead? This is bound to cause an issue when, for example, someone adds an explicit template specialization which now forces symbols in the `libraft` binary to include `cub` types. We have encountered this issue before and it caused a great deal of symbols conflicts for our users).

cpp/include/raft/distance/sparse_l2_nn.cuh
```diff
@@ -0,0 +1,114 @@
...
+ * @tparam DataT     data type
+ * @tparam OutT      output type to either store 1-NN indices and their minimum
+ *                   distances or store only the min distances. Accordingly, one
+ *                   has to pass an appropriate `ReduceOpT`
+ * @tparam IdxT      indexing arithmetic type
+ * @tparam ReduceOpT A struct to perform the final needed reduction operation
+ *                   and also to initialize the output array elements with the
+ *                   appropriate initial value needed for reduction.
+ *
+ * @param[out] min           will contain the reduced output (Length = `m`)
```
cjnolet, line 50: The order of arguments should be: `handle, params_struct, in, out, additional args`. The params_struct is optional for those algorithms that have a lot of parameters. 

Could we include a variant which a allows us to specify separate outputs for index and distance arrays? Ideally we would be able to have both, but this is something that has caused some problems when we want them separated and then need to take up additional memory to write the keyvaluepair output to additional arrays.

cjnolet (changes requested):
I don't know if I can express how happy I am to have this primitive (and eventually the corresponding masked-knn primitive). This is absolutely going to be used to provide speedups in across several algorithms. 

Most of the things I found were fairly minor and API related. I'm also hoping @tfeher will provide more feedback on some of the lower-level and CUDA-specific things.

cpp/include/raft/distance/sparse_l2_nn.cuh
ahendriksen: All done, apart from this one:

> Could we include a variant which a allows us to specify separate outputs for index and distance arrays?

This would be possible. Maybe we want to standardize on an interface between fused_l2_nn and masked_l2_nn?

cpp/test/distance/sparse_l2_nn.cu
ahendriksen: The benchmark is quite heavy as is. The current patterns have mainly been chosen to highlight performance characteristics:
- all_true: what happens when we can't skip any computations
- all_false: when we can skip all computation (how fast is skipping, not really interesting in the real world)
- checkerboard_64: what if each CUDA block can skip exactly half the computations (expected 2x speedup)
- checkerboard_4: what if some warps (not sure anymore actually..) can skip computations (expected speedup: minor)
- checkerboard: what if some threads in a warp can skip computations (expected speedup: none)

I agree that sampling from a distribution would be more useful to get a "real-world" feel for the end-to-end performance, but it does not directly help when comparing small changes.

cpp/include/raft/distance/detail/sparse_l2_nn.cuh
ahendriksen: Renamed to masked_nn.cuh (all other files as well)

cpp/bench/CMakeLists.txt
ahendriksen: The masked_nn benchmark is now in distance and the CMakeLlists now has the correct location.

cpp/include/raft/distance/detail/sparse_l2_nn.cuh
ahendriksen: To make deprecating the `l2` versions easier, I have renamed the files to the `masked_nn*.cuh` format.

cpp/include/raft/distance/sparse_l2_nn.cuh
ahendriksen: Done. To make it work, I had to remove `const` from `get_workspace_resource`, as performed in this commit: bda3bcad

cpp/test/distance/sparse_l2_nn.cu
ahendriksen: What is the definition of 'publication' here? I have updated the files to add copyright in the year 2023 when I have edited them. The files where originally published (as draft PR) in 2022.

tfeher (changes requested):
Hi Allard, thank you for this PR! I still looking into the kernels, but I would like to share the first batch of my comments.

About naming convention: @teju85 pointed out that this method is actually a special case of [SDDMM](https://ieeexplore.ieee.org/document/8638042), we might want to refer to that, and see if we can improve nomenclature.

cpp/test/distance/masked_nn.cu
```diff
@@ -0,0 +1,500 @@
...
+              << ", "
+                 "seed: "
+              << p.seed
+              << ", "
+                 "tol: "
+              << p.tolerance;
+  }
+};
+
+__global__ void init_adj(
```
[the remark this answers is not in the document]
tfeher: Have a look at this: https://github.com/rapidsai/raft/pull/1162

cpp/test/distance/masked_nn.cu
```diff
@@ -0,0 +1,500 @@
...
+typedef MaskedL2NNTest<double, true> MaskedL2NNTestD_Sqrt;
+TEST_P(MaskedL2NNTestD_Sqrt, Result)
+{
+  runTest(min.data());
+  ASSERT_TRUE(devArrMatch(
+    min_ref.data(), min.data(), params.m, CompareApproxAbsKVP<double>(params.tolerance), stream));
+}
+INSTANTIATE_TEST_CASE_P(MaskedL2NNTests, MaskedL2NNTestD_Sqrt, ::testing::ValuesIn(inputsd));
+
+/// This is to test output determinism of the prim
```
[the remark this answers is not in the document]
ahendriksen: I took this from the `fused_l2_nn` tests. I can't find why it was added there. Maybe @cjnolet can chime in? As the output data has to be initialized before running the core compute kernel, there might have been issues in the past. But these issues should be unearthed with 2 iterations. I will change it to 2 and also change it in the fusedL2NN test.

cpp/include/raft/distance/detail/masked_distance_base.cuh
```diff
@@ -0,0 +1,361 @@
...
+        if (block_adj == 0) { continue; }
+
+        // Determine which results, that are computed by this thread, have to
+        // be taken into account. This information is stored in a bitfield,
+        // thread_adj. If all results computed by this thread can be ignored,
+        // then we can also skip some computations (thread_adj == 0).
+
+        // We precompute this information because it is used in various
+        // locations to skip thread-local computations.
+        int thread_adj = compute_thread_adjacency(block_adj);
```
tfeher, line 152: Below you have guards at many points for `thread_adj !=0`. Do I see it correctly that if `thread_adj == 0`, then that thread does not perform any computation, just participates in the collective operations, like `__syncthreads`? If yes, could you mention this explicitly?

cpp/include/raft/distance/detail/masked_distance_base.cuh
```diff
@@ -0,0 +1,361 @@
...
+  }
+
+  DI uint32_t compute_thread_adjacency(const uint64_t block_adj)
+  {
+    uint32_t thread_adj = 0;
+#pragma unroll
+    for (int i = 0; i < P::AccRowsPerTh; ++i) {
+      const uint64_t read_mask  = 1ull << (this->accrowid + i * P::AccThRows);
+      const uint32_t write_mask = 1 << i;
+      if ((block_adj & read_mask) != 0) { thread_adj |= write_mask; }
```
tfeher, line 216: Could you add some explanation here?

cpp/include/raft/distance/detail/masked_distance_base.cuh
```diff
@@ -0,0 +1,361 @@
...
+        // be taken into account. This information is stored in a bitfield,
+        // thread_adj. If all results computed by this thread can be ignored,
+        // then we can also skip some computations (thread_adj == 0).
+
+        // We precompute this information because it is used in various
+        // locations to skip thread-local computations.
+        int thread_adj = compute_thread_adjacency(block_adj);
+
+        auto tile_idx_n       = idx_g == 0 ? 0 : group_idxs[idx_g - 1];
+        const auto tile_end_n = group_idxs[idx_g];
```
tfeher, line 155: Wouldn't it be better to call this `group_end_n`? To me the current tile is always the elements in range `tile_idx_n...tile_idx_n+P::Nblk-1`

cpp/include/raft/distance/detail/masked_nn.cuh
```diff
@@ -0,0 +1,308 @@
...
+#include <rmm/device_uvector.hpp>
+
+namespace raft {
+namespace distance {
+namespace detail {
+
+#if (ENABLE_MEMCPY_ASYNC == 1)
+#include <cuda_pipeline.h>
+using namespace nvcuda::experimental;
+#endif
```
tfeher, line 36: Is this used anywhere?

tfeher (changes requested):
Second batch of comments. Overall it looks good, it really helps that #837 has already taken care of some of the refactoring.

cpp/include/raft/distance/detail/masked_distance_base.cuh
```diff
@@ -0,0 +1,361 @@
...
+    const auto grid_stride_g = gridDim.x;
+    const auto grid_offset_g = blockIdx.x;
+
+    for (auto tile_idx_m = grid_offset_m; tile_idx_m < this->m; tile_idx_m += grid_stride_m) {
+      // Start loop over groups
+      for (auto idx_g = grid_offset_g; idx_g < this->num_groups; idx_g += grid_stride_g) {
+        // The __syncthreads() ensures that loading the block flag occurs at
+        // the same time in all threads of the block. Since all threads load
+        // the same address, this speeds up the code.
+        __syncthreads();
```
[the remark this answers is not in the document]
ahendriksen: I remember benchmarking with and without this `__syncthreads`. I am as surprised as you are that the version with this statement was faster.

cpp/include/raft/distance/sparse_l2_nn.cuh
cjnolet: Yes, that works. I don't think there's a need to keep the MR itself const there.

cpp/include/raft/distance/sparse_l2_nn.cuh
cjnolet: > This would be possible. Maybe we want to standardize on an interface between fused_l2_nn and masked_l2_nn

That sounds like a great idea. And I think there's uses for having both the array<pair> and (array, array) cases.

cpp/include/raft/distance/sparse_l2_nn.cuh
cjnolet: I updated this in https://github.com/rapidsai/raft/pull/1165 as well.

cpp/test/distance/sparse_l2_nn.cu
cjnolet: The year should reflect the time the files become a part of the release branch, not the time they were created as a draft.

ahendriksen:
> About naming convention: @teju85 pointed out that this method is actually a special case of [SDDMM](https://ieeexplore.ieee.org/document/8638042), we might want to refer to that, and see if we can improve nomenclature.

This looks quite similar indeed. We might have a slightly different way of expressing sparsity (not using CSR) and the output is a single vector and not a sparse matrix, but otherwise I quite like the "positive" connotation of "sampled" over "masked". For instance, the adjacency matrix would indicate which distance calculations are **sampled** rather than **not masked** (avoiding a double negation). In any case, I will discuss the differences and similarities of maskedL2NN to SDDM in the docstring (as Tamas and I discussed earlier today).

@cjnolet : what do you think? Should we go for `sampledL2NN` over `maskedL2NN`? I am willing to change the naming. I would really hate to change it and then back again, however. Your call! :+1:

cjnolet:
> @cjnolet : what do you think? Should we go for sampledL2NN over maskedL2NN?

I actually prefer that we stick with the `masked` term here. This is exactly a masked matrix multiplication (used in semirings, for example). I'd prefer to use the sampled terminology in a higher-level primitive that would be sampling to construct the adjacency matrix. I do agree, though, that it makes sense to at least reference the above mentioned paper. We should probably also reference masked matrix multiplication (which  also don't always accept sparse inputs for A and B).

cjnolet:
Ah, I notice the abstract of the SDMM paper does actually reference the semirings (e.g. graphblas) formulation of the problem as the initial motivation for their work. I think of their work as a layer above a masked matrix multiplication (whether dense or sparse). And yes, I'm excited about this primitive because it really is THAT reusable!

cpp/test/distance/masked_nn.cu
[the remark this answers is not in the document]
ahendriksen: I have added an issue: https://github.com/rapidsai/raft/issues/1171

cpp/test/distance/masked_nn.cu
```diff
@@ -0,0 +1,500 @@
...
+    auto exp = exp_h.get()[i];
+    auto act = act_h.get()[i];
+    if (!eq_compare(exp, act)) {
+      return ::testing::AssertionFailure()
+             << "actual=" << act.key << "," << act.value << " != expected=" << exp.key << ","
+             << exp.value << " @" << i;
+    }
+  }
+  return ::testing::AssertionSuccess();
+}
```
tfeher, line 363: These helper functions seem to be the same as the ones used by the fused_l2_nn test. Please remove code duplication by factoring them to a common test utils file. (Possibly a subtask for #1171).

tfeher (changes requested):
Hi Allard, this is my last batch of comments. Overall the PR looks good and I am looking forward to using the masked distance prim in our algorithms!

cpp/include/raft/distance/detail/masked_distance_base.cuh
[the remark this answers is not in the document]
ahendriksen: I just compared the benchmarks again. I think the `__syncthreads` can be safely removed.

cpp/include/raft/distance/detail/masked_distance_base.cuh
```diff
@@ -0,0 +1,361 @@
...
+  {
+  }
+
+  DI void run()
+  {
+    const auto grid_stride_m = (P::Mblk * gridDim.y);
+    const auto grid_offset_m = (P::Mblk * blockIdx.y);
+
+    const auto grid_stride_g = gridDim.x;
+    const auto grid_offset_g = blockIdx.x;
```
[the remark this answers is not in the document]
ahendriksen: I agree that we should be consistent, but think we should go the other way. These loop variables are not used by any member function (which is good) and therefore I think we should keep them here.


I have added a reminder in the refactoring PR to move the pairwise_distance_base member variables.

cpp/include/raft/distance/detail/masked_distance_base.cuh
ahendriksen: There is already [this comment](https://github.com/rapidsai/raft/blob/50d85329556d3c242f0a87a28ec40290a813b900/cpp/include/raft/distance/detail/masked_distance_base.cuh#L145), which is quite extensive. Does it not suffice?

cpp/test/distance/masked_nn.cu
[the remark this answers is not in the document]
ahendriksen: I have opend issue https://github.com/rapidsai/raft/issues/1171 to track this task.

cpp/test/distance/masked_nn.cu
ahendriksen: I have added this as a subtask of #1171 and have removed `CompareExactKVP` as it was not used in maskedL2NN tests.

cpp/include/raft/distance/detail/masked_nn.cuh
```diff
@@ -0,0 +1,308 @@
...
+                         IdxT tile_idx_n,
+                         IdxT tile_idx_m,
+                         IdxT tile_end_n) {
+    KVPReduceOpT pairRed_op(pairRedOp);
+
+#pragma unroll
+    for (int i = 0; i < P::AccRowsPerTh; ++i) {
+#pragma unroll
+      for (int j = 0; j < P::AccColsPerTh; ++j) {
+        acc[i][j] = regxn[i] + regyn[j] - (DataT)2.0 * acc[i][j];
```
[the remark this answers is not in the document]
ahendriksen: There are several granularities at which we can choose not to perform a distance computation:
- block
- warp (we do not do this explicitly, but may happen if all threads in a warp skip computation)
- thread
- per element in thread tile

Obviously, we skip computations at the block level. Also, if *all* distance computations in a thread can be skipped, we do so. Checking whether to skip a single distance computation in a thread requires some integer arithmetic (or would require keeping 4 booleans in registers). This turned out to be not worth the effort for several reasons:
1. In the best case, the computation is predicated off saving energy, but not time. 
2. In the worst case, the computation is interspersed with arithmetic to figure out if the computation has to take place, destroying performance. 

In summary, some computations are performed when they are not necessary. There is still the guarantee that a distance pair that has been masked off will not be written to the output, as checked here: https://github.com/rapidsai/raft/blob/50d85329556d3c242f0a87a28ec40290a813b900/cpp/include/raft/distance/detail/masked_nn.cuh#L108-L110)

cpp/include/raft/distance/detail/masked_distance_base.cuh
tfeher: One could debate "extensive" :). I would prefer if we explicitly state, at the end of that comment block that you refer to: "Even when the computation can be skipped for a thread, the thread still participates in synchronization operations."

cpp/include/raft/distance/detail/masked_nn.cuh
```diff
@@ -0,0 +1,308 @@
...
+
+  if (initOutBuffer) {
+    initKernel<DataT, OutT, IdxT, ReduceOpT>
+      <<<nblks, P::Nthreads, 0, stream>>>(min, m, maxVal, redOp);
+    RAFT_CUDA_TRY(cudaGetLastError());
+  }
+
+  auto fin_op = raft::identity_op{};
+
+  constexpr size_t shmemSize = P::SmemSize + ((P::Mblk + P::Nblk) * sizeof(DataT));
```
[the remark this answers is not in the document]
ahendriksen: Right now, it could. Since the function only supports using the L2 metric. However, for other metrics such as the the L1 metric that do not require storing the norms in shared memory, the  `((P::Mblk + P::Nblk) * sizeof(DataT))` could be removed. In addition, there is the correlation metric that requires two norms to be stored. 

All in all, I do agree that we could benefit from more abstraction and separation of concerns, but I think we should leave this task to PR #1142 . I have added a note there.

cpp/include/raft/distance/detail/masked_nn.cuh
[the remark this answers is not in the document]
tfeher: Thanks for the detailed explanation!

cpp/include/raft/distance/detail/masked_nn.cuh
```diff
@@ -0,0 +1,308 @@
...
+#pragma unroll
+      for (int i = 0; i < P::AccRowsPerTh; ++i) {
+        val[i] = {-1, maxVal};
+      }
+    };
+
+  IdxT lda = k, ldb = k, ldd = n;
+  MaskedDistances<true,
+                  DataT,
+                  DataT,
```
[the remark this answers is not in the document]
ahendriksen: I agree this is not pretty, but let's defer this to later when we have concrete use case.

cpp/include/raft/distance/detail/masked_nn.cuh
```diff
@@ -0,0 +1,308 @@
...
+  compress_to_bits_naive<<<compress_grid, dim3(32, 32), 0, stream>>>(
+    adj, num_groups, m, ws_adj64.data());
+
+  dim3 blk(P::Nthreads);
+  auto nblks            = raft::ceildiv<int>(m, P::Nthreads);
+  constexpr auto maxVal = std::numeric_limits<DataT>::max();
+  typedef raft::KeyValuePair<IdxT, DataT> KVPair;
+
+  // Accumulation operation lambda
+  auto core_lambda = [] __device__(DataT & acc, DataT & x, DataT & y) { acc += x * y; };
```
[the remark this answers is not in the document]
ahendriksen: I agree this is not pretty, but let's defer this to later when we have concrete use case.

cpp/include/raft/distance/detail/masked_distance_base.cuh
ahendriksen: Fair enough. I now see what information you were missing. I have added significantly  more documenation, also based on the questions that popped up in the review.

cpp/include/raft/distance/masked_nn.cuh
```diff
@@ -0,0 +1,152 @@
...
+ *
+ * @param handle             RAFT handle for managing expensive resources
+ * @param params             Parameter struct specifying the reduction operations.
+ * @param[in]  x             first matrix. Row major. Dim = `m x k`.
+ *                           (on device).
+ * @param[in]  y             second matrix. Row major. Dim = `n x k`.
+ *                           (on device).
+ * @param[in]  xn            L2 squared norm of `x`. Length = `m`. (on device).
+ * @param[in]  yn            L2 squared norm of `y`. Length = `n`. (on device)
+ * @param[in]  adj           A boolean adjacency matrix indicating for each
```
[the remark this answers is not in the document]
tfeher: We can add this later if we need.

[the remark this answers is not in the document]
ahendriksen: I will leave this for a future TODO. Right now, the compress_to_bits kernel is absolutely not a bottleneck. But it could be nice to save on memory.

cpp/include/raft/distance/detail/compress_to_bits.cuh
```diff
@@ -0,0 +1,49 @@
...
+  bool bit   = in[i * in_cols + j];
+  int bitpos = j % bits_per_element;
+
+  T bitfield = bit ? T(1) << bitpos : 0;
+
+  const size_t out_rows = raft::ceildiv(in_cols, bits_per_element);
+  const size_t out_cols = in_rows;
+  const size_t out_j    = i;
+  const size_t out_i    = j / bits_per_element;
+  if (out_i < out_rows && out_j < out_cols) { atomicOr(&out[out_i * out_cols + out_j], bitfield); }
```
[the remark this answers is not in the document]
ahendriksen: I have rewritten the implementation of compress_to_bits to use shared memory. This avoids the use of atomic ops.

ahendriksen:
Thank you both for the thorough review. I think the majority of your feedback has been incorporated into the current PR. 

**compress to bits**
I have added: 
- Doc string
- Tests
- Use of mdspan
- faster implementation with shared memory (avoiding byte-sized atomicOr)

I did not get around to: 
1. Exposing `compress_to_bits` in raft/utils. 
2. Adding an overload the accepts a bitfield. 

I hope the PR is good enough in its current form to accept.

ahendriksen:
In addition, I have run some benchmarks on a V100 to make sure that the current version is indeed fast enough. With the checkerboard_64 pattern (which skips 50% of distance computations) m=n=16K and k=32 we have that:

maskedL2NN is **1.7x faster** than fusedL2NN. fusedL2NN reaches 8.2 TFlop/s and maskedL2NN reaches 6.9 TFlop/s (84% of fusedL2NN), but since maskedL2NN can skip half the computations, it is still faster.

tfeher (approved):
Thanks Allard for addressing the issues! The PR looks good to me!

cjnolet:
@ahendriksen , the PR looks great. I have one additional as, though. Can you create Github issues for these remaining items just so it stays on our radar?

> I did not get around to:
>
> Exposing compress_to_bits in raft/utils.
> Adding an overload the accepts a bitfield.

cjnolet:
@ahendriksen one thing that I've noticed is that this PR and https://github.com/rapidsai/raft/pull/837 both seem to be consistently timing out in CI while there are some other PRs that don't seem to be timing out. This makes me a little nervous about immediately merging these right before burndown. In case this behavior is isolated to this PR, I want to be careful not to cause issues for other PRs during this release. 

We still have some time before code freeze to merge these, but it would help if you are able to verify locally that the build time (and ideally memory requirement during building) doesn't seem to be drastically affected by these changes. Maybe compare the build time for your changes against `branch-23.02` just to see? It could be something super simple- like a pre-compiled template instantiation might no longer be getting used which could be causing things to get re-compiled more times from scratch.

cjnolet (changes requested):
Holding off on merging right away so we can investigate the CI timeouts.

cjnolet:
@ahendriksen can you please update this PR to accept raft::device_resources everywhere instead of raft::handle_t?

ahendriksen:
> @ahendriksen can you please update this PR to accept raft::device_resources everywhere instead of raft::handle_t?

Done.

ahendriksen:
> 



> @ahendriksen , the PR looks great. I have one additional ask, though. Can you create Github issues for these remaining items just so it stays on our radar?

I have created issues:
- https://github.com/rapidsai/raft/issues/1196 
- https://github.com/rapidsai/raft/issues/1195

cjnolet (approved):

PR #839, opened 2022-09-22
[BUG] rng_state not including necessary cstdint

MatthiasKohl:
Just a one-line PR fixing a missing include in raft/random/rng_state.hpp


mhoemmen (approved):
Thanks for the fix!  I had baked this into PRs #830, #833, and #834, but it's worth having as its own PR.

PR #845, opened 2022-09-23
[REVIEW] Develop new mdspan-ified multi_variable_gaussian interface

mhoemmen:
Develop a new `multi_variable_gaussian` interface that uses `mdspan`.  The new interface uses free functions, rather than a class.

TODO:

- [x] Do not expose workspaces to the public API (see e.g., https://github.com/rapidsai/raft/pull/802#discussion_r978712609); use RMM instead.
- [x] Add a public API not in in the `detail` namespace, and test it.
- [x] File issues for refactoring and improving the implementation.

### DIFF
--- a/cpp/include/raft/random/detail/multi_variable_gaussian.cuh
+++ b/cpp/include/raft/random/detail/multi_variable_gaussian.cuh
@@ -19,6 +19,7 @@
 #include <cmath>
 #include <memory>
 #include <optional>
+#include <rmm/device_uvector.hpp>
 #include <raft/core/device_mdspan.hpp>
 #include <raft/core/handle.hpp>
 #include <raft/linalg/detail/cublas_wrappers.hpp>
@@ -300,7 +301,11 @@ multi_variable_gaussian_setup_token<ValueType> setup_multi_variable_gaussian(
   const raft::handle_t& handle, const int dim, multi_variable_gaussian_decomposition_method method);
 
 template <typename ValueType>
-std::size_t workspace_size(const multi_variable_gaussian_setup_token<ValueType>& token);
+multi_variable_gaussian_setup_token<ValueType> setup_multi_variable_gaussian(
+  const raft::handle_t& handle,
+  rmm::mr::device_memory_resource* mem_resource,
+  const int dim,
+  multi_variable_gaussian_decomposition_method method);
 
 // @param x[in] vector of dim elements
 // @param P[inout] On input, dim x dim matrix; overwritten on output
@@ -323,7 +328,11 @@ class multi_variable_gaussian_setup_token {
     multi_variable_gaussian_decomposition_method method);
 
   template <typename T>
-  friend std::size_t workspace_size(multi_variable_gaussian_setup_token<T>& token);
+  friend multi_variable_gaussian_setup_token<T> setup_multi_variable_gaussian(
+    const raft::handle_t& handle,
+    rmm::mr::device_memory_resource* mem_resource,
+    const int dim,
+    multi_variable_gaussian_decomposition_method method);
 
   template <typename T>
   friend void compute_multi_variable_gaussian(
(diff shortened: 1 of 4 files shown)


cjnolet:
> Do not expose workspaces to the public API (see e.g., Implement stats API with mdspan #802 (comment)); use RMM instead.

If you have time to support it, adding an optional `rmm::mr::device_memory_resource` (or even better, add our own interface that can wrap rmm's device and host memory resources) to the handle could be super helpful to get the workspace out of the public API. We can also use it on the ivf flat search API.

mhoemmen:
@cjnolet wrote:
> If you have time to support it, adding an optional `rmm::mr::device_memory_resource` (or even better, add our own interface that can wrap rmm's device and host memory resources) to the handle could be super helpful to get the workspace out of the public API. We can also use it on the ivf flat search API.

I'll be happy to add an optional `rmm::mr::device_memory_resource` parameter.  For adding it to the handle instead, do you have a sense for what that interface should look like?  I was thinking of a free function like this.

```c++
device_memory_resource get_default_memory_resource(const handle_t&);
```

The const reference is necessary because of all the interfaces that take `const handle&`, but storing the memory resource in the handle would call for the same `mutable` pattern as before.

cjnolet:
> The const reference is necessary because of all the interfaces that take const handle&, but storing the memory resource in the handle would call for the same mutable pattern as before.

I think that could work. To make this compatible w/ some of the ideas we've been tossing around for the more library agnostic raft handle, what do you think about sticking this in a file called `rmm_resource_handler.hpp`? Just thinking out loud here, but it would be nice if the handle ultimately just became a key/value store and primitives could request what they need from the handle. Eventually, we could have includes like `cublas_resource_handler.hpp`, which would lazy-load a cublas handle and store it on the raft handle when it's needed. Of course, this disregards the immutability of the const qualifier and might require some care (locks?) to avoid conflicts in concurrent updates. The handle is coupled to a single "user" stream, but what if these resources are updated in a worker stream?

The basic idea here would be that if no primitives are invoked that need cublas, there's no reason for the user to ever have to interact with it, and thus no reason to even bother creating a cublas handle at all.

mhoemmen:
@cjnolet wrote:
> Just thinking out loud here, but it would be nice if the handle ultimately just became a key/value store and primitives could request what they need from the handle.

I've filed this as a feature request: https://github.com/rapidsai/raft/issues/856

Thanks! : - )

cpp/include/raft/random/detail/multi_variable_gaussian.cuh
```diff
@@ -286,5 +291,159 @@ class multi_variable_gaussian_impl {
   ~multi_variable_gaussian_impl() { deinit(); }
 };  // end of multi_variable_gaussian_impl
 
+enum class multi_variable_gaussian_decomposition_method { CHOLESKY, JACOBI, QR };
```
cjnolet, line 294: Could we consider moving this to a file `random_types.hpp` or `multi_variable_gaussian_types.hpp` (though the former seems easier to find and use) for user visibility (and discoverability)? This also makes it easier for users to know what APIs are safe to expose through their public APIs.

cpp/include/raft/random/multi_variable_gaussian.cuh
```diff
@@ -59,6 +59,81 @@ class multi_variable_gaussian : public detail::multi_variable_gaussian_impl<T> {
...
+ *
+ * - `CHOLESKY`: Cholesky decomposition
+ * - `JACOBI`:
+
+ */
+using detail::multi_variable_gaussian_decomposition_method;
+using detail::multi_variable_gaussian_setup_token;
+
+template <typename ValueType>
+multi_variable_gaussian_setup_token<ValueType> setup_multi_variable_gaussian(
```
cjnolet, line 75: Very minor nitpick: What do you think of calling this `build_multivariable_gaussian_token` or make_multi_variable_gaussian_token` just to be more consistent w/ other APIs? `setup` works as well, and doesn't require the use of `token` in the suffic so I'm okay w/ that name as well.

cpp/include/raft/random/multi_variable_gaussian.cuh
```diff
@@ -59,6 +59,81 @@ class multi_variable_gaussian : public detail::multi_variable_gaussian_impl<T> {
   ~multi_variable_gaussian() { deinit(); }
 };  // end of multi_variable_gaussian
 
+/**
+ * @brief Matrix decomposition method for `compute_multi_variable_gaussian` to use.
+ *
+ * `compute_multi_variable_gaussian` can use any of the following methods.
+ *
+ * - `CHOLESKY`: Cholesky decomposition
+ * - `JACOBI`:
```
cjnolet, line 68: Is this incomplete?

cpp/include/raft/random/multi_variable_gaussian.cuh
```diff
@@ -59,6 +59,81 @@ class multi_variable_gaussian : public detail::multi_variable_gaussian_impl<T> {
...
+
+ */
+using detail::multi_variable_gaussian_decomposition_method;
+using detail::multi_variable_gaussian_setup_token;
+
+template <typename ValueType>
+multi_variable_gaussian_setup_token<ValueType> setup_multi_variable_gaussian(
+  const raft::handle_t& handle,
```
cjnolet, line 76: I like the overloads here- this should make it much easier to pull it from the handle.

cjnolet (changes requested):
Thanks for these wrappers, Mark! This looks great overall. Just minor things and I think it's ready to go.

cpp/include/raft/random/detail/multi_variable_gaussian.cuh
mhoemmen: Done; thanks!

cpp/include/raft/random/multi_variable_gaussian.cuh
mhoemmen: I renamed this to `build_multi_variable_gaussian_token`; thanks!

cpp/include/raft/random/multi_variable_gaussian.cuh
mhoemmen: It was!  Thanks for pointing this out! : - )  It took me a while to dig out the cuSOLVER documentation explaining the difference between the two algorithms -- I had guessed that one was Jacobi plane rotations, but needed confirmation, and LAPACK doesn't expose these.

cpp/include/raft/random/multi_variable_gaussian.cuh
mhoemmen: @cjnolet I've actually removed the public "setup" methods.  They should no longer be needed, since we now take a memory resource for the workspace.

cjnolet:
@gpucibot merge

cjnolet:
@gpucibot merge

PR #846, opened 2022-09-26
Mdspanifying (currently tested) `raft::matrix` 

cjnolet:


cpp/include/raft/cluster/single_linkage.cuh
```diff
@@ -48,11 +49,49 @@ void single_linkage(const raft::handle_t& handle,
...
+ * @param[in] c a constant used when constructing connectivities from knn graph. Allows the indirect
+ control of k. The algorithm will set `k = log(n) + c`
+ * @param[in] n_clusters number of clusters to assign data samples
+ */
+template <typename value_t, typename idx_t, LinkageDistance dist_type = LinkageDistance::KNN_GRAPH>
+void single_linkage(const raft::handle_t& handle,
+                    raft::device_matrix_view<const value_t, idx_t, row_major> X,
+                    raft::distance::DistanceType metric,
+                    linkage_output<idx_t>& out,
+                    int c,
```
mhoemmen, line 84: [optional] Given that the caller could easily mix up the order of `c` and `n_clusters`, two parameters that mean different things, would you consider supplying these parameters in a struct with named fields instead?

cpp/include/raft/cluster/single_linkage_types.hpp
```diff
@@ -27,23 +29,33 @@ enum LinkageDistance { PAIRWISE = 0, KNN_GRAPH = 1 };
...
+  idx_t m;
+  idx_t n_clusters;
+
+  idx_t n_leaves;
+  idx_t n_connected_components;
 
-  value_idx n_leaves;
-  value_idx n_connected_components;
+  // TODO: These will be made private in a future release
+  idx_t* labels;    // size: m
```
mhoemmen, line 42: [optional] Would you consider just storing these two arrays as `device_{matrix,vector}_view`?  Also, if `get_children()` always returns an mdspan with 2 columns, would you consider encoding that in the returned mdspan's `extents` type (e.g., `extents<idx_t, dynamic_extent, 2>`)?

cpp/include/raft/core/mdspan.hpp
```diff
@@ -222,4 +222,112 @@ RAFT_INLINE_FUNCTION auto unravel_index(Idx idx,
...
+template <class ElementType, class Extents, class Accessor>
+constexpr bool is_row_or_column_major(mdspan<ElementType, Extents, layout_right, Accessor> /* m */)
+{
+  return true;
+}
+
+template <class ElementType, class Extents, class Accessor>
+constexpr bool is_row_or_column_major(mdspan<ElementType, Extents, layout_stride, Accessor> m)
+{
+  return m.is_exhaustive();
```
mhoemmen, line 246: Excellent : - D

cpp/include/raft/core/mdspan.hpp
```diff
@@ -222,4 +222,112 @@ RAFT_INLINE_FUNCTION auto unravel_index(Idx idx,
...
+template <class ElementType, class Extents, class Accessor>
+constexpr bool is_row_major(mdspan<ElementType, Extents, layout_right, Accessor> /* m */)
+{
+  return true;
+}
+
+template <class ElementType, class Extents, class Accessor>
+constexpr bool is_row_major(mdspan<ElementType, Extents, layout_stride, Accessor> m)
+{
+  return m.is_exhaustive();
```
mhoemmen, line 270: [required] I don't think this is quite right.  Counterexample: `layout_stride::mapping{1, num_rows}` (which is column major).  This should be straightforward to fix, though.

```c++
return m.is_exhaustive() && m.stride(1) == typename Extents::index_type(1);
```

cpp/include/raft/core/mdspan.hpp
```diff
@@ -222,4 +222,112 @@ RAFT_INLINE_FUNCTION auto unravel_index(Idx idx,
...
+  return true;
+}
+
+template <class ElementType, class Extents, class Accessor>
+constexpr bool is_col_major(mdspan<ElementType, Extents, layout_right, Accessor> /* m */)
+{
+  return false;
+}
+
+template <class ElementType, class Extents, class Accessor>
```
mhoemmen, line 291: [required] I don't think this is quite right.  Counterexample: `layout_stride::mapping{num_cols, 1}` (which is row major).  This should be straightforward to fix, though.

```c++
return m.is_exhaustive() && m.stride(0) == typename Extents::index_type(1);
```

cpp/include/raft/core/mdspan.hpp
```diff
@@ -222,4 +222,112 @@ RAFT_INLINE_FUNCTION auto unravel_index(Idx idx,
...
+constexpr bool is_col_major(mdspan<ElementType, Extents, layout_stride, Accessor> m)
+{
+  return m.is_exhaustive();
+}
+
+template <class ElementType, class IndexType>
+constexpr bool is_matrix_view(mdspan<ElementType, matrix_extent<IndexType>> /* m */)
```
mhoemmen, line 298: [required] There are a couple issues with the `is_{matrix,vector,scalar}_view` predicates.

1. This specialization assumes the default Layout (`layout_right`) and Accessor (`default_accessor`).  (Not specifying default template arguments means that the specialization is _only_ for the defaults.)
2. `is_matrix_view` really means `Extents::rank() == 2`.  `is_vector_view` means `Extents::rank() == 1`.  `is_scalar_view` means `Extents::rank() == 0`.  That's really all you need to check.

For example:

```c++
template <class ElementType, class IndexType, size_t ... Exts, class Layout, class Accessor>
constexpr bool is_matrix_view(mdspan<ElementType, extents<IndexType, Exts...>, Layout, Accessor> /* m */)
{
  return sizeof...(Exts) == 2; // 1 for is_vector_view, 0 for is_scalar_view
}
```

mhoemmen (changes requested):
I didn't get all the way through this one, but I did notice some issues with the predicates.  Thanks for letting me review! : - )

cpp/include/raft/cluster/detail/single_linkage.cuh
```diff
@@ -54,7 +54,7 @@ void single_linkage(const raft::handle_t& handle,
                     size_t m,
                     size_t n,
                     raft::distance::DistanceType metric,
-                    linkage_output<value_idx, value_t>* out,
+                    linkage_output<value_idx>* out,
```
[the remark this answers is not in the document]
cjnolet: Thanks! Actually I didn't mean to include any raft::cluster changes in this PR. I had started looking at them, but meant to back those changes out (probably a 22.12 task)

cpp/include/raft/cluster/single_linkage_types.hpp
cjnolet: I'd actually like to remove this class in a future version and just accept the outputs directly. I can't remember why this was necessary but it was long before  we started focusing on end-user consumption.

cjnolet: (As for the reason I just added the mdspan getters in the meantime, I was trying not to break the existing API right away.)>

cpp/include/raft/cluster/detail/single_linkage.cuh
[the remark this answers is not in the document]
cjnolet: I take that back- after looking through how far I got, I decided it wasn't a bad idea to go ahead and include it.

cpp/include/raft/matrix/print.cuh
```diff
@@ -43,14 +43,20 @@ void print(const raft::handle_t& handle,
...
+ * @param[in] handle raft handle for managing resources
+ * @param[in] in input matrix with column-major layout
+ * @param[in] h_separator: horizontal separator character
+ * @param[in] v_separator: vertical separator character
  */
 template <typename m_t, typename idx_t>
-void print(raft::host_matrix_view<const m_t, idx_t, col_major> in)
+void print(const raft::handle_t& handle,
+           raft::host_matrix_view<const m_t, idx_t, col_major> in,
+           char h_separator = ' ',
```
mhoemmen, line 57: [optional] In the future, would you consider representing parameters like these, where it's easy to mix up the order, as a single struct?

```c++
struct separators {
  char horizontal = ' ';
  char vertical = ' ';
};
```

C++20 users can then use designated initializers to get the effect of named parameters:

```c++
print(handle, in, separators{.horizontal = '*', .vertical = '|'});
```

mhoemmen (approved):
EXCELLENT

cpp/include/raft/matrix/reciprocal.cuh
```diff
@@ -0,0 +1,71 @@
...
+ * @param scalar: every element is multiplied with scalar
+ * @param setzero round down to zero if the input is less the threshold
+ * @param thres the threshold used to forcibly set inputs to zero
+ * @{
+ */
+template <typename math_t, typename idx_t, typename layout>
+void reciprocal(const raft::handle_t& handle,
+                raft::device_matrix_view<const math_t, idx_t, layout> in,
+                raft::device_matrix_view<math_t, idx_t, layout> out,
+                math_t scalar,
```
divyegala, line 40: Should this be a `raft::host_scalar_view`?

cpp/include/raft/matrix/sqrt.cuh
```diff
@@ -0,0 +1,99 @@
...
+ * @brief Square root of every element in the input matrix
+ * @tparam math_t data-type upon which the math operation will be performed
+ * @tparam idx_t integer type used for indexing
+ * @tparam layout layout of the matrix data (must be row or col major)
+ * @param[in] handle: raft handle
+ * @param[in] in: input matrix and also the result is stored
+ * @param[out] out: output matrix. The result is stored in the out matrix
+ */
+template <typename math_t, typename idx_t, typename layout>
+void sqrt(const raft::handle_t& handle,
```
divyegala, line 35: There's a `raft::linalg::sqrt` that works on any shape now, because this is essentially just an element-wise op. Maybe we should remove this version later

cpp/include/raft/cluster/single_linkage.cuh
```diff
@@ -48,11 +51,53 @@ void single_linkage(const raft::handle_t& handle,
...
+  out_arrs.children = dendrogram.data_handle();
+  out_arrs.labels   = labels.data_handle();
+
+  single_linkage<idx_t, value_t, dist_type>(handle,
+                                            X.data_handle(),
+                                            static_cast<std::size_t>(X.extent(0)),
+                                            static_cast<std::size_t>(X.extent(1)),
+                                            metric,
+                                            &out_arrs,
+                                            c.has_value() ? c.value() : DEFAULT_CONST_C,
```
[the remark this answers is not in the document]
cjnolet: That's true the optional does have a non nullopt default, but while nullopt is a valid input, we still need to treat it as if it could be provided by a user.

cpp/include/raft/matrix/col_wise_sort.cuh
```diff
@@ -50,7 +51,105 @@ void sort_cols_per_row(const InType* in,
...
+};
+
+template <typename T>
+using sorted_keys_t = typename sorted_keys_alias<T>::type;
+}  // namespace sort_cols_per_row_impl
+
+/**
+ * @brief Overload of `sort_keys_per_row` to help the
+ *   compiler find the above overload, in case users pass in
+ *   `std::nullopt` for one or both of the optional arguments.
```
divyegala, line 135: I thought this overload was only needed when there are two optional arguments? @mhoemmen

cpp/include/raft/matrix/init.cuh
```diff
@@ -0,0 +1,45 @@
...
+ * @tparam math_t data-type upon which the math operation will be performed
+ * @tparam idx_t integer type used for indexing
+ * @tparam layout layout of the matrix data (must be row or col major)
+ * @param[in] handle: raft handle
+ * @param[in] in input matrix
+ * @param[out] out output matrix. The result is stored in the out matrix
+ * @param[in] scalar scalar value to fill matrix elements
+ */
+template <typename math_t, typename idx_t, typename layout>
+void fill(const raft::handle_t& handle,
```
divyegala, line 36: We should probably have another element wise function for every possible mdspan shape

divyegala (approved):
There's quite a few redundant element-wise functions in here. Can you create an issue to convert those to encapsulate any mdspan shape for later on?

cjnolet:
I mentioned this offline but it's still unclear to me what exactly the final state of matrix vs linalg should look like after consolidation. For example, I'm not completely convinced that these element wise ops should be in linalg because they aren't linear algebra, yet because they see element wise ops they might not necessarily belong in the matrix namespace either. We need to think about it a bit more, I think.

cjnolet:
@gpucibot merge

PR #848, opened 2022-09-27
[REVIEW] Do not use strcpy to copy 2 char

### LINKED ISSUE #847: strcpy of two char
There are a few parts of the SVD function that do `strcpy(&dst, &src)`, where each of `dst` and `src` is a `char`.  For example:

https://github.com/rapidsai/raft/blob/1dd2feb12b3e6edf519202e1ae6452cfe690bc38/cpp/include/raft/linalg/detail/svd.cuh#L71

This is incorrect for several reasons.

1. `strcpy` will attempt to read past the bounds of `src` in order to look for a null terminator (`\0` character).  Just doing that is UB.
2. If `strcpy` is lucky enough to find a null terminator, it will write _both_ `src` _and_ the null terminator (thus, two characters) to `&dst`.  This is UB because `dst` only points to one character.  It could clobber something on the stack (maybe even `src`).
3. There's no guarantee that the next character after `src` is a null terminator.  This means `strcpy` could clobber arbitrarily many bytes on the stack with random other bytes on the stack.

mhoemmen:
Given `char src` and `char dst`, one should copy them like `dst = src`, not like `strcpy(&dst, &src)`.

Using `strcpy` in this way invokes undefined behavior, and may cause stack corruption.

Fixes #847.

### DIFF
--- a/cpp/include/raft/linalg/detail/svd.cuh
+++ b/cpp/include/raft/linalg/detail/svd.cuh
@@ -66,15 +66,9 @@ void svdQR(const raft::handle_t& handle,
   char jobu  = 'S';
   char jobvt = 'A';
 
-  if (!gen_left_vec) {
-    char new_u = 'N';
-    strcpy(&jobu, &new_u);
-  }
+  if (!gen_left_vec) { jobu = 'N'; }
 
-  if (!gen_right_vec) {
-    char new_vt = 'N';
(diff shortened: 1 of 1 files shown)


cjnolet (approved):

cjnolet (approved):

cjnolet:
@gpucibot merge

PR #855, opened 2022-09-28
[REVIEW] Include knn.cuh in knn.cu benchmark source for finding brute_force_knn

teju85:
Current issue is that if you try to build only the benchmark code by doing `./build.sh bench`, the file `cpp/bench/spatial/knn.cu` fails to compile with the following error:
```
...
./cpp/bench/spatial/knn.cu(177): error: namespace "raft::spatial::knn" has no member "brute_force_knn"
          detected during:
            instantiation of "void raft::bench::spatial::knn<ValT, IdxT, ImplT>::run_benchmark(benchmark::State &) [with ValT=float, IdxT=int64_t, ImplT=raft::bench::spa$ial::brute_force_knn<float, int64_t>]"
...
```

The issue is that this file doesn't include `knn.cuh` (Courtesy: @achirkin). This PR fixes the error.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -218,7 +218,9 @@ fi
 
 if hasArg bench || (( ${NUMARGS} == 0 )); then
     BUILD_BENCH=ON
+    COMPILE_DIST_LIBRARY=ON
     ENABLE_NN_DEPENDENCIES=ON
+    COMPILE_NN_LIBRARY=ON
     CMAKE_TARGET="${CMAKE_TARGET};bench_raft"
 fi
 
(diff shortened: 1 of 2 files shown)


cjnolet (approved):

cjnolet:
rerun tests

cjnolet:
rerun tests

cjnolet:
@teju85 this is weird and I'm not sure why CI keeps timing out on this PR. I thought there was somehting wrong w/ CI at first but it doesn't seem to time out like this on the other PRs. maybe there's another template we need to consider instantiating here?

cjnolet:
@teju85 can you try also including `raft/distance/specializations.cuh` inside `raft/spatial/knn/specializations.cuh`? I think the knn specializations, especially the fused l2 knn, implicitly assume the needed distance specializations have been compiled already.

cjnolet:
@gpucibot merge

PR #878, opened 2022-10-04
ivf-pq post integration hotfixes

achirkin:
Fixes to ivf_pq::search:

  1. Fixed a typo in argument name of `select_cluster` (n_queries -> queries_batch) which led to illegal memory access for large batches.
  2. Removed unnecessary argument `max_batch_size` from the worker function.
  3. Replaced `<<<bracket-calling>>>` of dynamically selected kernels with `cudaLaunchKernel` invocations.
     The former led to the kernels silently being not called at all in some cases for no apparent reason (not reproducible in tests).

Fixes to ivf_pq::build:
  1. A missing include `raft/util/device_atomics.cuh` passed unnoticed due to transient dependencies.


tfeher (approved):
Thanks Artem for the fix! It looks good to me.

cjnolet:
@gpucibot merge

PR #887, opened 2022-10-04
Removing cub symbol from libraft-distance instantiation.

cjnolet:
The recent template instantiations for `fusedL2NN` included `cub::KeyValuePair` in the symbols. Unfortunately, the cub symbols end up including all of the cuda architectures upon which it is built in the symbol name. For example, if libraft-distance is built for 5 architectures then someone building locally using that libraft-distance package for only their architecture will get an undefined symbol error. The best solution here is to not allow symbols from cub to leak through any publicly exposed APIs (even if those APIs are template instantiations of implementaiton details). 

I'm also adding a new `raft::KeyValuePair` object which has a conversion constructor so we can still easily make use of `cub::KeyValuePair` inside RAFT functions which use our new KVP object.


seunghwak (approved):

BradReesWork (approved):

cjnolet:
@gpucibot merge

PR #891, opened 2022-10-05
ivf-pq integration: hotfixes

achirkin:
A second set of hotfixes for ivf-pq:

  1. `max_samples` is calculated from `cluster_offsets` to reduce the number of arguments and register pressure when it's not needed. The calculation on the kernel side was wrong and not matching the calculation on the host side (missing roundUp/128), which led to incorrect offsets inside the filled distance array.


cjnolet (approved):

cjnolet:
@gpucibot merge

cjnolet (approved):

cjnolet:
@gpucibot merge

PR #892, opened 2022-10-05
Output non-normalized distances in IVF-PQ and brute-force KNN

Nyrio:
Solves point 9 of #711 (the observed errors were due to comparing distances at different scales).

This PR does two things:

- Changes the naive BF KNN to not normalize int when converting to float and simply cast, resulting in distances in the same scale as the output of IVF-Flat (and solving errors observed in tests).
- Changes IVF-PQ to output distances in the original scale and not the normalized scale, consistently with IVF-Flat.


cpp/include/raft/spatial/knn/detail/ivf_pq_search.cuh
```diff
@@ -979,7 +1007,7 @@ struct ivfpq_compute_similarity {
  *   3. split the query batch into smaller chunks, so that the device workspace
  *      is guaranteed to fit into GPU memory.
  */
-template <typename ScoreT, typename LutT, typename IdxT>
+template <typename T, typename ScoreT, typename LutT, typename IdxT>
```
achirkin, line 1010: Rather than adding here an extra parameter, maybe it's better to define the `config` struct for half with the normalization constant == 1?

Nyrio: @achirkin I need this template parameter to know the normalization constant, there is no other way to know what was the original dataset type afaik. Unless I missed something?

cpp/include/raft/spatial/knn/detail/ivf_pq_search.cuh
```diff
@@ -393,11 +393,27 @@ void postprocess_neighbors(IdxT* neighbors,  // [n_queries, topk]
...
 
+template <typename T>
+HDI auto undo_normalization(const float& x, bool squared)
+  -> std::enable_if_t<std::is_same_v<T, float>, float>
+{
+  return x;
+};
+
+template <typename T>
+HDI auto undo_normalization(const float& x, bool squared)
```
achirkin, line 404: I think it would be less code if we didn't have this function at all? :) If at some point we add more metric types, we will have to continue adding more parameter here (power parameter, etc.). Also just logically `undo_normalization` would mean different things here for different metrics.

cpp/include/raft/spatial/knn/detail/ivf_pq_search.cuh
Nyrio: What I mean is: `ScoreT` is `float` or `half`, but `T` could be `int8_t`. In the post-processing step, we'll convert `float` to `float` but using the normalization constant of `int8_t`

achirkin (changes requested):
Thanks for the fix! I think we can make it a bit less complicated, see the couple comments

cpp/include/raft/spatial/knn/detail/ivf_pq_search.cuh
Nyrio: Alternately if we want to limit template instantiations, we can pass the constant. What do you think?

achirkin: Ah, right, I forgot the ScoreT cannot be int8_t/uint8_t; never mind!

achirkin: > Alternately if we want to limit template instantiations, we can pass the constant. What do you think?

No, I don't like that this constant would sort of mean different things. We can move the postprocessing step for the scores out of the worker code completely though!

Nyrio: > We can move the postprocessing step for the scores out of the worker code completely though!

Not really, `topk_dists` lives inside `ivfpq_search_worker` and uses `ScoreT`. Of course we could have two steps of post-processing but it wouldn't be efficient.

Nyrio:
There are quite a few IVF-PQ cases where this PR gives worse results than before, I'll look into those.

Nyrio:
Note: the worse test results that I observed in IVF-PQ are due to how `raft::CompareApprox` works: some mismatches were previously tolerated because the absolute error was lower than `eps`, which could only happen in the normalized scale but doesn't happen anymore in the original scale. So there is not much more we can do. Note that the tests are passing anyway.

Also, I am removing the "breaking" flag, since IVF-PQ was added in the same release.

cpp/include/raft/spatial/knn/detail/ivf_pq_search.cuh
```diff
@@ -1366,6 +1388,9 @@ inline void search(const handle_t& handle,
                       rot_queries.data() + uint64_t(index.rot_dim()) * offset_b,
                       neighbors + uint64_t(k) * (offset_q + offset_b),
                       distances + uint64_t(k) * (offset_q + offset_b),
+                      std::is_same_v<T, float>
```
achirkin, line 1391: I think, there's no need for this extra check of the target type?

cpp/include/raft/spatial/knn/detail/ivf_pq_search.cuh
```diff
@@ -405,23 +405,42 @@ void postprocess_distances(float* out,        // [n_queries, topk]
                            distance::DistanceType metric,
                            uint32_t n_queries,
                            uint32_t topk,
+                           double scaling_factor,
```
achirkin, line 408: Would you consider using `float` here and in the other signatures? We don't seem to be gaining anything by passing this factor as `double` here, since in all cases it's immediately casted to `float` anyway.

achirkin (changes requested):
Thanks, now the changes seem nice and non-intrusive to me. Just couple of tiny suggestions.

tfeher (approved):
Thanks Louis for the PR, it looks good to me!

achirkin (approved):
Thanks!

cjnolet:
@gpucibot merge

PR #896, opened 2022-10-05
[REVIEW] mdspan-ify uniformInt and normalTable

mhoemmen:
* Add overloads of `raft::random::uniformInt` and `raft::random::normalTable` that take mdspan.
* Add unit tests for each case.
* Improve documentation of `normalTable` to explain that the output table uses a row-major layout.

### DIFF
--- a/cpp/include/raft/random/rng.cuh
+++ b/cpp/include/raft/random/rng.cuh
@@ -75,6 +75,33 @@ void uniform(const raft::handle_t& handle,
 /**
  * @brief Generate uniformly distributed integers in the given range
  *
+ * @tparam OutputValueType Integral type; value type of the output vector
+ * @tparam IndexType Type used to represent length of the output vector
+ *
+ * @param[in] handle raft handle for resource management
+ * @param[in] rng_state random number generator state
+ * @param[out] out the output vector of random numbers
+ * @param[in] start start of the range
+ * @param[in] end end of the range
+ */
+template <typename OutputValueType, typename IndexType>
+void uniformInt(const raft::handle_t& handle,
+                RngState& rng_state,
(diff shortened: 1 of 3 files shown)


cpp/include/raft/random/rng.cuh
```diff
@@ -170,7 +199,70 @@ void normalInt(const raft::handle_t& handle,
...
+ * @tparam OutputValueType data type of output random number
+ * @tparam IndexType data type used to represent length of the arrays
+ *
+ * @param[in] handle raft handle for resource management
+ * @param[in] rng_state random number generator state
+ * @param[in] mu_vec mean vector (of length `out.extent(1)`)
+ * @param[in] sigma_vec Either the standard-deviation vector
+ *            (of length `out.extent(1)`) of each component,
+ *            or a scalar standard deviation for all components.
+ * @param[out] ptr the output table
```
cjnolet, line 214: I think this is intended to be `@param[out] out` now, right?

cjnolet (approved):
I think there's one misnamed argument error in the docs that might fail the CI build, but otherwise this looks great. Approving pending CI.

cpp/include/raft/random/rng.cuh
mhoemmen: @cjnolet Oh oof, that's right!  I need to get ready for a meeting, but I can fix this either here or in PR #897 -- please let me know which you prefer -- thanks!

cjnolet: No problem! I went ahead and just made the change since it's so tiny. Both PRs look great so since the changes are already in #897, I think it's fine to just merge that one over.

mhoemmen:
Superseded by PR #897; thanks for the review!

PR #898, opened 2022-10-06
[REVIEW] Update Python build instructions [skip-ci]

betatim:
This explicilty mentions that `nvcc` has to be installed manually and updates the build and pytest commands.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -10,6 +10,7 @@
     - [Googlebench](#gbench)
     - [C++ Using Cmake](#cpp_using_cmake)
     - [Python](#python)
+    - [Documentation](#docs)
 - [Using RAFT in downstream projects](#use_raft)
     - [Cmake Header-only Integration](#cxx_integration)
     - [Using Shared Libraries in Cmake](#use_shared_libs)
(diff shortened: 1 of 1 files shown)


wphicks (approved):

cjnolet:
Sorry @wphicks, I actually plan to include this in 22.10 since it's just a simple change.

BUILD.md
```diff
@@ -161,32 +161,24 @@ mamba activate raft_env_name
...
-```bash
-cd python/raft-dask
-python setup.py build_ext --inplace
-python setup.py install
-
-cd python/pylibraft
-python setup.py build_ext --inplace
-python setup.py install
+# to build pylibraft
+./build.sh libraft pylibraft --install --compile-libs
```
cjnolet, line 165: I think it's great to have the build.sh variants in here as well but i'd like to also keep the manual build instructions for developers to know that the other method can be used as well. I think it would be best to include both, and maybe move the setup.py can also be used to build the Python APIs manuall after the build.sh instructions?

cjnolet (changes requested):
Thanks for doing this, Tim! Just one little change but otherwise it looks great.

cjnolet (approved):

cjnolet:
Some issues we've encountered recently in CI required that we remove the explicit (re-)building of the python packages from `build.sh`. This also means that `build.sh docs` no longer builds the python packages, which are required to be built, in order to build the docs. I've added this to the BUILD docs.

cjnolet:
@gpucibot merge

betatim:
Thanks for the comments and merge!

PR #909, opened 2022-10-10
Integrate `accumulate_into_selected` from ANN utils into `linalg::reduce_rows_by_keys`

Nyrio:
`accumulate_into_selected` achieves much better performance than the previous implementation of `reduce_rows_by_keys` for large `nkeys` (`sum_rows_by_key_large_nkeys_kernel_rowmajor`). According to the benchmark that I added for this primitive, the difference is a factor of 240x for sizes relevant to IVF-Flat (and a factor of ~10x for smaller `nkeys`, e.g 64).

This is mostly because the legacy implementation, probably in an attempt to reduce atomic conflicts, assigned a key and a tile of the matrix to each block, and the block only reduces the rows corresponding to the assigned key. With a very large number of keys, e.g 1k, this results in blocks iterating over a large number of rows (possibly tens of thousands) and only reading and accumulating 1 in 1k rows.

This PR:

- Replaces `sum_rows_by_key_large_nkeys_rowmajor` with `accumulate_into_selected` (I didn't find any cases in which the old kernel performed better).
- Removes `accumulate_into_selected` from `ann_utils.cuh`.
- Fixes support for custom iterators in `reduce_rows_by_keys`.
- Uses the raft prims in `calc_centers_and_sizes`.

Perf notes:

- The original kmeans gets a 15-20% speedup for large numbers of clusters.
- The performance of `ivf_flat::build` stays the same as before.
- There are a bunch of extra steps since I separated the cluster size count from the reduction by key, but they are quite neglectable in comparison.

Question: the change breaks support for host-side-only arrays in `calc_centers_and_sizes`, is it actually a possibility? Should I add a branch and not use the raft prims when all arrays are host-side?

cc @achirkin @tfeher @cjnolet


cpp/bench/CMakeLists.txt
```diff
@@ -96,6 +96,7 @@ if(BUILD_BENCH)
             bench/linalg/add.cu
             bench/linalg/map_then_reduce.cu
             bench/linalg/matrix_vector_op.cu
+            bench/linalg/reduce_rows_by_key.cu
```
cjnolet, line 99: So happy to see all these new benchmarks being added

cpp/include/raft/spatial/knn/detail/ann_utils.cuh
```diff
@@ -383,6 +317,8 @@ __global__ void map_along_rows_kernel(
  * @brief Map a binary function over a matrix and a vector element-wise, broadcasting the vector
  * values along rows: `m[i, j] = op(m[i,j], v[i])`
  *
+ * @todo(lsugy): replace with matrix_vector_op
```
cjnolet, line 320: Can we add a GitHub issue for this and reference it here just to make sure we are tracking it?

cpp/include/raft/linalg/reduce_rows_by_key.cuh
```diff
@@ -54,21 +56,27 @@ void convert_array(IteratorT1 dst, IteratorT2 src, int n, cudaStream_t st)
...
+                        IdxT lda,
                         const KeysIteratorT d_keys,
                         const WeightT* d_weights,
                         char* d_keys_char,
-                        int nrows,
-                        int ncols,
-                        int nkeys,
-                        DataIteratorT* d_sums,
-                        cudaStream_t stream)
+                        IdxT nrows,
```
cjnolet, line 71: We are going to be deprecating the raw pointer APIs soon in favor of the new (more self-documenting) mdspan APIs. Do you see any reasons why we should prefer to keep the iterator-based APIs over the mdspan APIs?

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
```diff
@@ -255,13 +260,66 @@ void calc_centers_and_sizes(float* centers,
...
+  if (!reset_counters) {
+    temp_cluster_sizes.resize(n_clusters, stream);
+    temp_sizes = temp_cluster_sizes.data();
+  }
+
+  utils::mapping<float> mapping_op;
+  cub::TransformInputIterator<float, utils::mapping<float>, const T*> mapping_itr(dataset,
+                                                                                  mapping_op);
+
+  // todo(lsugy): use iterator from KV output of fusedL2NN
```
cjnolet, line 281: Can you create an issue for this and reference it here for tracking? Do you see any reason it would be more beneficial to use iterators here over the mdspan API?

cjnolet (changes requested):
Thanks again @Nyrio for this consolidation and optimization! Minor things again

cpp/include/raft/spatial/knn/detail/ann_utils.cuh
Nyrio: Oops I didn't mean to leave that comment here, but I already have a WIP PR for this: #911

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
Nyrio: The point of using an iterator here would be to avoid having one extra step after `fusedL2NN` to extract and cast the key (`fusedL2NN` can output key-value, value-only, but not key-only).

cpp/include/raft/linalg/reduce_rows_by_key.cuh
cjnolet: Yeah, I get the point of the "streamability" of the iterators vs raw pointers. Given that the mdspan is really just a very lightweight wrapper around any type (it can be but doesn't have to be a raw pointer) and we're just forwarding the underlying `data_handle()` to the function in the `detail` namespace, can't we also wrap the mdspan around an iterator?

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
```diff
@@ -231,22 +234,24 @@ constexpr inline auto calc_minibatch_size(uint32_t n_clusters,
  *    When set to `false`, this function may be used to update existing centers and sizes using
  *    the weighted average principle.
  * @param stream
+ * @param mr (optional) memory resource to use for temporary allocations
```
tfeher, line 237: Please update the docstring, stating that pointers have to be on the device.

tfeher (changes requested):
Thanks @Nyrio for the PR! I confirm that there is no need to handle host only pointers in `calc_centers_and_sizes`: only historically was that necessary, now all the calculation is done in device or managed memory for the IVF methods. Some nitpicks below, otherwise it looks good.

cpp/include/raft/linalg/reduce_rows_by_key.cuh
Nyrio: > mdspan is really just a very lightweight wrapper around any type (it can be but doesn't have to be a raw pointer)

Not exactly. As far as I understand, span or mdspan can only wrap around data that exists in memory (and continuously). `std::span` even has a member `data` which returns a pointer to the memory location of the first element. So it can wrap around simple iterators like `vector::begin` but not `cub::TransformInputIterator`.

cjnolet: mdspan and span are a little different in both form and function. we have been investing in mdspan more than span for our public APIs. mdspan itself does not require contiguity, though we do enforce that in many of the public APIs for the simple reason that we are using it as a facade for raw pointers. If we are expecting iterators, we could relax that constraint a bit when the underlying data_handle() is in fact an iterator. Also, @mhoemmen can correct me if I'm wrong but I don't believe the data backing an mdspan does need to exist in memory- I think it could even be a file pointer or a pointer to a remote data buffer so long as the accessor knows how to materialize it and it allows for random access. 

Im not trying to discourage the use of iterators in our public API and I definitely see the benefits of using them to apply functors lazily upon materialization to avoid copies / additional allocations. I'm hoping we can incorporate acceptance of iterators more consistently and broadly across our APIs and avoid having a only a couple functions throughout the codebase that accept iterators while most do not.

cjnolet: I'm also okay accepting this PR and adding a todo to figure out how we can accept iterator types more broadly across the new APIs so we can consolidate the dev experience a bit for our end-users.

Nyrio: Oh, I thought of `mdspan` as a multi-dimensional `std::span` but it seems that it's a bit more versatile.
So, to make sure I understand: each custom iterator would require a different accessor? And the mdspan type for iterator-accepting args will still be a template parameter, while args expecting raw pointers can use matrix or vector views?

mhoemmen: > Also, @mhoemmen can correct me if I'm wrong but I don't believe the data backing an mdspan does need to exist in memory- I think it could even be a file pointer or a pointer to a remote data buffer so long as the accessor knows how to materialize it and it allows for random access.

@cjnolet is correct (as usual!).  We say "mdspan doesn't need a backing span."

1. Elements don't need to exist in memory;
2. `data_handle_type` doesn't need to be `ElementType*`;
3. even if it is, or even if `data_handle_type` presents the syntax of an iterator, it doesn't need to act like one;
4. `reference` doesn't need to be `element_type&`; and
5. `access(p, i)` doesn't need to return `p[i]`.

One example of (3) is `MPI_Win` (which could be `void*`, but it's not a pointer to an array), a handle to remote memory.

tfeher (approved):
Thanks Louis for the update, it looks good to me!

cpp/include/raft/linalg/reduce_rows_by_key.cuh
mhoemmen: Also, it looks like all of RAFT's rank-1 mdspan would let one use `data_handle()`, `data_handle() + extent(0)` as an iterator range.

Nyrio:
@cjnolet If that's ok with you, can we merge this and mdspanify later? mdspanifying custom iterators requires helpers and types that we don't have yet.

cjnolet (approved):

cjnolet:
@gpucibot merge

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
```diff
@@ -285,7 +285,7 @@ void calc_centers_and_sizes(const handle_t& handle,
 
   // Compute weight of each cluster
   raft::cluster::detail::countLabels(
-    handle, labels, temp_sizes, (uint32_t)n_rows, (uint32_t)n_clusters, workspace);
+    handle, labels, temp_sizes, (int64_t)n_rows, (int64_t)n_clusters, workspace);
```
mhoemmen, line 288: Would you consider using C++ - style casts instead of C-style casts?

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
```diff
@@ -303,8 +307,8 @@ void calc_centers_and_sizes(const handle_t& handle,
     centers,
     centers,
     sizes_f,
-    (int64_t)dim,
-    (int64_t)n_clusters,
+    static_cast<int64_t>(dim),
```
mhoemmen, line 310: Thanks so much for doing this!  It tends to matter more for pointer conversions, but integer conversions can also be a source of bugs, so it's nice to be able to search for `static_cast`.

Nyrio: Yes, no problem. I don't always use `static_cast` for trivial conversions but I don't mind using it.

PR #911, opened 2022-10-10
Replace `map_along_rows` with `matrixVectorOp`

Nyrio:
The prim `matrixVectorOp` is more optimized than the ANN util function `map_along_rows`. However, in order to substitute the latter, I needed to add better support for differing matrix and vector types.

This PR adds support to arbitrary matrix and vector types in matrix-vector ops. The input and output matrices must have the same type but the vectors can each have different types.


Nyrio:
Please note that the use of `thrust::tuple` is a substitute for `std::tuple` due to some issues with the support of tuples in CUDA, in particular some functions like the assignment operator, `std::tie`, etc, don't have a `__device__` version. I left a TODO comment to switch to `cuda::std::tuple` if we have `libcudacxx` as a requirement in the future.

To get a feeling of how tuples and variadic arguments work together, you can look at this simpler example that I wrote when I was getting a grasp of this:

```cpp
template <typename Lambda, class Tuple, size_t...Is>
void run_op(Lambda op, Tuple&& args, std::index_sequence<Is...>) {
    op((std::get<Is>(args))...);
}

template <typename Lambda, typename... Args>
void f(Lambda op, int n, const Args*... args) {
    for(int i = 0; i < n; i++) {
        std::tuple<Args...> t = std::make_tuple((args[i])...);
        auto index = std::index_sequence_for<Args...>();
        run_op(op, t, index);
    }
}

int main() {
    std::vector<int> a = {1, 2, 3};
    std::vector<double> b = {4., 5., 6.};
    f([] (int x, double y) { std::cout << x * y << std::endl; }, 3, a.data(), b.data());
}
```

cpp/test/matrix/linewise_op.cu
```diff
@@ -216,7 +216,6 @@ struct LinewiseTest : public ::testing::TestWithParam<typename ParamsReader::Par
     std::vector<std::tuple<I, I>> dims;
```
achirkin, line 216: While you're at this, would you mind making a benchmark out of the main test loop here ([run()](https://github.com/rapidsai/raft/blob/4718e5bc73b0235c34581ed7f6f9c320d7e7b347/cpp/test/matrix/linewise_op.cu#L151) function)?
I've made this test before we had benchmarks and used the NVTX annotations to benchmark kernel run times in various cases. Would be nice to have it in a proper place :)

cpp/include/raft/matrix/detail/linewise_op.cuh
```diff
@@ -484,6 +517,7 @@ void matrixLinewiseVecRows(Type* out,
     const uint expected_grid_size = rowLen / raft::gcd(block_work_size, uint(rowLen));
     // Minimum size of the grid to make the device well occupied
     const uint occupy = getOptimalGridSize<BlockSize>();
+    // todo(lsugy): alignedLen instead of totalLen?
```
achirkin, line 520: Good point! Do any tests fail if we change this to `alignedLen`? Could be a typo, but could be an important undocumented quirk as well :)

cpp/include/raft/matrix/detail/linewise_op.cuh
```diff
@@ -527,7 +561,7 @@ struct MatrixLinewiseOp {
                   const bool alongLines,
                   Lambda op,
                   cudaStream_t stream,
-                  Vecs... vecs)
+                  const Vecs*... vecs)
```
achirkin, line 564: Would you consider splitting the matrix element `Type` into `InT`/`OutT` types for bonus points? I guess, the only thing we'd need to change in the logic then is how we check the alignment requirements for input/output.

achirkin (changes requested):
Very nice to see this done and I also like the fancier tuple approach! I only have mild concerns about performance, as this change increase the shared memory requests and possible ALU instruction counts. Worth checking if the register pressure changed; it tends to grow really quickly with the number of vector arguments.

cpp/include/raft/matrix/detail/linewise_op.cuh
```diff
@@ -21,11 +21,28 @@
...
 #include <algorithm>
+#include <thrust/tuple.h>
 
 namespace raft {
 namespace matrix {
 namespace detail {
 
+/** This type simplifies returning arrays and passing them as arguments */
+template <typename Type, int VecElems>
+struct VecArg {
```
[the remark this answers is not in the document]
Nyrio: This was precisely meant to replace `TxN` because of cases like `MatT`=4B and `VecT`=8B. If `VecBytes`=4B we can't use `TxN` as `VecElems` would be 0.

cpp/include/raft/matrix/detail/linewise_op.cuh
```diff
@@ -171,8 +189,10 @@ struct Linewise {
     }
     __syncthreads();
     {
-      Vec out;
-      *out.vectorized_data() = reinterpret_cast<typename Vec::io_t*>(shm)[threadIdx.x];
+      VecArg<VecT, VecElems> out;
```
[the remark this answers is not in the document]
Nyrio: I'll have to benchmark, but my intuition was that this isn't significant compared to global reads.

cpp/include/raft/matrix/detail/linewise_op.cuh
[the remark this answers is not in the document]
achirkin: I don't get it. Why wouldn't we parameterize TxN by type of the vector element, for each vector? Then if, for example, matrix elements are uint8_t, and two vector arguments are float and double, we'd be limited to `VecElems = 2`, and the size of reads 2, 8,  16 for the matrix and vector arguments correspondingly.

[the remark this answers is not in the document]
Nyrio: I think I see what you mean. What I tried previously that didn't work was to use `Linewise<VecT, IdxT, VecBytes, BS>::Vec` but what I should have used was `TxN_t<VecT, VecElems>`. The difference here is that `VecElems` will be `VecBytes / MatT` instead of `VecBytes / VecT`.

cpp/include/raft/matrix/detail/linewise_op.cuh
Nyrio: As it complicates the alignment logic a bit and we haven't seen a case yet where this is needed, I think it's better to keep that in the backlog.

cpp/include/raft/matrix/detail/linewise_op.cuh
Nyrio: All tests pass with the change.

cpp/test/matrix/linewise_op.cu
Nyrio: I've extended the matrix-vector op benchmark.

cpp/include/raft/matrix/detail/linewise_op.cuh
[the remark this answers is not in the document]
Nyrio: I've been running some benchmarks and the differences I see before/after are no more than noise.

cpp/include/raft/matrix/detail/linewise_op.cuh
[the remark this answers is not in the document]
Nyrio: As discussed offline, using `TxN` with the vector type but `VecElems` computed from the matrix type is currently not possible, because you can e.g have a situation where you want a vector type of 4 doubles, which isn't possible.

@achirkin suggested supporting more than one vector per `TxN` but I think it's outside of the scope of this PR and my change doesn't have any visible performance impact (reading the vector elements in shared mem is an insignificant operation).

cpp/include/raft/matrix/detail/linewise_op.cuh
```diff
@@ -309,20 +338,22 @@ __global__ void __launch_bounds__(BlockSize)
...
-  constexpr uint workSize = L::VecElems * BlockSize;
-  uint workOffset         = workSize;
-  __shared__ __align__(sizeof(Type) * L::VecElems)
-    Type shm[workSize * ((sizeof...(Vecs)) > 1 ? 2 : 1)];
+  constexpr uint workSize         = L::VecElems * BlockSize;
+  constexpr size_t maxVecItemSize = maxSizeOf<Vecs...>();
+  uint workOffset                 = workSize * maxVecItemSize;
+  __shared__ __align__(
+    maxVecItemSize *
+    L::VecElems) char shm[workSize * maxVecItemSize * ((sizeof...(Vecs)) > 1 ? 2 : 1)];
```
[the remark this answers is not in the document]
Nyrio: The primary reason for switching to the loop is not the alignment of the shared buffer, cf the first discussion above.

cjnolet (changes requested):
Changes look good from my end. Just fairly minor things as usual.

cpp/include/raft/linalg/detail/matrix_vector_op.cuh
```diff
@@ -22,80 +22,10 @@ namespace raft {
 namespace linalg {
 namespace detail {
 
-namespace {
```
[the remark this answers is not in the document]
Nyrio: It won't break cuML because all template types are inferred.
It could in theory break other projects if they provide the template list explicitly, but I'm not aware of any.

Nyrio:
I've solved the conflicts with #725 and #937 and done the changes requested by @cjnolet 

Note that I see failures in the linewise tests with padded spans, not only on this branch but also on `branch-22.12`. I have informed @mfoerste4 about those failures (I suspect missing boundary checks when reading the vectors).

mfoerste4:
> I've solved the conflicts with #725 and #937 and done the changes requested by @cjnolet
> 
> Note that I see failures in the linewise tests with padded spans, not only on this branch but also on `branch-22.12`. I have informed @mfoerste4 about those failures (I suspect missing boundary checks when reading the vectors).

@Nyrio , thanks for finding this. I added [#964](https://github.com/rapidsai/raft/pull/964) to fix the issue in the test setup.

cjnolet:
@Nyrio the fix to linewise op has been merged.

cjnolet:
rerun tests

Nyrio:
rerun tests

cpp/include/raft/matrix/detail/linewise_op.cuh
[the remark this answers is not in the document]
tfeher: @Nyrio could you add an issue about this, so that we don't forget it?

cpp/test/linalg/matrix_vector_op.cu
```diff
@@ -20,228 +20,229 @@
...
+MVTEST(MatVecOpTestF_i32_add2vec, inputs_i32, MV_EPS_F);
+MVTEST(MatVecOpTestF_i64_add1vec, inputs_i64, MV_EPS_F);
+MVTEST(MatVecOpTestF_i64_add2vec, inputs_i64, MV_EPS_F);
+MVTEST(MatVecOpTestD_i32_add1vec, inputs_i32, MV_EPS_D);
+MVTEST(MatVecOpTestD_i32_add2vec, inputs_i32, MV_EPS_D);
+MVTEST(MatVecOpTestD_i64_add1vec, inputs_i64, MV_EPS_D);
+MVTEST(MatVecOpTestD_i64_add2vec, inputs_i64, MV_EPS_D);
+
+/*
+ * This set of tests covers cases with different types.
```
tfeher, line 224: It would be nice to test the case of `int8_t` matrix with float input, so that we cover a different ratio of input data sizes. Since all checks pass here, this need not hold up the current PR, and could be implemented separately.

tfeher (approved):
Thanks @Nyrio for the PR! It looks good to me.

cpp/include/raft/matrix/detail/linewise_op.cuh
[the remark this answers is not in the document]
Nyrio: #990

cpp/include/raft/util/itertools.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Helpers inspired by the Python itertools library
+ *
+ */
+
+namespace raft::util::itertools {
+
+template <class S, typename... Args, size_t... Is>
+inline std::vector<S> product(std::index_sequence<Is...> index, const std::vector<Args>&... vecs)
```
cjnolet, line 30: Just a side question- is there a reason for explicitly inlining templated functions like this?

Nyrio: The point of inlining here was not to avoid linking errors, since it's a templated function as you said. My thinking was just that typically it's only going to be called once for a given set of template arguments (because the use case is mostly defining inputs/outputs for tests/benchmarks).

But since there is little benefit in inlining it anyway, I don't mind removing that keyword.

cjnolet: In general, we have been trying to remove the use of inline across the codebase unless absolutely necessary. IIRC, the use of incline (whether host or device) is just a suggestion and doesn't mean the compiler will always honor it, right? But I think if we don't need it then it could help in our quest to decrease compile times.

cjnolet: I think if it's something that's not intended to be invoked directly by the user then we should probably at least make an attempt to separate the private bits. If the in-line is absolutely needed then it's fine to keep it

Nyrio: On second thought the anonymous namespace wouldn't be appropriate since it's equivalent to making the function static, so I've separated it in raft fashion in a corresponding `detail` namespace, and kept it inline. I've also used the occasion to add documentation to the public API.

Nyrio:
@tfeher The last commit is for you. Re-requesting a review, so you can have a quick look at the new `int8_t`/`float` test.

cjnolet:
@gpucibot merge

PR #877, opened 2022-10-03
Mdspanifying remaining `raft::matrix` files

cjnolet:
- [x] Tests for `raft::matrix::argmax`
- [x] Tests for `raft::matrix::diagonal`
- [ ] Tests for `raft::matrix::triangular`
- [ ] Tests for `raft::matrix::reverse`
- [ ] Tests for `raft::matrix::slice`
- [ ] Tests for `raft::matrix::norm`
- [ ] Tests for `raft::linalg::matrix_vector` (moved from raft::matrix)

PR #912, opened 2022-10-11
Create cub-based argmin primitive and replace `argmin_along_rows` in ANN kmeans

Nyrio:
This PR follows up on [a suggestion](https://github.com/rapidsai/raft/pull/821#discussion_r984815136) from @cjnolet. The new `argmin` primitive is up to 5x faster than `argmin_along_rows` for dimensions relevant to ANN kmeans, and removes code duplication.

The reasons why it is faster are:

- `argmin_along_rows` often misses on doing a sequential reduction before the tree reduction, especially as it uses large block sizes, as much as 1024.
- CUB has a better reduction algorithm than the basic shared-mem reduction used in `argmin_along_rows`.
- If we switch the `argmin` prim to using the `cub::BLOCK_REDUCE_RAKING_COMMUTATIVE_ONLY` algorithm, we can get up to 30% further speedup! (I believe it's safe to use the commutative algorithm here since the offset is contained in the key-value pair so the reduction operation is commutative).

The speedup that I have measured for IVF-Flat build with the `InnerProduct` metric is around 15%.


cpp/include/raft/matrix/math.cuh
```diff
@@ -302,6 +302,20 @@ void ratio(
...
+/**
+ * @brief Argmin: find the row idx with minimum value for each column
+ * @param in: input matrix
+ * @param n_rows: number of rows of input matrix
+ * @param n_cols: number of columns of input matrix
+ * @param out: output vector of size n_cols
+ * @param stream: cuda stream
+ */
+template <typename math_t, typename out_t, typename idx_t = int>
+void argmin(const math_t* in, idx_t n_rows, idx_t n_cols, out_t* out, cudaStream_t stream)
```
cjnolet, line 314: We recently deprecated math.cuh so that we can split these files out and make the primitives easier for users to find. Let's add these to argmin.cuh and argmax.cuh

cpp/include/raft/spatial/knn/detail/ann_utils.cuh
```diff
@@ -151,66 +151,6 @@ inline void memzero(T* ptr, IdxT n_elems, rmm::cuda_stream_view stream)
   }
 }
 
-template <typename IdxT, typename OutT>
-__global__ void argmin_along_rows_kernel(IdxT n_rows, uint32_t n_cols, const float* a, OutT* out)
```
cjnolet, line 155: Definitely happy to things getting consolidated!

cpp/include/raft/matrix/math.cuh
```diff
@@ -310,8 +324,8 @@ void ratio(
  * @param out: output vector of size n_cols
  * @param stream: cuda stream
  */
-template <typename math_t, typename idx_t>
-void argmax(const math_t* in, int n_rows, int n_cols, idx_t* out, cudaStream_t stream)
+template <typename math_t, typename out_t, typename idx_t = int>
+void argmax(const math_t* in, idx_t n_rows, idx_t n_cols, out_t* out, cudaStream_t stream)
```
cjnolet, line 328: I redently wrote a quick test case for argmax because I realized it was missing. Let's make sure both of these are being tested. Also, ideally we would be adding new APIs that use mdspan and deprecating the raw pointer APIs where possible.

cjnolet (changes requested):
Thanks for these changes, @Nyrio! Definitely happy to see prims for both argmin and argmax. Mostly minor things.

cpp/bench/matrix/argmin.cu
```diff
@@ -0,0 +1,59 @@
...
+namespace raft::bench::linalg {
+
+struct ArgminParams {
+  int64_t rows, cols;
+};
+
+template <typename T, typename OutT>
+struct Argmin : public fixture {
+  Argmin(const ArgminParams& p)
+    : params(p), matrix(p.rows * p.cols, stream), indices(p.cols, stream)
```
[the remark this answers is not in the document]
Nyrio: No. But actually, I've seen a few other benchmarks that leave the data uninitialized. E.g `reduce.cu`. Do you think it would be better to fill it with uniform random data?

cpp/include/raft/matrix/math.cuh
cjnolet: I've closed #877 in favor if #937 but once that's in, this PR should use the new separate argmax header. You can also merge that into your branch if it makes it easier. It's still being reviewed but there shouldn't be any huge API changes (mostly tests and docs)

cpp/bench/matrix/argmin.cu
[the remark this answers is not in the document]
tfeher: Yes please.

cpp/include/raft/matrix/math.cuh
cjnolet: Just a small update. #937 should be merged soon.

cjnolet: It's been merged so we now have the argmax broken out into its own header and corresponding test.

Nyrio:
@lowener @cjnolet The new argmax header was slightly wrong, and the test was incorrect as well (see explanation below). Please have a detailed look at my latest change to see if you agree with the way I solved this.

The wrapper takes a row-major matrix view, but the implementation is in column-major semantics. As a result:

- The documentation of the wrapper was incorrect. In row-major semantics, we're finding the column with the minimum index for each row (in column-major it's the other way around).
- The dimension of the output was also incorrect. The wrapper checked that the size of the output was the same as the number of rows of the matrix, but the size of the output is N, which here was the number of columns. Note that I switched the arguments D and N passed to the implementation.
- The test passed mostly by chance, because it created a 4x3 row-major matrix, checking against the expected outputs for a 3x4 row-major matrix, but the errors in the test and wrapper canceled each other. Except that it was checking 4 rows in the `devArrMatch` despite both the actual and expected output only having 3 rows.

I've kept the column-major convention in the deprecated header, fixed the new ones for the row-major convention, and switched the implementation to drop row and columns in favor of N (number of reductions) and D (elements to reduce) and documented that.

lowener (approved):
Good catch, your changes on argmax make sense for me!

cjnolet (approved):
LGTM. Thanks for fixing the intermediate issues as well!

tfeher (approved):
Thanks Louis for addressing the issues, it looks good to me!

cjnolet:
@gpucibot merge

PR #917, opened 2022-10-14
Fix logger (vsnprintf consumes args)

Nyrio:
I noticed debug logs displayed nonsensical values so I had a look at the logger code and found it incorrect. Indeed `vsnprintf` consumes the arg list, so in order to call it twice, one must make a copy of the arg list.


cjnolet (approved):

cjnolet:
@gpucibot merge

PR #918, opened 2022-10-14
kmeans improvements: random initialization on GPU, NVTX markers, no batching when using fusedL2NN

Nyrio:
This PR:

- Fixes the random initialization bottleneck (gather indices were calculated on CPU though there was already a GPU implementation in place)
- Adds labels to the kmeans benchmark and NVTX ranges in the kmeans code, to simplify benchmarking and profiling
- Removes batching when `fusedL2NN` is used, since batching only serves the purpose of limiting memory use for the distance matrix


tfeher (approved):
Thanks @Nyrio for these updates, it looks good to me!

cjnolet:
@gpucibot merge

PR #919, opened 2022-10-14
Serialization of IVF Flat and IVF PQ

tfeher:
This PR implements serialization to file for  `ivf_pq::index` and `ivf_flat::index` structures.

Index building takes time, therefore downstream projects (like cuML) want to save the index (https://github.com/rapidsai/cuml/issues/4743). But downstream project should not depend on the implementation details of the index, therefore RAFT provides methods to serialize and deserialize the index.

This is still experimental:
- ideally we want to use a general serialization method for mdspan https://github.com/rapidsai/raft/pull/770,
- instead of directly saving to file, raft should provide a byte string and let the downstream project decide how to save it (e.g. pickle for cuML).

Python wrappers are provided for IVF-PQ to save/load the index.


cpp/CMakeLists.txt
```diff
@@ -46,7 +46,7 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 option(BUILD_SHARED_LIBS "Build raft shared libraries" ON)
 option(BUILD_TESTS "Build raft unit-tests" ON)
-option(BUILD_BENCH "Build raft C++ benchmark tests" OFF)
+option(BUILD_BENCH "Build raft C++ benchmark tests" ON)
```
cjnolet, line 49: Just a note that we probably want to turn this back off to alleviate surprises downstream

python/pylibraft/pylibraft/neighbors/ivf_pq/ivf_pq.pyx
```diff
@@ -722,3 +723,90 @@ def search(SearchParams search_params,
         raise ValueError("query dtype %s not supported" % queries_dt)
 
     return (distances, neighbors)
+
+
+@auto_sync_handle
+def save(filename, Index index, handle=None):
```
cjnolet, line 729: We should probably note very clearly in the pydocs that this feature is currently experimental and subject to change (so it's possible they might not be able to load what they saved should changes occur).

python/pylibraft/pylibraft/neighbors/ivf_pq/ivf_pq.pyx
```diff
@@ -722,3 +723,90 @@ def search(SearchParams search_params,
...
+
+    cdef string c_filename = filename.encode('utf-8') 
+
+    c_ivf_pq.save(deref(handle_), c_filename, deref(index.index))
+
+
+@auto_sync_handle
+def load(filename, handle=None):
+    """
+    Loads index from file.
```
cjnolet, line 773: Same here wrt to adding the experimental notes to the docs.

cpp/include/raft/neighbors/ivf_pq.cuh
```diff
@@ -196,6 +200,17 @@ inline void search(const handle_t& handle,
     handle, params, index, queries, n_queries, k, neighbors, distances, mr);
 }
 
+template <typename IdxT>
+void save(const handle_t& handle, const std::string& filename, const index<IdxT>& index_)
```
cjnolet, line 204: We should add doxygen here and make a strong note that these are experimental (here and ivf_flat) and so they are subject to change.

cpp/include/raft_runtime/neighbors/ivf_pq.hpp
```diff
@@ -74,4 +74,12 @@ RAFT_INST_BUILD_EXTEND(uint8_t, uint64_t)
 
 #undef RAFT_INST_BUILD_EXTEND
 
+void save(const handle_t& handle,
```
cjnolet, line 77: I know it's redundant but we should document these as well since the `raft::runtime` namespaces are likely to be in their own sections of the API docs.

cjnolet (changes requested):
I know this PR is still in draft and you intend to do more work on it to prepare it for merging but I wanted to provide a first-pass review just to help nudge it along.

tfeher (commented):
Thanks @cjnolet for the review! I have addressed the issues.

cjnolet (approved):
Changes LGTM, thanks @tfeher!

cjnolet:
@gpucibot merge

cjnolet:
rerun tests

PR #920, opened 2022-10-15
Moving kernel gramm primitives to `raft::distance::kernels`

cjnolet:


cpp/bench/distance/kernels.cu
```diff
@@ -0,0 +1,130 @@
...
+  KernelParams kernel_params;
+  bool is_row_major;
+};  // struct GramTestParams
+
+template <typename T>
+struct GramMatrix : public fixture {
+  GramMatrix(const GramTestParams& p)
+    : params(p), handle(stream), A(0, stream), B(0, stream), C(0, stream)
+  {
+    //    std::vector<std::string> kernel_names{"linear", "poly", "rbf", "tanh"};
```
divyegala, line 46: Is this supposed to be there?

cpp/include/raft/distance/detail/kernels/kernel_matrices.cuh
```diff
@@ -0,0 +1,376 @@
...
+   * K_ij = (gain*<x1_i, x2_k> + offset)^exponent
+   *
+   * @tparam math_t floating point type
+   * @tparam exp_t type of exponent
+   * @param exponent
+   * @param gain
+   * @param offset
+   * @param cublas_handle
+   */
+  PolynomialKernel(exp_t exponent, math_t gain, math_t offset, cublasHandle_t cublas_handle)
```
divyegala, line 143: Is there some sort of syntax error starting here? The coloring seems off from this point for the rest of the file

cjnolet: I don't think so. We explicitly instantiate this guy and it doesn't throw a compile error.

cjnolet: Ah, not the rbf kerbel instantiation is throwing an error because it's invoking kernel functions that don't actually exist! (They don't exist in cuml and haven't existed at least since we started doing the calver versioning). I think it could be that the factory is decoupling the compilation but I'm otherwise not sure how it's not throwing a compile error in cuml.

divyegala (approved):

cjnolet:
@gpucibot merge

Issue #374, opened 2021-11-15
[FEA] Move solvers from cuml to raft


cjnolet:
There are several solvers in cuML which can benefit algorithm development across the larger RAPIDS ecosystem by living in RAFT. For example, there is interest in exposing some of the solvers from RAFT to the `cupy.optimize` package. 

Current solvers in cuml which we have discussed moving over to RAFT:
1. L-BFGS (also to extend w/ trust region constraints for L-BFGS-B)
2. OWL-QN
3. Coordinate descent 
4. SGD

There have also been discussions of wanting the CG, Powell, and potentially Nelder-Meade simplex methods, though these would be targeted as future features after the above solvers have been moved over.

cjnolet:
Linking #669 #670 #671 #672

PR #923, opened 2022-10-17
Moving low-level iterative solver primitives to `raft::solver` 

cjnolet:
We've planned to move these for quite awhile and the codebase has finally been organized in a way that can accep them. The goal here is not to move over entire end-to-end GLM algorithms, but rather just move the underlying solvers (e.g. lgbfs/owl, sgd, cd, smo, lars, etc...) and then continue to clean up and refactor, consolidate and unify their APIs to accept more generic objectives.


cjnolet:
Closes https://github.com/rapidsai/raft/issues/374

PR #938, opened 2022-10-21
Calculate max cluster size correctly for IVF-PQ

tfeher:
This PR fixes the calculation of max cluster size. 

Previous calculation could return much larger than the actual values, which could lead to OOM while allocating temporary buffers while building index for large datasets.

### DIFF
--- a/cpp/include/raft/spatial/knn/detail/ivf_pq_build.cuh
+++ b/cpp/include/raft/spatial/knn/detail/ivf_pq_build.cuh
@@ -41,6 +41,7 @@
 #include <rmm/mr/device/pool_memory_resource.hpp>
 
 #include <thrust/binary_search.h>
+#include <thrust/extrema.h>
 #include <thrust/functional.h>
 #include <thrust/scan.h>
 #include <thrust/sequence.h>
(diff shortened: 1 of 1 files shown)


cpp/include/raft/spatial/knn/detail/ivf_pq_build.cuh
```diff
@@ -430,26 +431,21 @@ auto calculate_offsets_and_indices(IdxT n_rows,
...
-  auto max_cluster_size_dev = max_cluster_size_dev_buf.data();
-  update_device(max_cluster_size_dev, &max_cluster_size, 1, stream);
+  auto exec_policy = rmm::exec_policy(stream);
   // Calculate the offsets
   IdxT cumsum = 0;
   update_device(cluster_offsets, &cumsum, 1, stream);
   thrust::inclusive_scan(exec_policy,
                          cluster_sizes,
                          cluster_sizes + n_lists,
                          cluster_offsets + 1,
```
achirkin, line 441:
```suggestion
```

cpp/include/raft/spatial/knn/detail/ivf_pq_build.cuh
```diff
@@ -430,26 +431,21 @@ auto calculate_offsets_and_indices(IdxT n_rows,
...
-                         [max_cluster_size_dev] __device__(IdxT s, uint32_t l) {
-                           atomicMax(max_cluster_size_dev, l);
-                           return s + l;
-                         });
+                         [] __device__(IdxT s, uint32_t l) { return s + l; });
```
achirkin, line 442:
```suggestion
  thrust::inclusive_scan(
    exec_policy, cluster_sizes, cluster_sizes + n_lists, cluster_offsets + 1, thrust::plus<IdxT>{});
```
I guess, now we can use available functors for this (and force both arguments to be IdxT).

achirkin (approved):
Thanks for getting to the root of this, Tamas! Looks good.

cpp/include/raft/spatial/knn/detail/ivf_pq_build.cuh
tfeher: Makes sense, updated

cjnolet (approved):

cjnolet:
@gpucibot merge

Issue #940, opened 2022-10-21
[BUG] Build failures in ivf_pq_search.cuh


benfred:
**Describe the bug**

Building on my development box fails in `cpp/include/raft/spatial/knn/detail/ivf_pq_search.cuh` ..

The build failures seem to be related to C++17 `if constexpr` code. This causes some false positive  warnings about declared variables not being referenced, which cause build failures because of the `-Werror` flag.

I suspect this is because I'm using an older version of NVCC/ G++ than whats expected.

The specific errors are:
```
/home/ben/code/raft/cpp/include/raft/spatial/knn/detail/ivf_pq_search.cuh(659): error: variable "out_indices" was declared but never referenced
          detected during:
            instantiation of "void raft::spatial::knn::ivf_pq::detail::ivfpq_compute_similarity_kernel<OpT,IdxT,OutT,LutT,Capacity,PrecompBaseDiff,EnableSMemLut>(uint32_t, uint32_t, uint32_t, uint32_t, uint32_t, uint32_t, raft::distance::DistanceType, raft::neighbors::ivf_pq::codebook_gen, uint32_t, const float *, const float *, const uint8_t *, const IdxT *, const uint32_t *, const uint32_t *, const float *, const uint32_t *, LutT *, OutT *, IdxT *) [with OpT=uint8_t, IdxT=uint32_t, OutT=half, LutT=float, Capacity=0, PrecompBaseDiff=true, EnableSMemLut=false]" 
```

```
/home/ben/code/raft/cpp/include/raft/spatial/knn/detail/ivf_pq_search.cuh(635): error: variable "base_diff" was declared but never referenced
          detected during:
            instantiation of "void raft::spatial::knn::ivf_pq::detail::ivfpq_compute_similarity_kernel<OpT,IdxT,OutT,LutT,Capacity,PrecompBaseDiff,EnableSMemLut>(uint32_t, uint32_t, uint32_t, uint32_t, uint32_t, uint32_t, raft::distance::DistanceType, raft::neighbors::ivf_pq::codebook_gen, uint32_t, const float *, const float *, const uint8_t *, const IdxT *, const uint32_t *, const uint32_t *, const float *, const uint32_t *, LutT *, OutT *, IdxT *) [with OpT=uint64_t, IdxT=uint64_t, OutT=half, LutT=raft::spatial::knn::ivf_pq::detail::<unnamed>::fp8u_t, Capacity=0, PrecompBaseDiff=false, EnableSMemLut=true]" 
[...]

PR #941, opened 2022-10-21
Build fix for ivf_pq_search.cuh

benfred:
Closes https://github.com/rapidsai/raft/issues/940

### DIFF
--- a/cpp/include/raft/spatial/knn/detail/ivf_pq_search.cuh
+++ b/cpp/include/raft/spatial/knn/detail/ivf_pq_search.cuh
@@ -42,6 +42,19 @@
 
 #include <cuda_fp16.h>
 
+// Disable warnings like 'error: variable "base_diff" was declared but never referenced'
+// and 'error: missing return statement at end of non-void function'
+#define RAFT_PRAGMA(X) _Pragma(#X)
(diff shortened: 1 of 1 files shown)


cpp/include/raft/spatial/knn/detail/ivf_pq_search.cuh
```diff
@@ -42,6 +42,19 @@
 
 #include <cuda_fp16.h>
 
+// Disable warnings like 'error: variable "base_diff" was declared but never referenced'
+// and 'error: missing return statement at end of non-void function'
+#define RAFT_PRAGMA(X) _Pragma(#X)
```
cjnolet, line 47: I generally avoid supressing things like this because it undermines the warnings as errors flags that we've intentionally set on the built to fix these types of things. 

Tagging @robertmaynard for any potential ideas here. Perhaps if we can't figure out another way around this, and it really is happening because of the gcc version, we should consider supressing conditionally depending on the GCC version.

benfred:
The build failures seem to be related to my conda environment - which had a 11.6 CTK, even though I am using cuda 11.4. Upgrading to a clean conda environment fixed for me - making this PR unnecessary.

Issue #873, opened 2022-10-03
[BUG] `linalg::axpy` mdspan functions need gtests


cjnolet:
There were several syntax errors in this file which were not caught by the RAFT CI because we don't have proper tests for these. These were only found by a cuml build, which is not good.

PR #942, opened 2022-10-21
Add unittest for `linalg::axpy`

benfred:
Closes https://github.com/rapidsai/raft/issues/873
Closes https://github.com/rapidsai/raft/issues/944

### DIFF
--- a/cpp/include/raft/linalg/axpy.cuh
+++ b/cpp/include/raft/linalg/axpy.cuh
@@ -17,6 +17,7 @@
 #define __AXPY_H
 
 #pragma once
+
 #include "detail/axpy.cuh"
 
 #include <raft/core/device_mdspan.hpp>
@@ -67,8 +68,6 @@ void axpy(const raft::handle_t& handle,
  * @param [in] alpha raft::device_scalar_view
  * @param [in] x Input vector
  * @param [inout] y Output vector
- * @param [in] incx stride between consecutive elements of x
- * @param [in] incy stride between consecutive elements of y
  */
 template <typename InType,
           typename OutType,
@@ -78,9 +77,7 @@ template <typename InType,
 void axpy(const raft::handle_t& handle,
           raft::device_scalar_view<const typename InType::value_type, ScalarIdxType> alpha,
           InType x,
-          OutType y,
-          const int incx,
-          const int incy)
+          OutType y)
 {
   RAFT_EXPECTS(y.size() == x.size(), "Size mismatch between Output and Input");
 
@@ -88,9 +85,9 @@ void axpy(const raft::handle_t& handle,
                                           y.size(),
                                           alpha.data_handle(),
                                           x.data_handle(),
-                                          incx,
+                                          x.stride(0),
                                           y.data_handle(),
-                                          incy,
+                                          y.stride(0),
                                           handle.get_stream());
 }
 
@@ -104,8 +101,6 @@ void axpy(const raft::handle_t& handle,
  * @param [in] alpha raft::device_scalar_view
  * @param [in] x Input vector
  * @param [inout] y Output vector
- * @param [in] incx stride between consecutive elements of x
- * @param [in] incy stride between consecutive elements of y
  */
 template <typename InType,
           typename OutType,
@@ -115,9 +110,7 @@ template <typename InType,
 void axpy(const raft::handle_t& handle,
           raft::host_scalar_view<const typename InType::value_type, ScalarIdxType> alpha,
           InType x,
-          OutType y,
-          const int incx,
-          const int incy)
+          OutType y)
 {
   RAFT_EXPECTS(y.size() == x.size(), "Size mismatch between Output and Input");
 
@@ -125,9 +118,9 @@ void axpy(const raft::handle_t& handle,
                                            y.size(),
                                            alpha.data_handle(),
                                            x.data_handle(),
-                                           incx,
+                                           x.stride(0),
                                            y.data_handle(),
-                                           incy,
+                                           y.stride(0),
                                            handle.get_stream());
 }
 
--- a/cpp/include/raft/linalg/axpy.cuh
+++ b/cpp/include/raft/linalg/axpy.cuh
@@ -17,8 +17,10 @@
 #define __AXPY_H
 
 #pragma once
-
 #include "detail/axpy.cuh"
+
+#include <raft/core/device_mdspan.hpp>
+#include <raft/core/host_mdspan.hpp>
 
 namespace raft::linalg {
 
--- a/cpp/test/linalg/axpy.cu
+++ b/cpp/test/linalg/axpy.cu
@@ -38,17 +38,17 @@ auto make_strided_device_vector_view(ElementType* ptr, IndexType n, IndexType st
 {
   vector_extent<IndexType> exts{n};
   std::array<IndexType, 1> strides{stride};
-  auto layout = layout_stride::mapping<vector_extent<IndexType>>{exts, strides};
+  auto layout = typename LayoutPolicy::mapping<vector_extent<IndexType>>{exts, strides};
   return device_vector_view<ElementType, IndexType, LayoutPolicy>{ptr, layout};
 }
 
-template <typename InType, typename IdxType = int, typename OutType = InType>
+template <typename InType, typename IndexType = int, typename OutType = InType>
 struct AxpyInputs {
   OutType tolerance;
-  IdxType len;
+  IndexType len;
   InType alpha;
-  int incx;
-  int incy;
+  IndexType incx;
+  IndexType incy;
   unsigned long long int seed;
 };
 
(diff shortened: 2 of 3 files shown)


cpp/test/linalg/axpy.cu
```diff
@@ -0,0 +1,103 @@
+/*
+ * Copyright (c) 2019-2022, NVIDIA CORPORATION.
```
cjnolet, line 2: New files should only have the current year

cpp/test/linalg/axpy.cu
```diff
@@ -0,0 +1,103 @@
...
+    uniform(handle, r, x.data(), params.len, T(-1.0), T(1.0));
+    uniform(handle, r, y.data(), params.len, T(-1.0), T(1.0));
+    refy = rmm::device_uvector<T>(y, stream);
+
+    refy        = rmm::device_uvector<T>(y, stream);
+    int threads = 64;
+    int blocks  = raft::ceildiv<int>(params.len, threads);
+    naiveAxpy<T><<<blocks, threads, 0, stream>>>(params.len, params.scalar, x.data(), refy.data());
+
+    axpy(handle, params.len, &params.scalar, x.data(), 1.0, y.data(), 1.0, stream);
```
cjnolet, line 73: We need to test the new `mdspan` APIs (which will test the raw pointer functions as well since they just wrap them. Eventually the raw pointer APIs are going to go away and we'll just be using mdspan across the public APIs (and start using it in the detail APIs as well).

cpp/test/linalg/axpy.cu
```diff
@@ -0,0 +1,103 @@
...
+class AxpyTest : public ::testing::TestWithParam<UnaryOpInputs<T>> {
+ protected:
+  UnaryOpInputs<T> params;
+  rmm::device_uvector<T> refy;
+  rmm::device_uvector<T> y;
+
+ public:
+  AxpyTest()
+    : testing::TestWithParam<UnaryOpInputs<T>>(),
+      refy(0, rmm::cuda_stream_default),
```
cjnolet, line 45: The stream from the raft handle should be used everywhere.

cpp/test/linalg/axpy.cu
```diff
@@ -0,0 +1,103 @@
...
+  {
+    rmm::cuda_stream_default.synchronize();
+  }
+
+ protected:
+  void SetUp() override
+  {
+    params = ::testing::TestWithParam<UnaryOpInputs<T>>::GetParam();
+
+    raft::handle_t handle;
```
cjnolet, line 56: It might be easier to make this an instance variable so that it's available when `refy` and `y` are constructed.

cpp/test/linalg/axpy.cu
```diff
@@ -0,0 +1,103 @@
...
+
+    axpy(handle, params.len, &params.scalar, x.data(), 1.0, y.data(), 1.0, stream);
+
+    handle.sync_stream();
+  }
+
+  void TearDown() override {}
+};
+
+const std::vector<UnaryOpInputs<float>> inputsf  = {{0.000001f, 1024 * 1024, 2.f, 1234ULL}};
```
cjnolet, line 81: Can we add some more shapes w/ different powers of 2 and differen alphas to increase coverage?

cpp/test/linalg/axpy.cu
```diff
@@ -0,0 +1,103 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include <raft/linalg/axpy.cuh>
+
+#include "../test_utils.h"
+#include "unary_op.cuh"
```
cjnolet, line 19: Is this needed? I don't see `unary_op` used anywhere

cpp/test/linalg/axpy.cu
```diff
@@ -0,0 +1,103 @@
...
+  rmm::device_uvector<T> refy;
+  rmm::device_uvector<T> y;
+
+ public:
+  AxpyTest()
+    : testing::TestWithParam<UnaryOpInputs<T>>(),
+      refy(0, rmm::cuda_stream_default),
+      y(0, rmm::cuda_stream_default)
+  {
+    rmm::cuda_stream_default.synchronize();
```
cjnolet, line 48: `handle.sync_stream();`

cpp/test/linalg/axpy.cu
```diff
@@ -0,0 +1,103 @@
...
+    cudaStream_t stream = handle.get_stream();
+
+    raft::random::RngState r(params.seed);
+
+    rmm::device_uvector<T> x(params.len, stream);
+    y.resize(params.len, stream);
+
+    uniform(handle, r, x.data(), params.len, T(-1.0), T(1.0));
+    uniform(handle, r, y.data(), params.len, T(-1.0), T(1.0));
+    refy = rmm::device_uvector<T>(y, stream);
```
cjnolet, line 66: Rather than re-assigning, we should just resize. It looks like this might be duplicated on the line below as well.

cpp/test/linalg/axpy.cu
cjnolet: We should also test this with different strides.

cjnolet (changes requested):
Thanks for adding this test, @benfred! Not sure if you are finished but I figured it wouldn't hurt to do a quick first-pass review before I take off for the weekend.

cpp/test/linalg/axpy.cu
benfred: I've updated to use the new mdspan api's - but hit some problems trying to get this going with different strides =( https://github.com/rapidsai/raft/issues/944

cpp/test/linalg/axpy.cu
benfred: I was using the UnaryOpInputs for the test parameters - but have changed out to AxpyInputs to include a incx/incy stride, and removed this include

cpp/test/linalg/axpy.cu
benfred: It was definitely duplicated (good catch!) - but I do need to re-assign here, since `y` is both an input and output parameter and I need the same random input state in y/refy for both the `axpy` and `naiveAxpy` calls.

cpp/test/linalg/axpy.cu
```diff
@@ -0,0 +1,138 @@
...
+    int y_len = params.len * params.incy;
+    rmm::device_uvector<T> x(x_len, stream);
+    y.resize(y_len, stream);
+
+    uniform(handle, r, x.data(), x_len, T(-1.0), T(1.0));
+    uniform(handle, r, y.data(), y_len, T(-1.0), T(1.0));
+
+    // Take a copy of the random generated values in y for the naive reference implementation
+    // this is necessary since axpy uses y for both input and output
+    refy = rmm::device_uvector<T>(y, stream);
```
cjnolet, line 80: Instead of re-assigning this instance variable after construction, can we do `refy.resize()` and then use  `raft::copy(refy.data(), y.data())` it makes the intention a little more explicit. This is a common pattern that we use in these tests.

cjnolet:
Do you want to go ahead and make the changes for https://github.com/rapidsai/raft/issues/944 in this PR? I think the amount of changes is still reasonably small to correct the API and get the test supporting the different strides.

cpp/test/linalg/axpy.cu
benfred: sounds good - done in last commit

cpp/test/linalg/axpy.cu
```diff
@@ -0,0 +1,169 @@
...
+
+    if ((params.incx > 1) && (params.incy > 1)) {
+      axpy(handle,
+           make_host_scalar_view<const T>(&params.alpha),
+           make_strided_device_vector_view<const T>(x.data(), params.len, params.incx),
+           make_strided_device_vector_view<T>(y.data(), params.len, params.incy));
+    } else if (params.incx > 1) {
+      axpy(handle,
+           make_host_scalar_view<const T>(&params.alpha),
+           make_strided_device_vector_view<const T>(x.data(), params.len, params.incx),
```
cjnolet, line 108: Could we also test w/ column-major mdspan inputs (e.g. `raft::make_device_vector_view<const T, std::uint32_t, raft::col_major>(...)`)?

cpp/include/raft/linalg/axpy.cuh
```diff
@@ -113,24 +110,22 @@ template <typename InType,
 void axpy(const raft::handle_t& handle,
           raft::host_scalar_view<const typename InType::value_type, ScalarIdxType> alpha,
           InType x,
-          OutType y,
-          const int incx,
-          const int incy)
+          OutType y)
```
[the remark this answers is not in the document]
cjnolet: I'll add a Github issue to do this in a follow-on. I think we could scrape through and do it more generally as well.

cjnolet (approved):

cpp/include/raft/linalg/axpy.cuh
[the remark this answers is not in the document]
benfred: That sounds great - I think we should handle auto converting to host_scalar_view in a couple other spots than just the axpy code, so seems out of scope for this PR

cjnolet:
@gpucibot merge

PR #947, opened 2022-10-24
Adding optional handle to each public API function (along with example)

### LINKED ISSUE #946: [FEA] pylibraft functions should accept (optional) handle
The functions in pylibraft do not synchronize the stream when they are done processing so that multiple functions can be chained together without having to synchronize. However, the handle is still created internally to each function. We should expose the handle on the API as an option (None by default) so that a user can construct the handle and then pass it through a series of RAFT functions before calling synchronize on their own. 

We also need to document when the functions don't synchronize the streams explicitly.

cjnolet:
Closes #946


python/pylibraft/pylibraft/distance/fused_l2_nn.pyx
```diff
@@ -88,7 +91,15 @@ def fused_l2_nn_argmin(X, Y, output, sqrt=True):
                                       dtype=cp.float32)
         output = cp.empty((n_samples, 1), dtype=cp.int32)
 
-        fused_l2_nn_argmin(in1, in2, output)
+        # A single RAFT handle can optionally be reused across
```
dantegd, line 94: Quick comment: we might want to adopt a docstring generator like the one in cuML for common ones like this

dantegd (approved):

cjnolet:
@gpucibot merge

python/pylibraft/pylibraft/distance/fused_l2_nn.pyx
cjnolet: Yes I think centralizing this is a great idea. I'll start looking into the doc function in cuml

PR #962, opened 2022-10-27
Forward merging cuda-python version bump

cjnolet:


vyasr:
@cjnolet I suspect that this approach of combining multiple forward merges into one might write a slightly different history than we want since it includes changes _within_ Ashwin's original PR. I think what you want is to make a PR that just handles the current 22.10->22.12 merger (i.e. to replace #913), then let the cuda-python fix get merged into 22.10 and have the next forward merger handle that by itself.

Issue #805, opened 2022-09-06
[FEA] Expose `raft::linalg::dot` through public API


cjnolet:
Currently, the cublas `dot()` API is being hidden in `detail` but this is something that would be useful to expose through our linear algebra public API functions to facade users downstream away from the implementations and provide them a more consistent API vocabulary. 

We should wrap the cublas `dot()` functions in the `raft::linalg` namespace. 

cc @akifcorduk @katrinescheel

PR #968, opened 2022-10-31
Expose `linalg::dot` in public API

benfred:
Closes https://github.com/rapidsai/raft/issues/805


cpp/include/raft/linalg/dot.cuh
```diff
@@ -0,0 +1,70 @@
...
+          typename = raft::enable_if_input_device_mdspan<InputType2>,
+          typename = raft::enable_if_output_mdspan<OutputType>>
+void dot(const raft::handle_t& handle, InputType1 x, InputType2 y, OutputType out)
+{
+  RAFT_EXPECTS(x.size() == y.size(),
+               "Size mismatch between x and y input vectors in raft::linalg::dot");
+
+  // Right now the inputs and outputs need to all have the same value_type (float/double etc).
+  // Try to output a meaningful compiler error if mismatched types are passed here.
+  // Note: In the future we could remove this restriction using the cublasDotEx function
```
cjnolet, line 52: Should we just go ahead and wrap the cublasEx functions?

cjnolet (changes requested):
Thanks again for this PR! We are going to want to think about whether these (axpy, dot, etc...) should be accepting general mdspan or whether we should be constraining them to be vectors up front. 

It would also be nice to see the current vector factory functions made more flexible to enable strided layouts rather than adding new functions. 

These types of examples (using the existing device vector factory functions to create a strided vector) would be great to have in the quick start as well.

cpp/include/raft/linalg/dot.cuh
benfred: I created an issue so we can discuss further https://github.com/rapidsai/raft/issues/977 . 

Reading the docs a little closer, and it looks like even w/ cublasDotEx having different dtypes for the input/outputs isn't currently supported: https://docs.nvidia.com/cuda/cublas/index.html#cublas-dotEx - so it won't have much value for the dot API (though I could see a use for it myself with the gemm api w/ implicit and the mixed precision work I was talking about last week)

cpp/include/raft/linalg/dot.cuh
```diff
@@ -0,0 +1,70 @@
...
+ * @tparam OutputType  Either a host_scalar_view or device_scalar_view for the output
+ * @param[in] handle   raft::handle_t
+ * @param[in] x        First input vector
+ * @param[in] y        Second input vector
+ * @param[out] out     The output dot product between the x and y vectors
+ */
+template <typename InputType1,
+          typename InputType2,
+          typename OutputType,
+          typename = raft::enable_if_input_device_mdspan<InputType1>,
```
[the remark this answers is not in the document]
benfred: agreed - made the changes here so that both axpy and dot take device_vector_view's

cpp/include/raft/core/device_mdspan.hpp
```diff
@@ -276,4 +276,22 @@ auto make_device_vector_view(ElementType* ptr, IndexType n)
...
+ * @brief Create a 1-dim mdspan instance for device pointer, using a strided layout
+ * @tparam ElementType the data type of the vector elements
+ * @tparam IndexType the index type of the extents
+ * @tparam LayoutPolicy policy for strides and layout ordering
+ * @param[in] ptr on device to wrap
+ * @param[in] n number of elements in pointer
+ * @param[in] stride stride between elements
+ * @return raft::device_vector_view
+ */
+template <typename ElementType, typename IndexType = int, typename LayoutPolicy = layout_stride>
```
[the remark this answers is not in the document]
benfred: done

cpp/include/raft/core/device_mdspan.hpp
```diff
@@ -276,4 +276,22 @@ auto make_device_vector_view(ElementType* ptr, IndexType n)
...
+auto make_strided_device_vector_view(ElementType* ptr, IndexType n, IndexType stride)
```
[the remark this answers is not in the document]
benfred: I've updated make_device_vector_view to allow strided input here - let me know what you think.

cpp/include/raft/linalg/axpy.cuh
```diff
@@ -62,66 +62,61 @@ void axpy(const raft::handle_t& handle,
...
  * @param [in] x Input vector
  * @param [inout] y Output vector
  */
-template <typename InType,
-          typename OutType,
-          typename ScalarIdxType,
-          typename = raft::enable_if_input_device_mdspan<InType>,
-          typename = raft::enable_if_output_device_mdspan<OutType>>
+template <typename ElementType,
+          typename IndexType       = std::uint32_t,
```
cjnolet, line 71: We should consider allowing the compiler to infer these things. In general we have been removing the defaults everywhere unless absolutely necessary because they create confusion when a user actually wants to use a type other than the default (the compiler complains a valid overload doesn't exist unless the user explicitly provides the template arg)

cpp/include/raft/linalg/axpy.cuh
```diff
@@ -62,66 +62,61 @@ void axpy(const raft::handle_t& handle,
...
```
cjnolet, line 101: Same here wrt defaults. I know we have them in other places, but the goal has been to remove them as we see them (where it's possible)

cpp/include/raft/linalg/dot.cuh
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * @brief Computes the dot product of two vectors.
+ * @param[in] handle   raft::handle_t
+ * @param[in] x        First input vector
+ * @param[in] y        Second input vector
+ * @param[out] out     The output dot product between the x and y vectors.
+ * @note The out parameter can be either a host_scalar_view or device_scalar_view
+ */
+template <typename ElementType,
+          typename IndexType       = std::uint32_t,
```
cjnolet, line 37: Same for defaults here (and across the pr)

cpp/include/raft/linalg/dot.cuh
```diff
@@ -0,0 +1,59 @@
...
+          typename ScalarIndexType = std::uint32_t,
+          typename LayoutPolicy1   = layout_c_contiguous,
+          typename LayoutPolicy2   = layout_c_contiguous>
+void dot(const raft::handle_t& handle,
+         raft::device_vector_view<const ElementType, IndexType, LayoutPolicy1> x,
+         raft::device_vector_view<const ElementType, IndexType, LayoutPolicy2> y,
+         raft::device_scalar_view<ElementType, ScalarIndexType> out)
```
cjnolet, line 44: As discussed previously with @lowener, im okay adding an overload that accepts a host scalar by value. At least it provides the same user experience until the implicit conversion mechanism is ready.

cjnolet (changes requested):
Changes are looking great! Remaining things are very minor.

cpp/include/raft/core/device_mdspan.hpp
```diff
@@ -266,14 +266,23 @@ auto make_device_matrix_view(ElementType* ptr, IndexType n_rows, IndexType n_col
...
  * @param[in] n number of elements in pointer
+ * @param[in] stride the stride between consecutive elements in the vector. Setting to a value
+ *            other than 1 requires LayoutPolicy to be set to layout_stride
  * @return raft::device_vector_view
  */
 template <typename ElementType,
           typename IndexType    = std::uint32_t,
           typename LayoutPolicy = layout_c_contiguous>
-auto make_device_vector_view(ElementType* ptr, IndexType n)
+auto make_device_vector_view(ElementType* ptr, IndexType n, IndexType stride = 1)
```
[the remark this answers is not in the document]
benfred: If I'm understanding you correctly - you're thinking we can just pass the layout mapping to the `make_device_vector_view` function directly , and add a new factory function for creating this layout mapping?

I took a stab at that in the last commit - unfortunately, I couldn't get a single `make_device_vector_view` function to compile successfully with being passed both a IndexType with the number of elements and the Mapping with the strided layout (was getting compile errors in various other raft functions that I hadn't updated). However, I could get it to work with adding an overload - which is whats in the last commit.   Do you have any suggestions on how to clean this up =) ? 

I'll add something to the tutorial / docs once we're happy with the API -

cpp/include/raft/linalg/dot.cuh
benfred: I've added an overload that takes a host_scalar_view - and a host scalar by pointer too (though maybe we only should add the host_scalar_view overload to be consistent with other functions like axpy?)

cpp/include/raft/linalg/dot.cuh
```diff
@@ -55,5 +54,65 @@ void dot(const raft::handle_t& handle,
...
+ */
+template <typename ElementType,
+          typename IndexType,
+          typename ScalarIndexType,
+          typename LayoutPolicy1,
+          typename LayoutPolicy2>
+void dot(const raft::handle_t& handle,
+         raft::device_vector_view<const ElementType, IndexType, LayoutPolicy1> x,
+         raft::device_vector_view<const ElementType, IndexType, LayoutPolicy2> y,
+         ElementType* out)
```
cjnolet, line 103: I think for the host output, we probably should drop this overload. Sorry for being confusing here. I think it makes more sense to accept a host scalar by value for functions like axpy where the scalar is an input. For output on host, I think we should stick to the mdspan scalar wrappers.

cpp/include/raft/core/device_mdspan.hpp
```diff
@@ -266,23 +266,42 @@ auto make_device_matrix_view(ElementType* ptr, IndexType n_rows, IndexType n_col
...
+}
+
+/**
+ * @brief Construct a strided vector layout mapping
+ * @tparam IndexType the index type of the extents
+ * @params[in] n the number of elements in the vector
+ * @params[in] stride the stride between elements in the vector
+ */
+template <typename IndexType>
+auto make_vector_strided_layout(IndexType n, IndexType stride)
```
cjnolet, line 301: I think this can be more reusable than just 1d. Of course, I don't mind having a specialized function for vectors, but perhaps we could add a more generalized function  like `make_strided_layout(extents, strides)` and then have `make_vector_strided_layout()` invoke `make_strided_layout`? Just thinking about reuse here. We can have strided layouts that aren't necessarily c- or f-contiguous.

cpp/include/raft/linalg/dot.cuh
benfred: removed in latest commit

cpp/test/linalg/axpy.cu
```diff
@@ -100,18 +88,22 @@ class AxpyTest : public ::testing::TestWithParam<AxpyInputs<T>> {
     if ((params.incx > 1) && (params.incy > 1)) {
       axpy(handle,
            make_host_scalar_view<const T>(&params.alpha),
-           make_strided_device_vector_view<const T>(x.data(), params.len, params.incx),
-           make_strided_device_vector_view<T>(y.data(), params.len, params.incy));
+           make_device_vector_view<const T, IndexType, layout_stride>(
```
cjnolet, line 91: Are we testing this with a device scalar view input as well?

cpp/test/linalg/dot.cu
```diff
@@ -0,0 +1,155 @@
...
+
+    rmm::device_uvector<T> x(x_len, stream);
+    rmm::device_uvector<T> y(y_len, stream);
+    uniform(handle, r, x.data(), x_len, T(-1.0), T(1.0));
+    uniform(handle, r, y.data(), y_len, T(-1.0), T(1.0));
+
+    naiveDot<<<256, 256, 0, stream>>>(
+      params.len, x.data(), params.incx, y.data(), params.incy, refoutput.data());
+
+    auto out_view = make_device_scalar_view<T, IndexType>(output.data());
```
cjnolet, line 83: Can we test this with both device and host scalar views since we support it?

cpp/include/raft/core/device_mdspan.hpp
```diff
@@ -268,12 +268,55 @@ auto make_device_matrix_view(ElementType* ptr, IndexType n_rows, IndexType n_col
...
+ */
+template <typename ElementType, typename IndexType, typename LayoutPolicy = layout_c_contiguous>
+auto make_device_vector_view(
+  ElementType* ptr,
+  const typename LayoutPolicy::template mapping<vector_extent<IndexType>>& mapping)
+{
+  return device_vector_view<ElementType, IndexType, LayoutPolicy>{ptr, mapping};
+}
+
+template <typename ExtentType, typename StrideType>
```
cjnolet, line 294: This needs to be documented

cpp/test/linalg/axpy.cu
benfred: we weren't - but I've added tests for both the device/host scalar inputs in the last commit

cpp/include/raft/core/device_mdspan.hpp
benfred: done

cjnolet (approved):
Looks great, thanks again @benfred!

cjnolet:
@gpucibot merge

PR #969, opened 2022-11-01
Remove unnecessary requirements for raft-dask.

vyasr:
numpy is a requirement for pylibraft but not raft-dask (and numba isn't required for either).


cjnolet:
@gpucibot merge

PR #970, opened 2022-11-01
IVF-PQ Python wrappers

tfeher:
This PR adds python wrappers to IVF-PQ.


tfeher:
There are a few items that still need to be addressed, but otherwise it is ready for the first review @cjnolet 
- [ ] handle `codebook_kind` parameter
- [ ] handle `lut_dtype` and `internal_distance_dtype` search params
- [ ] Add more test parameters combinations
- [ ] Add xfail tests

achirkin:
What would you think about re-exporting default values of C++ params structs? Like we did a while ago in cuml/linearSVM (https://github.com/rapidsai/cuml/issues/4381).
This would require a bit of extra work to make it nice, e.g. partially port the metaclass definition to allow inserting default values into python docstrings. But I think it still makes sense in the long run. I suspect we may want to change the default parameter values in future (at least those that affect performance, but don't affect the data layout/search results).

tfeher:
@achirkin Thanks for the suggestion! yes it occurred to me to return to this idea. I would dedicate a follow up PR for this task.

tfeher:
There is one issue pending: forwarding the C++ exceptions from `build` and `search` to Python. Using the standard method leads to the following error message:
```
pylibraft/neighbors/ivf_pq.cxx:7155:45: error: no matching function for call to 'raft::neighbors::ivf_pq::index<long unsigned int>::index()'
 7155 |   raft::neighbors::ivf_pq::index<uint64_t>  __pyx_t_20;
```
This happens  because Cython tries define a variable for the return value, and the return value type (`index`) does not have a default constructor. I will look into a workaround.

python/pylibraft/pylibraft/neighbors/ivf_pq.pyx
```diff
@@ -0,0 +1,543 @@
...
+        if exp_cols is not None and cai["shape"][1] != exp_cols:
+            raise ValueError("Incorrect number of columns, expected {} got {}" \
+                                 .format(exp_cols, cai["shape"][1]))
+
+        if exp_rows is not None and cai["shape"][0] != exp_rows:
+            raise ValueError("Incorrect number of rows, expected {} , got {}" \
+                                .format(exp_rows, cai["shape"][0]))
+        
+
+class IvfPq:
```
cjnolet, line 191: Overall I'm really excited about this new API!

pylibraft is supposed to be very lightweight wrappers around the c++ layer, though, and as a result? I'd prefer if instead of following cuml/scikit-learn's "estimator" design, we design this so it directly corresponds to RAFT's C++ API.

This would make the ivfpq index itself a class with flat python functions in the pylibraft.neighbors.ivfpq package to build(), extend(), and search() the index. 

I think this also makes a more clear separation between cuml's python API and pylibraft, since cuml can continue to follow the "estimator" design style and not directly mirror pylibraft's API.

cjnolet (changes requested):
Thanks for adding this feature to pylibraft! I think the feature mostly looks ready to go, but I'd like to keep pylibraft's APIs consistent and make sure they all directly correspond to the c++ APIs they are wrapping.

cjnolet:
Closes https://github.com/rapidsai/raft/issues/998

python/pylibraft/pylibraft/neighbors/ivf_pq.pyx
tfeher: That is a good point. Wrapping the index in its own class might help resolve the issue I am having with the exception handling. I will update the PR accordingly.

tfeher: Refactored, please have a look.

cjnolet: Looks great! I'll give it a more thorough review tomorrow

python/pylibraft/pylibraft/neighbors/ivf_pq.pyx
```diff
@@ -0,0 +1,547 @@
...
+    cdef c_ivf_pq.index_params params
+
+    def __init__(self, *, 
+                 n_lists=1024, 
+                 metric="l2_expanded",
+                 kmeans_n_iters=20, 
+                 kmeans_trainset_fraction=0.5,
+                 pq_bits=8,
+                 pq_dim=0,
+                 codebook_kind="per_subspace",
```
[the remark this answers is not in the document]
tfeher: You are right, "subspace" or "cluster" is simpler. I had the general question, whether to use strings, or the enums which we are also exporting eg `ivf_pq.PER_SUBSPACE`. The same question applies for the metric parameter. I will go with the strings for now, since that feels more user friendly. Let me know if you would prefer to have the enums.

python/pylibraft/pylibraft/neighbors/ivf_pq.pyx
```diff
@@ -0,0 +1,547 @@
...
+from pylibraft.common import Handle
+from pylibraft.common.handle cimport handle_t
+from rmm._lib.memory_resource cimport device_memory_resource
+
+cimport pylibraft.neighbors.c_ivf_pq as c_ivf_pq
+
+from pylibraft.neighbors.c_ivf_pq cimport index_params
+from pylibraft.neighbors.c_ivf_pq cimport search_params
+
+def is_c_cont(cai):
```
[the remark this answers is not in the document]
tfeher: I have moved it into testing/utils. I was not sure whether you want to keep that file for pytest utilities, or general utilities. Let me know if you prefer to place the helper into a different file.

python/pylibraft/pylibraft/neighbors/ivf_pq.pyx
```diff
@@ -0,0 +1,547 @@
...
+
+
+def extend(Index index, new_vectors, new_indices, handle=None):
+    """
+    Extend an existing index with new vectors.
+        
+        
+    Parameters
+    ----------
+    index : ivf_pq.Index
```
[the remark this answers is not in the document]
tfeher: Yes, I have added the example.

tfeher (commented):
Thanks @cjnolet for the review! I have addressed the issues. There is a problem with the extend test, I am looking into it.

tfeher:
Fixed the problem with `extend`.

python/pylibraft/pylibraft/neighbors/ivf_pq.pyx
```diff
@@ -0,0 +1,547 @@
...
+            The number of course clusters to select for the fine search.
+        lut_dtype: default = ivf_pq.CUDA_R_32F (float)
+            Data type of look up table to be created dynamically at search time. The use of 
+            low-precision types reduces the amount of shared memory required at search time, so
+            fast shared memory kernels can be used even for datasets with large dimansionality.
+            Note that the recall is slightly degraded when low-precision type is selected.
+            Possible values [CUDA_R_32F, CUDA_R_16F, CUDA_R_8U]
+        internal_distance_dtype: default = ivf_q.CUDA_R_32F (float)
+            Storage data type for distance/similarity computation.
+            Possible values [CUDA_R_32F, CUDA_R_16F]
```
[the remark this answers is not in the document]
cjnolet: Are you referring to switching from dtypes like np.float32 to "float"? In general I think the two are used pretty interchangeably.

python/pylibraft/pylibraft/neighbors/ivf_pq.pyx
```diff
@@ -0,0 +1,652 @@
...
+        dataset = cp.random.random_sample((n_samples, n_features), dtype=cp.float32)
+
+        # Build index
+        handle = Handle()
+        index = ivf_pq.build(ivf_pq.IndexParams(), dataset, handle=handle)
+
+        # Search using the built index
+        queries = cp.random.random_sample((n_queries, n_features), dtype=cp.float32)
+        k = 10
+        neighbors = cp.empty((n_queries, k), dtype=cp.uint64)
```
cjnolet, line 567: I'm hoping that after https://github.com/rapidsai/raft/pull/1008 is merged, we can make the explicit output arguments optional (eg setting them to None by default) so that this function will return a tuple or device_ndarray in the default case (indices, distances).

cjnolet (commented):
Thanks again for writing this! I'm really excited to get it in for 22.12. It's al

python/pylibraft/pylibraft/neighbors/ivf_pq.pyx
cjnolet: We can do that change later, though, and don't need to hold up this PR for it.

python/pylibraft/pylibraft/neighbors/ivf_pq.pyx
```diff
@@ -0,0 +1,652 @@
...
+
+    return index 
+
+
+cdef class SearchParams:
+    cdef c_ivf_pq.search_params params
+
+    def __init__(self, *, n_probes=20, 
+                 lut_dtype=CUDA_R_32F, 
+                 internal_distance_dtype=CUDA_R_32F):
```
cjnolet, line 477: Can we use the numpy dtype format here? Aside from the obvious implications on the user experience, I'd also like to hide cuda-specific things from the public APIs as much as possible in general. Even on the c/c++ side, we might consider eventually creating our own enum that users interact with, even if the cuda code underneath uses these types directly.

python/pylibraft/pylibraft/neighbors/ivf_pq.pyx
```diff
@@ -0,0 +1,652 @@
...
+        """
+        IVF-PQ search parameters
+
+        Parameters
+        ----------
+        n_probes: int, default = 1024
+            The number of course clusters to select for the fine search.
+        lut_dtype: default = ivf_pq.CUDA_R_32F (float)
```
cjnolet, line 485: We should accept numpy dtypes here and convert them into the corresponding cuda data types (applies to all parts that a user sets explicitly)

python/pylibraft/pylibraft/neighbors/ivf_pq.pyx
```diff
@@ -0,0 +1,652 @@
...
+        internal_distance_dtype: default = ivf_q.CUDA_R_32F (float)
+            Storage data type for distance/similarity computation.
+            Possible values [CUDA_R_32F, CUDA_R_16F]
+        
+        """
+
+        self.params.n_probes = n_probes
+        self.params.lut_dtype = lut_dtype
+        self.params.internal_distance_dtype = internal_distance_dtype
+        # self.params.shmem_carveout = self.shmem_carveout # TODO(tfeher): enable if #926 adds this
```
cjnolet, line 500: We are going to want to consider eventually hiding this somehow behind more environment-specific set of parameters. This is leaking cuda-specific stuff into the public APIs (not something I think we need to change in this PR, but something we should be thinking about with Grace coming, and potentially having to support CPU code paths in the future)

cjnolet (changes requested):
Looking great! I definitely think this is going to make it into 22.12, just really one little thing- hiding the cuda data types behind np.dtype for API consistency

python/pylibraft/pylibraft/neighbors/ivf_pq.pyx
```diff
@@ -0,0 +1,652 @@
...
+    
+    if handle is None:
+        handle = Handle()
+    cdef handle_t* handle_ = <handle_t*><size_t>handle.getHandle()
+        
+    idx = Index()
+        
+    if dataset_dt == np.float32:
+        c_ivf_pq.build(deref(handle_),      
+                       index_params.params,
```
achirkin, line 341: Do you think it would make sense to wrap this into interruptible (https://github.com/rapidsai/raft/blob/branch-22.12/python/pylibraft/pylibraft/common/interruptible.pyx#L36-L54) to allow ctrl+c the build/search? Also it may be useful to set release the GIL for all member of the index as in https://github.com/rapidsai/cuml/blob/branch-22.12/python/cuml/svm/linear.pyx#L47

cjnolet: @achirkin the `handle.sync()` in the Python API invokes `handle.sync_stream()` in the C++ layer. Isn't that enough to wrap this in the interruptible call? Or do we need to add `nogil` to all of the cython externs in order for it to work?

achirkin: I think, it's not enough, but also we'd need to do the same `interruptible` wrapper for the `handle.sync_stream()` to be able to reliably cancel the task.

Here's my intuition:
-  The cpu thread gets blocked by any synchronization call. To make it interruptible, we need this point to `interruptible::synchronize` (which is the case for `handle.sync_stream()`).
- To make a program to respond to ctrl+c, we need to install an interrupt handler; this is what [`with cuda_interruptible():`](https://github.com/rapidsai/raft/blob/branch-22.12/python/pylibraft/pylibraft/common/interruptible.pyx#L36-L54) snipped does. At the moment, raft's python wrapper on the `handle.sync_stream()` does not have it (but it should!).
- If the C++-side function does not contain any synchronization points, there's nothing to interrupt. If it was the case for ivf-pq functions, calling the `handle.sync_stream()` (patched with the interrupt handler) would indeed be enough for the interruptible behavior, because the program would exit the ifv-pq functions and wait on the latter. At the moment ivf-pq has one synchronization point in `search` (very fast, shouldn't be a problem) and plenty of them in `build` (this is what we really want to be able to cancel on ctrl+c).
- Honestly, I'm not entirely sure if releasing the GIL for any of these C++ calls is necessary. Yet, even if it's not the case, I think there may be some benefits in setting everything we can to `nogil` (perhaps, for running concurrent computations?..)

tfeher (commented):
Thanks for the reviews, I have updated the PR!

cjnolet (approved):
LGTM. Thanks again @tfeher!

cjnolet:
@gpucibot merge

PR #973, opened 2022-11-01
Update raft tests to compile with C++17 features enabled

robertmaynard:
The raft tests require C++17 features but doesn't request this compiler mode, so compilation will fail with compilers that default to 11 or 14.


robertmaynard:
Encountered locally when building with gcc-9 which defaults to c++14

cjnolet (approved):
Thanks for finding and fixing this!

robertmaynard:
@gpucibot merge

PR #976, opened 2022-11-01
Use new rapids-cmake functionality for rpath handling.

vyasr:
The new functionality creates a single source of truth defining the location at which a library will be, allowing rapids-cmake to handle the necessary relative paths in the RPATH and avoiding the current error-prone approach that requires each CMakeLists.txt to correctly assign the relative path.

### DIFF
--- a/python/pylibraft/pylibraft/distance/CMakeLists.txt
+++ b/python/pylibraft/pylibraft/distance/CMakeLists.txt
@@ -22,8 +22,5 @@ rapids_cython_create_modules(
     CXX
     SOURCE_FILES "${cython_sources}"
     LINKED_LIBRARIES "${linked_libraries}"
+    ASSOCIATED_TARGETS cuml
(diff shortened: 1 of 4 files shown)


python/pylibraft/pylibraft/distance/CMakeLists.txt
```diff
@@ -22,8 +22,5 @@ rapids_cython_create_modules(
     CXX
     SOURCE_FILES "${cython_sources}"
     LINKED_LIBRARIES "${linked_libraries}"
+    ASSOCIATED_TARGETS cuml
```
cjnolet, line 25: Is this a copy/paste error?

vyasr: And here I was thinking I was careful... thanks :)

cjnolet (approved):

vyasr:
@gpucibot merge

PR #980, opened 2022-11-02
Fix pylibraft docstring example code

benfred:
Fix the pylibraft example code given in the docstrings so that they run without exceptions.


cjnolet (approved):

cjnolet:
@gpucibot merge

PR #986, opened 2022-11-03
Provide constants and infrastucture for CUDA-free usage

wphicks:
For several downstream applications (notably the FIL backend for Triton), it is useful to be able to make use of components of RAFT in a CUDA-free environment. This PR sets up infrastructure to help with conditional CUDA-free implementation of RAFT components, including:

- A `DISABLE_CUDA` CMake option for disabling CUDA components in eligible headers
- A constexpr boolean `CUDA_ENABLED` for determining if CUDA is enabled
- A custom exception `cuda_unsupported` for attempting to use a CUDA-only feature when CUDA has been disabled


robertmaynard (changes requested):
This option shouldn't be encoded in `cpp/CMakeLists.txt` as that means once a package of raft is built it can't be switched between CUDA-free and CUDA-required, which the code would fully support.

Instead `cpp/include/raft/core/device_support.hpp` should check for RAFT_DISABLE_CUDA which non CUDA consumer would provide

wphicks:
> This option shouldn't be encoded in cpp/CMakeLists.txt as that means once a package of raft is built it can't be switched between CUDA-free and CUDA-required, which the code would fully support.

I believe that's precisely what we want. For instance, this might (one day) allow us to provide a slimmed-down CPU-only install of pylibraft/libraft for downstream consumers like CPU-only cuML. I _do_ think we should probably switch the identifier from an opt-in to an opt-out, though. Then, downstream consumers can either set `RAFT_DISABLE_CUDA` themselves or use `DISABLE_CUDA` in their CPM args if they're compiling RAFT themselves. Does that make sense, or is there a downside I'm missing?

wphicks:
> Since you have target_compile_definitions(raft INTERFACE RAFT_ENABLE_GPU) downstream consumers of a raft package ( say from conda ) wouldn't be able to change the mode it has. Which means we would need separate conda packages for GPU and non-GPU support if we encode onto the CMake raft targets this choice.

Yes, this is exactly what we want. For applications which only need non-CUDA functionality from RAFT (e.g. CPU-only cuML), a slimmed-down non-GPU conda package could be created which does not include anything CUDA related.

cpp/CMakeLists.txt
```diff
@@ -227,6 +228,13 @@ target_compile_definitions(raft::raft INTERFACE $<$<BOOL:${RAFT_NVTX}>:NVTX_ENAB
   ]=])
 endif()
 
+##############################################################################
+# - CUDA-free build support --------------------------------------------------
+
+if (NOT DISABLE_CUDA)
+  target_compile_definitions(raft INTERFACE RAFT_ENABLE_GPU)
```
wphicks, line 235: Forgot to update this to use RAFT_DISABLE_CUDA; will update.

robertmaynard:
> > Since you have target_compile_definitions(raft INTERFACE RAFT_ENABLE_GPU) downstream consumers of a raft package ( say from conda ) wouldn't be able to change the mode it has. Which means we would need separate conda packages for GPU and non-GPU support if we encode onto the CMake raft targets this choice.
> 
> Yes, this is exactly what we want. For applications which only need non-CUDA functionality from RAFT (e.g. CPU-only cuML), a slimmed-down non-GPU conda package could be created which does not include anything CUDA related.

Right but from a `find_package(raft)` standpoint you can't distinguish from finding a GPU enabled vs disabled version of the package. So either the package found needs to support both modes, or you need to have different package names.

wphicks:
> Right but from a find_package(raft) standpoint you can't distinguish from finding a GPU enabled vs disabled version of the package. So either the package found needs to support both modes, or you need to have different package names.

Ah, that makes sense! My take on it was that the consumers would control whether they want a CPU-only version of RAFT elsewhere. I.e. Their conda environment would have the CPU-only conda package or they would provide `DISABLE_CUDA` in their CPM args if they're consuming it that way.

I'm not entirely opposed to changing the package name when `DISABLE_CUDA` is set, but I think I would find that a bit surprising as a downstream consumer. Do we have examples of how other libraries handle this? XGBoost, for instance, does not change the package name based on its `USE_CUDA` variable, but maybe you have examples that are more comparable to a package like RAFT?

robertmaynard:
> This still doesn't solve the problem of things like mdspan, though, which compile device functions optionally based on the presence of nvcc, however I wonder if maybe the best way to approach the CPU build from that context would be to hide or turn off the presence of CUDA or the nvcc path so cmake can't find it (@robertmaynard could this be something that's specific in the components?).

The mdspan header enables/disables that behavior based on the compiler vendor. So no change is needed from the build-system. When compiling c++ files you won't activate the cuda bits.

cjnolet:
> The mdspan header enables/disables that behavior based on the compiler vendor. So no change is needed from the build-system. When compiling c++ files you won't activate the cuda bits.

Ah, that's right. Because we could just use `cpp` vs `cu` extensions to tell the compiler that we do or do not want to use nvcc, right?

robertmaynard:
> > The mdspan header enables/disables that behavior based on the compiler vendor. So no change is needed from the build-system. When compiling c++ files you won't activate the cuda bits.
> 
> Ah, that's right. Because we could just use `cpp` vs `cu` extensions to tell the compiler that we do or do not want to use nvcc, right?

Correct

wphicks:
> As an example, I've seen other libraries such as FAISS and GraphBLAS do this successfully by providing libgraphblas.so for the default, non-CUDA enabled stuff, and libgraphblascuda.so for the CUDA-enabled APIs.

That seems reasonable. Do we want `libraft.so` and `libraftcuda.so` or would we rather have the default be CUDA-enabled and use `libraftcpu.so` or `libraftnocuda.so` or something like that?

cjnolet:
@wphicks I'm not opposed to having `libraft_cpu.so` and `libraft_gpu.so` and I'm not even opposed to having them both build by default (and thus making them both opt-out). 

Just to back up for a moment to share some context- the current set of RAFT shared libs were separated into `libraft_nn.so` and `libraft_distance.so` mostly so that we could better isolate dependencies (and make them optional for users who didn't need them). The dependencies in libraft_distance.so have since become "standard" RAPIDS C++ dependencies and we're actively working to remove the FAISS dependency. Ultimately, the goal by 23.02/23.04 once FAISS is removed, is to consolidate the current shared libs into a single `libraft` shared lib (upon which I'm personally learning towards using the convention `libraft_{device_type}.so`. 

@robertmaynard @wphicks @divyegala Thoughts?

PR #991, opened 2022-11-07
Provide CPU/GPU interoperable device_id type

wphicks:
Building on #985 and #986, this PR provides a `device_id` type templated on `device_type` which can be used in CPU/GPU interoperable APIs which must accept a device ID. The primary purpose of this type is to allow for API compatibility and code reuse between CPU and GPU implementations of the same functionality.


wphicks:
rerun tests

wphicks:
Looks like a network issue killed the last run.

PR #995, opened 2022-11-08
Changing Overloads for GCC 11/12 bug

divyegala:


cpp/include/raft/stats/contingency_matrix.cuh
```diff
@@ -142,17 +142,25 @@ void contingencyMatrix(const T* groundTruth,
  * @param[in]  ground_truth: device 1-d array for ground truth (num of rows)
  * @param[in]  predicted_label: device 1-d array for prediction (num of columns)
  * @param[out] out_mat: output buffer for contingency matrix
- * @param[in]  min_label: Optional, min value in input ground truth array
- * @param[in]  max_label: Optional, max value in input ground truth array
+ * @param[in]  opt_min_label: Optional, min value in input ground truth array
```
cjnolet, line 145: Thinking about the docs a bit here, this seems a bit ambiguous now that we aren't explicitly constraining the std::optional  type in the arguments. Can you update this to specify explicitly that a `std::optional` is required here? This applied throughout the PR.

cpp/include/raft/random/rng.cuh
```diff
@@ -730,10 +749,10 @@ void laplace(const raft::handle_t& handle,
  *   the CUDA stream on which to run.
  * @param[inout] rng_state Pseudorandom number generator state.
  * @param[in] in Input vector to be sampled.
- * @param[in] wts Optional weights vector.
+ * @param[in] weights_opt std::optional weights vector.
  *        If not provided, uniform sampling will be used.
  * @param[out] out Vector of samples from the input vector.
- * @param[out] outIdx If provided, vector of the indices
+ * @param[out] outIdx_opt std::optional vector of the indices
```
cjnolet, line 755: Sorry to be nitpicky but it just occurred to me that we also erase the template param type here. Can we specify that in the docs too? (e.g. std::optional<raft::device_vector_view<const WeightsT, IdxT>>).

cjnolet (approved):
LGTM!

cjnolet:
rerun tests

cjnolet:
@gpucibot merge

PR #1000, opened 2022-11-09
[FEA] Add randomized svd from cusolver

lowener:


cpp/include/raft/linalg/detail/rsvd.cuh
```diff
@@ -32,6 +32,97 @@ namespace raft {
...
+    "raft::linalg::randomized_svd(%d, %d, %d)", n_rows, n_cols, k);
+
+  RAFT_EXPECTS(k < std::min(n_rows, n_cols), "k must be < min(n_rows, n_cols)");
+  RAFT_EXPECTS((k + p) < std::min(n_rows, n_cols), "k + p must be < min(n_rows, n_cols)");
+  RAFT_EXPECTS(!gen_U || (U != nullptr), "computation of U vector requested but found nullptr");
+  RAFT_EXPECTS(!gen_V || (V != nullptr), "computation of V vector requested but found nullptr");
+  cudaStream_t stream          = handle.get_stream();
+  cusolverDnHandle_t cusolverH = handle.get_cusolver_dn_handle();
+  cusolverDnParams_t dn_params = nullptr;
+  RAFT_CUSOLVER_TRY(cusolverDnCreateParams(&dn_params));
```
cjnolet, line 59: We should only be interacting w/ cusolver through the cusolver wrappers so this should be creating the params right in the function itself.

cpp/include/raft/linalg/rsvd.cuh
```diff
@@ -18,13 +18,94 @@
...
+ * @param[in]  in:      input matrix in col-major format. On exit this matrix is destroyed
+ *                      [dim = n_rows * n_cols]
+ * @param[out] S:       array of singular values of input matrix.
+ *                      [dim = k]
+ * @param[out] U:       optional left singular values of input matrix. Use std::nullopt to not
+ * generate it. [dim = n_rows * k]
+ * @param[out] V:       optional right singular values of input matrix. Use std::nullopt to not
+ * generate it. [dim = k * n_cols]
+ * @param[in]  k:       Rank of the k-SVD decomposition of matrix in. Number of singular values to
+ * be computed. The rank is less than min(m,n).
```
cjnolet, line 41: I think this should say the rank must be less than or equal to min(m, n).

cpp/include/raft/linalg/rsvd.cuh
```diff
@@ -18,13 +18,94 @@
...
+ * @param[in]  p:       Oversampling. The size of the subspace will be (k + p). (k+p) is less than
+ * min(m,n). (Recommended to be at least 2*k)
+ * @param[in]  niters:  Number of iteration of power method. (2 is recommended)
+ */
+template <typename math_t, typename idx_t>
+void randomized_svd(const raft::handle_t& handle,
+                    raft::device_matrix_view<math_t, idx_t, raft::col_major> in,
```
cjnolet, line 48: This should probably be `const math_t`. Even if cusolver ends up needing to change it, we should still enforce const in the APIs and cast it away before invoking cusolver (and document it as such in the doxygen).

cpp/test/linalg/randomized_svd.cu
```diff
@@ -0,0 +1,246 @@
...
+      left_eig_vectors_ref.data(), left_eig_vectors_ref_h, params.n_row * params.k, stream);
+    raft::update_device(
+      right_eig_vectors_ref.data(), right_eig_vectors_ref_h, params.k * params.n_col, stream);
+    raft::update_device(sing_vals_ref.data(), sing_vals_ref_h, params.k, stream);
+
+    randomized_svd(handle,
+                   raft::make_device_matrix_view<T, uint32_t, raft::col_major>(
+                     data.data(), params.n_row, params.n_col),
+                   raft::make_device_vector_view<T, uint32_t>(sing_vals_act.data(), params.k),
+                   std::make_optional(raft::make_device_matrix_view<T, uint32_t, raft::col_major>(
```
cjnolet, line 108: We probably want to test this will `std::nullopt` for each of the arguments as well.

cjnolet (changes requested):
I'm definitely happy to see this getting wrapped within the RAFT APIs. It's been a long time coming. Just a few fairly minor things.

cpp/include/raft/linalg/rsvd.cuh
lowener: According to cuSolver doc:
`Rank of the k-SVD decomposition of matrix A. rank is less than min(m,n). `

cjnolet: My point here is that this should read "rank must be..." rather than "rank is". In practice, it doesn't make much sense for the rank to be the number of columns in the original data set anyways, but we should be letting the user know what the expected rank "should" be,not what it "is". Does that make sense?

lowener: Yes thanks for the clarification!

cjnolet:
@lowener so you mind retargeting this to 23.02?

lowener:
rerun tests

ajschmidt8:

cjnolet (approved):
LGTM. Great work @lowener!

cjnolet:
@lowener, it would be great to finally get this merged for 23.06. Can you open a PR to cuml to verify these changes?

cjnolet (approved):

cjnolet (approved):

PR #1001, opened 2022-11-10
Add `except +` to cython extern cdef declarations

benfred:
RAFT is using exceptions for error's, but these weren't always being declared to cython in the `cdef extern` blocks. This means that if the function were to throw an exception, the python program would core dump on an uncaught c++ exception - instead of translating the exception into a python exception.


cjnolet (approved):
Thanks for adding these!

cjnolet:
@gpucibot merge

PR #989, opened 2022-11-04
Update cuSparse wrappers to avoid deprecated functions

wphicks:

PR #1003, opened 2022-11-10
Revert "Update cuSparse wrappers to avoid deprecated functions (#989)"

cjnolet:
The changes in https://github.com/rapidsai/raft/pull/989 were merged into RAFT before being fully verified in cuML and it appears to have introduced a bug that caused some googletests to break. I'm reverting this in the meantime to unblock cuml while we fix the resulting bug in cuml.


cjnolet:
I'm going to close this because we already have this branch pinned in cuml so we don't actually need to merge these changes.

PR #1005, opened 2022-11-10
Remove make_mdspan template for memory_type enum

wphicks:
This prevents requiring full specification for all template parameters in downstream `make_mdspan` usage. A more complete discussion of how we want to use `memory_type` in the public API can be deferred to a later PR.


cjnolet (approved):

wphicks:
@gpucibot merge

cjnolet:
I understand this is removing a public API function, but it was just added yesterday, so other than that, this isn't *really*  a breaking change, right?

wphicks:
> I understand this is removing a public API function, but it was just added yesterday, so other than that, this isn't really a breaking change, right?

That's correct.

Issue #1002, opened 2022-11-10
[FEA] Promote TestDeviceBuffer to user-facing APIs


cjnolet:
The TestDeviceBuffer in pylibraft was created initially for testing so that we could build some type of cuda array interface compliant wrapper around an rmm device buffer to pass into the functions for testing without having to introduce a dependency like cupy. 

However, the functions in pylibraft could be made much more pythonic and easy to use if the functions themselves were able to allocate and return the output buffers. For users who still want to allocate the buffers themselves, we could still accept them as an optional argument. Scipy/cupy have similar conventions like this as well.

Consider the following:

1. pw_dists  = pairwise_distance(...) 

2. pairwise_distance(..., out=pw_dists)

We could rename it to something like pylibraft.common.DeviceBuffer. The nice thing is that cupy could directly consume this buffer through the cuda array interface so it would play nicely with whatever library is used downstream (cupy, numba, etc...)

PR #1008, opened 2022-11-10
Moving TestDeviceBuffer to `pylibraft.common.device_ndarray`

cjnolet:
This allows the `pylibraft` APIs to process outputs in place or allocate and return new device memory for outputs in a way which benefits from `__cuda_array_interface__` compliance. Ultimatley, it allows us to stay decoupled from libraries like cupy while still affording the user the benefits of not having to explicitly allocate outputs up front. 

Closes https://github.com/rapidsai/raft/issues/1002


python/pylibraft/pylibraft/cluster/experimental/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+
+# Set the list of Cython files to build
+set(cython_sources kmeans.pyx)
+set(linked_libraries raft::raft raft::distance)
+
+# Build all of the Cython targets
+rapids_cython_create_modules(
+  CXX
+  SOURCE_FILES "${cython_sources}"
+  LINKED_LIBRARIES "${linked_libraries}" MODULE_PREFIX cluster_experimental_
```
benfred, line 23: How does this cluster_experimental module differ from the cluster module?

benfred (approved):
I really like this change - I think its a solid improvement to the usability of the python bindings, while still not requiring cupy

python/pylibraft/pylibraft/cluster/experimental/CMakeLists.txt
cjnolet: Ha great find! I think this file was checked in by mistake.

cjnolet:
Thanks @benfred! 

I'm thinking we should also provide a factory function that accepts the shape, order, and dtype and creates an empty numpy array (internally) to construct the `device_ndarray()`. Maybe something the caller can invoke like this:

```python
dev_ndarr = device_ndarray.empty((50, 10), dtype=np.float32, order="C")
```

Obviously, there's a drawback to using the device_ndarray since it requires a host/device copy. However, I think that's okay because we are giving the option of ease of use (eg have the function automatically return results) and speed (eg users can still explicitly pass in the outputs to keep everything on device and avoid using the device_ndarray)

cjnolet:
JFYI, I added the `pylibraft.device_ndarray.empty()` staticmethod and updated its use in `pairwise_distance`.

python/pylibraft/pylibraft/common/device_ndarray.py
```diff
@@ -0,0 +1,147 @@
...
+                Shape of the empty array, e.g., (2, 3) or 2.
+        dtype : data-type, optional
+                Desired output data-type for the array, e.g, numpy.int8.
+                Default is numpy.float32.
+        order : {'C', 'F'}, optional (default: 'C')
+                Whether to store multi-dimensional dat ain row-major (C-style)
+                or column-major (Fortran-style) order in memory
+        """
+        arr = np.empty(shape, dtype=dtype, order=order)
+        return device_ndarray(arr)
```
benfred, line 74: This maybe extremely nit-picky - but I prefer using `@classmethod` instead of `@staticmethod` for alternate constructors in python, since `@classmethod` works with subclasses: 

```suggestion
    @classmethod
    def empty(cls, shape, dtype=np.float32, order="C"):
        """
        Return a new device_ndarray of given shape and type, without
        initializing entries.

        Parameters
        ----------
        shape : int or tuple of int
                Shape of the empty array, e.g., (2, 3) or 2.
        dtype : data-type, optional
                Desired output data-type for the array, e.g, numpy.int8.
                Default is numpy.float32.
        order : {'C', 'F'}, optional (default: 'C')
                Whether to store multi-dimensional dat ain row-major (C-style)
                or column-major (Fortran-style) order in memory
        """
        arr = np.empty(shape, dtype=dtype, order=order)
        return cls(arr)
```

benfred (approved):
this looks good! I see your point about the extra h2d memory copy w/ `device_ndarray.empty` - but I agree that users can always provide pre-allocated arrays to get around this

divyegala (approved):

cjnolet:
@gpucibot merge

PR #1013, opened 2022-11-14
Add wheel builds

vyasr:
This PR enables building wheels for pylibraft and raft-dask.


vyasr:
I've merged #1032 into this PR to test if it's sufficient to fix wheel builds, but I'm assuming that PR will get merged into 22.12 before this one will so that part of the diff should vanish.

cjnolet (approved):
LGTM

.github/workflows/wheels.yml
```diff
@@ -0,0 +1,75 @@
+name: RAFT wheels
+
+on:
+  push:
+    branches:
+      - 'pull-request/[0-9]+'
```
ajschmidt8, line 6: We don't have the GPU resources currently to run the build / test workflows on every pull request. We'll need to either run a reduced build / test matrix for PRs or configure this workflow to run on a different event type.

vyasr: I've been waiting for each PR to get approved before removing this so that everything up to the final set of changes is tested. Since it looks like we're ready I'll go ahead and update.

cjnolet:
rerun tests

vyasr:
rerun tests

vyasr:
rerun tests

ajschmidt8 (approved):

vyasr:
@gpucibot merge

PR #1019, opened 2022-11-15
IVF-Flat: make adaptive-centers behavior optional

achirkin:
Add an indexing parameter `adaptive_centers` (false by default) to control whether `index.centers()` should be kept up-to-date with the cluster data or remain immutable.

### DIFF
--- a/cpp/include/raft/neighbors/ivf_flat_types.hpp
+++ b/cpp/include/raft/neighbors/ivf_flat_types.hpp
@@ -86,7 +86,7 @@ struct index : ann::index {
   {
     return metric_;
   }
-  /** Whethe `centers()` change upon extending the index (ivf_pq::extend). */
+  /** Whether `centers()` change upon extending the index (ivf_pq::extend). */
   [[nodiscard]] constexpr inline auto adaptive_centers() const noexcept -> bool
   {
     return adaptive_centers_;
--- a/cpp/include/raft/neighbors/ivf_flat_types.hpp
+++ b/cpp/include/raft/neighbors/ivf_flat_types.hpp
@@ -24,6 +24,7 @@
 #include <raft/util/integer_utils.hpp>
 
 #include <optional>
+#include <type_traits>
 
 namespace raft::neighbors::ivf_flat {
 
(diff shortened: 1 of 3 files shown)


achirkin:
NB: breaking change (changed default behavior and the signature of one of the constructors).

cpp/include/raft/neighbors/ivf_flat_types.hpp
```diff
@@ -200,10 +219,15 @@ struct index : ann::index {
   ~index()                          = default;
 
   /** Construct an empty index. It needs to be trained and then populated. */
-  index(const handle_t& handle, raft::distance::DistanceType metric, uint32_t n_lists, uint32_t dim)
+  index(const handle_t& handle,
+        raft::distance::DistanceType metric,
+        uint32_t n_lists,
+        bool adaptive_centers,
```
cjnolet, line 225: Rather than making this an option that persists with the index, why not just make it an optional argument on the `extend()` function? That would enable the user to determine whether it should be done outside of the creation of the index. It's really not an index option, but a design detail based on a specific use-case / usage-pattern.

achirkin: Although I see it sounds more intuitive to keep the parameter in the `extend`, I'd disagree for two reasons:

1. The update-centers routine does not go through the whole data (which would be costly), but only aggregates newly added data. For this to work, the `centers()` must always be up-to-date. Hence, we cannot allow doing an "updating" `extend` after a "non-updating" one. Also from the logical point of view, this would break both alternative invariants (centers are neither constant nor up-to-date). 
2. I hope to preserve the public API the same for all our (ANN) models: `build` gets all parameters through the struct, `extend` never accepts extra parameters (and gets  them from the index). Besides, the `build` function would need to take `update_centers` as well, because it optionally calls `extend`.

cpp/include/raft/neighbors/ivf_flat_types.hpp
```diff
@@ -37,6 +38,19 @@ struct index_params : ann::index_params {
   uint32_t kmeans_n_iters = 20;
   /** The fraction of data to use during iterative kmeans building. */
   double kmeans_trainset_fraction = 0.5;
+  /**
+   * By default (adaptive_centers = false), the cluster centers are trained in `ivf_flat::build`,
+   * and never modified in `ivf_flat::extend`. As a result, you may need to retrain the index
+   * from scratch after adding (`ivf_flat::extend`) a few times new data, distribution of which
```
cjnolet, line 44:
```suggestion
   * from scratch after invoking (`ivf_flat::extend`) a few times with new data, the distribution of which
```

cpp/include/raft/neighbors/ivf_flat_types.hpp
```diff
@@ -37,6 +38,19 @@ struct index_params : ann::index_params {
...
+   * is not well-represented by the original training set.
```
cjnolet, line 45:
```suggestion
   * is no longer representative of the original training set.
```

cpp/include/raft/neighbors/ivf_flat_types.hpp
```diff
@@ -37,6 +38,19 @@ struct index_params : ann::index_params {
...
+   *
+   * The alternative behavior (adaptive_centers = true) is to update cluster centers every time
```
cjnolet, line 47:
```suggestion
   * The alternative behavior (adaptive_centers = true) is to update the cluster centers for new data when
```

cpp/include/raft/neighbors/ivf_flat_types.hpp
```diff
@@ -37,6 +38,19 @@ struct index_params : ann::index_params {
...
+   * the new data is added. In this case, `index.centers()` are always exactly the centroids of
```
cjnolet, line 48:
```suggestion
   * it is added. In this case, `index.centers()` are always exactly the centroids of
```

cjnolet (changes requested):
LGTM. I just have some minor updates to the description of the new option.

cjnolet (approved):
LGTM. Thanks again Artem!

cjnolet:
@gpucibot merge

Issue #972, opened 2022-11-01
[FEA] Expose `pylibraft` function to compute the kmeans inertia


cjnolet:
This function is already exposed on RAFT's public APIs and just needs to be wrapped w/ pylibraft so it can be used in Python. This is in support of the Scikit-learn plugins PoC.

PR #1028, opened 2022-11-17
Expose cluster_cost to python

benfred:
Add cython bindings for the cluster_cost function, to allow computing inertia from python.

Closes https://github.com/rapidsai/raft/issues/972

### DIFF
--- a/cpp/src/distance/cluster_cost.cuh
+++ b/cpp/src/distance/cluster_cost.cuh
@@ -34,11 +34,11 @@ void cluster_cost(const raft::handle_t& handle,
   rmm::device_uvector<ElementType> x_norms(n_samples, handle.get_stream());
   rmm::device_uvector<ElementType> centroid_norms(n_clusters, handle.get_stream());
   raft::linalg::rowNorm(
-    x_norms.data(), X, n_samples, n_features, raft::linalg::L2Norm, true, handle.get_stream());
+    x_norms.data(), X, n_features, n_samples, raft::linalg::L2Norm, true, handle.get_stream());
   raft::linalg::rowNorm(centroid_norms.data(),
                         centroids,
-                        n_clusters,
                         n_features,
+                        n_clusters,
                         raft::linalg::L2Norm,
                         true,
                         handle.get_stream());
@@ -51,8 +51,8 @@ void cluster_cost(const raft::handle_t& handle,
                                      x_norms.data(),
                                      centroid_norms.data(),
                                      n_samples,
-                                     n_features,
                                      n_clusters,
+                                     n_features,
                                      (void*)workspace.data(),
                                      true,
                                      true,
--- a/cpp/src/distance/cluster_cost.cuh
+++ b/cpp/src/distance/cluster_cost.cuh
@@ -1,0 +1,79 @@
+/*
+ * Copyright (c) 2022, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <raft/cluster/kmeans.cuh>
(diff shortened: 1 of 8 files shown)


python/pylibraft/pylibraft/cpp/kmeans.pxd
```diff
@@ -0,0 +1,67 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# cython: profile=False
+# distutils: language = c++
+# cython: embedsignature = True
+# cython: language_level = 3
+
+from pylibraft.common.handle cimport handle_t
```
cjnolet, line 21: I'm liking this convention of putting the wrapper cython definitions in a `cpp` directory.

python/pylibraft/pylibraft/test/test_kmeans.py
```diff
@@ -88,3 +88,19 @@ def test_compute_new_centroids(
...
+@pytest.mark.parametrize("dtype", [np.float32, np.float64])
+def test_cluster_cost(n_rows, n_cols, n_clusters, dtype):
+
+    X = np.random.random_sample((n_rows, n_cols)).astype(dtype)
+    X_device = device_ndarray(X)
+
+    centroids = X[:n_clusters]
+    centroids_device = device_ndarray(centroids)
+
+    # TODO: compute inertia naively, make sure is close
```
cjnolet, line 105: We can use our pairwise distances to compute this naively :-P

cjnolet (changes requested):
The changes LGTM. Just need assertions in the pytest.

python/pylibraft/pylibraft/test/test_kmeans.py
benfred: done - thanks for the pointer!

benfred:
I've marked this ready to review - even though I don't think the results being returned here are actually correct because of https://github.com/rapidsai/raft/issues/1036

cpp/src/distance/cluster_cost.cuh
```diff
@@ -0,0 +1,79 @@
...
+  auto distances = raft::make_device_vector<ElementType, IndexType>(handle, n_samples);
+  thrust::transform(
+    handle.get_thrust_policy(),
+    min_cluster_distance.data_handle(),
+    min_cluster_distance.data_handle() + n_samples,
+    distances.data_handle(),
+    [] __device__(const raft::KeyValuePair<IndexType, ElementType>& a) { return a.value; });
+
+  rmm::device_scalar<ElementType> device_cost(0, handle.get_stream());
+  raft::cluster::kmeans::cluster_cost(
```
cjnolet, line 70: I think we're doing an additional computation here that can be avoided by running the `fusedL2NNMinReduce` w/ `sqrt=False`. If we have the squared distances already then we should just be able sum them up to get the `inertia` score.

cjnolet: Could do a simple L1 norm computation for the reduction, I guess. https://github.com/rapidsai/raft/blob/d6df557eeb8388c74bc0135c0a32681f887b8bb8/cpp/include/raft/linalg/norm.cuh#L109

benfred: thats a good point - and I've made the change to fusedL2NNMinReduce w/ sqrt=False.   I checked the cluster cost, and its already only doing a sum reduction - so using sqrt=True was wrong here.  

This also lets us avoid the issue in #1036 - afaict you are right and this issue is only with sqrt=True (all the tests pass at least now =) ).

python/pylibraft/pylibraft/test/test_kmeans.py
```diff
@@ -88,3 +88,31 @@ def test_compute_new_centroids(
     actual_centers = new_centroids_device.copy_to_host()
 
     assert np.allclose(expected_centers, actual_centers, rtol=1e-6)
+
+
+@pytest.mark.parametrize("n_rows", [8])
```
cjnolet, line 93: Might be a good idea to test this with a slightly larger number like 100.

cjnolet (approved):
LGTM. Just a very small comment but I'm pre-approving since it's pretty minor and we could do it in a follow-on if needed.

python/pylibraft/pylibraft/test/test_kmeans.py
benfred: nice catch! I had reduced it for debugging - 

its back at 100 in the last commit

benfred:
@gpucibot merge

cjnolet:
cc @betatim

cjnolet:
@gpucibot merge

Issue #981, opened 2022-11-02
[FEA] Use doctest for testing python docstrings


benfred:
**Is your feature request related to a problem? Please describe.**

The example code in the python docstrings isn't being automatically tested right now.  This can cause the example docstrings to break over time as the codebase is updated (https://github.com/rapidsai/raft/pull/980)

We should add tests to make sure that the example code blocks in our python docstrings can run successfully

**Describe the solution you'd like**

We can run the example docstrings with our unittests using doctest via the `--doctest-modules` option in pytest. This is consistent with other rapids projects like cudf

**Additional context**

This requires changing the format of the docstring code blocks to the `numpydoc` format ( https://github.com/rapidsai/cudf/issues/652 )

PR #1030, opened 2022-11-18
Fixing broken doc functions and improving coverage

cjnolet:


cjnolet:
rerun tests

cjnolet:
rerun tests

cpp/include/raft/random/rng.cuh
```diff
@@ -722,6 +722,8 @@ template <typename T>
 using weight_t = typename weight_alias<T>::type;
 }  // namespace sample_without_replacement_impl
 
+//@{
```
benfred, line 725: Should we add something like `@defgroup sample without replacement calls` here to indicate what this group is?

docs/source/cpp_api/mdspan.rst
```diff
@@ -216,13 +222,10 @@ Device Vocabulary
...
 ----------------
 
-.. doxygenfunction:: raft::make_device_mdspan
-    :project: RAFT
-
 .. doxygenfunction:: raft::make_device_matrix_view
     :project: RAFT
 
-.. doxygenfunction:: raft::make_device_vector_view
+.. doxygenfunction:: raft::make_device_vector_view(ElementType* ptr, IndexType n)
```
benfred, line 228: This might be a dumb question - but why do we need to update this function to include the parameters, but not the make_device_matrix_view function right above it (or make_device_scalar_view below it)?  It seems we should be consistent here

python/pylibraft/pylibraft/distance/pairwise_distance.pyx
```diff
@@ -118,51 +118,63 @@ def distance(X, Y, out=None, metric="euclidean", p=2.0, handle=None):
     --------
     To compute pairwise distances on cupy arrays:
 
-    >>> import cupy as cp
-    >>> from pylibraft.common import Handle
-    >>> from pylibraft.distance import pairwise_distance
-    >>> n_samples = 5000
-    >>> n_features = 50
-    >>> in1 = cp.random.random_sample((n_samples, n_features),
+    .. code-block:: python
```
benfred, line 121: I appreciate that this is consistent with all the other docstrings in raft right now - but I think we should probably be standardizing the other way on the numpydoc format so that we can use doctest to test these docstrings automatically (#981)

benfred (approved):
looks good! I like that this expands the docs coverage - and we're now including all the pylibraft modules in the sphinx docs. 

Left a couple minor comments below,

cpp/include/raft/random/rng.cuh
cjnolet: I'd like to move this out into its own file in 23.02. Sphinx seems to be having trouble resolving it and I'm not sure if it's related to sphinx, breathe, or doxygen. I'll create an issue for it.

docs/source/cpp_api/mdspan.rst
cjnolet: Mostly to keep this consistent with the other APIs. doxygen has been getting really confused with some overloads, even when the overloads aren't necessarily ambiguous so I've been specifying the params in the meantime. Another reason I've been specifying the params is so that we can isolate the mdspan-based APIs in the docs and ignore the pointer-based APIs in the meantime.

I'm hoping we can eventually find a better way to handle this.

python/pylibraft/pylibraft/distance/pairwise_distance.pyx
cjnolet: Got it. In that case I definitely think we should use the other format. Let me create an issue for that for 23.02 as well. It's just getting too close to code freeze to do that for 22.12.

benfred: yeah - I don't think we should change this for 22.12, but would be nice to get in for 23.02 .

cpp/include/raft/random/rng.cuh
cjnolet: Thanks for the suggestion here! Using `defgroup` did fix the ambiguity issues I was facing w/ the `sample_with_replacement` function so I went ahead and added that here. I'm thinking we should use defgroup more throughout the codebase in general to isolate and document only the `mdspan`-based APIs everywhere.

cjnolet: I'll leave the issue open though, since the other piece is moving `sample_without_replace` into its own file.

tfeher (approved):
Hi Corey, nice to see these updates! I have some suggestions below, but most of these can be implemented in a follow up PR. Overall the current PR looks good to me, pre-approving.

cjnolet:
@tfeher thank you for the detailed review (and for doing it on such short notice)! Since code freeze is tomorrow (which is a US holiday), I’m definitely going to create a follow-on PR for all of these items, but I’ll target it for 23.02.

cjnolet:
@gpucibot merge

PR #1032, opened 2022-11-18
Make ucx linkage explicit and add a new CMake target for it

### LINKED ISSUE #1031: [FEA] Remove `dlopen` of ucx from std_comms
We initially used dlopen in the std_comms because ucx wasn't yet fully integrated into the rapids software stack and so it was harder to link directly against ucx. Now that we can do so, it's making the build process unnecessarily confusing and complicated, since we directly like against ucx in the python package but then still dlopen in the c++ layer, expecting ucx to be magically available on the system ld path. 

This should be a fairly trivial change, but will also require downstream projects to link against ucx directly as well in their c++ code. Thankfully, as @vyasr pointed out, the new version of ucx supports find_package.

vyasr:
This PR removes the dlopen logic for libucp in ucp_helper.hpp in favor of calling the relevant APIs directly. It also adds a new CMake component `raft::distributed` that can be used by dependent libraries to indicate the dependency on parts of raft that require UCX.

While it does not change any public APIs, I have marked this PR as breaking since it does mean that any C++ code linking to UCX must now ensure that UCX is available at link time. It is no longer sufficient to make the library available at runtime.

Resolves #1031.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -205,6 +205,12 @@ endif()
 add_library(raft INTERFACE)
 add_library(raft::raft ALIAS raft)
 
+rapids_find_package(
+  ucx REQUIRED
+  BUILD_EXPORT_SET raft-exports
+  INSTALL_EXPORT_SET raft-exports
+)
+
 target_include_directories(
   raft INTERFACE "$<BUILD_INTERFACE:${RAFT_SOURCE_DIR}/include>" "$<INSTALL_INTERFACE:include>"
 )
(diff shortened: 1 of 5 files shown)


vyasr:
rerun tests

cjnolet:
Thanks for fixing this, @vyasr.  Now we just need to add ucx as an explicit link dependency of cuml and cugraph

vyasr:
> Thanks for fixing this, @vyasr. Now we just need to add ucx as an explicit link dependency of cuml and cugraph

Done.

cjnolet:
@robertmaynard this adds a new optional component for `distributed` (which currently adds ucx but will add NCCL in the future). Any chance we can get your blessing on these changes?

cjnolet (approved):
LGTM, thanks @vyasr!

cjnolet:
@gpucibot merge

cjnolet:
@gpucibot merge

PR #939, opened 2022-10-21
Add cutlass 3xTF32,DMMA based L2/cosine distance kernels for SM 8.0 or higher

mdoijade:
-- 3xTF32 cutlass based L2 exp/cosine kernel provides 3.5x speedup for fp32 inputs compared to existing pairwise distance kernel for ampere or higher.
-- DMMA cutlass based implementation for L2 exp/cosine provides 2.6x speedup instead of existing double precision FMA pipeline based kernel.
-- add cutlass as header only dependency to RAFT.

PR #1033, opened 2022-11-18
Suppres ptxas stack size warning in Debug mode

tfeher:
PR #939 introduced CUTLASS dependency. When compiled in debug mode, this leads to the following error:

```
ptxas error   : Stack size for entry function '_ZN12raft_cutlass6KernelINS_...' cannot be statically determined
```

This would be normally just a warning, but we treat warnings as errors. This PR disables the warning in Debug mode.


cjnolet (approved):

cjnolet:
@gpucibot merge

PR #1035, opened 2022-11-18
IVF-PQ: use device properties helper

achirkin:
Use raft handle's lazy-loading helper `get_device_properties` instead of explicitly calling `cudaGetDeviceProperties` on every kernel launch, which is a costly operation.


tfeher (approved):
Thanks for this improvement, LGTM!

cjnolet:
@gpucibot merge

PR #1038, opened 2022-11-22
Add ANN refinement method

tfeher:
This PR implements refinement for approximate nearest neighbor search.

Refinement is a post processing step for ANN search, it follows an ANN search that returned `k0` neighbor candidates,
and select `k` out of these candidates. The selection by calculating exact distances from the original dataset.

Refinement can increase accuracy. It is useful for ANN methods that quantize the dataset and therefore loose accuracy during distance calculation (e.g. IVF-PQ).


sean-frye:
rerun tests

Ethyling:
rerun tests

Ethyling:
rerun tests

cpp/include/raft/spatial/knn/detail/ivf_flat_build.cuh
```diff
@@ -296,4 +302,76 @@ inline auto build(
...
+ *
+ * @param[in] handle the raft handle
+ * @param[inout] refinement_index
+ * @param[in] dataset device pointer to dataset vectors, size [n_rows, dim]. Note that n_rows is
+ *   not known to this function, but each candidate_idx has to be smaller than n_rows.
+ * @param[in] candidate_idx device poniter to neighbor candidates, size [n_queries, n_candidates]
+ * @param[in] n_candidates  of neighbor_candidates
+ */
+template <typename T, typename IdxT>
+inline void build_refinement_index(const handle_t& handle,
```
achirkin, line 323: Maybe rename this to something like `fill_refinement_index` to avoid a false association with the `build` method? My first thought was to compare it against `build` function, whereas it should be compared against "extend".

Ethyling:
rerun tests

Ethyling:
rerun tests

robertmaynard (approved):
CMake changes LGTM

cpp/include/raft/neighbors/refine.cuh
```diff
@@ -0,0 +1,99 @@
...
+ * @param[in] handle the raft handle
+ * @param[in] dataset host matrix that stores the dataset [n_rows, dims]
+ * @param[in] queries host matrix of the queries [n_queris, dims]
+ * @param[in] neighbor_candidates host matrix with indices of candidate vectors [n_queries,
+ *   n_candidates], where n_candidates >= k
+ * @param[out] indices host matrix that stores the refined indices [n_queries, k]
+ * @param[out] distances host matrix that stores the refined distances [n_queries, k]
+ * @param[in] metric distance metric to use. Euclidean (L2) is used by default
+ */
+template <typename idx_t, typename data_t, typename distance_t, typename matrix_idx>
```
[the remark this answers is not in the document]
tfeher: Good point. We could probably just have mdspan with the corresponding extent. But I have some issues with type deduction.  I plan to address this in a follow up PR.

tfeher (commented):
Thanks Artem for the review, I have updated the PR!

> Since ivf-flat is wired to accept the interleaved data, we end up reading the vectors twice, which does not sound very efficient. 

That is right. Actually the index building is the time consuming part because it need to gather the neighbor vectors based on their indices, so we do not have a good access pattern. Streaming over the already gathered data much faster.

> Could a naive segmented distance kernel followed by a select_topk invocation be faster and even less code? 

Yes, we shall explore that in a follow up work.

cpp/include/raft/neighbors/detail/refine.cuh
```diff
@@ -0,0 +1,232 @@
...
+  // - We run IVF flat search with n_probes=1 to select the best k elements of the candidates.
+  rmm::device_uvector<uint32_t> fake_coarse_idx(n_queries, handle.get_stream());
+
+  thrust::sequence(
+    handle.get_thrust_policy(), fake_coarse_idx.data(), fake_coarse_idx.data() + n_queries);
+
+  raft::neighbors::ivf_flat::index<data_t, idx_t> refinement_index(
+    handle, metric, n_queries, false, dim);
+
+  raft::spatial::knn::ivf_flat::detail::fill_refinement_index(handle,
```
cjnolet, line 109: Just a little note: I'm hoping to remove the raft::spatial::knn namespace altogether in 23.02 and officially move all the detail headers over to raft::neighbors. Hopefully we can get all the deprecated files removed as well.

cjnolet (approved):
LGTM

cjnolet:
@gpucibot merge

PR #1040, opened 2022-11-22
[HOTFIX] Fix fusedL2NN bug that can happen when the same point appears in both x and y

Nyrio:
Solves #1036 

Even when computing a sum of squares, the distance from a point to itself can apparently be `-0.0` in which case the square root is `nan` and comparisons are broken.


Nyrio:
cc @benfred @cjnolet

benfred (approved):
lgtm -  Thanks for the quick fix @Nyrio !

cjnolet (approved):

benfred:
@gpucibot merge

PR #1046, opened 2022-11-24
New helper: using_mapped_memory_t

achirkin:
Add a RAII-style helper that checks if a given pointer is available on the current device. If that is not the case, it maps the host memory to the device memory for the lifetime of the struct.

This helper is especially useful when the host memory comes from `mmap`, as it allows to avoid an extra host allocation and a host<->host memory transfer. It does not help with huge allocations though: the passed buffer must fit within the device memory.


achirkin:
Note, this PR improves an already existing helper and moves it out of the `detail` namespace.

cpp/include/raft/util/cudart_utils.hpp
```diff
@@ -537,6 +538,83 @@ inline auto get_pool_memory_resource(rmm::mr::device_memory_resource*& mr, size_
...
+  using_mapped_memory_t(PtrT ptr, size_t size)
+  {
+    if (ptr == nullptr) { return; }
+    cudaPointerAttributes attr;
+    RAFT_CUDA_TRY(cudaPointerGetAttributes(&attr, ptr));
+    if (attr.devicePointer != nullptr) {
+      dev_ptr_ = attr.devicePointer;
+    } else {
+      host_ptr_ = reinterpret_cast<void*>(const_cast<uncv_ptr_t>(ptr));
+      RAFT_CUDA_TRY(cudaHostRegister(host_ptr_, size, choose_flags()));
```
tfeher, line 578: 1. Since using mapped memory has a perf impact, it would be good to emit a log message.
2. It would be great if the error message could inform the user about possible workaround (e.g. using managed memory).

tfeher (changes requested):
Thanks Artem for the PR! The code where the new wrapper is used looks much cleaner. It looks good overall, I just have a suggestion about logging and error reporting.

cpp/include/raft/util/cudart_utils.hpp
tfeher: I would prefer option 1, maybe a new file in `raft/util/`?

cjnolet: Ultimately I'd like to see a builder in raft/core for this if it's part of the vocabulary that users should be interacting with directly.

However, I'd also like to see it consolidated with mdarray. For example, I think it would plug into the existing mdarray API nicely as a new container strategy  and I believe it would be able to make use of the host_device_accessor to represent the ability to be read from both host and device.

cjnolet:
@achirkin, I'm making my way through https://github.com/rapidsai/raft/pull/1056/ and I notice use of `with_mapped_memory_t` is removed [here](https://github.com/rapidsai/raft/pull/1056/files#diff-d86ebde7fe8da2962578ea518d33caf74c817ce7d53fd39dd5073ceb1913f02aL1043). 

Are you still planning to update this using the mdarray wrapper that was discussed offline?

achirkin:
> Are you still planning to update this using the mdarray wrapper that was discussed offline?

Yes, I'm on it! (I've just come back from holidays) Even though I removed the dependency on this in #1056 by managing the copying manually, I still believe this is a useful feature at various stages of porting new algorithms to GPU.

cjnolet:
@achirkin I recall we had a discussion awhile back internally, which seemed to have surfaced a reasonable design for this feature. Are you still planning to work on this or is this PR just too stale at this point?

PR #1048, opened 2022-11-24
Fix concurrency issues in k-means++ initialization

Nyrio:
The cub calls were previously launched on the default stream whereas the rest was launched in the stream attached to the raft handle, and there was a missing synchronization after a D2H copy of the index that is used immediately after.


cjnolet:
@gpucibot merge

Issue #1024, opened 2022-11-17
[ENH] Move template arguments of `L1Op`, `L2Op` etc to `operator()` for automatic type inferrence


Nyrio:
Follows [a suggestion](https://github.com/rapidsai/raft/pull/979#discussion_r1024461911) by @achirkin who believes this would be less error-prone and would simplify the experience on the caller side.

PR #1049, opened 2022-11-25
Unify use of common functors

Nyrio:
The same functors are redefined in many different files. I believe this change can help with compilation times, executable size, and consistency.


Nyrio:
Notes regarding #1024 since it's related:

- If I remove the template arguments of all these functors, it will be a breaking change. I also have the option of setting them to void and throwing a deprecation warning if set, then we can remove them later.
- Any thoughts on whether there are cases where we don't want the types to be inferred? Some prims have different `InT`/`OutT` but many of these functors are just implicitly casting the output anyway.

cpp/include/raft/util/cuda_utils.cuh
```diff
@@ -508,44 +508,170 @@ HDI double myATanh(double x)
...
 
-template <typename Type>
+template <typename InT, typename OutT = InT>
 struct Sum {
-  HDI Type operator()(Type a, Type b) { return a + b; }
+  HDI OutT operator()(InT a, InT b) const { return a + b; }
+};
+
+template <typename InT, typename OutT = InT>
+struct Subtract {
```
achirkin, line 562: Since the effort is to unify the use of these structs and encourage their use, I'd suggest we revisit https://github.com/rapidsai/raft/issues/1024 again and move the template arguments into the operator() where possible. I'd leave at the struct level only in non-inferable cases, like the non-type template parameters or return type.


Although I see it can be tempting to do the arithmetic op and convert data type in one go, I think it's not healthy to have the `OutT` different from the `InT` in the arithmetic functions like these. Instead, we could provide more "functors" like the [`ann_utils/mapping` ](https://github.com/rapidsai/raft/blob/branch-22.12/cpp/include/raft/spatial/knn/detail/ann_utils.cuh#L158) plus the function composition (something like this):
```
template <typename OuterF, typename InnerF>
struct compose {
  OuterF fo;
  InnerF fi;
  template <typename... Args>
  HDI auto operator()(Args... args) const { return fo(fi(args...); }
};
```
It could be worth checking if the resulting compiled code is efficient though...

Nyrio: I agree, hence my comment above. Mainly we need to figure out the proper way of deprecating template args for the functors that already existed in the API. Regarding the input/output types, I mostly agree it's not really useful to have different input/output types if it only does an implicit cast anyway. For non-trivial conversions, we could add a composition operator like your `compose` structure. For trivial casts, I think it's unnecessary.

Nyrio: That's one possibility, yes. And we can also use the occasion to improve the naming scheme (`Sum` -> `add_op`, `L2Op` -> `square_op`, `L1Op` -> `abs_op` etc ; it's all open for discussion)

@cjnolet What do you think about this? (deprecating the old functors and replacing them with new type-inferred ones with different names).

Nyrio: @achirkin I've added automatic type inference and deprecated the existing functors.

Nyrio:
(sorry for the irrelevant review triggers, I updated the target branch)

cpp/include/raft/spatial/knn/detail/ivf_pq_search.cuh
```diff
@@ -422,9 +422,8 @@ void postprocess_distances(float* out,        // [n_queries, topk]
         out,
         in,
         len,
-        [scaling_factor] __device__(ScoreT x) -> float {
-          return scaling_factor * scaling_factor * float(x);
-        },
+        raft::scalar_op<raft::binary_compose_op<raft::mul_op, raft::cast_op<float>>, float>(
+          scaling_factor * scaling_factor),
```
Nyrio, line 426: Just a note on this for reviewers: it's an interesting case because automatic type inference leads to ambiguity (CUDA doesn't know which operator to use to multiply `half` and `float`). So it's a good occasion to try out composition, which arguably doesn't really have advantages here as opposed to passing a custom lambda unless by chance the same composition is used somewhere else in the code. But it does work, perhaps at the expense of readability.

cpp/include/raft/util/cuda_utils.cuh
```diff
@@ -508,44 +508,307 @@ HDI double myATanh(double x)
...
+};
+/** @} */
+
+/**
+ * @defgroup Functors Commonly used functors.
+ * The optional index argument is mostly to be used for MainLambda in reduction kernels
+ * @{
+ */
+
+struct identity_op {
```
cjnolet, line 583: The original operator functions were jammed into `cuda_utils` back before we had built out a formal API for primitives and now that we are revamping and once again centralizing these, I think we should consider moving them (and protecting the HDI w/ another macro that is safe to use w/ gcc). 

I'm thinking something like `raft/core/operators.hpp` might be a great spot to put them more front-and-center. I'd also like see the operators in the sparse pairwise distances API consolidated there as well.

cjnolet: This would be the macro to replace `HDI`, btw, which would be safe to use in both CUDA and non-cuda environments: https://github.com/rapidsai/raft/blob/11c51059c401d344097b86d9b9ab4a6f464e9410/cpp/include/raft/core/detail/macros.hpp#L40.

Nyrio: Ok, there's one circular dependency issue to solve, though: the functors need `cuda_utils.cuh`, and `cuda_utils.cuh` has helpers that use the functors (e.g `warpReduce` when `reduce_op` isn't passed).

My suggestion would be to move all the reduction helpers to another header, e.g `device_utils.cuh` which already has reduction helpers. If we want to keep the deprecated ones for a while, we can replace `add_op` with a custom lambda there.

achirkin: +1 for `raft/core/operators.hpp` would be a perfect spot for the new operators! Also, I'd add `constexpr` to all constructors/methods.

Nyrio: I'll also replace `mySqrt` and similar functions in the functors then, since the new header won't depend on `cuda_utils.cuh`

cpp/include/raft/core/operators.hpp
```diff
@@ -0,0 +1,262 @@
...
+template <typename Type>
+using scalar_pow_op = scalar_op<pow_op, Type>;
+
+template <typename OuterOpT, typename InnerOpT>
+struct unary_compose_op {
+  OuterOpT outer_op;
+  InnerOpT inner_op;
+
+  template <typename Type>
+  constexpr RAFT_INLINE_FUNCTION auto operator()(Type a) const
```
achirkin, line 242: Are there any problems with allowing variable number of arguments in the input op here (as I tried to draft in https://github.com/rapidsai/raft/pull/1049#discussion_r1033378266)?

cpp/include/raft/core/operators.hpp
```diff
@@ -0,0 +1,262 @@
...
+  constexpr const_op(ScalarT s) : scalar{s} {}
+
+  template <typename InT>
+  constexpr RAFT_INLINE_FUNCTION ScalarT operator()(InT unused) const
+  {
+    return scalar;
+  }
+};
+
+template <typename ComposedOpT, typename ScalarT>
```
achirkin, line 204: Would you consider swapping the template arguments here? I have a feeling there's a greater chance one would want to force the type of the scalar, but keep the `ComposedOpT` inferred (then, perhaps the order of the field should be swapped as well). Also, maybe rename the latter to `BinaryOpT`, so that it's clear we expect the binary operator as the argument.

cpp/include/raft/core/operators.hpp
```diff
@@ -0,0 +1,262 @@
...
+struct scalar_op {
+  ComposedOpT composed_op;
+  const ScalarT scalar;
+
+  constexpr scalar_op(ScalarT s) : scalar{s} {}
```
achirkin, line 209: I'm not sure: does having this constructor prevent the default two-parameter constructor from existence? If yes, maybe it makes sense to add it here as well.

achirkin, line 209: Maybe add the default  initializer for `composed_op` here as well, for readability? ( `scalar_op(..): scalar{s}, composed_op{}`)

cpp/include/raft/core/operators.hpp
Nyrio: Oops, you're right, could definitely be useful when wrapping around ops that take an optional index. And if the outer op needs the index, the binary compose operator can be used.

Nyrio:
I'm also thinking that we could have an operator to apply an arbitrary number of compositions (e.g `outer(inner1(a), inner2(b), inner3(c), ...)`). But let's be lazy about this and only add it if the need arises.

cpp/include/raft/spatial/knn/detail/ivf_pq_search.cuh
achirkin: What if you do instead unary_compose_op(scalar_mul_op{ scaling_factor * scaling_factor }, raft::cast_op<float>{}). does it compile?

cpp/include/raft/core/operators.hpp
```diff
@@ -0,0 +1,262 @@
...
+template <typename Type>
+using scalar_pow_op = scalar_op<pow_op, Type>;
+
+template <typename OuterOpT, typename InnerOpT>
+struct unary_compose_op {
+  OuterOpT outer_op;
+  InnerOpT inner_op;
+
+  template <typename... Args>
+  constexpr RAFT_INLINE_FUNCTION auto operator()(Args... args) const
```
achirkin, line 242: Do you think we should care about having the perfect forwarding here? Certainly, most of the time we just pass arithmetic types here, but maybe once in a while a struct passed by reference could come up. Hence passing it by value wouldn't be optimal.

cpp/include/raft/core/operators.hpp
```diff
@@ -0,0 +1,262 @@
...
+using scalar_div_op = scalar_op<div_op, Type>;
+
+template <typename Type>
+using scalar_div_checkzero_op = scalar_op<div_checkzero_op, Type>;
+
+template <typename Type>
+using scalar_pow_op = scalar_op<pow_op, Type>;
+
+template <typename OuterOpT, typename InnerOpT>
+struct unary_compose_op {
```
achirkin, line 237: I'm ok with this name `unary_compose_op`, but maybe it could also make sense to change it to just `compose_op`? My intuition here is that (1) users may mistakenly think that the `inner_op` is limited to one argument (2) it may be used all over the place, so shorter name could pay off :)

cpp/include/raft/core/operators.hpp
```diff
@@ -0,0 +1,262 @@
...
+namespace raft {
+
+/**
+ * @defgroup Functors Commonly used functors.
+ * The optional index argument is mostly to be used for MainLambda in reduction kernels
+ * @{
+ */
+
+struct identity_op {
+  template <typename Type, typename IdxType = int>
```
achirkin, line 33: Would it make sense to replace the `IdxType` here and in the other operators, which do not use the second argument, with something like the this?:
```
template <typename Type, typename UnusedArgs...>
constexpr RAFT_INLINE_FUNCTION auto operator()(Type in, UnusedArgs...) const { ... }
```
I think, the latter would be more flexible and explicit in the fact that it does not use and does not need any sorts of indices.

cpp/include/raft/core/operators.hpp
Nyrio: I don't see how `ComposedOpT` could be inferred. The only argument to the constructor is the scalar, so if C++ was a bit smarter it could infer `ScalarT` using C++17's CTAD, but unfortunately, the other template argument seems to break that.

Nyrio: I'll see about having two constructors, then. The one-arg constructor is convenient for the aliases:

```c++
template <typename Type>
using scalar_add_op = scalar_op<add_op, Type>;

template <typename Type>
using scalar_sub_op = scalar_op<sub_op, Type>;

...
```

However, I'm still unsure in which case `ComposedOpT` will be inferred and `ScalarT` won't? With a two-arg constructor, both are inferred.

achirkin: I'm assuming the case when the scalar you provide is of different type and you want it be implicitly converted.

Nyrio: Nothing that can't be achieved with a cast:

```c++
uint32_t offset;
scalar_op<size_t> op1(add_op, offset);
scalar_op op2(add_op, (size_t)offset);
```

I'm wondering if, in a future version, C++ will be clever enough to infer `ScalarT` when constructing like:

```c++
scalar_op<add_t> op(offset);
```

achirkin: Sure it can be achieved with the cast; it's just my personal preference to like typing `scalar_op<size_t>(1, f)` instead of  `scalar_op(f, size_t(1))`.
Btw, c-style casts are discouraged by our clang-format; hence in your case one would either use constructor `scalar_op op2(add_op, size_t(offset));` or static cast `scalar_op op2(add_op, static_cast<size_t>(offset))`.

cpp/include/raft/core/operators.hpp
Nyrio: Good point! I've been quite conservative here but I like your idea better.

cpp/include/raft/spatial/knn/detail/ivf_pq_search.cuh
Nyrio: Nope, but let me add some constructors, guard with `enable_if` when necessary, and I can make it possible ;)

cpp/include/raft/core/operators.hpp
Nyrio: What I have observed so far is that if the inner op takes references, as `Args...` are inferred as references, structs are passed through with zero copy. But it doesn't hurt to add `std::forward` so I will. If you have an example of a case where that makes a difference, I'd be curious to read it.

Consider the following example, the output is 37:

```c++
template <typename T>
void f(T& a, const T& b) {
    a = b;
}

template<typename... Args>
void g(Args... args) {
    f(args...);
}

int main() {
    int a = 13;
    f(a, 37);
    std::cout << a << std::endl;
}
```

Nyrio: You're right, we need perfect forwarding here to pass args by reference, I'll add that.

cpp/include/raft/distance/detail/chebyshev.cuh
```diff
@@ -72,7 +72,7 @@ static void chebyshevImpl(const DataT* x,
 
   // Accumulation operation lambda
   auto core_lambda = [] __device__(AccT & acc, DataT & x, DataT & y) {
-    const auto diff = raft::L1Op<AccT, IdxT>()(x - y);
+    const auto diff = raft::myAbs(x - y);
```
cjnolet, line 75: It sould be nice to see this become a `raft::abs_op` and the one below it become `raft::max_op`. Any reason for now consoliding these as well?

cpp/include/raft/core/operators.hpp
```diff
@@ -0,0 +1,292 @@
...
+template <typename Type>
+using scalar_div_op = scalar_op<Type, div_op>;
+
+template <typename Type>
+using scalar_div_checkzero_op = scalar_op<Type, div_checkzero_op>;
+
+template <typename Type>
+using scalar_pow_op = scalar_op<Type, pow_op>;
+
+template <typename OuterOpT, typename InnerOpT>
```
achirkin, line 243: I saw we'd got at least one case where we nest several `compose_op`s. Then, bonus points for making this arbitrary chain composition :)
```C++
template <typename... OpsT>
struct compose_op {
  const std::tuple<OpsT...> ops;

  constexpr explicit compose_op(OpsT... ops) : ops(ops...) {}

  template <typename... Args>
  constexpr RAFT_INLINE_FUNCTION auto operator()(Args&&... args) const
  {
    return compose<sizeof...(OpsT)>(std::forward<Args>(args)...);
  }

 private:
  template <size_t RemOps, typename... Args>
  constexpr RAFT_INLINE_FUNCTION auto compose(Args&&... args) const
  {
    if constexpr (RemOps > 0) {
      return compose<RemOps - 1>(std::get<RemOps - 1>(ops)(std::forward<Args...>(args)...));
    } else {
      return identity_op{}(std::forward<Args...>(args)...);
    }
  }
};
```
Then, the composition in `ivf_pq_search:434` would look much simpler:
```C++
compose_op{scalar_mul_op<float>{scaling_factor}, sqrt_op{}, cast_op<float>{}}
```

achirkin (changes requested):
Thanks for the changes! Could you please run `clang-tidy` over the `include/raft/core/operators.hpp` and fix (or suppress when necessary) the warnings? I see a bunch of them, mostly `modernize-use-trailing-return-type`.

cpp/include/raft/core/operators.hpp
```diff
@@ -0,0 +1,292 @@
...
+  constexpr compose_op(OuterOpT out_op, InnerOpT in_op) : outer_op{out_op}, inner_op{in_op} {}
+
+  template <typename... Args>
+  constexpr RAFT_INLINE_FUNCTION auto operator()(Args&&... args) const
+  {
+    return outer_op(inner_op(std::forward<Args>(args)...));
+  }
+};
+
+template <typename OuterOpT, typename InnerOpT1, typename InnerOpT2 = raft::identity_op>
```
achirkin, line 264: As you suggested, we could generalize this one as well (NB: add include the tuple header):
```C++
template <typename OuterOpT, typename... ArgOpsT>
struct map_args_op {
  const OuterOpT outer_op;
  const std::tuple<ArgOpsT...> arg_ops;

  constexpr explicit map_args_op(OuterOpT outer_op, ArgOpsT... arg_ops)
    : outer_op{outer_op}, arg_ops{arg_ops...}
  {
  }

  template <typename... Args>
  constexpr RAFT_INLINE_FUNCTION auto operator()(Args&&... args) const
  {
    constexpr size_t kNumOps = sizeof...(ArgOpsT);
    static_assert(kNumOps == sizeof...(Args),
                  "The number of arguments does not match the number of mapping operators");
    return map_args(std::make_index_sequence<kNumOps>{}, std::forward<Args>(args)...);
  }

 private:
  template <size_t... I, typename... Args>
  constexpr RAFT_INLINE_FUNCTION auto map_args(std::index_sequence<I...>, Args&&... args) const
  {
    return outer_op(std::get<I>(arg_ops)(std::forward<Args>(args))...);
  }
};
```

cpp/include/raft/distance/detail/chebyshev.cuh
Nyrio: Should we not replace all these calls with the standard equivalents, e.g `myAbs` becomes `std::abs` which works on both host and device? The current definition of `myAbs` will compile with nvcc but not gcc.

cpp/include/raft/core/operators.hpp
```diff
@@ -0,0 +1,292 @@
...
+
+  template <typename InT>
+  constexpr RAFT_INLINE_FUNCTION auto operator()(InT a) const
+  {
+    return composed_op(a, scalar);
+  }
+};
+
+template <typename Type>
+using scalar_add_op = scalar_op<Type, add_op>;
```
achirkin, line 226: What would you think about renaming these `scalar_xxx_op` into something like `xxx_const_op`? I see the intuition of using the word scalar since we're using these operators mostly in kernels that map vectors and matrices. Yet, if we look at the `operators.hpp` as an independent piece of the api, the word `const` seems to fit better, since all of the operators here are, in fact, defined on scalars :) .
I'm not insisting on this, though.

achirkin: Then, perhaps, it would also make sense to rename and slightly generalize `scalar_op` to something like this:
```C++
template <size_t ArgPos, typename ConstT, typename MultiArgOpT>
struct plug_const_op {...}
```
(or just hide it into `detail` namespace)

Nyrio: When considering the operation at the level of the primitive you're using it for, you have a vector or matrix on the left hand, and a scalar on the right hand. I think the term scalar is well understood in that context.

Of course, it's a matter of personal preference and mine is to keep the `scalar` names, since there is also a `const_op` that returns a constant no matter the input. But I don't have a strong opinion on this, so if others weigh in on this I might change my mind :)

achirkin: > When considering the operation at the level of the primitive you're using it for, you have a vector or matrix on the left hand, and a scalar on the right hand. I think the term scalar is well understood in that context.

Sure, but my point here is that, since this is going to be a part of public api, the users are not obliged to pass these operators to `unary_op` , `matrix_vector_op`, etc. They could pass it somewhere less obvious, or technically even use them on the host side directly.

Nyrio: I don't like the idea that these ops could be used directly, I think they should only be used as arguments to functor-based prims.

```c++
c = raft::add{}(a, b);  // don't do this!
c = a + b;              // do this
```

cpp/include/raft/distance/detail/chebyshev.cuh
cjnolet: My thinking was just to wrap the functions w/ the `raft::` namespace (even if just calling down to the std::) to keep the APIs consistent and provide a facade to centralized future changes as needed. So long as there's no perf impact for doing so.

Nyrio:
@achirkin I integrated your improvements of the compose ops, adding default constructors when all op types are default-constructible.

Regarding `modernize-use-trailing-return-type`, I don't think that the whole codebase follows that convention yet, and it's probably worth discussing whether we should. There are cases where it's useful (e.g to use `decltype` in the return type) but also cases where it's unnecessarily verbose. If you want, you can have a chat about this with @MatthiasKohl who [disabled this rule in cugraph-ops](https://github.com/rapidsai/cugraph-ops/blob/branch-23.02/.clang-tidy#L108).

One example where I find this rule pretty stupid is:

```c++
auto f(...) -> auto
{
  ...
}
```

achirkin:
> One example where I find this rule pretty stupid is:

You don't need to type that, clang-tidy won't complain if you use auto on the left side with nothing on the right side :-D

cpp/include/raft/distance/detail/chebyshev.cuh
Nyrio: I see. One objection that I have about putting them in the same header as the functors, though, is that they achieve different purposes: the functors are meant to be callables used as lambdas / passed to prims, not to be called directly. `myXxx` (->`raft::xxx`) is used directly as a device/type-agnostic wrapper around some math functions. It could be confusing to mix them.

Nyrio:
> You don't need to type that, clang-tidy won't complain if you use auto on the left side with nothing on the right side :-D

Mmmm, it did though 😄

```
operators.hpp:102:39: warning: use a trailing return type for this function [modernize-use-trailing-return-type]
  constexpr RAFT_INLINE_FUNCTION auto operator()(const T1& a, const T2& b) const
```

cpp/include/raft/distance/detail/chebyshev.cuh
cjnolet: That sounds fine to me. What about a header like `raft/core/math.hpp`? Again, we don't have to do it in this PR, but it's nice seeing all of these APIs consolidated and put in more obvious places that are easier to document.

achirkin:
Which version of llvm/clang-tidy do you use? Mine is `11.1.0` and it does not complain about the `auto` cases.

cpp/include/raft/distance/detail/chebyshev.cuh
achirkin: Would it make sense to replace all cases with `std::xxx` alternatives where possible, as @Nyrio suggested, and only add new functions to raft when there are no suitable standard alternatives? I'm afraid otherwise a user may be confused as to why they should prefer raft version over std for some basic primitives. Or, as a compromise, we can just re-export some of these primitives (e.g. `namespace raft { using std::max; }`).

cjnolet: I'm definitely not against re-exporting into the raft namespace but my suggestion is mostly conditional on whether there's ever going to be a perf hit from using std vs some custom accelerated intrinsic or function that invokes CUDA-specific instructions. In that case, we could still export the functions into the RAFT namespace but explicltly define new raft variants when necessary. Overall, I'm trying to provide as consistent an experience as possible in the raft namespace to prevent users from having to refer to the docs to figure out which things are covered under RAFT and which things are not. not saying we need to export everything in std math, but the things which we frequently use would be useful to keep in the `raft::` namespace even if only as a facade to enable us to centralize changes in the future.

Nyrio: > whether there's ever going to be a perf hit from using std vs some custom accelerated intrinsic

With nvcc, writing `std::abs` uses the `std` version on the host side, and the device intrinsics on the device side. So no perf hit compared to the current code. (with Clang I've seen compilation errors because it says we're trying to use host functions in device code). Artem's suggestion is simply an easier way to expose these names as part of the `raft` namespace without unnecessary wrappers, but that still gives up the option to write different variants later.

Nyrio:
> Which version of llvm/clang-tidy do you use? Mine is 11.1.0 and it does not complain about the auto cases. Also, try passing -extra-arg=-std=c++17 to clang-tidy if you don't.

12.0.0 and I am passing `-extra-arg=-std=c++17`.

achirkin:
Oh, I can replicate your problem now. Add `-p cpp/build` (the folder where `compile_commands.json` is) to the parameters of clang-tidy as well. This will make it aware of actual types and the warning won't show up anymore. The full command line I have in my vscode setup:
```
clang-tidy  -extra-arg=-std=c++17 -p ${fileWorkspaceFolder}/cpp/build ${file}
```

Nyrio:
I changed the return types to `auto` to silence those warnings, but there is another problematic clang-tidy warning that I absolutely disagree with:

```
operators.hpp:276:29: warning: use default member initializer for 'ops' [modernize-use-default-member-init]
  const std::tuple<OpsT...> ops;
                            ^
```

Here, it might not be possible to default-construct `ops`, that's why the default constructor is protected with `enable_if`. Or am I misunderstanding something?

cpp/include/raft/distance/detail/chebyshev.cuh
Nyrio: It's probably better to make this change in a separate PR.

tfeher (changes requested):
Thanks Louis for this work, it is great to see these helper function used uniformly!

Looks great overall, and the main issue I see is adding some comments to the distance methods about potential danger when low precision types are used.

cpp/include/raft/distance/detail/cosine.cuh
```diff
@@ -248,10 +247,13 @@ void cosineAlgo1(Index_ m,
   InType* row_vec = workspace;
   if (pA != pB) {
     row_vec += m;
-    raft::linalg::rowNorm(col_vec, pA, k, m, raft::linalg::L2Norm, isRowMajor, stream, norm_op);
-    raft::linalg::rowNorm(row_vec, pB, k, n, raft::linalg::L2Norm, isRowMajor, stream, norm_op);
+    raft::linalg::rowNorm(
```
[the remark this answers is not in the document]
Nyrio: Well, this code doesn't compile at all if `AccType != InType`.

```c++
AccType* workspace,
...
InType* col_vec = workspace;
```

Since it affects the other distance functions and it's not related to this PR, I'll open a separate issue rather than adding comments.

cpp/include/raft/core/operators.hpp
Nyrio: Alright, I am outnumbered. I've renamed accordingly.

Nyrio:
I've added full test coverage of the operators on both host and device.
I had to move some utils from `test_utils.h` to a new file `test_utils.cuh` in the process to compile host-only tests with a host compiler.

cpp/include/raft/core/operators.hpp
```diff
@@ -0,0 +1,354 @@
...
+ *
+ * @tparam OpsT Any number of operation types.
+ */
+template <typename... OpsT>
+struct compose_op {
+  const std::tuple<OpsT...> ops;
+
+  template <typename TupleT = std::tuple<OpsT...>,
+            typename CondT  = std::enable_if_t<std::is_default_constructible_v<TupleT>>>
+  constexpr compose_op()
```
achirkin, line 281: It's just occurred to me that the two constructors would clash if we try to initialize `compose_op` with zero argument operators. For the sake of completeness it could be a proper identity operator otherwise. Do we really need the default constructor here at all?

achirkin (approved):
Thanks @Nyrio, for the tests! Looks good to me. Here's just one comment that does not really need to be addressed.

cpp/include/raft/core/operators.hpp
Nyrio: I don't think that zero operators is a valid use case for this structure, and I think that this default constructor adds value. We could maybe add a `static_assert` for the sake of clear error reporting...

achirkin: We're surely talking about very hypothetical things now, but one could imagine calling this struct by another template function in the user code, where zero arguments (functions) means identity. It makes a good monoid of unary functions :)

Nyrio: Surprisingly, the structure seems to compile just fine when `Args...` is empty. The second constructor is used. But if I comment out the second constructor, the default one is used. It doesn't see this as multiple definitions.

Nyrio: According to the [Overload resolution](https://en.cppreference.com/w/cpp/language/overload_resolution) rules, section _Best viable function_:

> 4) or, if not that, F1 is a non-template function while F2 is a template specialization

The non-template constructor unambiguously takes precedence over the template one.

Nyrio:
I don't think the CI failures have much to do with my changes.

```
/opt/conda/envs/rapids/bin/gtests/libraft/CLUSTER_TEST:
error while loading shared libraries: libnvJitLink.so.12: cannot open shared object file: No such file or directory
```

robertmaynard:
> I don't think the CI failures have much to do with my changes.
> 
> ```
> /opt/conda/envs/rapids/bin/gtests/libraft/CLUSTER_TEST:
> error while loading shared libraries: libnvJitLink.so.12: cannot open shared object file: No such file or directory
> ```

Something is pulling in and linking to CUDA 12. libnvJitLink is a new library introduced in 12.

Nyrio:
Perhaps this?:

```
libcusparse:      12.0.0.76-0                              nvidia
```

robertmaynard:
> Perhaps this?:
> 
> ```
> libcusparse:      12.0.0.76-0                              nvidia
> ```

Yes that would do it. @cjnolet Is the presumption we wait on GHA or is there an open PR that fixes the current env?

cjnolet:
@robertmaynard, the plan is to submit the fix that @dantegd has been testing in cuml to pin the libcusolver and libcusparse packages so the offending cuda 12 binary is never installed/linked.

cjnolet:
rerun tests

ajschmidt8:

cjnolet (approved):
LGTM

cjnolet:
@gpucibot merge

PR #1052, opened 2022-11-29
ANN refinement Python wrapper

tfeher:
This PR adds Python wrapper for the ANN refinement method. Refinement can work both with device dataset and host dataset.

### DIFF
--- a/python/pylibraft/pylibraft/common/mdspan.pxd
+++ b/python/pylibraft/pylibraft/common/mdspan.pxd
@@ -43,3 +43,14 @@ cdef extern from "raft/core/device_mdspan.hpp" \
     cdef device_matrix_view[ElementType, IndexType, LayoutPolicy] \
         make_device_matrix_view[ElementType, IndexType, LayoutPolicy](
             ElementType* ptr, IndexType n_rows, IndexType n_cols)
+
+
+cdef extern from "raft/core/host_mdspan.hpp" \
+        namespace "raft" nogil:
+
+    cdef cppclass host_matrix_view[ElementType, IndexType, LayoutPolicy]:
+        pass
+
+    cdef host_matrix_view[ElementType, IndexType, LayoutPolicy] \
+        make_host_matrix_view[ElementType, IndexType, LayoutPolicy](
+            ElementType* ptr, IndexType n_rows, IndexType n_cols)
--- a/python/pylibraft/pylibraft/common/mdspan.pxd
+++ b/python/pylibraft/pylibraft/common/mdspan.pxd
@@ -1,0 +1,43 @@
+#
+# Copyright (c) 2022, NVIDIA CORPORATION.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# cython: profile=False
+# distutils: language = c++
+# cython: embedsignature = True
+# cython: language_level = 3
+
+import pylibraft.common.handle
+
+from cython.operator cimport dereference as deref
+
+cdef extern from "raft/thirdparty/mdspan/include/experimental/__p0009_bits/layout_stride.hpp" \
+    namespace "std::experimental":
+    cdef cppclass layout_right:
+        pass
+
+
+cdef extern from "raft/core/mdspan_types.hpp" \
+        namespace "raft":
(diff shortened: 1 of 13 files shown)


tfeher:
Todo
- [ ] support refinement using host dataset

python/pylibraft/pylibraft/common/mdspan.pxd
```diff
@@ -0,0 +1,55 @@
...
+# distutils: language = c++
+# cython: embedsignature = True
+# cython: language_level = 3
+
+import pylibraft.common.handle
+
+from cython.operator cimport dereference as deref
+
+
+cdef extern from "raft/thirdparty/mdspan/include/experimental/__p0009_bits/layout_stride.hpp" namespace "std::experimental":  # noqa: E501
```
cjnolet, line 26: cc @benfred for awareness. This looks very similar to https://github.com/rapidsai/raft/pull/1016 but keeps the `LayoutPolicy`, which will prove to be useful generally.

cjnolet: The two `pxd` files are very similar otherwise, so I don't think there's much concern for conflicts across the two.

python/pylibraft/pylibraft/neighbors/refine.pyx
```diff
@@ -0,0 +1,461 @@
...
+    if k is None:
+        if indices is not None:
+            k = indices.__array_interface__["shape"][1]
+        elif distances is not None:
+            k = distances.__array_interface__["shape"][1]
+        else:
+            raise ValueError("Argument k must be specified if both indices "
+                             "and distances arg is None")
+
+    n_queries = queries.__array_interface__["shape"][0]
```
cjnolet, line 407: it would be nice to see a corresponding wrapper for host functions (e.g. `ai_wrapper` maybe?) just to keep the APIs similar between device and host. We don't need to do that in this PR, but it would certainly clean this up a tad and minimize potential errors from things like string typos.

python/pylibraft/pylibraft/neighbors/refine.pyx
```diff
@@ -0,0 +1,461 @@
...
+    else:
+        return _refine_host(dataset, queries, candidates, k, indices,
+                            distances, metric, handle)
+
+
+def _refine_device(dataset, queries, candidates, k, indices, distances,
+                   metric, handle):
+    cdef handle_t* handle_ = <handle_t*><size_t>handle.getHandle()
+
+    cdef device_matrix_view[float, uint64_t, row_major] dataset_view_fp32
```
cjnolet, line 308: It would be a little cleaner just to create these on the fly when invoking `c_refine`. (Cython makes this unnecessarily messy because it doesn't allow defining things within loops or conditionals.

python/pylibraft/pylibraft/neighbors/refine.pyx
```diff
@@ -0,0 +1,461 @@
...
+    candidates : array interface compliant matrix, shape (n_queries, k0)
+        dtype [float, int8, uint64_t]
+    k : int
+        Number of neighbors to search (k <= k0). Optional if indices or
+        distances arrays are given (in which case their second dimension
+        is k).
+    indices :  Optional CUDA array interface compliant matrix shape
+                (n_queries, k), dtype uint64_t. If supplied, neighbor
+                indices will be written here in-place. (default None)
+        Supported dtype [float, int8, uint8]
```
cjnolet, line 239: Should the supported dtypes be for the distances?

cjnolet (changes requested):
I think this looks good. Just a couple very minor things, really. If you think we should push them off to a follow-on, I don't mind approving.

python/pylibraft/pylibraft/neighbors/refine.pyx
```diff
@@ -0,0 +1,461 @@
...
+
+    Parameters
+    ----------
+    index_params : IndexParams object
+    dataset : array interface compliant matrix, shape (n_samples, dim)
+        Supported dtype [float, int8, uint8]
+    queries : array interface compliant matrix, shape (n_queries, dim)
+        Supported dtype [float, int8, uint8]
+    candidates : array interface compliant matrix, shape (n_queries, k0)
+        dtype [float, int8, uint64_t]
```
achirkin, line 231: These are candidate indices, right? Looks like you expose only `uint64_t` indices here.

python/pylibraft/pylibraft/neighbors/refine.pyx
achirkin: I believe this belongs nor here (`uint64_t`) neither to distances (`float` only).

python/pylibraft/pylibraft/neighbors/refine.pyx
tfeher: That is a good point. I will submit a follow up PR for this.

tfeher (commented):
Thanks @cjnolet and @achirkin for the reviews, I have updated the PR.

cjnolet (approved):

cjnolet:
@gpucibot merge

PR #1054, opened 2022-11-30
Specify correct global targets names to raft_export

robertmaynard:
The raft_nn and raft_distance targets are exported with the names nn and distance, therefore those are the names that need to be marked as GLOBAL.


cjnolet:
@gpucibot merge

PR #1057, opened 2022-11-30
switch mma instruction shape to 1684 from current 1688 for 3xTF32 L2/cosine kernel

mdoijade:
-- switch mma instruction shape to 1684 from current 1688 as it is always faster for all the inputs tried from DISTANCE_BENCH for L2 and cosine distances.
-- the speedup in best case is 1.37x, and at minimum it is 1.05x faster.

### DIFF
--- a/cpp/include/raft/distance/detail/pairwise_distance_gemm.h
+++ b/cpp/include/raft/distance/detail/pairwise_distance_gemm.h
@@ -66,7 +66,7 @@ struct PairwiseDistanceGemm {
   /// Warp-level tile size (concept: GemmShape)
   // This code section describes the size of MMA op
   using InstructionShape =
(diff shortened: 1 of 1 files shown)


mdoijade:
[cutlass_m16n8k4_vs_m16n8k8_3xtf32_pairwise_dist_kernel.xlsx](https://github.com/rapidsai/raft/files/10125379/cutlass_m16n8k4_vs_m16n8k8_3xtf32_pairwise_dist_kernel.xlsx)

attaching the L2/Cosine perf from DISTANCE_BENCH, for m16n8k4 vs   m16n8k8, for 3xTF32(fp32) kernel

tfeher (approved):
Thanks Mahesh, LGTM!

cjnolet:
@gpucibot merge

PR #1060, opened 2022-12-01
Add support for 64bit svdeig

lowener:
`raft::linalg::eigDC` already supports 64bit.
This will be useful for https://github.com/rapidsai/cuml/issues/4906


lowener:
rerun tests

PR #1065, opened 2022-12-01
Bifurcate Dependency Lists [skip-gpuci]

ajschmidt8:
This PR uses the [`rapids-dependency-file-generator`](https://github.com/rapidsai/dependency-file-generator/) to handle sourcing dependencies.

Similarly to https://github.com/rapidsai/rmm/pull/1073/ and https://github.com/rapidsai/cudf/pull/11674, this PR introduces a GitHub Action that enforces consistency between the new `dependencies.yaml` file and the generated conda environment for developers.


cjnolet (approved):
LGTM. Thanks AJ!

dependencies.yaml
```diff
@@ -0,0 +1,124 @@
...
+          - cmake>=3.23.1,!=3.25.0
+          - cuda-python >=11.7.1,<12.0
+          - cython>=0.29,<0.30
+          - ninja
+          - scikit-build>=0.13.1
+      - output_types: [conda]
+        packages:
+          - c-compiler
+          - cxx-compiler
+          - rapids-build-env=23.02.*
```
bdice, line 36: @ajschmidt8 Did you say this package was no longer needed on a different PR?

dependencies.yaml
```diff
@@ -0,0 +1,124 @@
...
+          - matrix:
+              arch: aarch64
+            packages:
+              - gcc_linux-aarch64=9.*
+              - sysroot_linux-aarch64==2.17
+  develop:
+    common:
+      - output_types: [conda, requirements]
+        packages:
+          - clang=11.1.0
```
bdice, line 54: I think we can drop clang/clang-tools, which means the entire `develop` list could be deleted. I don't think these are used anywhere. clang-format (the original reason this was added, afaik) is now handled via pre-commit.

dependencies.yaml
```diff
@@ -0,0 +1,124 @@
...
+      - output_types: [conda, requirements]
+        packages:
+          - breathe
+      - output_types: [requirements]
+        packages:
+          - sphinx_markdown_tables
+      - output_types: [conda]
+        packages:
+          - doxygen>=1.8.20
+          - rapids-doc-env=23.02.*
```
bdice, line 85: Likewise, is this still needed?

dependencies.yaml
```diff
@@ -0,0 +1,124 @@
...
+          - sphinx-markdown-tables
+  notebook:
+    common:
+      - output_types: [conda]
+        packages:
+          - rapids-notebook-env=23.02.*
```
bdice, line 91: Still needed?

dependencies.yaml
```diff
@@ -0,0 +1,124 @@
...
+    common:
+      - output_types: [conda]
+        packages:
+          - rmm=23.02.*
+          - dask==2022.11.1
+          - distributed==2022.11.1
+          - ucx>=1.13.0
+          - ucx-py=0.30.*
+          - ucx-proc=*=gpu
+          - libfaiss>=1.7.0
```
bdice, line 117: I noticed a really annoying issue recently in rapids-compose. It merges the dependency lists / conda channels, and ends up placing the `pytorch` channel (which is used in cudf tests) above `conda-forge`. This in turn means that `libfaiss` is pulled from `pytorch` and it ignores the `faiss-proc=*=cuda` package below, which only affects `libfaiss` from `conda-forge`. 🙃 Just something to be aware of. We may need to specify `libfaiss>=1.7.0=cuda*` or something like that, in addition to `faiss-proc=*=cuda`. Another solution would be to fix the conda-forge packages of `pytorch` and never use the `pytorch` channel for any dependencies.

bdice (changes requested):

dependencies.yaml
ajschmidt8: Yes. For this PR, the intention is just to replace the manually managed `conda/environment` files with the dfg. I figured we can remove these packages in the follow-up GHA PR.

If we remove them before, it could break the environment. In the GHA PR, we can remove them and then identify which dependencies we need to re-add after they're removed.

dependencies.yaml
ajschmidt8: I believe @cjnolet said that they do/plan to run `clang`.

I notice this script among others https://github.com/rapidsai/raft/blob/branch-23.02/cpp/scripts/run-clang-tidy.py

bdice: I think only `clang-tools` is required for providing the `clang-tidy` binary. No scripts are calling `clang` as far as I can tell.

dependencies.yaml
bdice: Okay, that sounds fine to me. We'll clean up later.

bdice (approved):
It seems like this is a valid port of the existing environment to use `dependencies.yaml`, but it leaves us with some cleanup tasks for later in the GitHub Actions migration.

dependencies.yaml
cjnolet: We are in the process of updating our code so we can support clang. I'd prefer to leave in the dependency in the meantime (we do have an experimental clang compile script in there that we're working on making build successfully).

bdice: Okay. We can resolve this.

dependencies.yaml
vyasr: I just ran into the same issue and fixed it in the old env files, but then I realized this PR was ready to go. I'll make a PR with the additional pinning, which is how I fixed it in my case and it worked fine.

vyasr: #1078

PR #1066, opened 2022-12-02
Consolidating mdspan cython defitions

cjnolet:
We recently had two PRs merged (one in 22.12 and another in 23.02) that independently added cython defitions around the mdspan factory functions. 

I've consolidated them in this PR and also exposed the remaining needed bits for specifying layouts in cython as well.


python/pylibraft/pylibraft/cpp/kmeans.pxd
```diff
@@ -33,6 +33,9 @@ from pylibraft.cpp.optional cimport optional
 cdef extern from "raft_distance/kmeans.hpp" \
         namespace "raft::cluster::kmeans::runtime":
 
+    ctypedef const float const_float
+    ctypedef const double const_double
```
benfred, line 37: do we need these typedefs here?   It doesn't seem like they are being used

cjnolet: Nope, those need to be removed

benfred (approved):
looks great!

cjnolet:
I've added these changes to 1055 already. Closing.

PR #1069, opened 2022-12-05
IVF-PQ: tweak launch configuration

achirkin:
Add one more criterion for reducing the block size when selecting the optimal launch configuration. This fixes a few benchmark cases where the performance has been slightly worse compared to the reference implementation.


achirkin:
NB: with this update, there still have been observed couple cases where the reference implementation is faster by 5-9%; in all of them the raft ivf-pq kernel was faster, and the slowdown came from the other part - select-top-k block kernel (selecting the candidate clusters for search).

tfeher (approved):

cjnolet:
@gpucibot merge

Issue #825, opened 2022-09-14
[FEA] Kmeans auto-find K


cjnolet:
Some important workflows require the ability to auto-find k using a measure of residual (spread of point distances across all centroids) and dispersion (spread of centroids in relation to each other).

This requires an objective which maximizes the cluster to cluster distances while minimizing the point to cluster spread as much as possible. We should be able to do this fairly easily, especially with our new consolidated k-means implementations.

PR #1070, opened 2022-12-06
Initial port of auto-find-k

cjnolet:
This is a port of https://github.com/rapidsai/cuml/issues/818 (originally from NVGraph) which uses the Calinski-Harabasz score 
to find the optimal value of k. 

Todo:
- [x] create histogram of cluster sizes
- [x] add googletests 
- [x] expose public API

Closes https://github.com/rapidsai/raft/issues/825

cc @jeaton32


cjnolet:
rerun tests

cpp/include/raft/cluster/detail/kmeans_auto_find_k.cuh
```diff
@@ -0,0 +1,219 @@
...
+                                                                labels.view(),
+                                                                residual,
+                                                                n_iter);
+
+      detail::countLabels(
+        handle, labels.data_handle(), clusterSizes.data_handle(), n, mid, workspace);
+
+      resultsView[mid]           = residual[0];
+      clusterDispertionView[mid] = raft::stats::cluster_dispersion(
+        handle, centroids_const_view, cluster_sizes_view, std::nullopt, n);
```
benfred, line 168: This block of code appears 3x in this function  (here, and for calculating left/right values above)- maybe we should refactor to remove the duplicate code? I think we could pull out to a lambda fairly easily, and then call that

like 
```c++
auto calculate_dispersion = [&](int idx) { 
... 
};
calculate_dispersion(mid);
```

benfred (commented):
This looks really good! 

Some minor comments:

cpp/include/raft/cluster/detail/kmeans_auto_find_k.cuh
cjnolet: Good point! I'll fix this.

benfred (approved):
lgtm!

PR #1077, opened 2022-12-08
Export NCCL dependency as part of raft::distributed.

vyasr:
This is a follow-up to #1032 to include NCCL in raft::distributed so that all communications libraries are now cleanly encapsulated in a single interface dependency, allowing us to remove it from the raft-dask Python build and simplifying the process for other consumers.


cjnolet:
@vyasr I'm really happy for this change. Can you also update the README.md and `docs/source/build.md`?

robertmaynard (approved):

PR #1078, opened 2022-12-08
Ensure that we get the cuda version of faiss.

vyasr:
If we try to use the current dependency spec in a situation where other channels are enabled that contain the non-CUDA version of FAISS we need to ensure that the CUDA-enabled package is downloaded.


bdice:
See this comment for additional details: https://github.com/rapidsai/raft/pull/1065#discussion_r1041470397

cjnolet:
We are actively working to remove the FAISS dependency (aiming for 23.02) so hopefully we shouldn't need to do this dance for much longer. 

cc @benfred @teju85 @dantegd

raydouglass (approved):

cjnolet:
@gpucibot merge

PR #1083, opened 2022-12-08
Exclude changelog from pre-commit spellcheck

benfred:
Exclude the changelog from the pre-commit codespell check- this is to prevent CI failures after release when the changelog is updated, as suggested here https://github.com/rapidsai/cudf/pull/12097#pullrequestreview-1180342650

This also exports the `__version__` for pylibraft / raft_dask, which was missing previously

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -97,6 +97,8 @@ repos:
         rev: v2.1.0
         hooks:
               - id: codespell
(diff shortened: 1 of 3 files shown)


benfred:
This is to avoid failures like https://github.com/rapidsai/raft/pull/1080 -

cjnolet (approved):

cjnolet:
@gpucibot merge

PR #1091, opened 2022-12-12
Separating more namespaces into easier-to-consume sections

cjnolet:


python/pylibraft/pylibraft/config.py
```diff
@@ -0,0 +1,37 @@
...
+    instead automatically convert the output to the specified output type,
+    depending on availability of the requested type.
+
+    Parameters
+    ----------
+    output : str or callable. str can be either
+             { "raft", "cupy", or "torch" }.
+             default = "raft". callable should accept
+             pylibraft.common.device_ndarray
+             as a single argument and return the converted type.
```
benfred, line 32: copying this suggestion from the other PR =) 

```suggestion
        Set output format for RAFT functions.

        Calling this function will change the output type of RAFT functions.
        By default RAFT returns a `pylibraft.common.device_ndarray` for arrays
        on GPU memory. Calling `set_output_as` allows you to have RAFT return
        arrays as cupy arrays or pytorch tensors instead. You can also have
        RAFT convert the output to other frameworks by passing a callable to
        do the conversion here.

        Notes
        -----
        Returning arrays in cupy or torch format requires you to install
        cupy or torch.

        Parameters
        ----------
        output : { "raft", "cupy", "torch" } or callable
            The output format to convert to. Can either be a str describing the
            framework to convert to, or a callable that accepts a
            device_ndarray and returns the converted type.
```

cjnolet: Doh! I just realized I never merged over the branch from the PR I closed

benfred (approved):
lgtm!

README.md
```diff
@@ -156,6 +156,18 @@ import torch
...
+When the corresponding library has been installed and available in your environment, this conversion can also be done automatically by all RAFT compute APIs by setting a global configuration option:
+```python
+import pylibraft.config
+pylibraft.config.set_output_as("cupy")  # All compute APIs will return cupy arrays
+pylibraft.config.set_output_as("torch") # All compute APIs will return torch tensors
+```
+
+You can also specify a `callable` that accepts a `pylibraft.common.device_ndarray` and performs a custom conversion. The following example converts all output to `numpy` arrays:
+```python
+pylibraft.config.set_output_as(lambda device_ndarray: return device_ndarray.copy_to_host())
```
dantegd, line 168: This is awesome

cpp/include/raft/core/cublas_macros.hpp
```diff
@@ -108,6 +122,9 @@ inline const char* cublas_error_to_string(cublasStatus_t err)
     }                                                                \
   } while (0)
 
+/**
+ * @}
+ */
```
dantegd, line 127: Unrelated but was wondering if we have a gh issue for the rename mentioned just below?

docs/source/cpp_api/neighbors_ball_cover.rst
```diff
@@ -0,0 +1,19 @@
+Random Ball Cover
+=================
+
+This page provides C++ class references for the publicly-exposed elements of the neighbors package.
```
dantegd, line 4: Was wondering, for all cases, why is it necessary to repeat this string in all the sub `.rst` files of each group?

dantegd (approved):
Looks fantastic!

docs/source/cpp_api/neighbors_ball_cover.rst
cjnolet: It's being copied for now only so that the pages are consistent, but I have it as a personal TODO to go through these and customize each brief heading description for both the top-level and subsections. (I could keep adding to this PR but it's so massive already).

cjnolet: I just removed the headings from all the subsections.

cjnolet:
@gpucibot merge

cjnolet:
@gpucibot merge

PR #1093, opened 2022-12-13
Adding auto-conversion configuration to pylibraft

cjnolet:
This enables a user who might already have cupy and/or pytorch installed to be able to set a configuration globally in raft to automatically convert `pylibraft.common.device_ndarray` to corresponding pytorch or cupy arrays through cuda array interface. Thist just makes the interop much easier than having to document for the users how to convert between each individual library. 

It would be nice if in the the future we also either added dlpack support or provided a conversion strategies for JAX and tensorflow.

### DIFF
--- a/python/pylibraft/pylibraft/config.py
+++ b/python/pylibraft/pylibraft/config.py
@@ -21,17 +21,26 @@ class config:
     @classmethod
     def set_output_as(self, output):
         """
-        RAFT functions which normally return outputs with memory on device will
-        instead automatically convert the output to the specified output type,
-        depending on availability of the requested type.
+        Set output format for RAFT functions.
+
+        Calling this function will change the output type of RAFT functions.
+        By default RAFT returns a `pylibraft.common.device_ndarray` for arrays
+        on GPU memory. Calling `set_output_as` allows you to have RAFT return
+        arrays as cupy arrays or pytorch tensors instead. You can also have
+        RAFT convert the output to other frameworks by passing a callable to
+        do the conversion here.
+
+        Notes
+        -----
+        Returning arrays in cupy or torch format requires you to install
+        cupy or torch.
 
         Parameters
         ----------
-        output : str or callable. str can be either
-                 { "raft", "cupy", or "torch" }.
-                 default = "raft". callable should accept
-                 pylibraft.common.device_ndarray
-                 as a single argument and return the converted type.
+        output : { "raft", "cupy", "torch" } or callable
+            The output format to convert to. Can either be a str describing the
+            framework to convert to, or a callable that accepts a
+            device_ndarray and returns the converted type.
         """
         if output not in SUPPORTED_OUTPUT_TYPES and not callable(output):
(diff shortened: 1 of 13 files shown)


python/pylibraft/pylibraft/config.py
```diff
@@ -0,0 +1,38 @@
...
+        instead automatically convert the output to the specified output type,
+        depending on availability of the requested type.
+
+        Parameters
+        ----------
+        output : str or callable. str can be either
+                 { "raft", "cupy", or "torch" }.
+                 default = "raft". callable should accept
+                 pylibraft.common.device_ndarray
+                 as a single argument and return the converted type.
```
benfred, line 34: I think the parameter formatting should be slightly different here  (https://numpydoc.readthedocs.io/en/latest/format.html#parameters) - and docstrings should start with a summary string on the first line

I also think the description text here could be a bit clearer.

How about something like :

```suggestion
        Set output format for RAFT functions.

        Calling this function will change the output type of RAFT functions.
        By default RAFT returns a `pylibraft.common.device_ndarray` for arrays
        on GPU memory. Calling `set_output_as` allows you to have RAFT return
        arrays as cupy arrays or pytorch tensors instead. You can also have
        RAFT convert the output to other frameworks by passing a callable to
        do the conversion here.

        Notes
        -----
        Returning arrays in cupy or torch format requires you to install
        cupy or torch.

        Parameters
        ----------
        output : { "raft", "cupy", "torch" } or callable
            The output format to convert to. Can either be a str describing the
            framework to convert to, or a callable that accepts a
            device_ndarray and returns the converted type.
``` 
?

python/pylibraft/pylibraft/test/test_config.py
```diff
@@ -0,0 +1,54 @@
...
+        return device_ndarray.empty((m, n))
+    elif t == tuple:
+        return device_ndarray.empty((m, n)), device_ndarray.empty((m, n))
+    elif t == list:
+        return [device_ndarray.empty((m, n)), device_ndarray.empty((m, n))]
+
+
+@pytest.mark.parametrize(
+    "out_type",
+    [["cupy", cupy.ndarray], ["raft", pylibraft.common.device_ndarray]],
```
benfred, line 36: Might be worth testing out the `callable` path here too?  I think we could test out setting the output as numpy pretty easy:

```suggestion
    [["cupy", cupy.ndarray], ["raft", pylibraft.common.device_ndarray], [lambda arr: arr.copy_to_host(), np.ndarray]],
```

(note that I haven't tested this out at all - and I'd be wary of applying the suggestion directly =)

benfred (commented):
The change here looks good - and I really like the functionality that this brings us . 

Some minor comments here:

python/pylibraft/pylibraft/test/test_config.py
cjnolet: Looks like this works!

cjnolet:
Closing this in favor of #1091

PR #1094, opened 2022-12-13
Fix race condition in `raft::random::discrete`

Nyrio:
The stream needs to be passed to `cub::DeviceScan`.


cjnolet:
rerun tests

cjnolet:
@gpucibot merge

PR #1098, opened 2022-12-14
Cleanup faiss includes

benfred:
This mainly removes un-needed faiss includes - but also removes the `KeyValueWarpSelect` function which isn't being used anywhere, and replaces `faiss::gpu::KeyValuePair` with `cub::KeyValuePair` in warp_select_faiss.cuh


cpp/include/raft/spatial/knn/detail/warp_select_faiss.cuh
```diff
@@ -17,41 +17,7 @@
...
-  {
-  }
-
-  /// Inequality operator
-  __host__ __device__ __forceinline__ bool operator!=(const KeyValuePair& b)
-  {
-    return (value != b.value) || (key != b.key);
-  }
-};
+using cub::KeyValuePair;
```
cjnolet, line 20: Doesn't need to hold up this PR but it would be nice to use raft::KeyValuePair here for consistency. I've been converting the cub::keyvaluepair to the raft version everywhere for consistency except where the cub version is absolutely necessary.

cjnolet (approved):
Just one little thing and we don't have to hold up the PR for it.

cjnolet:
@gpucibot merge

PR #1099, opened 2022-12-14
Use `GenPC` (Permuted Congruential) as the default random number generator everywhere

Nyrio:
`GenPC` is already the default generator used by `RngState` when not passed to the constructor: https://github.com/rapidsai/raft/blob/a9e060d354917114bb8baa5de9c55ef917f203af/cpp/include/raft/random/rng_state.hpp#L48-L52

But the default value for the `GeneratorType` remained `GenPhilox` in some prims, such as `make_blobs` and `make_regression`. PC is faster and more reliable, so it should be the default there too.

Also, whenever possible, the generator type shouldn't be hardcoded, it should either use the default (construct `RngState` with the seed only) or an argument (as is the case with `make_blobs` and `make_regression`).

_Note: this will effectively modify many test inputs, so be aware of that when comparing results prior to and following the change._

### DIFF
--- a/cpp/include/raft/random/detail/rng_impl_deprecated.cuh
+++ b/cpp/include/raft/random/detail/rng_impl_deprecated.cuh
@@ -39,7 +39,7 @@ namespace detail {
 
 class RngImpl {
  public:
-  RngImpl(uint64_t seed, GeneratorType _t = GenPhilox)
+  RngImpl(uint64_t seed, GeneratorType _t = GenPC)
     : state{seed, 0, _t},
       type(_t),
       // simple heuristic to make sure all SMs will be occupied properly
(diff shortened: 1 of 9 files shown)


cpp/include/raft/random/detail/rng_impl_deprecated.cuh
```diff
@@ -39,7 +39,7 @@ namespace detail {
 
 class RngImpl {
  public:
-  RngImpl(uint64_t seed, GeneratorType _t = GenPhilox)
```
MatthiasKohl, line 42: not sure I would change the deprecated interface, since that might affect other libraries still using this. Those libraries should move on eventually anyway, and then they'll have the new default, plus we'll remove this interface at some point.

Nyrio: Sure, just reverted the change on the deprecated interface.

cjnolet (approved):

cjnolet:
@gpucibot merge

PR #1100, opened 2022-12-14
Speedup `make_blobs` by up to 2x by fixing inefficient kernel launch configuration

Nyrio:
The kernel generates two elements per iteration and attempts to write the second element with an offset equal to the grid stride. However, the grid stride is currently computed to be greater than the length of the generated array, so this second value is never used. By using a grid stride of half the array size, we speed up the kernel by nearly 2x in some cases (see perf charts in the PR comments).

_Note: this will effectively modify many test inputs, so be aware of that when comparing results prior to and following the change._

### DIFF
--- a/cpp/bench/random/make_blobs.cu
+++ b/cpp/bench/random/make_blobs.cu
@@ -25,6 +25,12 @@ struct make_blobs_inputs {
   bool row_major;
 };  // struct make_blobs_inputs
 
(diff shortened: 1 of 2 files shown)


Nyrio:
This is a before/after benchmark:

2022-12-14_make_blobs

cjnolet (approved):

cjnolet:
@gpucibot merge

PR #1103, opened 2022-12-14
Update handle docstring

dantegd:


benfred (commented):
Thanks for the fix! It looks like I missed testing out the pylibraft.common module when I added the doctests =(

Can you also add a line here  https://github.com/rapidsai/raft/blob/branch-23.02/python/pylibraft/pylibraft/test/test_doctests.py#L95 so that we include this docstring in the tests?

dantegd:
@benfred just added the pylibraft.common objects to the test (Handle and Stream) indeed :)

cjnolet:
@gpucibot merge

PR #1076, opened 2022-12-07
Add GitHub Actions Workflows.

bdice:
This PR adds GitHub Actions workflows to `raft`.

**Wait until after the 22.12 release is complete (and potentially until other repos are ready to migrate to GHA) to merge.**

### Task list

Coverage required for this PR:
- [x] C++ tests
  - [x] Two test failures on ARM? `Result/18: SWoRTests/SWoRTestD; /opt/conda/conda-bld/work/cpp/test/random/sample_without_replacement.cu:269`
- [x] Python tests
- [x] Codecov
- [x] Style checks

Future work:
- Build Metrics Report (copy from cudf when it is added)
- JUnit output annotations? https://github.com/marketplace/actions/junit-report-action
- Reduce exit code handling verbosity (use global variable with `trap` function)
- Set up `rapids-dependency-file-generator` pre-commit hook
- Docs builds

PR #1109, opened 2022-12-17
Fixing the sample-without-replacement test failures

vinaydes:
The sample-without-replacement test fails on A100 when `largeWeight` parameter is enabled, observed in PR #1076. The failure is caused as the sample with largest weight might not always be the first element getting selected in the sampled array. To fix this issue, this PR proposes to search the sample with large weight in the first three elements of sampled array.

### DIFF
--- a/cpp/test/random/sample_without_replacement.cu
+++ b/cpp/test/random/sample_without_replacement.cu
@@ -214,7 +214,11 @@ const std::vector<SWoRInputs<float>> inputsf = {{1024, 512, -1, 0.f, GenPhilox, 
         << "repeated index @i=" << i << " idx=" << val;                                            \
       occurrence.insert(val);                                                                      \
     }                                                                                              \
-    if (params.largeWeightIndex >= 0) { ASSERT_EQ(h_outIdx[0], params.largeWeightIndex); }         \
+    if (params.largeWeightIndex >= 0) {                                                            \
+      ASSERT_TRUE((h_outIdx[0] == params.largeWeightIndex) ||                                      \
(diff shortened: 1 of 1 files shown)


dantegd (approved):
Fix LGTM

cjnolet:
@gpucibot merge

cpp/test/random/sample_without_replacement.cu
```diff
@@ -214,7 +214,11 @@ const std::vector<SWoRInputs<float>> inputsf = {{1024, 512, -1, 0.f, GenPhilox,
         << "repeated index @i=" << i << " idx=" << val;                                            \
       occurrence.insert(val);                                                                      \
     }                                                                                              \
-    if (params.largeWeightIndex >= 0) { ASSERT_EQ(h_outIdx[0], params.largeWeightIndex); }         \
+    if (params.largeWeightIndex >= 0) {                                                            \
+      ASSERT_TRUE((h_outIdx[0] == params.largeWeightIndex) ||                                      \
+                  (h_outIdx[1] == params.largeWeightIndex) ||                                      \
+                  (h_outIdx[3] == params.largeWeightIndex));                                       \
```
benfred, line 220: This looks like you're checking against the 1st, 2nd and 4th element's here - is this by design? 

Based off the PR description i would have expected this line to check the 3rd element instead of the 4th

vinaydes: Good catch. I was checking first five elements first and the deleted wrong indices. Let me correct that. Thanks.

PR #1111, opened 2022-12-19
Decoupling raft handle from underlying resources 

cjnolet:
This implements a design idea a few of us have been kicking around for a little while now to help decouple underlying resources from the raft handle and also allow users to never have to explicitly include headers for resources that are never used (such as cublas, cusolver, cusparse, comms, etc...). 

This effectively breaks the existing raft::handle_t into separate headers for the various resources it contains, providing functions that can be individually included and invoked on a `raft::resources`. This still allows us to write something like a `raft::device_resources` (and also allows us to maintain API compatibility in the meantime by backing the existing `raft::handle_t` with a `raft::resources`. 

One of the major goals of this PR is to also enable a handle to be used outside of just cuda resources and to allow for unused resources to not need to be loaded nor compiled at all into user code downstream. 

Follow-on work after this PR will include:
1. Updating all of RAFT's public functions to accept `raft::resources` and using the individual resource accessors instead of assuming `device_resources` everywhere. 
2. Deprecating the `handle_t` in favor of the more explicit `device_resources`


cpp/include/raft/comms/detail/test.hpp
```diff
@@ -142,7 +142,9 @@ bool test_collective_reduce(const handle_t& handle, int root)
  */
 bool test_collective_allgather(const handle_t& handle, int root)
 {
+  printf("Getting comms_t from handle\n");
```
divyegala, line 145: To be removed

cpp/include/raft/comms/detail/test.hpp
```diff
@@ -142,7 +142,9 @@ bool test_collective_reduce(const handle_t& handle, int root)
...
   comms_t const& communicator = handle.get_comms();
+  printf("Done.\n");
```
divyegala, line 147: Same as above

cpp/include/raft/core/resource/thrust_policy.hpp
```diff
@@ -0,0 +1,64 @@
...
+  thrust_policy_resource(rmm::cuda_stream_view stream_view)
+    : thrust_policy_(std::make_shared<rmm::exec_policy>(stream_view))
+  {
+  }
+  void* get_resource() override { return thrust_policy_.get(); }
+
+  ~thrust_policy_resource() override {}
+
+ private:
+  std::shared_ptr<rmm::exec_policy> thrust_policy_;
```
divyegala, line 33: This does not need to be a `shared_ptr`

cpp/include/raft/core/resources.hpp
```diff
@@ -0,0 +1,107 @@
...
+      resources_.insert(std::make_pair(resource_type, factory->make_resource()));
+    }
+    return reinterpret_cast<res_t*>(resources_.at(resource_type).get()->get_resource());
+  }
+
+ private:
+  mutable std::mutex mutex_;
+  mutable std::unordered_map<resource::resource_type, std::shared_ptr<resource::resource_factory>>
+    factories_;
+  mutable std::unordered_map<resource::resource_type, std::shared_ptr<resource::resource>>
```
divyegala, line 104: Convert the `shared_ptr` to a `unique_ptr` since we never return the `shared_ptr` to the user, just the raw pointer. That means the lifetime of this resource is entirely managed by the creating container of this resource, and thus no extra bookkeeping needs to be done

divyegala (changes requested):

divyegala (approved):

wphicks (changes requested):
Really excited to see this change go in! Just one performance issue that I noticed, but otherwise it looks great.

cpp/include/raft/core/resource/sub_comms.hpp
```diff
@@ -0,0 +1,74 @@
...
+namespace raft::resource {
+class sub_comms_resource : public resource {
+ public:
+  sub_comms_resource() {}
+  void* get_resource() override { return &communicators_; }
+
+  ~sub_comms_resource() override {}
+
+ private:
+  std::unordered_map<std::string, std::shared_ptr<comms::comms_t>> communicators_;
```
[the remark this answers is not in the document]
cjnolet: That's fair. Looking at [this blog](http://scottmeyers.blogspot.com/2015/09/should-you-be-using-something-instead.html) from Scott Meyers, it looks like a linear search was actually the fastest for very small numbers of elements. That being said, an ordered vector and lower_bound doesn't take that much more effort so I think we should just use that.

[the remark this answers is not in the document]
benfred: Do we need to use a linear_search? Since the resource_type is a small integer - can't we just allocate a vector with the max number of resources, and then just index into it instead (trading off a small amount of memory space for lookup speed)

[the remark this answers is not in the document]
benfred: Do we need to use a linear_search? Since the resource_type is a small integer, we could just allocate a vector with the maximum size of the resource_type - and then index into that instead.

docs/source/developer_guide.md
```diff
@@ -8,11 +16,239 @@ The process for working on a CUDA/C++ feature which might span RAFT and one or m
...
+void ivf_pq_search(raft::resources const& res, raft::ivf_pq_params const&params, raft::ivf_pq_index const & index,
+raft::device_matrix<value_t, idx_t, raft::row_major> queries,
+raft::device_matrix<value_t, idx_t, raft::row_major> out_inds,
+raft::device_matrix<value_t, idx_t, raft::row_major> out_dists);
+}
+```
+
+### Other functions on state
+
+These guidelines also mean that it is the responsibility of C++ API to expose methods to load and store (aka marshalling) such a data structure. Further continuing the IVF-PQ example,  the following methods could achieve this:
```
dantegd, line 126:
```suggestion
These guidelines also mean that it is the responsibility of C++ API to expose methods to load and store (aka marshaling) such a data structure. Further continuing the IVF-PQ example,  the following methods could achieve this:
```

dantegd (approved):
Love the design and implementation

docs/source/developer_guide.md
cjnolet: Ha, this looks like a conflict between UK english vs US english. I personally prefer `marshalling` as the verb form of `marshal`. It just feels right to me. However, it formally looks like another spelling is indeed `marshaling`.

cpp/include/raft/core/resource/sub_comms.hpp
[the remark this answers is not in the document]
cjnolet: > Do we need to use a linear_search? Since the resource_type is a small integer, we could just allocate a vector with the maximum size of the resource_type - and then index into that instead.

I certainly don't mind this approach while we know that the number of potential resources could be quite small. My initial concern is when this object gets used more generically outside of the resource_type enum but we can revisit when we come to that point.

cpp/include/raft/core/resources.hpp
```diff
@@ -0,0 +1,107 @@
...
+      resources_.insert(std::make_pair(resource_type, factory->make_resource()));
+    }
+    return reinterpret_cast<res_t*>(resources_.at(resource_type).get()->get_resource());
+  }
+
+ private:
+  mutable std::mutex mutex_;
+  mutable std::unordered_map<resource::resource_type, std::shared_ptr<resource::resource_factory>>
+    factories_;
+  mutable std::unordered_map<resource::resource_type, std::unique_ptr<resource::resource>>
```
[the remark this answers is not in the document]
cjnolet: @wphicks @benfred, I've converted the `unordered_map` in `resources.hpp` to a vector and mplemented Ben's idea to size the vectors to the number of potential resources up front to speed up the lookups for now. These are indeed core places which are expected to be invoked many times over the lifespan of a typical applicaiton.

I'd like to hold off on changing the sub_comms resource, though. For one, the `get_subcomm` function is invoked very infrequently and when it is invoked, it's done only once or twice for the lifetime of an entire application. Also, given that we're using a variable sized key (`std::string`) to index the `unordered_map`, I'm not sure we're going to see much speedup from an ordered vector in that case. I'd like to investigate it further but it's not going to be a bottleneck like the `unordered_map` use in `resources.hpp`.

wphicks (approved):
Looks great!

cpp/include/raft/core/resources.hpp
```diff
@@ -0,0 +1,132 @@
...
+#include <algorithm>
+#include <mutex>
+#include <raft/core/logger.hpp>
+#include <string>
+#include <vector>
+
+namespace raft {
+
+template <typename T>
+using pair_res = std::pair<resource::resource_type, std::shared_ptr<T>>;
```
divyegala, line 28: These should be in the class body, and not at the namespace level so these alias names don't leak

cpp/include/raft/core/resource/cuda_stream.hpp
```diff
@@ -0,0 +1,95 @@
...
+  }
+  resource_type get_resource_type() override { return resource_type::CUDA_STREAM_VIEW; }
+  resource* make_resource() override { return new cuda_stream_resource(stream); }
+
+ private:
+  rmm::cuda_stream_view stream;
+};
+
+/**
+ * Load a cudaStream_t from a res (and populate it on the res
```
achirkin, line 59: Is the naming `cudaStream_t` intentional, or should it be `cuda_stream_view` here?

cpp/include/raft/core/resource/cuda_stream.hpp
```diff
@@ -0,0 +1,95 @@
...
+inline rmm::cuda_stream_view get_cuda_stream(resources const& res)
+{
+  if (!res.has_resource_factory(resource_type::CUDA_STREAM_VIEW)) {
+    res.add_resource_factory(std::make_shared<cuda_stream_resource_factory>());
+  }
+  return *res.get_resource<rmm::cuda_stream_view>(resource_type::CUDA_STREAM_VIEW);
+};
+
+/**
+ * Load a cudaStream_t from a res (and populate it on the res
```
achirkin, line 73: Same as above `cudaStream_t`/`cuda_stream_view`?

cpp/include/raft/core/resource/cuda_event.hpp
```diff
@@ -0,0 +1,64 @@
...
+ */
+class cuda_stream_sync_event_resource_factory : public resource_factory {
+ public:
+  resource_type get_resource_type() override { return resource_type::CUDA_STREAM_SYNC_EVENT; }
+  resource* make_resource() override { return new cuda_event_resource(); }
+};
+
+/**
+ * Load a cudaStream_t from a res (and populate it on the res
+ * if needed).
```
achirkin, line 52:
```suggestion
 * Load a stream-sync event from a res (and populate it on the res if needed).
 * NB: this is a "singleton" event used for synchronizing streams in the "singleton" stream pool in
 * the same resource.
```

I wonder though, would it make sense to somehow hide the stream_sync_event from the user, so it wouldn't interfere with the stream_pool functionality?

cpp/include/raft/core/resource/cuda_stream_pool.hpp
```diff
@@ -0,0 +1,173 @@
...
+ * @brief ask stream pool to wait on last event in main stream
+ */
+inline void wait_stream_pool_on_stream(const resources& res)
+{
+  printf("waiting on stream pooll\n");
+  printf("stream pool size: %ld\n", get_stream_pool_size(res));
+  cudaEvent_t event = get_cuda_stream_sync_event(res);
+  printf("After get_cuda_stream_sync_event\n");
+  RAFT_CUDA_TRY(cudaEventRecord(event, get_cuda_stream(res)));
+  printf("After cuda event record\n");
```
achirkin, line 168: Can we replace the printfs with the debug logging at this point?

achirkin (commented):
Spotted a few minor things while I was getting used to the new code.

cpp/include/raft/core/resource/cuda_event.hpp
cjnolet: While is this is considered part of the public APIs, it is not required that a user interact with the stream sync event directly. From their perspective, they'd just be invoking the "sync stream" function on the cuda_stream resource.

achirkin: My point is that it's public, and it's not clear that the user shouldn't use this event for other purposes. The user may decide to use this event  for other purposes to avoid the need to manage the event lifetime manually. This might interfere with the stream pool internals.

cjnolet: That's fair. I've moved the factory and accessor for the cuda stream sync event resource.

cpp/test/random/multi_variable_gaussian.cu
```diff
@@ -90,8 +91,10 @@ class MVGTest : public ::testing::TestWithParam<MVGInputs<T>> {
   {
   }
 
+ protected:
   void SetUp() override
   {
+    printf("Settingup test\n");
```
benfred, line 97: Is this a left-over debugging statement? 
```suggestion
```

cpp/test/random/multi_variable_gaussian.cu
```diff
@@ -100,10 +103,13 @@ class MVGTest : public ::testing::TestWithParam<MVGInputs<T>> {
     corr      = params.corr;
     tolerance = params.tolerance;
 
+    printf("Getting cublas handl\n");
```
benfred, line 106:
```suggestion
```

cpp/test/random/multi_variable_gaussian.cu
```diff
@@ -100,10 +103,13 @@ class MVGTest : public ::testing::TestWithParam<MVGInputs<T>> {
...
     auto cublasH   = handle.get_cublas_handle();
     auto cusolverH = handle.get_cusolver_dn_handle();
     auto stream    = handle.get_stream();
 
+    printf("Got handles\n");
```
benfred, line 111:
```suggestion
```

benfred (approved):
lgtm!

PR #1113, opened 2022-12-20
Add public API and tests for hierarchical balanced k-means

Nyrio:


cjnolet:
@Nyrio i use the "assignee" in the PRs to mark the author of the PR so it's more easy to search the PRs. If you'd like to request @tfeher and @achirkin for review, please use the "reviewers" section of the PR.

Nyrio:
> i use the "assignee" in the PRs to mark the author of the PR so it's more easy to search the PRs.

I'm not sure to understand: GitHub can also filter by author.

gh_author

cjnolet:
@Nyrio 

This is the method we've been using and it's been helpful thus far. Please request reviewers using the "Reviewers" section.

cpp/include/raft/core/operators.hpp
```diff
@@ -130,10 +130,10 @@ struct div_op {
 };
 
 struct div_checkzero_op {
-  template <typename Type>
-  constexpr RAFT_INLINE_FUNCTION auto operator()(const Type& a, const Type& b) const
+  template <typename T1, typename T2>
+  constexpr RAFT_INLINE_FUNCTION auto operator()(const T1& a, const T2& b) const
```
achirkin, line 134: I think, it would be healthier to use the `map_args_op{div_checkzero_op{}, identity_op{}, cast_op<T1>{}}` on the call site rather than changing the semantics here.

cpp/include/raft/core/operators.hpp
```diff
@@ -361,5 +372,24 @@ struct map_args_op {
...
+  }
+  constexpr write_only_op(ComposedOpT o) : composed_op{o} {}
+
+  template <typename OutT, typename... Args>
+  constexpr RAFT_INLINE_FUNCTION void operator()(OutT* out, Args&&... args) const
+  {
+    *out = composed_op(std::forward<Args>(args)...);
+  }
+};
+
```
achirkin, line 393: I see, this operator should be useful for `linalg::writeOnlyUnaryOp`. Yet, I'd argue against adding this, and using the `writeOnlyUnaryOp` in general. The name of that helper is a little misleading and passing/dereferencing a pointer is unnecessary in many use-cases.  I've been using that helper for a while, because I was too lazy to propose an alternative that wouldn't need dereferencing a pointer.

Do you know if there exists a convenient helper in thrust to fill a buffer using an unary function on an index range? If not, maybe it's time to add something like this to raft (or just leave the current use-case a lambda)?

achirkin (changes requested):
Thanks, @Nyrio, I'm looking forward to see this merged! Here are a few small comments on auxiliary changes while I'm looking through the main files.

cpp/include/raft/core/operators.hpp
achirkin: Oh, wait, I see many of the other binary operators are heterogeneous in their argument types as well. Unless we don't change all of them to use the same type for both arguments, it does not make much sense to hold off this change as well.
Maybe add the explicit type declaration `-> T1` to them, so that the users wouldn't need to try hard to remember the C++ arithmetics type inference rules?

cpp/include/raft/matrix/gather.cuh
```diff
@@ -44,13 +47,13 @@ namespace raft::matrix {
...
-            int map_length,
-            MatrixIteratorT out,
+template <typename InputIteratorT, typename MapIteratorT, typename OutputIteratorT, typename IndexT>
+void gather(const InputIteratorT in,
+            IndexT D,
+            IndexT N,
+            const MapIteratorT map,
+            IndexT map_length,
+            OutputIteratorT out,
             cudaStream_t stream)
```
achirkin, line 57: Not necessarily for this PR, but do you think it's feasible to implement a version of `gather` (and, maybe, `gather_if`) that takes the host pointer as the input `in`? e.g. load the `in` data into gpu in batches using `raft::copy` and then apply the `MapIteratorT` on it and write into `out` in a kernel.
I believe, it can be very handy for processing large datasets in algorithms like `ivf-pq`: copy a small part of a huge dataset with indices coming from random sampling (select the training set) or clustering (select the subset by the label).

cpp/include/raft/matrix/gather.cuh
```diff
@@ -30,10 +30,13 @@ namespace raft::matrix {
 /**
  * @brief  gather copies rows from a source matrix into a destination matrix according to a map.
  *
```
achirkin, line 32: Could you please add a small use-case example in the docs for each new function in the public namespace?

cpp/include/raft/matrix/gather.cuh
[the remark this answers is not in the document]
achirkin: Also, does `gather` compare well against `copy_selected` in terms of speed?

achirkin (changes requested):
Couple things regarding `gather`.

cpp/include/raft/core/operators.hpp
Nyrio: I prefer not to cast explicitly if you wouldn't when writing the operation in plain C++. And I think most people wouldn't cast an integer divisor to float when the dividend is float.

Regarding the explicit type declaration, I disagree, it can be either `T1` or `T2` based on the types. E.g if one is float and the other integer, the result is always float. Hence my writing `return T1{0} / T2{1};` to avoid making any assumption.

cpp/include/raft/core/operators.hpp
Nyrio: I agree we could have a write-only helper a bit more along the lines of how unary op is working, i.e the op returns the value instead of taking a pointer and writing the output in memory. As long as `writeOnlyUnaryOp` is in use, the helper `write_only_op` can be useful IMO.

And yes, I believe we can also simply use `thrust::generate`:

```c++
template <typename DerivedPolicy,
          typename ForwardIterator,
          typename Generator>
__host__ __device__ void generate(
    const thrust::detail::execution_policy_base< DerivedPolicy > & exec,
    ForwardIterator first,
    ForwardIterator last,
    Generator gen);
```

Nyrio: Actually [`thrust::tabulate`](https://nvidia.github.io/thrust/api/groups/group__transformations.html#function-tabulate) seems to do the job. `thrust::generate` is only for generators that don't take arguments (e.g constant, random)

Nyrio: Btw the documentation for `write_only_unary_op` is incorrect:

https://github.com/rapidsai/raft/blob/9944b3a8b83bfd6cd8298a73cd175298e168d264/cpp/include/raft/linalg/unary_op.cuh#L126-L127

We have the following options:

- Change the behavior of `write_only_unary_op` and/or `writeOnlyUnaryOp` to actually take a functor with the signature `InType func(const InType& val);` (breaking change)
- Create another helper with a different name wrapping around `thrust::tabulate` (or our own custom kernel)
- Stop using `writeOnlyUnaryOp` and use `thrust::tabulate` directly.

cc @cjnolet, what do you think about this?

cpp/include/raft/matrix/gather.cuh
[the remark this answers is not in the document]
Nyrio: I found a lot of issues with `matrix::gather`, from compilation errors to poor performance, so I opened a separate PR: #1126

Nyrio:
I fixed a bug where mapping functions could be applied on already mapped data, which is problematic if the mapping function isn't the identity when input/output types are the same.

cpp/include/raft/matrix/gather.cuh
Nyrio: I assume you were referring to `build_clusters` and `calc_centers_and_sizes` in the other header? I haven't added any new function to `gather.cuh`.

cpp/include/raft/cluster/kmeans_balanced_types.hpp
```diff
@@ -0,0 +1,51 @@
...
+struct kmeans_balanced_params {
+  /**
+   * Number of training iterations
+   */
+  uint32_t n_iters = 20;
+
+  /**
+   * Metric to use for distance computation.
+   */
+  raft::distance::DistanceType metric = raft::distance::DistanceType::L2Expanded;
```
cjnolet, line 37: K-means isn't really useful without an L2-based metric. We should probably mention the accepted `DistanceType`s in the docs here.

cpp/include/raft/cluster/kmeans_balanced.cuh
```diff
@@ -0,0 +1,341 @@
...
+ *                        datatype. If DataT == MathT, this must be the identity.
+ */
+template <typename DataT, typename MathT, typename IndexT, typename MappingOpT = raft::identity_op>
+void fit(handle_t const& handle,
+         kmeans_balanced_params const& params,
+         raft::device_matrix_view<const DataT, IndexT> X,
+         raft::device_matrix_view<MathT, IndexT> centroids,
+         MappingOpT mapping_op = raft::identity_op())
+{
+  logger::get(RAFT_NAME).set_level(params.verbosity);
```
cjnolet, line 60: Perhaps we should expect that the distance matches what is supported here too, and that k is a reasonable value (e.g. 0<k<n).

cpp/include/raft/spatial/knn/detail/ivf_flat_build.cuh
```diff
@@ -134,15 +134,18 @@ inline auto extend(const handle_t& handle,
...
-                                   new_labels.data(),
-                                   orig_index.metric(),
-                                   stream);
+  raft::cluster::kmeans_balanced_params kmeans_params;
+  kmeans_params.metric     = orig_index.metric();
+  auto new_vectors_view    = raft::make_device_matrix_view<const T, IdxT>(new_vectors, n_rows, dim);
+  auto orig_centroids_view = raft::make_device_matrix_view<const float, IdxT>(
+    orig_index.centers().data_handle(), n_lists, dim);
+  auto labels_view = raft::make_device_vector_view<LabelT, IdxT>(new_labels.data(), n_rows);
+  raft::cluster::kmeans_balanced::predict(handle,
```
cjnolet, line 143: Looks great!

cpp/include/raft/util/cudart_utils.hpp
```diff
@@ -42,7 +42,7 @@
 #include <mutex>
 
 ///@todo: enable once logging has been enabled in raft
-//#include "logger.hpp"
+// #include "logger.hpp"
```
cjnolet, line 45: We can just remove this comment now.

cjnolet (changes requested):
Super excited to have this in `raft::cluster`. Mostly API-ish things.

cpp/test/cluster/kmeans_balanced.cu
```diff
@@ -0,0 +1,235 @@
...
+}
+
+const auto inputsf_i32 = get_kmeans_balanced_inputs<float, int>();
+const auto inputsd_i32 = get_kmeans_balanced_inputs<double, int>();
+const auto inputsf_i64 = get_kmeans_balanced_inputs<float, int64_t>();
+const auto inputsd_i64 = get_kmeans_balanced_inputs<double, int64_t>();
+
+#define KB_TEST(test_type, test_name, test_inputs)         \
+  typedef RAFT_DEPAREN(test_type) test_name;               \
+  TEST_P(test_name, Result) { ASSERT_TRUE(score == 1.0); } \
```
tfeher, line 180: I am not so familiar with the adjusted_rand_score. Is it reasonable to expect score 1.0? If yes, could you add an explicit note about it?

tfeher (changes requested):
Hi Louis, thank for this PR! It is nice to have the balanced hierarchical k-means algorithm publicly available. It looks good in general, I just have a few smaller comments.

cpp/include/raft/cluster/kmeans_balanced.cuh
```diff
@@ -0,0 +1,341 @@
...
+ * @param[in]  mapping_op     (optional) Functor to convert from the input datatype to the
+ *                            arithmetic datatype. If DataT == MathT, this must be the identity.
+ */
+template <typename DataT,
+          typename MathT,
+          typename IndexT,
+          typename LabelT,
+          typename CounterT,
+          typename MappingOpT = raft::identity_op>
+void calc_centers_and_sizes(handle_t const& handle,
```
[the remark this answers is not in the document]
Nyrio: I was wondering about this for both `calc_centers_and_sizes` and `build_clusters`. They are used in the `knn` namespace so I'm not sure whether it was better to allow the `knn` namespace to use `cluster::detail` or add a public API for those in `cluster`

cpp/test/cluster/kmeans_balanced.cu
Nyrio: That is a discussion we had earlier following failures in the k-means tests. An adjusted rand score of 1.0 means that the clustering must be exactly the same as the reference. It's reasonable to expect this in most cases, but there can be malformed datasets where two blobs are close to each other and therefore some points in the reference might be mislabeled.

If the input generator is deterministic, we don't need to worry about this. But as it is currently not (depends on the GPU), we could potentially run into one of these malformed datasets if we run on a GPU that we haven't run on before. Alternatively, we could fix the blob centers, but it makes it harder to change the test parameters.

Anyway, there is te same problem with the k-means tests, it's not new in this PR, and whatever change we do in one test should be mirrored in the other.

cpp/include/raft/cluster/kmeans_balanced_types.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+  /**
+   * Verbosity level. Note that it will not log any level greater than RAFT_ACTIVE_LEVEL
+   */
+  int verbosity = RAFT_LEVEL_INFO;
```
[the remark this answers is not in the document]
cjnolet: Yeah, I think we probably don't want to have thi sin the C++ API (our original kmeans was written before we had started really standardizing on things and making our public APIs more consistent). I also propose that we make this params object  either extend `kmeans_types.hpp` (if that makes sense... it might not) or have them share a common base class (maybe `kmeans_base_params`?) just so we can centralize the params they both share.

cpp/include/raft/matrix/gather.cuh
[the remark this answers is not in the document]
Nyrio: > Do I understand it right, that since you split MatrixIteratorT into InputIteratorT and OutputIteratorT, we can use this in place of knn::details:utils::copy_selected? Then, could you please remove copy_selected and replace the use sites?

@achirkin I just noticed that in IVF-PQ it is used with `ld_src != ld_dst`. `matrix::gather` doesn't support that yet. I'll replace all the use sites I can, but we'll have to keep `copy_selected` for now if we don't want to block the PR.

Nyrio: @achirkin Quick thoughts about this (definitely not for this PR):

- Approach 1: do multiple passes, only copy a row if the source is in the current batch. Problem: all the input dataset will be copied to the device at some point (though not concurrently), and also performance gets worse the more passes you have / the smaller the batches.
- Approach 2: gather on the host. Not completely stupid if the number of elements to gather is small.
- Approach 3: revert the map, sort it by source id, and do multiple passes, only looking at the part of the map that corresponds to the current input batch.

cpp/include/raft/cluster/kmeans_balanced.cuh
```diff
@@ -0,0 +1,341 @@
...
+          typename CounterT,
+          typename MappingOpT>
+void build_clusters(handle_t const& handle,
+                    const kmeans_balanced_params& params,
+                    raft::device_matrix_view<const DataT, IndexT> X,
+                    raft::device_matrix_view<MathT, IndexT> centroids,
+                    raft::device_vector_view<LabelT, IndexT> labels,
+                    raft::device_vector_view<CounterT, IndexT> cluster_sizes,
+                    MappingOpT mapping_op               = raft::identity_op(),
+                    rmm::mr::device_memory_resource* mr = nullptr,
```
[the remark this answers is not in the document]
Nyrio: @cjnolet This function is called by IVF-PQ which creates its own temporary pool resources (managed and device) with a controlled initial size. We can't really / don't want to change the global resource, right? Nor can we rely on it being set as we want.

We should probably not be exposing `build_clusters` nor `calc_centers_and_sizes` in the public API at all but the problem is that they are used by `knn` so it would break encapsulation to have them only in `cluster::detail`.

I'm not sure what to do. Either we expose the resource, or we allow `knn` to compromise on encapsulation. Unless you know a better way to pass the temporary pool from IVF-PQ to this function?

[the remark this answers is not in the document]
cjnolet: We run into issues when we're creating our own pools "behind the scenes" which aren't directly controllable by the user. We can always provide documentation to help the power user make more informed decisions but when users are building a workflow and computing the amount of memory needed to make their calls into RAFT (or the downstream libraries like cuml), they most often will 
1) just use the default RMM resource, 
2) set a global pool to get a speedup (which most often will be some large percentage of their available GPU memory), or 
3) set a managed / managed pool resource to get the benefits of the pool but oversubscribe their GPU to avoid the OOMs. 

If users are invoking several RAFT functions in a row (such as many cuml algorithms do), it's going to make memory management very complicated, if not close to impossible, when we're creating resources and allocating pools behind their back. Thus, we need to afford them the flexibility of being able to specify these things so, for example, the user who sets a pool w/ 90% of their available memory isn't scratching their head when they go to compute the nearest neighbors and encounter an OOM. It's much more straightforward to suggest they can get a perf boost by setting a pooled allocator for the `workspace_resource()` than to tell them they have no control over this.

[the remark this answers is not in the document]
Nyrio: Oh wait, you're suggesting that IVF-PQ shouldn't set its own pool?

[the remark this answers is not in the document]
cjnolet: > Oh wait, you're suggesting that IVF-PQ shouldn't set its own pool?

Yes, I am suggesting that for the longer-term. 

In the near-term, I propose that we add a copy constructor to the handle which accepts an optional `workspace_resource` and does a shallow-copy of the underlying array of existing resources, setting the new resource on the new instance. This will be useful to users so they can invoke different functions that can make use of different pool resources but still give them the control over their configuration (essentially I'm just saying we remove the explicit `mr` argument from this function and use the one already on the `handle_t`, whilst having the existing IVF-pq implementation use a new `handle_t` if it's going to use a new pool). Eventually we will also enable the user to control the resources in ivf-pq as well.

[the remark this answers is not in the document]
cjnolet: In the longer longer term, I'd like for _all_ of RAFT's primitive functions to use the workspace_resource on the given `handle_t`, if one has been set (and fall back to the default resource set on the device). But that's probably a ways out.

Nyrio:
Please note that for simplicity I have merged @cjnolet's branch #1165 into this one.

@achirkin Can you please have a look at the explanation of the algorithm that I wrote in the oxygen for `raft::cluster::kmeans_balanced::fit` and tell me if you'd change anything?

cpp/include/raft/cluster/kmeans_balanced.cuh
```diff
@@ -0,0 +1,366 @@
...
+ * @tparam DataT Type of the input data.
+ * @tparam MathT Type of the centroids and mapped data.
+ * @tparam IndexT Type used for indexing.
+ * @tparam MappingOpT Type of the mapping function.
+ * @param[in]  handle     The raft handle
+ * @param[in]  params     Structure containing the hyper-parameters
+ * @param[in]  X          Training instances to cluster. The data must be in row-major format.
+ *                        [dim = n_samples x n_features]
+ * @param[out] centroids  The generated centroids [dim = n_clusters x n_features]
+ * @param[in]  mapping_op (optional) Functor to convert from the input datatype to the arithmetic
```
achirkin, line 71: Is there a technical requirement to have the identity mapping when `DataT` == `MathT`? At the first glance, the mapping_op sounds like a useful feature even in the case when the types match. If you have a static check for the equality of the types somewhere, perhaps you can replace/complement it with comparing the `MappingOpT` to `identity_op` (static as well).

cpp/include/raft/cluster/kmeans_balanced.cuh
```diff
@@ -0,0 +1,366 @@
...
+ *
+ * The number of mesoclusters is chosen by rounding the square root of the number of clusters. E.g
+ * for 512 clusters, we would have 23 mesoclusters. The number of fine clusters per mesocluster is
+ * chosen proportionally to the number of points in each mesocluster.
+ *
+ * This variant of k-means uses random initialization and a fixed number of iterations, though
+ * iterations can be repeated if the balancing step moved the centroids.
+ *
+ * Additionally, this algorithm supports quantized datasets in arbitrary types but the core part of
+ * the algorithm will work with a floating-point type, hence a conversion function can be provided
```
achirkin, line 48: Just to make sure I'm up to date with the changes. Does that mean `MathT` must be floating point (`double` or `float`)? Or only `float`? (I vaguely remember a `predict_float_core` function here). Do we plan to support `half` here?
If we don't plan to support `half` in here, you can even use the `std::is_floating_point` static_assert here to be more pedantic :)

cpp/include/raft/cluster/detail/kmeans_balanced.cuh
```diff
@@ -0,0 +1,1084 @@
...
+  size_t mem_per_row = 0;
+  switch (metric) {
+    // fusedL2NN only needs one integer per row for a mutex.
+    case distance::DistanceType::L2Expanded:
+    case distance::DistanceType::L2SqrtExpanded: {
+      mem_per_row += sizeof(int);
+    } break;
+    // Other metrics require storing a distance matrix.
+    default: {
+      mem_per_row += sizeof(float) * n_clusters;
```
achirkin, line 205: Should it be sizeof(MathT)?

cpp/include/raft/cluster/detail/kmeans_balanced.cuh
```diff
@@ -0,0 +1,1084 @@
...
+
+  // build fine clusters
+  auto [mesocluster_size_max, fine_clusters_nums_max, fine_clusters_nums, fine_clusters_csum] =
+    arrange_fine_clusters(n_clusters, n_mesoclusters, n_rows, mesocluster_sizes);
+
+  if (mesocluster_size_max * n_mesoclusters > 2 * n_rows) {
+    RAFT_LOG_WARN("build_hierarchical: built unbalanced mesoclusters");
+    RAFT_LOG_TRACE_VEC(mesocluster_sizes, n_mesoclusters);
+    RAFT_LOG_TRACE_VEC(fine_clusters_nums.data(), n_mesoclusters);
+  }
```
achirkin, line 1036: Please merge the changes from https://github.com/rapidsai/raft/pull/1161, that's an important workaround to avoid out-of-memory when the clusters are very unbalanced.

achirkin (changes requested):
The public API looks good! Have you tried to compile and run the clusters on `double`? Also, I did a small workaround to the build_fine_clusters very recently - please, have a look and merge the changes https://github.com/rapidsai/raft/pull/1161

Nyrio:
> Have you tried to compile and run the clusters on double?

That's covered in the new unit test.

cpp/include/raft/cluster/kmeans_balanced.cuh
Nyrio: There are a few reasons to have this assumption:

- We don't want to have to copy the dataset when unnecessary + affect the batch size.
- There are functions that can take the data transformed or not yet. The way to know whether it has been transformed already at the moment is the type of data. I ran into an issue earlier where I was transforming it twice. Note that this wasn't an issue with the ANN `mapping<float>` precisely because it behaves like an identity if the type is already float.

I'm not saying it impossible, just unnecessarily complex and not presently needed.

cpp/include/raft/cluster/kmeans_balanced.cuh
Nyrio: `MathT = double` is supported and tested. Hence the renaming of `predict_float_core`. `half` could be supported in the future but there are a bunch of other prims you'll have to add `half` support for before you can do that.

cpp/include/raft/cluster/detail/kmeans_balanced.cuh
Nyrio: I just fixed that, perhaps double-check that I didn't forget anything? Thanks

cpp/include/raft/cluster/kmeans_balanced.cuh
achirkin: I don't see exactly where is it unnecessarily complex? This just sounds like a rather artificial constraint.

cpp/include/raft/cluster/detail/kmeans_balanced.cuh
achirkin: Yep, looks ok.

cpp/include/raft/cluster/kmeans_balanced.cuh
Nyrio: Unnecessary because we don't have any use for this yet; the other k-means algorithm doesn't support it either. And though it's not particularly hard to implement, it requires some changes and additional checks, and proper testing too, and I don't want to hold this PR for this.

I think that we can always do this change later. As the identity is currently assumed when the type is the same, it's fair to consider it non-breaking to change that behavior in the future (a user should really not be passing a non-identity op if the type is the same).

cjnolet (approved):
@nyrio, can you please update this PR to accept raft::device_resources instead of raft::handle_t everywhere?

tfeher (approved):
Thanks Louis for the updates, the PR looks good to me!

cpp/include/raft/cluster/detail/kmeans_balanced.cuh
```diff
@@ -74,16 +79,17 @@ constexpr static inline float kAdjustCentersWeight = 7.0f;
...
-                         const kmeans_balanced_params& params,
-                         const MathT* centers,
-                         IdxT n_clusters,
-                         IdxT dim,
-                         const MathT* dataset,
-                         const MathT* dataset_norm,
-                         IdxT n_rows,
-                         LabelT* labels,
-                         rmm::mr::device_memory_resource* mr)
+inline std::enable_if_t<std::is_floating_point_v<MathT>> predict_core(
```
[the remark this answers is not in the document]
Nyrio: `int8` dataset with floating-point centroids, or all `int8`? The former means repeating the conversion or cast operation many times in the distance kernels, and for the latter, we have to evaluate if there could be a loss of quality in the clustering, or if learning could somehow get stuck?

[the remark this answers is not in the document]
achirkin: (I'm not sure) Aren't the distance kernels always memory-bound? If so, I don't think the conversion (which is usually a simple multiply + cast) would noticeably hurt performance.

cpp/include/raft/cluster/kmeans_balanced.cuh
achirkin: Ok, nevermind.

cpp/test/cluster/kmeans_balanced.cu
```diff
@@ -0,0 +1,235 @@
...
+template <typename MathT, typename IdxT>
+std::vector<KmeansBalancedInputs<MathT, IdxT>> get_kmeans_balanced_inputs()
+{
+  std::vector<KmeansBalancedInputs<MathT, IdxT>> out;
+  KmeansBalancedInputs<MathT, IdxT> p;
+  p.kb_params.n_iters = 20;
+  p.kb_params.metric  = raft::distance::DistanceType::L2Expanded;
+  p.tol               = MathT{0.0001};
+  std::vector<std::tuple<size_t, size_t, size_t>> row_cols_k = {
+    {1000, 32, 5}, {1000, 100, 20}, {10000, 32, 10}, {10000, 100, 50}, {10000, 500, 100}};
```
achirkin, line 163: Would you consider adding a single or couple tests for the memory footprint? e.g. set a small number of clusters and training iterations, but the size in the orders of magnitude of the GPU memory. To have at some understanding of the out-of-memory error limits.

cpp/include/raft/cluster/kmeans_balanced.cuh
```diff
@@ -0,0 +1,365 @@
...
+ * @param[in]  mapping_op (optional) Functor to convert from the input datatype to the arithmetic
+ *                        datatype. If DataT == MathT, this must be the identity.
+ */
+template <typename DataT, typename MathT, typename IndexT, typename MappingOpT = raft::identity_op>
+void fit(const raft::device_resources& handle,
+         kmeans_balanced_params const& params,
+         raft::device_matrix_view<const DataT, IndexT> X,
+         raft::device_matrix_view<MathT, IndexT> centroids,
+         MappingOpT mapping_op = raft::identity_op())
+{
```
achirkin, line 80: If you require the mapping to be identity when `DataT == MathT`, consider adding a static assert:
```C++
static_assert(!std::is_same_v<MathT, DataT> || !std::is_same_v<MappingOpT, raft::identity_op>,
  "Only the identity mapping is supported when MathT == DataT")
```

achirkin (approved):
LGTM, just couple minor things below.

cpp/include/raft/cluster/kmeans_balanced.cuh
Nyrio: No, this would break the ANN methods that pass `mapping<float>` which behaves like the identity when the input type is float. We want *an* identity op, not *the* `identity_op`.

cpp/test/cluster/kmeans_balanced.cu
Nyrio: I've added a test with more rows but I can't push it close to the memory limits, as the tests are supposed to run on smaller GPUs as well. We can always edit these parameters manually when testing for this locally.

PR #1114, opened 2022-12-21
Allow host dataset for IVF-PQ

tfeher:
This PR enables building (or extending) an IVF-PQ index using data in host memory.

### DIFF
--- a/python/pylibraft/pylibraft/common/ai_wrapper.py
+++ b/python/pylibraft/pylibraft/common/ai_wrapper.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2022, NVIDIA CORPORATION.
+# Copyright (c) 2022-2023, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/python/pylibraft/pylibraft/common/ai_wrapper.py
+++ b/python/pylibraft/pylibraft/common/ai_wrapper.py
@@ -27,32 +27,32 @@ class ai_wrapper:
 
     def __init__(self, ai_arr):
         """
-        Constructor accepts a CUDA array interface compliant array
+        Constructor accepts an array interface compliant array
 
         Parameters
         ----------
-        ai_arr : CUDA array interface array
+        ai_arr : array interface array
         """
         self.ai_ = ai_arr.__array_interface__
 
     @property
     def dtype(self):
         """
-        Returns the dtype of the underlying CUDA array interface
+        Returns the dtype of the underlying array interface
         """
         return np.dtype(self.ai_["typestr"])
 
     @property
     def shape(self):
         """
-        Returns the shape of the underlying CUDA array interface
+        Returns the shape of the underlying array interface
         """
         return self.ai_["shape"]
 
     @property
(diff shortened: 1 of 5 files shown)


tfeher:
Note for reviewer: `ai_wrapper` class is currently just a copy of `cai_wrapper`, and one shall reduce the redundant code in a follow up PR. I have created a separate issue for this: #1115.

python/pylibraft/pylibraft/common/ai_wrapper.py
```diff
@@ -0,0 +1,89 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+import numpy as np
+
+from pylibraft.common import input_validation
+
+
+class ai_wrapper:
```
cjnolet, line 21: Could we have the `cai_wrapper` extend the `ai_wrapper`? This way we can define the common methods once and just have the `cai_wrapper` implement the `copy_to_host()` (and maybe the `ai_wrapper` could implement a `copy_to_device()`?)

python/pylibraft/pylibraft/neighbors/ivf_pq/ivf_pq.pyx
```diff
@@ -465,16 +474,23 @@ def extend(Index index, new_vectors, new_indices, handle=None):
         handle = Handle()
     cdef handle_t* handle_ = <handle_t*><size_t>handle.getHandle()
 
-    vecs_cai = cai_wrapper(new_vectors)
+    try:
+        vecs_cai = cai_wrapper(new_vectors)
```
cjnolet, line 478: We don't need to do so in this PR but I wonder if we should eventually encapsulate this inside a simple factory function that dispatches to the correct wrapper based on whether it's host or device? I can just see this pattern getting repetitive if we end up using it in more algorithms in the future.

cjnolet (changes requested):
Looks great, just a couple small things.

python/pylibraft/pylibraft/common/ai_wrapper.py
tfeher: Changed `cai_wrapper` to be a child class of `ai_wrapper`. 

The `copy_to_host` method is only part of  `device_ndarray`, but not `cai_wrapper`. As I understand the `[c]ai_wrapper` classes just provide convenience for inspecting the array properties, but they do not need to provide copy operation for the data.

python/pylibraft/pylibraft/neighbors/ivf_pq/ivf_pq.pyx
tfeher: I have added a factory function.

tfeher (commented):
Thanks @cjnolet for the comments, I have updated the PR!

python/pylibraft/pylibraft/common/ai_wrapper.py
cjnolet: Ah yes, my mistake. I guess then the only difference really is the name of array interface that's being called so even better to have it this way.

cjnolet (approved):
Changes LGTM

PR #1116, opened 2022-12-21
Build wheels alongside conda CI

sevagh:
This PR adds pip wheel CI to the Conda CI, instead of having them work separately.


.github/workflows/pr.yaml
```diff
@@ -47,3 +51,38 @@ jobs:
...
+      skbuild-configure-options: "-DRAFT_BUILD_WHEELS=ON -DDETECT_CONDA_ENV=OFF -DFIND_RAFT_CPP=OFF"
+  wheel-tests-pylibraft:
+    needs: wheel-build-pylibraft
+    secrets: inherit
+    uses: rapidsai/shared-action-workflows/.github/workflows/wheels-manylinux-test.yml@upkeep/manylinux-improvements
+    with:
+      build_type: pull-request
+      package-name: pylibraft
+      test-unittest: "python -m pytest -v ./python/pylibraft/pylibraft/test"
+      test-smoketest: "" #"python ci/wheel_smoke_test_pylibraft.py"
```
vyasr, line 70: @cjnolet do you have any recommendations on what we could use as a reasonable smoke test for pylibraft and raft-dask? Here are the current examples for [cudf](https://github.com/rapidsai/cudf/pull/12427/files#diff-f2d06f7697ad781f29c9c3322ac1f522feaba7d01d761d3fd086721d9ab872fb) and [rmm](https://github.com/rapidsai/rmm/pull/1182/files#diff-6f7e4916411f089af7bb37a031ef77d2a63b2d365a9df81d78116af7fb68c048). For context, for PR CI we will build packages on both arm and amd architectures, but to conserve GPU CI resources we will only be running the full test suite on amd and will instead only run a small smoke test on arm. The full test suite will be run for arm on nightly builds.

.github/workflows/build.yaml
```diff
@@ -52,3 +52,50 @@ jobs:
       branch: ${{ inputs.branch }}
       date: ${{ inputs.date }}
       sha: ${{ inputs.sha }}
+  wheel-build-pylibraft:
+    secrets: inherit
+    uses: rapidsai/shared-action-workflows/.github/workflows/wheels-manylinux-build.yml@main
+    with:
+      build_type: ${{ inputs.build_type || 'branch' }}
+      repo: rapidsai/raft
```
ajschmidt8, line 60: linking to this comment to keep things DRY

- https://github.com/rapidsai/cuml/pull/5109#discussion_r1063938257

ajschmidt8 (changes requested):

vyasr:
@sevagh to fix the raft-dask NCCL issue during the CMake configure, after [this line](https://github.com/rapidsai/raft/blob/branch-23.02/python/raft-dask/CMakeLists.txt#L53) add
```
find_package(NCCL REQUIRED)
```

.github/workflows/pr.yaml
cjnolet: I think running the raft [pairwise distance](https://github.com/rapidsai/raft/blob/d86610d19a0d368d637a9551a13ddfcb59d1937e/python/pylibraft/pylibraft/test/test_distance.py#L44) tests for a single distance (row-major order, not in-place) should be a good smoke test for pylibraft. 

For raft-dask, we could get away w/ 1 [p2p test](https://github.com/rapidsai/raft/blob/d86610d19a0d368d637a9551a13ddfcb59d1937e/python/raft-dask/raft_dask/test/test_comms.py#L250) for UCX and 1 [collectives test](https://github.com/rapidsai/raft/blob/d86610d19a0d368d637a9551a13ddfcb59d1937e/python/raft-dask/raft_dask/test/test_comms.py#L201) for NCCL.

ajschmidt8 (approved):

.github/workflows/pr.yaml
sevagh: I adapted your suggested unit tests into two smoke tests: https://github.com/rapidsai/raft/pull/1116/commits/c75fb2858c70cfedf7b0e6384b00656b7de4acc4

vyasr (approved):
Looks great!

Issue #803, opened 2022-09-06
[FEA] Replace FAISS `bfknn()` with a version using RAFT primitives


cjnolet:
A large goal for RAFT is to remove the dependency on FAISS in order to remain as lightweight as possible. We have a version of IVF-Flat, which was added in 22.08. We are adding a version of IVF-PQ for 22.10 and have a KNN with a fused k-select which works well when k<64.

What is needed in order to become fully independent of FAISS is for the `bfknn()` call from FAISS to be ported to RAFT primitives. This call is mostly a tiled pairwise distance computation with a k-select performed over each tile. The port here should be fairly straightforward, but we will need to support all of the distances that FAISS supports and make sure there is no loss in performance along the way. 

In addition, FAISS' k-select has proven to be highly optimized and performant and we make heavy use of it throughout RAFT and cuML. We will need to make sure we have a suitable replacement in one of the k-select functions in RAFT. If it turns out we are better of continuing to use the k-select from FAISS, we are probably better off just copying the current version (it hasn't changed much at all over the years) directly into RAFT and providing the proper attribution / licensing info.

Issue #799, opened 2022-09-02
[FEA] Port k-select from FAISS


cjnolet:
In order to officially remove the FAISS dependency from RAFT, we are going to need to port over the k-select from FAISS (called block select pair and potentially warp select pair). We might be able to copy over this logic and provide appropriate attribution and licensing in RAFT.

Issue #798, opened 2022-09-02
[FEA] Port brute force knn from FAISS to RAFT primitives 


cjnolet:
As we are working to integrate RAFT into FAISS, we need to remove all dependence on FAISS to avoid a circular dependency problem. Since the code is pretty close to optimal for moderate values of k, we might be better off just copying the k-selection code into RAFT for the time being (providing proper attribution to FAISS, another with the appropriate licenses, of course).

Essentially, we need to port the `bfknn` to use RAFT primitives. Successful implementation of this issue will include all of the distances which are currently supported by `bfknn` and replacement of `bfknn` from FAISS in our `brute_force_knn` with our own version. Ideally, we would also provide more customization around the batch sizes in order to make our API a little more flexible.

PR #1120, opened 2023-01-04
Update builds for CUDA `11.8` and Python `3.10`

ajschmidt8:
This PR updates the `raft` CI workflows to build against the CUDA `11.8` / Python `3.10` [branch](https://github.com/rapidsai/shared-action-workflows/tree/cuda-118) of the `shared-action-workflows` repository.


bdice:
This PR is blocked until we can get ARM + CUDA builds of `faiss` version 1.7.1 or newer (see here: https://github.com/conda-forge/faiss-split-feedstock/pull/58) or remove the `faiss` dependency entirely (several open issues, including #798, #799, #803, ...).

The existing ARM + CUDA builds of `faiss` 1.7.0 do not support GPU architectures newer than 80, which we need to support (86 and 90).

jakirkham:
PR ( https://github.com/conda-forge/faiss-split-feedstock/pull/58 ) added ARM builds (without CUDA). So just CPU builds. That said, this does seem to cover the changes we had in other places before. Though please double check that.

PR ( https://github.com/conda-forge/faiss-split-feedstock/pull/62 ) adds CUDA builds. Based on Bradley's comment above, think we should check that the recipe is doing what we want with [CUDA arches]( https://github.com/conda-forge/faiss-split-feedstock/blob/1c7f8991c3fd09fe520eac0f14438e679ac62ab7/recipe/build-lib.sh#L14-L15 ). There is a second point about build time, which I've outlined in more detail in comment ( https://github.com/conda-forge/faiss-split-feedstock/pull/62#issuecomment-1374307900 ). Feel free to raise anything else I've missed in that PR.

bdice (approved):

raydouglass (approved):

AyodeAwe (approved):

cjnolet (approved):
Thank you for also updating the build docs!

Issue #1058, opened 2022-11-30
[BUG] IVF flat incorrect distance values with int8 data


tfeher:
**Describe the bug**
The distance values returned with IVF flat search method might be incorrect in certain cases

IVF flat search is used by ANN refinement (if the dataset is in device memory) #1052. The Python tests of refinement revealed this problem. This only happens with int8 data (and not with uint8). IVF flat search has specific kernels for different data types, this might point towards a bug in the int8 specialization.

Note: the returned neighbor indices are correct, only the distance values show this problem.

**Steps/Code to reproduce bug**
To reproduce the issue:
1. Uncomment the following lines
https://github.com/tfeher/raft/blob/c34952ba42ca313e280f89fe4a0bc6a495a03852/python/pylibraft/pylibraft/test/test_refine.py#L127-L128

2. Run the refinement test
```bash
pytest pylibraft/test/test_refine.py -k test_refine_dtypes
```
Output:
```
>       assert np.mean(diff) < eps
E       assert 0.41265658 < 0.001
E        +  where 0.41265658 = <function mean at 0x14cd30085e50>(array([[0.46209675, 0.3915514 , 0.41272217, 0.41572857, 0.40346152,\n        0.37558016, 0.39662588, 0.36034888, 0.3658...373, 0.33194792, 0.20105657,\n        0.38698354, 0.3945595 , 0.40543446, 0.3377944 , 0.3658476 ]],\n      dtype=float32))
E        +    where <function mean at 0x14cd30085e50> = np.mean

pylibraft/test/test_ivf_pq.py:79: AssertionError
========================================================================================================== short test summary info ==========================================================================================================
FAILED pylibraft/test/test_refine.py::test_refine_dtypes[device-int8-l2_expanded-True-100] - assert 0.41855982 < 0.001
FAILED pylibraft/test/test_refine.py::test_refine_dtypes[device-int8-l2_expanded-True-1024] - assert 0.41032004 < 0.001
FAILED pylibraft/test/test_refine.py::test_refine_dtypes[device-int8-l2_expanded-True-37] - assert 0.40580684 < 0.001
FAILED pylibraft/test/test_refine.py::test_refine_dtypes[device-int8-l2_expanded-False-100] - assert 0.41458178 < 0.001
FAILED pylibraft/test/test_refine.py::test_refine_dtypes[device-int8-l2_expanded-False-1024] - assert 0.41299787 < 0.001
[...]

PR #1122, opened 2023-01-05
Fix `euclidean_dist` in IVF-Flat search

Nyrio:
Solves #1058 

This is a tricky bug so the fix deserves some explanation. The previous implementation of `euclidean_dist` was the following in vectorized cases, where `x` and `y` are `int32` vectors of 4 `int8` each and `acc` is a single `int32` number to accumulate the distance in:

```c++
// Compute vectorized absolute differences independently.
const auto diff = static_cast<int32_t>(__vabsdiffs4(x, y));
// Square, reduce, and add to the accumulator.
acc = dp4a(diff, diff, acc);
```

Now consider the following case:

```c++
x = 0x80; // -128, 0, 0, 0
y = 0x7f; //  127, 0, 0, 0
```

The difference between -128 and 127 is 255, represented as `FF` (`__vabsdiffs4` is smart enough not to compute `abs(a-b)` which would result in `01`). However, if we call the signed version of `dp4a`, `FF` is cast from `int8` to `int32` as `FFFFFFFF` (or -1). The square of -1 is 1, which is added to `acc` (instead of 65025).

As the output of `__vabsdiffs4` is correct when considered as an unsigned number, and as addition is the same for signed and unsigned in 2's complement (and `acc` is positive anyway), the easiest fix is to use the unsigned version of `dp4a`, which will cast overflowed differences properly to 32 bits. The previous code simply becomes:

```c++
const auto diff = __vabsdiffs4(x, y);
acc = dp4a(diff, diff, static_cast<uint32_t>(acc));
```

-----

Additionally, to avoid underflows in the non-vectorized unsigned case, I replaced the subtraction with `__usad` (absolute difference of unsigned numbers). Note that using the subtraction was correct anyway, because the addition/subtraction is the same for unsigned and signed integers, as well as the least significant half of the multiplication (which is the part that is stored), and the square of a number is also the square of its opposite. Consider:

```c++
uint32_t a = 10;
uint32_t b = 20;
uint32_t c = a - b; // fffffff6, i.e -10 or 4294967286
uint32_t d = c * c; // (ffffffec)00000064, i.e 100
```

### DIFF
--- a/cpp/include/raft/spatial/knn/detail/ivf_flat_search.cuh
+++ b/cpp/include/raft/spatial/knn/detail/ivf_flat_search.cuh
@@ -899,7 +899,7 @@ struct euclidean_dist<Veclen, int8_t, int32_t> {
   {
     if constexpr (Veclen > 1) {
       const auto diff = __vabsdiffs4(x, y);
-      acc             = __dp4a(diff, diff, static_cast<uint32_t>(acc));
+      acc             = dp4a(diff, diff, static_cast<uint32_t>(acc));
     } else {
       const auto diff = x - y;
       acc += diff * diff;
--- a/cpp/include/raft/spatial/knn/detail/ivf_flat_search.cuh
+++ b/cpp/include/raft/spatial/knn/detail/ivf_flat_search.cuh
@@ -898,11 +898,8 @@ struct euclidean_dist<Veclen, int8_t, int32_t> {
   __device__ __forceinline__ void operator()(int32_t& acc, int32_t x, int32_t y)
   {
     if constexpr (Veclen > 1) {
-      const uint32_t offset = 0x80808080;
-      const auto offset_x   = __vadd4(x, offset);
-      const auto offset_y   = __vadd4(y, offset);
-      const auto diff       = __vabsdiffu4(offset_x, offset_y);
-      acc = __dp4a(diff, diff, static_cast<uint32_t>(acc));
+      const auto diff = __vabsdiffs4(x, y);
+      acc             = __dp4a(diff, diff, static_cast<uint32_t>(acc));
     } else {
       const auto diff = x - y;
       acc += diff * diff;
--- a/cpp/include/raft/spatial/knn/detail/ivf_flat_search.cuh
+++ b/cpp/include/raft/spatial/knn/detail/ivf_flat_search.cuh
@@ -886,8 +886,7 @@ struct euclidean_dist<Veclen, uint8_t, uint32_t> {
       const auto diff = __vabsdiffu4(x, y);
       acc             = dp4a(diff, diff, acc);
     } else {
-      // Note: diff can underflow but the multiplication is still correct in that case.
-      const auto diff = x - y;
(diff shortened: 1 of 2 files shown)


cpp/include/raft/spatial/knn/detail/ivf_flat_search.cuh
```diff
@@ -897,8 +898,8 @@ struct euclidean_dist<Veclen, int8_t, int32_t> {
   __device__ __forceinline__ void operator()(int32_t& acc, int32_t x, int32_t y)
   {
     if constexpr (Veclen > 1) {
-      const auto diff = static_cast<int32_t>(__vabsdiffs4(x, y));
-      acc             = dp4a(diff, diff, acc);
+      const auto diff = __vabsdiffs4(x, y);
+      acc             = dp4a(diff, diff, static_cast<uint32_t>(acc));
```
achirkin, line 902: So, do I understand the problem right, that `dp4a.s32.s32` interprets each byte as `int8_t`, whereas `__vabsdiffs4` produces four `uint8_t` values (thus, the outputs above 127 interpreted as negative)? And the solution is to do `dp4a.u32.u32`, which interprets components correctly as `uint8_t`? Then, could you please maybe add a short note here that we really want the unsigned overload of `dp4a`?

cpp/include/raft/spatial/knn/detail/ivf_flat_search.cuh
```diff
@@ -886,6 +886,7 @@ struct euclidean_dist<Veclen, uint8_t, uint32_t> {
       const auto diff = __vabsdiffu4(x, y);
       acc             = dp4a(diff, diff, acc);
     } else {
+      // Note: diff can underflow but the multiplication is still correct in that case.
```
achirkin, line 889: Now that you said it out loud, it feels a little uncomfortable. Maybe use `x > y ? x - y : y - x` or `__usad ( unsigned int  x, unsigned int  y, unsigned int  z)`?

achirkin (changes requested):
Wow, that must have been a tough investigation, good job! You've explained the problem very nice here, but maybe it's worth adding a short gist of it (or a link) in the code?

cpp/include/raft/spatial/knn/detail/ivf_flat_search.cuh
Nyrio: `x > y ? x - y : y - x` is less efficient than casting to signed.

I could go with `__usad`, it switches one instruction for another: `IADD3` to `VABSDIFF.U32`.

tfeher (approved):
Thanks Louis for fixing this problem, the PR looks good to me!

Could you update the second part of the PR description? You say
> added comment in the unsigned, non-vectorized implementation

But that has changed during the review.

Nyrio:
> Could you update the second part of the PR description?

Ah, yes, done.

cjnolet (approved):
Changes LGTM. Thanks @Nyrio!

achirkin (approved):
LGTM!

PR #1125, opened 2023-01-06
Adding ability to use an existing stream in the pylibraft Handle

### LINKED ISSUE #1123: Is there a way create a handle from an external cuda stream? Or get the pointer to the raft stream for use by other libraries?
I have an application that is using cupy for matrix manipulation.  I would like to use the `fused_l2_nn_argmin` function to find nearest vectors.   I would like to utilize  a cuda stream for asynchronous operation.   I would like to use the same cuda stream for the cupy oprerations and the raft operations.

  I see the the `fused_l2_nn_argmin`  can accept a `handle` argument so that the operation can be run on a stream.   I would like to be able to use the same stream for the cupy calculations and the raft handle.   Is there a way to create a raft handle from a cupy stream? This would be similar to the `cupy.cuda.ExternalStream` function.

Alternatively, is there a way to get the pointer to the raft Stream so that it could be passed to the `cupy.cuda.ExternalStream()` function.

In cupy I can do the following

```python
str_ptr = cp.cuda.runtime.streamCreate()
cp_str= cp.cuda.ExternalStream(str_ptr)
cp_str.use():
    # Matrix operations using the cupy stream
   fused_l2_nn_argmin(in1, in2, output)  # raft operation will use default stream

```

In raft I can do
```python
raft_stream=Stream()
raft_handle=Handle(raft_stream)
str_ptr = cp.cuda.runtime.streamCreate()
cp_str= cp.cuda.ExternalStream(str_ptr)
cp_str.use():
    # Matrix operations using the cupy stream
    fused_l2_nn_argmin(in1, in2, output, handle=raft_handle) # this will use the raft_stream, 
```

What I would like to do (method 1):
```python
str_ptr = cp.cuda.runtime.streamCreate()
cp_str= cp.cuda.ExternalStream(str_ptr)
raft_external_stream = Stream(str_ptr)
raft_handle= Handle(raft_external_stream)

cp_str.use():
    # Matrix operations using the cupy stream
   fused_l2_nn_argmin(in1, in2, output, handle=raft_handle) # this will use the same stream

```
Method 2:
``` python
raft_stream=Stream()
raft_handle=Handle(raft_stream)
cp_str=  cp.cuda.ExternalStream(raft_stream.get_ptr())  # wrap the raft stream as a cupy stream
cp_str.use():
    # Matrix operations using the raft stream
    fused_l2_nn_argmin(in1, in2, output, handle=raft_handle) # this will use the raft_stream, 

```
Since cuML can use a raft handle, this would be more broadly applicable to any application that is utilizing cuda streams and a mix of cupy and rapids.

Any help will be appreciated, thanks.

cjnolet:
Closes #1123

### DIFF
--- a/python/pylibraft/pylibraft/test/test_distance.py
+++ b/python/pylibraft/pylibraft/test/test_distance.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2022, NVIDIA CORPORATION.
+# Copyright (c) 2022-2023, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -17,7 +17,7 @@ import numpy as np
 import pytest
 from scipy.spatial.distance import cdist
 
-from pylibraft.common import Handle, device_ndarray
+from pylibraft.common import Handle, Stream, device_ndarray
 from pylibraft.distance import pairwise_distance
 
 
@@ -64,9 +64,10 @@ def test_distance(n_rows, n_cols, inplace, metric, order, dtype):
     input1_device = device_ndarray(input1)
     output_device = device_ndarray(output) if inplace else None
 
-    handle = Handle()
+    s2 = Stream()
+    handle = Handle(stream=s2)
     ret_output = pairwise_distance(
-        input1_device, input1_device, output_device, metric
+        input1_device, input1_device, output_device, metric, handle=handle
     )
     handle.sync()
 
--- a/ci/checks/copyright.py
+++ b/ci/checks/copyright.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2020-2022, NVIDIA CORPORATION.
+# Copyright (c) 2020-2023, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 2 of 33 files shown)


cjnolet:
rerun tests

python/pylibraft/pylibraft/test/test_distance.py
```diff
@@ -64,9 +64,10 @@ def test_distance(n_rows, n_cols, inplace, metric, order, dtype):
     input1_device = device_ndarray(input1)
     output_device = device_ndarray(output) if inplace else None
 
-    handle = Handle()
+    s2 = Stream()
+    handle = Handle(stream=s2)
```
dantegd, line 68: why is this change necessary for this particular test? Not sure if it adds something now that there is a `test_handle`

dantegd (changes requested):
Just had some very small questions/comments

python/pylibraft/pylibraft/test/test_distance.py
cjnolet: This one is mostly there just to verify that using an existing stream doesn't cause any crazy errors to the output of the function. I only added it to this test because it doesn't really seem like it needs to be in all the tests and i didn't want to pollute the simplicity of the handle tests by re-writing the pairwise distance test in its entirety. We could remove it, but it seems pretty benign (and it's also helpful).

dantegd (approved):

PR #1134, opened 2023-01-10
Fix documentation author

bdice:
Fixes docs to mark NVIDIA as the author.


cjnolet (approved):
LGTM! Thanks for making this change!

PR #1136, opened 2023-01-11
Add raft::void_op functor

ahendriksen:
Follow up on PR #1049. Adds a void_op functor for lambdas that are unused.

### DIFF
--- a/cpp/include/raft/core/operators.hpp
+++ b/cpp/include/raft/core/operators.hpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, NVIDIA CORPORATION.
+ * Copyright (c) 2022-2023, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -37,6 +37,14 @@ struct identity_op {
   constexpr RAFT_INLINE_FUNCTION auto operator()(const Type& in, UnusedArgs...) const
   {
     return in;
+  }
+};
+
+struct no_op {
+  template <typename Type, typename... UnusedArgs>
+  constexpr RAFT_INLINE_FUNCTION void operator()(const Type& in, UnusedArgs...) const
+  {
+    return;
   }
 };
 
--- a/cpp/include/raft/core/operators.hpp
+++ b/cpp/include/raft/core/operators.hpp
@@ -40,9 +40,9 @@ struct identity_op {
   }
 };
 
-struct no_op {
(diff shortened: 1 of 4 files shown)


cpp/include/raft/core/operators.hpp
```diff
@@ -40,6 +40,14 @@ struct identity_op {
   }
 };
 
+struct no_op {
+  template <typename Type, typename... UnusedArgs>
+  constexpr RAFT_INLINE_FUNCTION void operator()(const Type& in, UnusedArgs...) const
```
Nyrio, line 45: Why keep `const Type& in`? The variadic arguments should be sufficient.

cpp/include/raft/core/operators.hpp
```diff
@@ -40,6 +40,14 @@ struct identity_op {
   }
 };
 
+struct no_op {
```
Nyrio, line 43: Nitpick, but I'm not so sure about naming it `no_op`, due to potential confusion with the old `Nop` which has been replaced by `identity_op`.

What about `void_op` / `empty_op`?

cpp/include/raft/core/operators.hpp
```diff
@@ -40,6 +40,14 @@ struct identity_op {
...
+  template <typename Type, typename... UnusedArgs>
+  constexpr RAFT_INLINE_FUNCTION void operator()(const Type& in, UnusedArgs...) const
+  {
+    return;
+  }
```
achirkin, line 48: Perhaps, we could remove the first argument and the return statement here?

achirkin (changes requested):
Thanks for the suggestion! One thing regarding the implementation below

cpp/include/raft/core/operators.hpp
achirkin: Oops, never mind, @Nyrio was faster with the comments :)

cpp/include/raft/core/operators.hpp
ahendriksen: Oops. Fixed.

cpp/include/raft/core/operators.hpp
ahendriksen: Sure! is now void_op

cpp/include/raft/core/operators.hpp
ahendriksen: Thanks! I removed the first argument, but kept the return for symmetry with all the other operators in the file :)

Nyrio (approved):

achirkin (approved):

cjnolet (approved):

PR #1137, opened 2023-01-11
Adding workspace resource

cjnolet:
This will default to `rmm::mr::get_current_device_resource()` in the event no explicit workspace resource has been set. It's using raw pointers right now, but that may be okay as the RMM memory resource API seems to promote that over shared pointers (or any dereferencing at all).


cjnolet:
cc @achirkin @Nyrio @tfeher

benfred (approved):
looks good!

PR #1138, opened 2023-01-12
Add L2SqrtExpanded support to ivf_pq

benfred:


achirkin (approved):
Thanks for the PR, looks nice and straightforward. Interesting that we've got the post-processing step of taking the square root already covered earlier.

python/pylibraft/pylibraft/neighbors/ivf_pq/ivf_pq.pyx
```diff
@@ -135,9 +134,11 @@ cdef class IndexParams:
         n_list : int, default = 1024
             The number of clusters used in the coarse quantizer.
         metric : string denoting the metric type, default="l2_expanded"
-            Valid values for metric: ["l2_expanded", "inner_product"], where
-            - l2_expanded is the equclidean distance without the square root
+            Valid values for metric: ["l2_expanded", "inner_product",
```
cjnolet, line 137: Doesn't need to be done in this PR, but I think `l2_expanded` is equivalent to `sqeuclidean` in this context and we should probably accept that as an option as well (since it's likely to be more familiar to users) .

cjnolet (approved):
LGTM!

python/pylibraft/pylibraft/neighbors/ivf_pq/ivf_pq.pyx
benfred: https://github.com/rapidsai/raft/pull/1160

PR #1140, opened 2023-01-13
Changing handle_t to device_resources everywhere

cjnolet:
This is the step 1 of a two-step change to eventually have all of our public API functions use `raft::resources` instead of `raft::handle_t`. Step 1 is to use `device_resources` everywhere. This is not a breaking change since `handle_t` extends `device_resources`. Step 2 is to scrape through and use `raft::resources`, which will require adding explicit includes for each resources and using accessor functions everywhere rather than methods on the `device_resources`. No immediate rush for the second step but I'd like to at least have the documentation adjusted to use our new nomenclature everwhere.


benfred (commented):
looks good! 

Some minor questions / comments:
* I see the `handle_t` is still used in the pylibraft bindings - should we change pylibraft over to use device_resources in this PR too? 
* There is a`#include <raft/handle.hpp>` in `linalg/strided_reduction.cuh` that should be changed to include the device_resources header instead

cjnolet:
Thanks for reviewing @benfred!

> I see the handle_t is still used in the pylibraft bindings - should we change pylibraft over to use device_resources in this PR too?

The `raft::handle_t` was intentially made a subclass of `raft::device_resources` so that we can change everything internally without breaking user APIs. I suppose we could just use device resources underneath for now and make python's `Handle()` subclass `device_resources` the same way it is doing in the C++ layer.

python/pylibraft/pylibraft/common/handle.pyx
```diff
@@ -32,7 +32,7 @@ from .cuda import CudaRuntimeError
 
 
 cdef class DeviceResources:
-    """
+    '''
```
benfred, line 35: This is super nit-picky - but the previous `"""` is inline with the `black` python styling we are using here (this would fail code formatting if it were a pure python file instead of a a cython file - and is only passing because black doesn't handle cython).

benfred (approved):
lgtm!

python/pylibraft/pylibraft/common/handle.pyx
cjnolet: yep, I'll change this. One of the many things I had tried to get the doctest strings works

PR #1142, opened 2023-01-13
Simplify distance/detail to make is easier to dispatch to different kernel implementations

ahendriksen:
The pairwise distance metrics are quite varied. The table below summarizes the differences, in terms of 

- Epilog : whether the metric has a non-empty epilog operation.
- Uses norms: whether the metric requires precalculation of the norms of the vectors.
- Has params: whether the norm has additional parameters. The L2 metric, for instance, has the `sqrt` boolean parameter that determines whether to calculate the squared or actual distance.
- Pre- & post-processing: For some metrics, the norms have to be precalculated. For other metrics, the input matrices are transformed before the kernel launch, and "untransformed" after.
- Expensive inner loop: some metrics use `pow`, `log` or other expensive functions in the inner loop. 
- Depends on row-major: the calculation of some metrics depend on whether the input is row-major. 
- CUTLASS: some metrics have an implementation using CUTLASS and tensor cores.

<table border="2" cellspacing="0" cellpadding="6" rules="groups" frame="hsides">

<colgroup>
<col  class="org-left" />

<col  class="org-left" />

<col  class="org-left" />

<col  class="org-left" />

<col  class="org-left" />

<col  class="org-left" />

<col  class="org-left" />

<col  class="org-left" />
</colgroup>
<thead>
<tr>
<th scope="col" class="org-left">Metric</th>
<th scope="col" class="org-left">Epilog</th>
<th scope="col" class="org-left">Uses norms</th>
<th scope="col" class="org-left">Has params</th>
<th scope="col" class="org-left">Pre- &amp; post-processing</th>
<th scope="col" class="org-left">Expensive inner loop</th>
<th scope="col" class="org-left">Depends on row-major</th>
<th scope="col" class="org-left">CUTLASS</th>
</tr>
</thead>

<tbody>
<tr>
<td class="org-left">Canberra</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">x</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
</tr>

<tr>
<td class="org-left">Chebyshev (Linf)</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
</tr>

<tr>
<td class="org-left">Correlation</td>
<td class="org-left">x</td>
<td class="org-left">x (twice)</td>
<td class="org-left">x (many)</td>
<td class="org-left">compute norms</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">x</td>
<td class="org-left">&#xa0;</td>
</tr>

<tr>
<td class="org-left">Cosine</td>
<td class="org-left">x</td>
<td class="org-left">x</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">compute norms</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">x</td>
</tr>

<tr>
<td class="org-left">Hamming</td>
<td class="org-left">x</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">x (k)</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
</tr>

<tr>
<td class="org-left">Hellinger</td>
<td class="org-left">x</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">sqrt and square</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
</tr>

<tr>
<td class="org-left">Jensen Shannon</td>
<td class="org-left">x</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">x</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
</tr>

<tr>
<td class="org-left">KL divergence</td>
<td class="org-left">x</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">x (row major, x == y)</td>
<td class="org-left">yes</td>
<td class="org-left">x</td>
<td class="org-left">x</td>
<td class="org-left">&#xa0;</td>
</tr>

<tr>
<td class="org-left">L1</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
</tr>

<tr>
<td class="org-left">L2 expanded</td>
<td class="org-left">x</td>
<td class="org-left">x</td>
<td class="org-left">x (sqrt)</td>
<td class="org-left">compute norms</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">x</td>
</tr>

<tr>
<td class="org-left">L2 unexpanded</td>
<td class="org-left">x</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">x (sqrt)</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
</tr>

<tr>
<td class="org-left">Minkowski (Lp)</td>
<td class="org-left">x</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">x (p)</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">x</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
</tr>

<tr>
<td class="org-left">Russel-Rao</td>
<td class="org-left">x</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">x (k, 1/k)</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
<td class="org-left">&#xa0;</td>
</tr>
</tbody>
</table>

To keep the complexity that results from all these differences in check, there are several layers between the public API and the kernel launch, each with their own responsibility. 

## Before
1. `raft::distance::pairwise_distance` takes distance type as a run-time argument and dispatches to `raft::distance::detail::pairwise_distance_impl`.
2. `raft::distance::detail::pairwise_distance_impl` allocates workspace as necessary and calls `raft::distance::detail::distance`
3. `raft::distance::detail::distance` defines a default final operation (the identity) and calls an overload of itself.
4. `raft::distance::detail::distance` (with `fin_op`) initializes a `DistanceImpl` zero-sized struct with the correct template arguments and runs the `.run()` method of the struct.
5. `raft::distance::detail::DistanceImpl<DistanceType>.run()` calls `raft::distance::detail::XX_Impl`.
6. `raft::distance::detail::XX_Impl` has the following responsibilities:
   - Pre-compute norms if necessary
   - Transform input if necessary
   - If metric supports a CUTLASS operation, dispatch if necessary.
   - Swap inputs if column-major.
   - Based on runtime parameter `row_major` dispatch to function template `raft::distance::detail::XX<bool row_major>`
7. `raft::distance::detail::XX` based on alignment of input data dispatch to function template `raft::distance::detail::XX_Impl<int veclen>` (different overload of previous `raft::distance::detail::XX_Impl`)
8. `raft::distance::detail::XX_Impl` has the following responsibilities:
   - Define `core_op` and `epilog_op`
   - Define `use_norms`
   - Launch kernel `pairwiseDistanceMatKernel` with correct launch parameters

**Observations**: 
- Steps 6 and 7 both convert a runtime value to a compile time constant (row-major layout and alignment). 
- Step 7 is repeated (copy pasted) for each metric.
- Steps 7 and 8 do a lot of different things and the steps in between do relatively little.
- Steps 1-5 do fairly little (but require a lot of boilerplate)

**Proposal**:

1. Collect as much of the runtime behavior of each metric in a `distance_op` that [contains](https://github.com/ahendriksen/raft/blob/wip-refactor-distance/cpp/include/raft/distance/detail/distance_ops/canberra.cuh):
    - The core_op
    - The epilog_op
    - The required shared memory
    - Whether the inner loop is expensive (and thus loop unrolling should be curtailed)
2. Collect the runtime -> compile-time dispatch in one location ([dispatch.cuh](https://github.com/ahendriksen/raft/blob/486393eff4e0cf1d45ab9d7990b64d607e835d70/cpp/include/raft/distance/detail/pairwise_matrix/dispatch.cuh#L70))
3. Collect kernel launching in one [location](https://github.com/ahendriksen/raft/blob/486393eff4e0cf1d45ab9d7990b64d607e835d70/cpp/include/raft/distance/detail/pairwise_matrix/kernel_sm60.cuh#L108)
4. Remove some of the boilerplate in steps 1-5.

## After
1. `raft::distance::pairwise_distance` takes distance type as a run-time argument, allocates workspace as necessary, and dispatches to `raft::distance::detail::distance`.
2. `raft::distance::detail::distance` defines a default final operation (the identity) and calls an overload of itself.
3. `raft::distance::detail::distance` (with `fin_op`) calls an overload of `raft::distance::detail::distance_impl` for the correct distance type.
4. `raft::distance::detail::distance_impl` has the following responsibilities:
   - Pre-compute norms if necessary
   - Initialize distance op with parameters as necessary, see below for more information.
   - Transform input if necessary
   - If metric supports a CUTLASS operation, dispatch if necessary.
   - Dispatch to `raft::distance::detail::distance_matrix_dispatch`
5. `raft::distance::detail::distance_matrix_dispatch` has the following responsibilities:
   - swap x, y matrices if column major
   - dispatch to correct kernel based on run-time parameters `row_major` and `vec_len`
   - Determine kernel policy based on parameters
   - Call `raft::distance::detail::pairwise_matrix`
6. `raft::distance::detail::pairwise_matrix` launches the `raft::distance::detail::pairwise_matrix_kernel` with the correct launch parameters.

**Distance_op**
 `raft::distance::detail::ops::XX_distance_op` [[example]](https://github.com/ahendriksen/raft/blob/wip-refactor-distance/cpp/include/raft/distance/detail/distance_ops/canberra.cuh) has the following responsibilities:
   - Take any parameters (sqrt, k, etc)
   - Define `core_op` and `epilog_op`
   - Define `use_norms`, `expensive_inner_loop`, and `shared_mem_size()`.

Still TODO:

- [x] Rename Minkowski and Chebyshev to Lp and Linf.
- [x] Do something with this note in the comments: "if workspace is passed as nullptr, this will return in worksize, the number of bytes of workspace required", which is wrong.
- [x] Add a mechanism to limit duplicate compilation when a CUTLASS kernel is available. This is done in follow up PR #1295.
- [x] Some distance_ops have additional template parameters. This must be cleared up.


ahendriksen:
Reminder/TODO

make grid stride loop variables local variables instead of member variables. 

See: https://github.com/rapidsai/raft/pull/838#discussion_r1085195377

ahendriksen:
TODO: https://github.com/rapidsai/raft/pull/838#discussion_r1085900953

Also look at fused and maskedL2NN for shared memory calculations.

ahendriksen:
@cjnolet , @tfeher : Although there is still some minor cleanup to be done, I think the PR is ready for a broad strokes review. I have updated the PR message to give a broad overview of what has changed.

cjnolet:
> raft::distance::pairwise_distance takes distance type as a run-time argument, allocates workspace as necessary, and dispatches to raft::distance::detail::distance.

Does this mean we'll be removing the explicit template type for each distance (and thus reducing the number of templates that need to be instantiated and speeding up compile times)?

ahendriksen:
We still have the distance template argument (either in the form of the enum or a distance op type) in some places, but quite a lot of duplication has been removed.

ahendriksen:
The CI error does not seem related to the changes in the PR: 

```
2023-02-22T19:37:49.5526382Z   Error compiling Cython file:
2023-02-22T19:37:49.5533415Z   ------------------------------------------------------------
2023-02-22T19:37:49.5539691Z   ...
2023-02-22T19:37:49.5546449Z   # cython: language_level = 3
2023-02-22T19:37:49.5552762Z 
2023-02-22T19:37:49.5559417Z 
2023-02-22T19:37:49.5566630Z   from libcpp.memory cimport shared_ptr, unique_ptr
2023-02-22T19:37:49.5573013Z 
2023-02-22T19:37:49.5580739Z   from rmm._lib.cuda_stream_pool cimport cuda_stream_pool
```

Is there a way to rerun the CI without pushing an empty commit?

vyasr:
@ahendriksen if you click on the "Details" button it will take you to the actions tab where you can use the "Re-run jobs" dropdown. See https://docs.github.com/en/actions/managing-workflow-runs/re-running-workflows-and-jobs for more info. I've queued up a rerun.

ahendriksen:
Thanks! And thanks for linking to the docs. I will have a look to see what else I am missing.

cpp/include/raft/distance/detail/distance_ops/template.cuh
```diff
@@ -0,0 +1,60 @@
...
+
+namespace raft::distance::detail::ops {
+
+// Describes the computation the template distance
+//
+// Fill in the TODO items.
+
+template <typename DataT, typename AccT, typename IdxT>
+struct template_distance_op {
+  TODO member;
```
cjnolet, line 29: Oh I see- so this file is just meant to be a guide for a future developer and never meant to be included or used directly. I wonder if we should should include this in the docs instead of putting it in the code. Maybe we could start creating links off of the developer guide for guides related to individual namespaces?

cpp/include/raft/distance/distance.cuh
```diff
@@ -252,71 +250,63 @@ void pairwise_distance(raft::resources const& handle,
                        bool isRowMajor = true,
                        Type metric_arg = 2.0f)
 {
+  cudaStream_t stream = raft::resource::get_cuda_stream(handle);
```
cjnolet, line 253: It's great to see this! We should also deprecate and remove the functions that accept an optional workspace and use `raft::resource::get_workspace_resource()` (`raft/core/resource/device_memory_resource.hpp`) to instantiate a `device_uvector_policy` (currently `raft/core/detail/device_mdarray.hpp` but will become `raft/core/device_container_policy.hpp` once https://github.com/rapidsai/raft/pull/1279 is merged).

cpp/test/CMakeLists.txt
```diff
@@ -118,19 +118,19 @@ if(BUILD_TESTS)
...
-    test/distance/dist_euc_unexp.cu
-    test/distance/dist_eucsqrt_exp.cu
     test/distance/dist_hamming.cu
     test/distance/dist_hellinger.cu
     test/distance/dist_inner_product.cu
     test/distance/dist_jensen_shannon.cu
     test/distance/dist_kl_divergence.cu
     test/distance/dist_l1.cu
-    test/distance/dist_minkowski.cu
+    test/distance/dist_l2_exp.cu
```
cjnolet, line 129: The rename look great! Definitely on board w/ being more consistent across APIs.

cpp/include/raft/distance/distance.cuh
```diff
@@ -252,71 +250,63 @@ void pairwise_distance(raft::resources const& handle,
...
+
+  auto dispatch = [&](auto distance_type) {
+    auto worksize = getWorkspaceSize<distance_type(), Type, Type, Type, Index_>(x, y, m, n, k);
+    workspace.resize(worksize, stream);
+    detail::distance<distance_type(), Type, Type, Type, Index_>(
```
cjnolet, line 258: I guess we still need to know some things ahead of time for loop unrolling. I'm wondering if there's still any way we could capture (and consolidate) those differences in templates so we can eventually get to the point where the distances themselves no longer need to be separate templates?

cjnolet (changes requested):
I really like the idea of abstracting each unique distance formula into composable `ops`. In this case, not only does it allow us to specify and react differently to various performance conditions, but it allows us to capture the differences of the computations themselves in a single place. This is very much what we did w/ the sparse APIs as well- they are a series of composable operations that can be executed according to their needs (in the sparse case that's binary vs dot-product-based vs full-pairwise evaluation). 


I think this looks great. I'd like to see the functions in the public API that accept explicit workspaces deprecated because the memory resource should be getting propagated through the `raft::resources` instance (and be controllable by the user). I think we have an opportunity to here to expand our developer guide as well. Otherwise, I'm completely on board w/ this change.

cpp/include/raft/distance/distance.cuh
ahendriksen: That sounds like a good idea. Let's keep that a separate PR. That way we can keep this PR as non-breaking (as well as it's two dependent PRs).

cpp/include/raft/distance/detail/distance_ops/template.cuh
ahendriksen: Yes indeed. I agree that adding more documentation would be really great. I would suggest holding off a bit till the design solidifies following PRs #1295 and #1307.

cpp/include/raft/distance/detail/pairwise_matrix/dispatch.cuh
```diff
@@ -0,0 +1,208 @@
...
+ *                   the KernelPolicy.
+ * @param f          Lambda that takes two std::integral_constant parameters representing
+ *                   row_major and vec_len.
+ */
+template <typename F>
+void dispatch(bool row_major, int vec_len, F&& f)
+{
+  if (row_major) {
+    switch (vec_len) {
+      case 4: f(std::bool_constant<true>(), vec_len_constant<4>()); break;
```
tfeher, line 74: Currently we assume that the input type is 4 bytes. What if we have `int8_t` input in the future? In that case `vec_len` can be larger than 4, but we would land in the default case. Maybe better to take `case 4` even in those cases? (Or just add a note here about this issue.)

cpp/include/raft/distance/detail/pairwise_matrix/dispatch.cuh
```diff
@@ -0,0 +1,208 @@
...
+  size_t byte_alignment = min(align_x, align_y);
+
+  // Since alignment is in bytes, it could be smaller than sizeof(DataT).
+  // Handle this (unlikely) case here.
+  RAFT_EXPECTS(sizeof(DataT) <= byte_alignment,
+               "Input matrix must be aligned to size of elements.");
+
+  // Compute number of elements that can be loaded in one instruction
+  // without causing misalignent errors.
+  int vec_len_aligned = (byte_alignment % sizeof(DataT) == 0) ? byte_alignment / sizeof(DataT) : 1;
```
tfeher, line 194: Shall we have a helper function to avoid repetition?

cpp/include/raft/distance/detail/pairwise_matrix/kernel_sm60.cuh
```diff
@@ -0,0 +1,135 @@
+/*
```
tfeher, line 1: Why the suffix `_sm60`?

cpp/include/raft/distance/detail/pairwise_matrix/kernel_sm60.cuh
```diff
@@ -0,0 +1,135 @@
...
+                                                                              IdxT lda,
+                                                                              IdxT ldb,
+                                                                              IdxT ldd,
+                                                                              OutT* dOutput,
+                                                                              opT distance_op,
+                                                                              FinOpT fin_op)
+{
+  extern __shared__ char smem[];
+
+  // Wrap operator back into lambdas. This is temporary and should be removed. (TODO)
```
tfeher, line 48: This is done to avoid changing  `PairwiseDistances`?

tfeher (changes requested):
Thanks Allard for this PR! This looks great! I am very happy to see the reduction of code duplication. I have just a few smaller questions.

cpp/include/raft/distance/detail/distance_ops/correlation.cuh
```diff
@@ -0,0 +1,122 @@
...
+    __syncthreads();
+
+#pragma unroll
+    for (int i = 0; i < Policy::AccRowsPerTh; ++i) {
+      regx2n[i] = sx2Norm[i * Policy::AccThRows + (threadIdx.x / Policy::AccThCols)];
+    }
+#pragma unroll
+    for (int i = 0; i < Policy::AccColsPerTh; ++i) {
+      regy2n[i] = sy2Norm[i * Policy::AccThCols + (threadIdx.x % Policy::AccThCols)];
+    }
```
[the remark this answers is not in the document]
ahendriksen: That would be ideal indeed. Unfortunately that would require constructing an instance of the PairwiseDistances struct. In the future, we can hopefully also refactor pairwise distance base to make use of free functions instead of inheritance. Then this would be feasible (perhaps some care must be taken to avoid circular dependencies?)

cpp/include/raft/distance/detail/distance_ops/kl_divergence.cuh
```diff
@@ -0,0 +1,94 @@
...
+  // some ops such as correlation_distance_op use more.
+  template <typename Policy>
+  constexpr size_t shared_mem_size()
+  {
+    return Policy::SmemSize;
+  }
+
+  DI void core(AccT& acc, DataT& x, DataT& y) const
+  {
+    // TODO: make sure that these branches get hoisted out of main loop.. Could
```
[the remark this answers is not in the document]
ahendriksen: No, but we have a lot of work coming up that will require benchmarking to make sure we do not have regressions. If the performance of KL divergence is terrible. It will come up.

cpp/include/raft/distance/detail/distance_ops/correlation.cuh
```diff
@@ -0,0 +1,122 @@
...
+  {
+    // Note how we can sneakily get a pointer to shared memory here, to store
+    // more data. If the implementation of PairwiseDistanceMatKernel ever
+    // changes, this will be where we find the bugs.
+    extern __shared__ char smem[];
+
+    DataT regx2n[Policy::AccRowsPerTh], regy2n[Policy::AccColsPerTh];
+
+    DataT* sx2Norm =
+      (DataT*)(&smem[Policy::SmemSize + (Policy::Mblk + Policy::Nblk) * sizeof(DataT)]);
```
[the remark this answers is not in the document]
ahendriksen: Yes, that is where the "norms" are stored that are loaded by "load_norms". This data is produced here: https://github.com/rapidsai/raft/blob/e7a8e89be466f898a9c446f4a9f60a38e2f40b33/cpp/include/raft/distance/detail/distance.cuh#L162

cpp/include/raft/distance/detail/distance_ops/correlation.cuh
```diff
@@ -0,0 +1,122 @@
...
+  template <typename Policy>
+  DI void epilog(AccT acc[Policy::AccRowsPerTh][Policy::AccColsPerTh],
+                 DataT* regxn,
+                 DataT* regyn,
+                 IdxT gridStrideX,
+                 IdxT gridStrideY) const
+  {
+    // Note how we can sneakily get a pointer to shared memory here, to store
+    // more data. If the implementation of PairwiseDistanceMatKernel ever
+    // changes, this will be where we find the bugs.
```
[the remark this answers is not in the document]
ahendriksen: I agree. It would be even better if the shared memory layout is defined in a struct. This will require some refactoring of pairwise_distance_base though, which is extremely difficult to get right.

cpp/include/raft/distance/detail/pairwise_matrix/dispatch.cuh
ahendriksen: I have added a note.

cpp/include/raft/distance/detail/pairwise_matrix/kernel_sm60.cuh
ahendriksen: I am planning to add a version for Hopper (ie SM90). This mimics the code organization in CUTLASS. Since SM60 is the minimum supported version of RAFT, I named the kernel SM60.

cpp/include/raft/distance/detail/pairwise_matrix/dispatch.cuh
ahendriksen: Sure! The header is split up to improve compile times in PR #1307 . I will pick it up there.

tfeher (approved):
Thanks Allard for the updates! The PR looks good to me!

cpp/include/raft/distance/detail/distance_ops/template.cuh
cjnolet: Can you open an issue for this and reference the issue in the code please? That way we don't lose sight of it with everything else going on.

cpp/include/raft/distance/distance.cuh
cjnolet: Created an issue to keep this one on the radar: https://github.com/rapidsai/raft/issues/1320

cpp/include/raft/distance/detail/pairwise_matrix/kernel_sm60.cuh
cjnolet: Rather than just putting a todo in the code, can we create a tracking issue for this and reference it here in the code please?

ahendriksen: I have added issue https://github.com/rapidsai/raft/issues/1323.

cjnolet:
Just a heads up- we need to merge [this](https://github.com/rapidsai/cuml/pull/5264) cuml PR before this RAFT PR is merged, otherwise cuml will break downstream.

ahendriksen:
Good catch. I should be more careful with the non-breaking label. Although that file has been deprecated for a couple of months now.

cjnolet:
> Good catch. I should be more careful with the non-breaking label. Although that file has been deprecated for a couple of months now.
No problem. To be fair, we had updated cuml in 23.02 to remove most of its uses of deprecated headers but this one slipped through the cracks. Before we merged this I just wanted to do a quick grep to be sure.

cjnolet (approved):
LGTM!

Issue #555, opened 2022-03-14
[FEA] Remove deprecated files


cjnolet:
There are several files which have been deprecated over the 22.04 release to give us an opportunity to update consuming projects before we remove them.

We should remove them in either the 22.06 or 22.08 release cycles. These files are marked with "@deprecated" tags.

PR #1145, opened 2023-01-16
Remove deprecated headers

lowener:
Related to #555 
The previous PR (#1041) was too old and impacted too many files. This one is simpler and address most of the deprecated files.


cjnolet (approved):
LGTM!

Issue #1025, opened 2022-11-17
[ENH] Deprecate `myMax` and similar functions


Nyrio:
Follows [this discussion](https://github.com/rapidsai/raft/pull/979#discussion_r1024464274). @achirkin and @cjnolet can perhaps say more about the reasons behind deprecating these functions and why they existed in the first place?

Another concern: right now we can't really compile with Clang anyway, but I seem to remember using `std::max` in device functions doesn't compile with Clang?

achirkin:
`std::max` is a constexpr [since C++14](https://en.cppreference.com/w/cpp/algorithm/max). Could it have something to do with that?

ahendriksen:
I believe that `std::max` and `fmaxf` are not defined the same way. 

In any case, they produce different instruction sequences. 

eg:
```
#include <algorithm>

typedef float T;

__global__ void kernel(T* x, T* y, T* z) {
    z[0] = std::max(x[0], y[0]);
    __syncthreads();
    z[1] = fmaxf(x[1], y[1]);
}
```

Produces 

```
// std::max
 FSETP.GEU.AND P0, PT, R0, R9, PT 
 FSEL R9, R9, R0, !P0 
// fmaxf:
 FMNMX R11, R0, R11, !PT 
```

See: https://godbolt.org/z/YnocKz4j1

Nyrio:
@ahendriksen You are correct. NVCC adds some CUDA intrinsics to the `std` namespace so that they are defined on device (such as `sqrt`), in which case the generated code would be exactly the same. But that's not the case for `std::max` which behaves differently from the intrinsic (and can be used in device functions due to being defined as `constexpr`).

I'm leaning towards using the `std::` versions on host and the CUDA intrinsics on device, for Clang compatibility, using conditional compilation, e.g:

```c++
template <typename T>
constexpr RAFT_INLINE_FUNCTION auto sqrt(const T& x)
{
    #ifdef __CUDA_ARCH__
    return ::sqrt(x);
    #else
    return std::sqrt(x);
    #endif
}
```

Nyrio:
Btw, the [Math API documentation](https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH.html#group__CUDA__MATH) says:

> Note also that due to implementation constraints, certain math functions from std:: namespace may be callable in device code even via explicitly qualified std:: names. However, such use is discouraged, since this capability is unsupported, unverified, undocumented, not portable, and may change without notice.

[...]

PR #1146, opened 2023-01-16
Improvement of the math API wrappers

Nyrio:
Solves #1025 

Provides a centralized collection of host- and device-friendly wrappers around common math operations, with generalizations when useful. Deprecates former `myXxx` wrappers.

Those wrappers are mostly intended to future-proof the API as well as simplify the definition of host-device functions.


cjnolet (changes requested):
So nice to see all these `my*` types being removed, the math functions consolidated, and now even tests for the use of the wrappers in both host and device functions. 

My feedback is mostly for documentation consistency, otherwise I think the changes look great.

cjnolet (approved):
Looks great! Thanks again for consolidating these

PR #1147, opened 2023-01-17
Catch signal handler change error

tfeher:
The interruptible signal handle was incompatible with dask, it would trigger the following error:
```
Exception: "ValueError('signal only works in main thread of the main interpreter')"
```

This PR fixes the problem by catching the error and keeping the original signal handler in that case.


tfeher:
Credits to @achirkin who provided the patch, thanks!

cjnolet (approved):

PR #1148, opened 2023-01-17
balanced-k-means: fix a too large initial memory pool size

achirkin:
`calc_minibatch_size` decides on the batch size under assumption that the workspace shouldn't exceed 1GB. It takes into account that fewer extra buffers are needed when the data type `T` is float. However, we don't take this into account when setting the initial memory pool size immediately after calculating `max_minibatch_size`. As a result, under some conditions, the algorithm attempts to allocate more memory than available. This PR sets the limit of the initial pool size to 1GB to fix the issue.

### DIFF
--- a/cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
+++ b/cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
@@ -18,9 +18,6 @@
 
 #include "ann_utils.cuh"
 
-#include <thrust/gather.h>
-#include <thrust/transform.h>
-
 #include <raft/cluster/detail/kmeans_common.cuh>
 #include <raft/common/nvtx.hpp>
 #include <raft/core/cudart_utils.hpp>
@@ -44,6 +41,11 @@
 #include <rmm/mr/device/device_memory_resource.hpp>
 #include <rmm/mr/device/managed_memory_resource.hpp>
 #include <rmm/mr/device/per_device_resource.hpp>
+
+#include <thrust/gather.h>
+#include <thrust/transform.h>
+
+#include <tuple>
 
 namespace raft::spatial::knn::detail::kmeans {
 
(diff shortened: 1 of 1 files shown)


achirkin:
Tagging @Nyrio for awareness, since you're working on balanced k-means.

Nyrio:
Wouldn't it be better to update the calculation of the initial pool size to avoid assuming that we need to store the distance matrix? Similar to how we calculate the batch size.

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
```diff
@@ -974,7 +974,9 @@ void build_hierarchical(const handle_t& handle,
   rmm::mr::device_memory_resource* device_memory = nullptr;
   IdxT max_minibatch_size =
     calc_minibatch_size(n_clusters, n_rows, dim, metric, std::is_same_v<T, float>);
-  auto pool_guard = raft::get_pool_memory_resource(device_memory, max_minibatch_size * dim * 4);
+  auto pool_guard = raft::get_pool_memory_resource(
+    device_memory,
+    std::min<size_t>(size_t(dim * 4) * max_minibatch_size, 1000lu * 1000lu * 1000lu));
```
tfeher, line 979: `calc_minibatch_size` could return the `size_per_row`. It would be better to use that in place of `(dim*4)`. That way, if we actually do not need a large pool, then we will not allocate that.

tfeher (changes requested):

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
achirkin: Yes, makes sense!

tfeher (approved):
Thanks Artem for the update, LGTM!

PR #1150, opened 2023-01-17
Remove faiss bfKnn call from fused_l2_knn unittest

benfred:


cjnolet (approved):
LGTM! Great to see a naive case that's completely decoupled from the brute-force API.

PR #1152, opened 2023-01-18
Fix various build errors

hcho3:
Fix various build errors I encountered as I tried to build RAFT locally on my workstation. (Command used: `./build.sh -g raft-dask pylibraft libraft tests bench --compile-libs`)

* Add `gtest` as a link dependency of the C++ benchmark suite, to fix the error
```
[266/332] Building CUDA object CMakeFiles/NEIGHBORS_BENCH.dir/bench/neighbors/refine.cu.o
FAILED: CMakeFiles/NEIGHBORS_BENCH.dir/bench/neighbors/refine.cu.o
In file included from /home/phcho/Desktop/raft/cpp/bench/neighbors/../../test/neighbors/../test_utils.cuh:19,
from /home/phcho/Desktop/raft/cpp/bench/neighbors/../../test/neighbors/ann_utils.cuh:28,
from /home/phcho/Desktop/raft/cpp/bench/neighbors/../../test/neighbors/refine_helper.cuh:18,
from /home/phcho/Desktop/raft/cpp/bench/neighbors/refine.cu:39:
/home/phcho/Desktop/raft/cpp/bench/neighbors/../../test/neighbors/../test_utils.h:22:10: fatal error:
gtest/gtest.h: No such file or directory
22 | #include <gtest/gtest.h>
   |          ^~~~~~~~~~~~~~~
compilation terminated. 
```

* Explicitly specify the namespace for `alignTo`.
* Cast pointers into an integral type prior to passing it to `alignTo`.
* When calling `areSameAlignOffsets()`, pass the underlying pointers of the mdspan objects. Passing an mdspan to `areSameAlignOffsets()` is an error.

### DIFF
--- a/cpp/bench/CMakeLists.txt
+++ b/cpp/bench/CMakeLists.txt
@@ -1,5 +1,5 @@
 # =============================================================================
-# Copyright (c) 2022, NVIDIA CORPORATION.
+# Copyright (c) 2022-2023, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
 # in compliance with the License. You may obtain a copy of the License at
--- a/cpp/bench/CMakeLists.txt
+++ b/cpp/bench/CMakeLists.txt
@@ -32,6 +32,7 @@ function(ConfigureBench)
     PRIVATE raft::raft
             $<$<BOOL:${ConfigureBench_DIST}>:raft::distance>
             $<$<BOOL:${ConfigureBench_NN}>:raft::nn>
+            GTest::gtest
             benchmark::benchmark
             Threads::Threads
             $<TARGET_NAME_IF_EXISTS:OpenMP::OpenMP_CXX>
--- a/cpp/include/raft/matrix/detail/linewise_op.cuh
+++ b/cpp/include/raft/matrix/detail/linewise_op.cuh
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021-2022, NVIDIA CORPORATION.
+ * Copyright (c) 2021-2023, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/cpp/include/raft/matrix/detail/linewise_op.cuh
+++ b/cpp/include/raft/matrix/detail/linewise_op.cuh
@@ -796,7 +796,7 @@ struct MatrixLinewiseOp {
                   "layout for in and out must be either padded row or col major");
 
     // also statically assert padded matrix alignment == 2^i*VecBytes
-    assert(raft::Pow2<VecBytes>::areSameAlignOffsets(in, out));
+    assert(raft::Pow2<VecBytes>::areSameAlignOffsets(in.data_handle(), out.data_handle()));
 
     if (alongLines)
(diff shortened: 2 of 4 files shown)


lowener (approved):
Nice fix, I have been bothered by some of these errors recently as well!

cpp/bench/CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ function(ConfigureBench)
     PRIVATE raft::raft
             $<$<BOOL:${ConfigureBench_DIST}>:raft::distance>
             $<$<BOOL:${ConfigureBench_NN}>:raft::nn>
+            GTest::gtest
```
achirkin, line 35: The benchmark shouldn't depend on gtest. I think, we should try to break the offending dependency instead (which is not so trivial).

cpp/include/raft/matrix/detail/linewise_op.cuh
```diff
@@ -796,7 +796,7 @@ struct MatrixLinewiseOp {
                   "layout for in and out must be either padded row or col major");
 
     // also statically assert padded matrix alignment == 2^i*VecBytes
-    assert(raft::Pow2<VecBytes>::areSameAlignOffsets(in, out));
+    assert(raft::Pow2<VecBytes>::areSameAlignOffsets(in.data_handle(), out.data_handle()));
```
achirkin, line 799: Normally, we use `RAFT_EXPECTS` instead of the plain `assert`. Seems like this one slipped through unnoticed. While you're at it, could you please fix this as well?

achirkin (changes requested):
Thanks for the PR!

cpp/bench/CMakeLists.txt
achirkin: ...Or maybe we could live with the gtest dependency in benchmarks while we solve the problem with the header dependencies? Any thoughts, @cjnolet ?

cjnolet: 100% agreed w/ @achirkin here. I was going to type up a similar comment but he beat me to it!

cjnolet: > The benchmark shouldn't depend on gtest. I think, we should try to break the offending dependency instead (which is not so trivial).

My previous comment was in response to this, btw. I don't think we should be crossing the gtest and benchmark dependencies. We should try and pull out the shared bits into a separate header (and maybe try to find a good place in the `include/raft/`?) for the shared bits but keep gtest-specific things in test and benchmark-specific things in bench.

cjnolet: I'm also okay opening a github issue about this and keeping the gtest dependency in here for now but adding a comment here that references the issue.

robertmaynard (approved):
Given that @cjnolet says we are good with opening an issue about refactoring the GTest dependency required by the benchmarks I am approving the CMake changes.

cpp/include/raft/matrix/detail/linewise_op.cuh
hcho3: Done

cpp/bench/CMakeLists.txt
hcho3: Opened #1153

achirkin (approved):

cjnolet (approved):

PR #1154, opened 2023-01-18
Use rapids-cmake new COMPONENT exporting feature

robertmaynard:
Switches raft over to the upcoming raft_export(`COMPONET`) support. 

This allows raft to not require to support custom exporting logic


robertmaynard:
> LGTM. I notice this is still a draft. Is it ready to go into 23.02 or are you still working on it?

Still a draft as the rapids-cmake feature hasn't been merged yet. I expect we are going to have to slip this to 23.04 given that rapids-cmake is in burndown

cjnolet:
@robertmaynard just pinging to see how you feel about this PR. Do you think this is going to make it for 23.04 or should we push?

robertmaynard:
> @robertmaynard just pinging to see how you feel about this PR. Do you think this is going to make it for 23.04 or should we push?

Aiming to have the rapids-cmake changes merged tomorrow

cjnolet (approved):

cpp/CMakeLists.txt
```diff
@@ -629,19 +594,31 @@ if(compiled IN_LIST raft_FIND_COMPONENTS)
...
-include(cmake/modules/raft_export.cmake)
-raft_export(
-  INSTALL raft COMPONENTS compiled distributed EXPORT_SET raft-exports GLOBAL_TARGETS raft compiled
-  distributed NAMESPACE raft:: DOCUMENTATION doc_string FINAL_CODE_BLOCK code_string
+rapids_export(
+  INSTALL raft
+  EXPORT_SET raft-exports
+  COMPONENTS ${raft_components}
+  COMPONENTS_EXPORT_SET ${raft_export_sets}
+  GLOBAL_TARGETS raft compiled distributed
```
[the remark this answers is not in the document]
robertmaynard: Making distributed a build time flag would mean we would need 2 conda builds to make packages. If the goal is to have everything generated in a single conda build we can't have components gated behind CMake options.

> Could we just always create the raft::compiled target so that it'll never actually build anything if a user never links against it?

So would that mean we never have a version of raft on conda that doesn't have pre-compiled libraries?

cpp/CMakeLists.txt
```diff
@@ -637,19 +602,31 @@ if(compiled IN_LIST raft_FIND_COMPONENTS)
 endif()
 ]=]
 )
+set(raft_components compiled distributed)
+set(raft_export_sets raft-compiled-exports raft-distributed-exports)
+if(TARGET raft_lib)
+  list(APPEND raft_components compiled)
+  list(APPEND raft_export_sets raft-compiled-lib-exports)
```
vyasr, line 609: At this point I'm a bit confused by the different targets and export sets that we have in raft. Why do we need both the `raft_compiled` and `raft_lib` targets? It used to be necessary for the nn/dist targets so that calling code could use either component independently as header-only or precompiled, but now why can't the `raft` target just be the header-only target and `raft_compiled` always be precompiled? And that stems from my other question, why do we need both `raft-compiled-exports` and `raft-compiled-lib-exports`?

cpp/CMakeLists.txt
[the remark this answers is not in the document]
vyasr: > So would that mean we never have a version of raft on conda that doesn't have pre-compiled libraries?

I think this would directly go against the goals of #1386

vyasr (commented):
The component changes themselves look good, but we raised some important questions about exactly what the packages should look like now that we need to resolve.

cpp/CMakeLists.txt
cjnolet: @vyasr from my understanding at this point, the answer is "we don't need both `raft_compiled` and `raft_lib` anymore, but at the current time they aren't really causing harm, so we aren't in an immediate rush to remove them". @robertmaynard please correct me if that assertion is wrong.

robertmaynard: The `raft_compiled` target is created even when `RAFT_COMPILE_LIBRARY` is turned off, so one solution for proper invariant behavior when generating the libraft-headers and libraft was to have the current organization.

If we are okay with making the `raft::compiled` target conditional all this logic can be removed and we could have a single target `raft_compiled` and export set ( `raft-compiled-exports` ). If we are good with this, I would be happy to open a PR that does these changes.

PR #1155, opened 2023-01-18
Make cutlass use static ctk

sevagh:
Cutlass links to the CTK in ways that cause problems for downstream pip wheel builds (especially in cugraph).

This might help.


robertmaynard (approved):

sevagh:
I used this branch (pull-request/1155) in the cugraph PR's PINNED_TAG for raft: https://github.com/rapidsai/cugraph/pull/3096 

That's the wheel (likely) suffering from this cutlass problem because it ends up without `libcusolver` symbols and fails to work.

PR #1135, opened 2023-01-11
IVF-PQ: tighten the test criteria

achirkin:
Make the recall reporting a bit more verbose and try to tighten the `min_recall` for various test cases. This should help spot any regressions in future and improve our understanding of ivf-pq performance for various inputs.

PR #1156, opened 2023-01-19
ANN tests: make the min_recall check strict

achirkin:
In #1135, we adjusted the min_recall values to report if any regressions happen in ivf-pq. However, `eval_neighbours` function, which is used in several ANN test suites, doesn't fail unless the regression is really large (it prints a warning if the calculated recall is "slightly" smaller than the expected recall).
In this PR, I make `eval_neighbours` always fail if the calculated recall is smaller than the expected recall. Slightly adjust the tests and do a small refactoring along the way.


achirkin:
NB: some of the affected algorithm are not deterministic due to concurrency etc.; I've successfully run the test suite on my workstation 35 times, but we should keep an eye on how it behaves on the CI machines as well.

cjnolet:
My recent merge commit broke the build. Just leaving a note here that I'll fix this.

achirkin:
Did a dozen tests on a100 and v100, seems to be passing with the latest adjustments.

PR #1160, opened 2023-01-19
Use squeuclidean for metric name in ivf_pq python bindings

benfred:
Use sqeuclidean instead of l2_expanded for the distance name in the ivf_pq python bindings. This matches both sklearn, and the RAFT pairwise_distance api - and should be less confusing for our users


cjnolet (approved):
LGTM!

PR #1161, opened 2023-01-20
Protect balanced k-means out-of-memory in some cases

achirkin:
There's no guarantee that our balanced k-means implementation always produces balanced clusters. In the first stage, when mesoclusters are trained, the biggest cluster can grow larger than half of all input data. This becomes a problem at the second stage, when in `build_fine_clusters`, the mesocluster data is copied in a temporary buffer. If size is too big, there may be not enough memory on the device. A quick workaround:

 1. Expand the error reporting (RAFT_LOG_WARN)
 2. Artificially limit the mesocluster size in the event of highly unbalanced clustering

### DIFF
--- a/cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
+++ b/cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
@@ -839,6 +839,10 @@ inline auto arrange_fine_clusters(uint32_t n_clusters,
  *  As a result, the fine clusters are what is returned by `build_hierarchical`;
  *  this function returns the total number of fine clusters, which can be checked to be
  *  the same as the requested number of clusters.
+ *
+ *  Note: this function uses at most `fine_clusters_nums_max` points per mesocluster for training;
+ *  if one of the clusters is larger than that (as given by `mesocluster_sizes`), the extra data
+ *  is ignored and a warning is reported.
  */
 template <typename T, typename IdxT, typename LabelT>
 auto build_fine_clusters(const handle_t& handle,
--- a/cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
+++ b/cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
@@ -1034,8 +1034,8 @@ void build_hierarchical(const handle_t& handle,
   auto [mesocluster_size_max, fine_clusters_nums_max, fine_clusters_nums, fine_clusters_csum] =
     arrange_fine_clusters(n_clusters, n_mesoclusters, n_rows, mesocluster_sizes);
 
-  const auto mesocluster_size_max_balanced =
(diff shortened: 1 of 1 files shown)


cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
```diff
@@ -880,8 +880,8 @@ auto build_fine_clusters(const handle_t& handle,
   uint32_t n_clusters_done = 0;
   for (uint32_t i = 0; i < n_mesoclusters; i++) {
     uint32_t k = 0;
-    for (IdxT j = 0; j < n_rows; j++) {
-      if (labels_mptr[j] == (LabelT)i) { mc_trainset_ids[k++] = j; }
+    for (IdxT j = 0; j < n_rows && k < mesocluster_size_max; j++) {
+      if (labels_mptr[j] == LabelT(i)) { mc_trainset_ids[k++] = j; }
     }
     if (k != mesocluster_sizes[i])
```
tfeher, line 886: I suggest we omit warning when `k` is limited due to `mesocluster_size_max`.
```suggestion
    if (k != mesocluster_sizes[i] && k < mesocluster_size_max)
```

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
```diff
@@ -1036,10 +1030,19 @@ void build_hierarchical(const handle_t& handle,
...
     arrange_fine_clusters(n_clusters, n_mesoclusters, n_rows, mesocluster_sizes);
 
-  if (mesocluster_size_max * n_mesoclusters > 2 * n_rows) {
-    RAFT_LOG_WARN("build_hierarchical: built unbalanced mesoclusters");
+  const auto mesocluster_size_max_balanced =
+    uint32_t(size_t{n_rows * 2lu} / std::max<size_t>(n_mesoclusters, 1lu));
+  if (mesocluster_size_max > mesocluster_size_max_balanced) {
+    RAFT_LOG_WARN(
+      "build_hierarchical: built unbalanced mesoclusters (max_mesocluster_size == %u > %u). "
+      "At most %u points will be used for training within each mesocluster. "
```
tfeher, line 1038: Please add this line to the docstring of build_fine_clusters

tfeher (approved):
Thanks Artem for the PR, this is a good workaround for the imbalanced training issue. LGTM.

tfeher (approved):
Thanks Artem for the PR, this is a good workaround for the imbalanced training issue. LGTM.

cpp/include/raft/spatial/knn/detail/ann_kmeans_balanced.cuh
achirkin: I think, it's better to leave the warning, because it also gives an extra information about which clusters are the offending clusters and emphasizes the importance of the previous warning about the unbalanced mesoclusters. In the logs, they go one after another and it's rather easy to see the link between the two.

PR #1165, opened 2023-01-24
Enable shallow copy of `handle_t`'s resource pointers with different workspace_resource 

cjnolet:
This effectively affords users the flexibility to shallow copy a handle and it's underlying vectors and change out only the `workspace_resource` so that they can, for example, configure multiple different pools or managed pools for workspace resources. 

cc @Nyrio RE: hierarchical k-means API.

### DIFF
--- a/cpp/test/core/handle.cpp
+++ b/cpp/test/core/handle.cpp
@@ -268,4 +268,35 @@ TEST(Raft, WorkspaceResource)
   delete pool_mr;
 }
 
+TEST(Raft, WorkspaceResourceCopy)
+{
+  auto stream_pool = std::make_shared<rmm::cuda_stream_pool>(10);
+
+  handle_t handle(rmm::cuda_stream_per_thread, stream_pool);
+
+  auto pool_mr = new rmm::mr::pool_memory_resource(rmm::mr::get_current_device_resource());
+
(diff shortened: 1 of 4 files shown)


cpp/test/core/handle.cpp
```diff
@@ -268,4 +268,77 @@ TEST(Raft, WorkspaceResource)
...
+  ASSERT_EQ(handle.get_stream_pool_size(), copied_handle.get_stream_pool_size());
+
+  // Sanity check to make sure non-corresponding streams are not equal
+  ASSERT_NE(handle.get_stream_pool().get_stream(0).value(),
+            copied_handle.get_stream_pool().get_stream(1).value());
+
+  for (size_t i = 0; i < handle.get_stream_pool_size(); ++i) {
+    ASSERT_EQ(handle.get_stream_pool().get_stream(i).value(),
+              copied_handle.get_stream_pool().get_stream(i).value());
+  }
```
divyegala, line 292: These lines could be separated out into a common function since they are shared by all three tests, right? Something like `test_handle_equivalence(handle_t const& original_handle, handle_t const& copied_handle)`

divyegala (approved):
Pre-approving, just a suggestion for the tests

PR #1169, opened 2023-01-24
Use CTK 118/cp310 branch of wheel workflows

sevagh:
This PR builds wheels using the cuda-118 branch of the workflows, which bumps CTK 11.5.1 to CTK 11.8.0 and cp39 to cp310.

### DIFF
--- a/python/raft-dask/setup.cfg
+++ b/python/raft-dask/setup.cfg
@@ -46,4 +46,4 @@ skip=
 
 [options]
 packages = find:
-python_requires = >=3.7,<=3.10
+python_requires = >=3.7
--- a/python/raft-dask/setup.cfg
+++ b/python/raft-dask/setup.cfg
@@ -46,4 +46,4 @@ skip=
 
 [options]
 packages = find:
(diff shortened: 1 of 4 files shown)


ajschmidt8 (approved):

python/raft-dask/setup.cfg
```diff
@@ -46,4 +46,4 @@ skip=
 
 [options]
 packages = find:
-python_requires = >=3.7,<3.10
+python_requires = >=3.7
```
bdice, line 49: Oops. Meant to update this in https://github.com/rapidsai/raft/pull/1120. I missed it because this is using both `setup.cfg` and `setup.py`. I didn't think of that possibility -- I only updated `setup.py` and `pyproject.toml` files in the CUDA 11.8 / Python 3.10 updates for conda packages.

This suggestion would align with the rest of RAPIDS.
```suggestion
python_requires = >=3.8,<3.11
```

cjnolet (approved):
LGTM

PR #1173, opened 2023-01-25
[REVIEW] Serializer for mdspans

### LINKED ISSUE #1017: [FEA] Tracker for mdspan serialization and copy
Following up with #752.

Design choices
* Provide efficient serializers for mdspan's with C and Fortran layouts. Mdspan's can have an arbitrary layouts which makes it hard to serialize efficiently.
* Provide a copy operator so that users can transform their mdspans into C or Fortran layouts.
  - In general, provide a generic fallback implementation (possibly suboptimal), and provide specialized versions of `copy` that are optimized for certain use cases.
* Adopt [the NumPy format](https://numpy.org/doc/1.23/reference/generated/numpy.lib.format.html) for easy inter-op with NumPy and other array libraries.
* Also expose `__array_interface__` for zero-copy serialization

TODO's

- [ ] Implement a copy operation between two mdspan's
  * [x] Implement fallback copy operation to copy elements between two mdspan's on the host memory
  * [ ] Implement fallback copy operation to copy elements between two mdspan's on the device memory
  * [ ] Implement wrappers for copying Fortran-, C-layout, or strided arrays between the host and device memory
- [ ] Implement serializer for mdspan's with Fortran and C layouts
  * [ ] Serialize to `std::ostream`
  * [ ] Serialize to container iterator
  * [ ] Serialize via `__array_interface__`

hcho3:
Closes #1017

Implement a serializer for mdspan using the NumPy format. Row- and column-major layouts are supported.

TODOs
- [x] Revise IVF to use the new mdspan serializer
- [x] Add gtest to cover serializing device mdspans
- [x] Implement serializer for scalars
- [x] Rename header to `raft/core/serialize.hpp`
- [x] Use `device_resources` instead of `handle_t`
- [x] Use 64-byte alignment
- [x] Serialize scalars as 0D NumPy array
- [x] Add version field to ANN serialization
- [x] Test mdspan serializer in the Python layer


cpp/include/raft/core/mdspan_serializer.hpp
```diff
@@ -0,0 +1,117 @@
+/*
+ * Copyright (c) 2023, NVIDIA CORPORATION.
```
cjnolet, line 2: What do you think about `raft/core/serialize.hpp`?

cpp/include/raft/spatial/knn/detail/ivf_flat_build.cuh
```diff
@@ -405,15 +406,15 @@ void save(const handle_t& handle, const std::string& filename, const index<T, Id
   write_scalar(of, index_.metric());
```
cjnolet, line 406: It would be nice if all of these `write_*` could be moved to `raft/core/serialize.hpp` (and renamed to `serialize_*` and `deserialize_*`).

cpp/include/raft/core/mdspan_serializer.hpp
hcho3: @cjnolet Do we have serializers for other RAFT types?

cjnolet: Not yet, but when we do (such as serializing scalars, key/value pairs, etc...) then we can stick them in this header.

hcho3:
@cjnolet I introduced a few macros to replace global constant variables, as the latter caused a link error due to symbol duplication. Is this okay? Macros are not hygienic and will not be scoped by namespaces. The alternative is to add new source file(s). What's the policy like for adding source files? Is RAFT meant to be mostly header only?

cjnolet:
> Macros are not hygienic and will not be scoped by namespaces

I think what you've done should be okay since they are at least prefixed w/ `RAFT_NUMPY` thus lowering the potential for collision. 

> What's the policy like for adding source files? Is RAFT meant to be mostly header only?

RAFT's APIs are all header-only and the source files are used for 1) template specializations that users can link against to avoid recompiling expensive headers, and 2) runtime APIs that can be accessed on host. This means the serializer code needs to be header-only at its core but if we find that the functions are expensive to compile, we can always add an explicit template specialization for them.

hcho3:
Got it, thanks for clarification

cpp/include/raft/core/serialize.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+namespace raft {
+
+template <typename ElementType, typename Extents, typename LayoutPolicy, typename AccessorPolicy>
+inline void serialize_mdspan(
+  const raft::handle_t& handle,
+  std::ostream& os,
+  const raft::host_mdspan<ElementType, Extents, LayoutPolicy, AccessorPolicy>& obj)
+{
+  static_assert(std::is_same_v<LayoutPolicy, raft::layout_c_contiguous> ||
```
tfeher, line 39: Oh, now I see that here are the asserts I was missing from numpy_serializer. I think it would make sense to move them closer to the actual code which relies on these assumptions.

cpp/include/raft/core/serialize.hpp
```diff
@@ -0,0 +1,142 @@
...
+  // Copy to host before serializing
+  // For contiguous layouts, size() == product of dimensions
+  std::vector<typename obj_t::value_type> tmp(obj.size());
+  cudaStream_t stream = handle.get_stream();
+  raft::update_host(tmp.data(), obj.data_handle(), obj.size(), stream);
+  using inner_accessor_type = typename obj_t::accessor_type::accessor_type;
+  auto tmp_mdspan =
+    raft::host_mdspan<ElementType, Extents, LayoutPolicy, raft::host_accessor<inner_accessor_type>>(
+      tmp.data(), obj.extents());
+  detail::numpy_serializer::serialize(handle, os, tmp_mdspan);
```
tfeher, line 65: Make sure the data has arrived
```
      handle.sync_stream();
```

tfeher (changes requested):
Hi @hcho3, thanks for this work! I am really excited to have this in raft! It looks great overall, I have just a few comments.

cpp/include/raft/core/detail/mdspan_numpy_serializer.hpp
```diff
@@ -0,0 +1,442 @@
...
+               version_major,
+               version_minor);
+}
+
+inline void write_header(std::ostream& os, const header_t& header)
+{
+  std::string header_dict     = header_to_string(header);
+  std::size_t preamble_length = RAFT_NUMPY_MAGIC_STRING_LENGTH + 2 + 2 + header_dict.length() + 1;
+  RAFT_EXPECTS(preamble_length < 255 * 255, "Header too long");
+  std::size_t padding_len = 16 - preamble_length % 16;
```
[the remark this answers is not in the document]
hcho3: My bad. I was looking at [an old version of the NumPy format](https://numpy.org/doc/1.13/neps/npy-format.html) and didn't realize they later [increased the alignment requirement to 64](https://github.com/numpy/numpy/pull/9025). Let me update the code.

tfeher (approved):
Thanks @hcho3 for the update! The PR looks good to me.

hcho3:
Can I get another round of review?

cpp/include/raft/spatial/knn/detail/ivf_pq_build.cuh
```diff
@@ -1365,7 +1367,12 @@ auto build(raft::device_resources const& handle,
   }
 }
 
-static const int serialization_version = 1;
+// Serialization version 2
+// No backward compatibility yet; that is, can't add additional fields without breaking
+// backward compatibility.
+// TODO(hcho3) Implement next-gen serializer for IVF that allows for expansion in a backward
+//             compatible fashion.
+static const int serialization_version = 2;
```
achirkin, line 1376: We've already had a problem with ivf-pq once, that `index` members changed but we forgot to update the load/save functions. Would be nice to detect these changes when updating the code, but I'm not aware of any good way to do this in C++.
As a very crude approximation, what do you think about adding something like this to `load` and `save` functions?
```C++
static_assert(sizeof(ivf_pq::index<uint64_t>) == {literal}, "The size of the index struct has changed since the last update; paste in the new size and consider updating the save/load logic");
```

cpp/include/raft/spatial/knn/detail/ivf_pq_build.cuh
```diff
@@ -1365,7 +1367,12 @@ auto build(raft::device_resources const& handle,
...
```
achirkin, line 1375: Maybe also add a `constexpr` while you're at it?

cpp/include/raft/core/mdspan.hpp
```diff
@@ -33,6 +33,9 @@ template <typename ElementType,
           typename AccessorPolicy = std::experimental::default_accessor<ElementType>>
 using mdspan = std::experimental::mdspan<ElementType, Extents, LayoutPolicy, AccessorPolicy>;
 
+template <class IndexType, std::size_t Rank>
+using dextents = std::experimental::dextents<IndexType, Rank>;
```
achirkin, line 37: This seems to be only used in tests, is it? If so, do we need to re-export this definition?

achirkin (changes requested):
I'm excited finally to get this in raft! A few minor things in the in the comments.

Also, I'm concerned a little bit regarding the naming. Atm we use `serialize_xx`, `deserialize_xx` and also `load`,`save`. This looks a bit inconsistent. Would you consider changing all of these to `serialize_xx`/`deserialize_xx`? Or even just `serialize`/`deserialize` (and rely on overload machinery)?

cpp/include/raft/core/detail/mdspan_numpy_serializer.hpp
```diff
@@ -0,0 +1,487 @@
...
+};
+
+template <class T>
+struct is_complex : std::false_type {
+};
+template <class T>
+struct is_complex<std::complex<T>> : std::true_type {
+};
+
+template <typename T, typename std::enable_if_t<std::is_floating_point_v<T>, bool> = true>
```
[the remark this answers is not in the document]
hcho3: Good point, we'd need to create a specialization specifically for `half` type.

hcho3:
> Would you consider changing all of these to `serialize_xx`/`deserialize_xx`? Or even just `serialize`/`deserialize` (and rely on overload machinery)?

I'm in favor of `serialize_xx`/`deserialize_xx`, because using the same `serialize` name for both scalars and mdspans may confuse the compiler and cause a surprising behavior in overload resolution. That is, when we declare
```
template <typename T>
void serialize(std::ostream& os, const T& value)
```
to serialize scalars, the function may be accidentally called with an mdspan (by setting `T=mdspan<...>`). 

I have rename all functions to `serialize_xxx` / `deserialize_xxx`.

cpp/include/raft/spatial/knn/detail/ivf_pq_build.cuh
hcho3: Done

achirkin (approved):
LGTM!

cjnolet (approved):
I’m pre-approving so we don’t need to hold up the release for these but just a couple small things we should probably open a follow-on for (in 23.04)

cpp/include/raft/core/serialize.hpp
```diff
@@ -0,0 +1,167 @@
...
+
+/**
+ * Collection of serialization functions for RAFT data types
+ */
+
+namespace raft {
+
+template <typename ElementType, typename Extents, typename LayoutPolicy, typename AccessorPolicy>
+inline void serialize_mdspan(
+  const raft::device_resources&,
```
divyegala, line 35: Can we use the base-class `raft::resources` for all these functions? `device_resources` is not needed with `host_mdspan` overloads

divyegala (approved):
Pre-approving, the PR looks really great! Just one comment, could you please file an issue or add a TODO for my comment to be tackled later?

raydouglass (approved):

PR #1188, opened 2023-01-26
Add function to convert mdspan to a const view

lowener:
`make_const_mdspan` is a helper function to convert `mdspan<T>` into `mdspan<const T>`.
I added examples of it's usage
@mhoemmen @Nyrio


Nyrio (approved):
Thanks for this!

cpp/include/raft/core/mdspan.hpp
```diff
@@ -304,4 +304,24 @@ RAFT_INLINE_FUNCTION auto unravel_index(Idx idx,
...
 
+/**
+ * @brief Create a copy of the given mdspan with const element type
+ * @tparam mdspan_type Expected type raft::host_mdspan or raft::device_mdspan
+ * @param mds raft::host_mdspan or raft::device_mdspan object
+ * @return raft::host_mdspan or raft::device_mdspan with vector_extent
+ *         depending on AccessoryPolicy
+ */
+template <typename mdspan_type, typename = enable_if_mdspan<mdspan_type>>
+auto make_const_mdspan(mdspan_type mds)
```
cjnolet, line 315: This is great! Can you add the corresponding `doxygenfunction` to the docs (`docs/cpp_api/mdspan_mdspan.rst`) as well?

mhoemmen:
Thanks for writing this!

This facility serves an analogous purpose as `std::view::as_const`.

https://en.cppreference.com/w/cpp/ranges/as_const_view

The use case example here

https://en.cppreference.com/w/cpp/ranges/constant_range

looks very much like the use cases you would like to support.

cpp/include/raft/core/mdspan.hpp
```diff
@@ -304,4 +304,24 @@ RAFT_INLINE_FUNCTION auto unravel_index(Idx idx,
...
+{
+  using const_element_t = std::add_const_t<typename mdspan_type::element_type>;
+  using const_accessor_t =
```
mhoemmen, line 318: The definition of `const_accessor_t` unexpectedly type-erases the user's accessor type.  This could be surprising to users, and it could also prevent valid code from building.  For example, users might write their own accessor `my_custom_accessor`, and a function `device_function` that takes it.

```c++
template<class T>
void device_function(mdspan<const T, dextents<size_t, 2>, layout_left, my_custom_accessor<const T>> m);
```

`make_const_mdspan` would unexpectedly change the accessor.  This would break stdblas, for example, since it uses custom accessors.  Users might also be surprised that their `device_accessor` gets type-erased into `host_device_accessor`.

The fix is to add a customization point `accessor_of_const`, which defines how to get an accessor that is the "const version" of a given accessor.

```c++
template<class ElementType>
default_accessor<std::add_const_t<ElementType>>
accessor_of_const(default_accessor<ElementType> a)
{
  return {a};
}

template<class ElementType>
host_accessor<default_accessor<std::add_const_t<ElementType>>>
accessor_of_const(host_accessor<default_accessor<ElementType>> a)
{
  return {a};
}

template<class ElementType>
device_accessor<default_accessor<std::add_const_t<ElementType>>>
accessor_of_const(device_accessor<default_accessor<ElementType>> a)
{
  return {a};
}

template<class ElementType>
host_device_accessor<default_accessor<std::add_const_t<ElementType>>>
accessor_of_const(host_device_accessor<default_accessor<ElementType>> a)
{
  return {a};
}
```

Users could specialize this for their own accessor types.  Then, `make_const_mdspan` would be defined like this.

```
auto make_const_mdspan(mdspan<ElementType, Extents, Layout, Accessor> mds)
{
  auto acc_c = accessor_of_const(mds.accessor()); 
  return mdspan<std::add_const_t<ElementType>, Extents, Layout, decltype(acc_c)>{
    mds.data_handle(),
    mds.mapping(),
    acc_c};
}
```

mhoemmen: If I remember right, `host_device_accessor` is just an alias to some type that has `bool` template parameters for "host accessible" and "device accessible"; if so, that would make the specialization logic simpler.

The point is that you only specialize `accessor_of_const` for accessor types where you know how to get the accessor-of-const.  Not every accessor type is required to have a natural accessor-of-const conversion.  If it does, it may not necessarily be exposed as a converting constructor.  (Think of accessors that view open files or remote PGAS memory, for example.)

cjnolet (approved):

cpp/include/raft/core/mdspan.hpp
```diff
@@ -304,24 +304,53 @@ RAFT_INLINE_FUNCTION auto unravel_index(Idx idx,
...
+}
+
 /**
  * @brief Create a copy of the given mdspan with const element type
- * @tparam mdspan_type Expected type raft::host_mdspan or raft::device_mdspan
- * @param mds raft::host_mdspan or raft::device_mdspan object
- * @return raft::host_mdspan or raft::device_mdspan with vector_extent
- *         depending on AccessoryPolicy
+ *
+ * @tparam ElementType the data type of the mdspan elements
```
mhoemmen, line 340: Optionally, would you consider the following small documentation change?

```suggestion
 * @tparam ElementType the const-qualified data type of the mdspan elements
```

"The data type of the mdspan elements" is `mdspan::value_type`.  The "element type" (`mdspan::element_type`) is the cv-qualified data type.  The "v" part of "cv qualified" isn't likely to be relevant, so I didn't mention it above.

cpp/include/raft/core/mdspan.hpp
```diff
@@ -304,24 +304,53 @@ RAFT_INLINE_FUNCTION auto unravel_index(Idx idx,
...
+ * @tparam MemType the type of memory where the elements are stored.
+ * @param a host_device_accessor
+ * @return host_device_accessor<std::experimental::default_accessor<std::add_const_t<ElementType>>, MemType> 
+ */
+template<class ElementType, memory_type MemType>
+host_device_accessor<std::experimental::default_accessor<std::add_const_t<ElementType>>, MemType>
+accessor_of_const(host_device_accessor<std::experimental::default_accessor<ElementType>, MemType> a)
+{
+  return {a};
```
mhoemmen, line 334: This is the correct way to do it, as long as `host_device_accessor` has an element type converting constructor like `default_accessor` does (see [the first constructor here](http://eel.is/c++draft/mdspan.accessor.default.members)).  A straightforward unit test would fail to compile without that constructor.

mhoemmen (approved):
Thanks for considering this design change!  I find that this makes the code clean and easy to extend.

Optionally, if you want users to be able to add support for their custom accessors, you might want to consider a customization point object or `tag_invoke` design for `accessor_of_const`.

cpp/include/raft/core/mdspan.hpp
cjnolet: This is a great point, @mhoemmen. It's nice to see the new function being used in the k-means tests but we *should* probably have a dedicated testcase for this (ideally in the mdspan cpp test file).

mhoemmen (approved):

PR #1190, opened 2023-01-26
Fix for svd API

lowener:


cpp/include/raft/linalg/svd.cuh
```diff
@@ -332,38 +340,52 @@ void svd_eig(
...
          U.has_value(),
          handle.get_stream());
 }
 
+template <typename ValueType, typename IndexType, typename UType>
+void svd_eig(const raft::handle_t& handle,
+             raft::device_matrix_view<const ValueType, IndexType, raft::col_major> in,
+             raft::device_vector_view<ValueType, IndexType> S,
+             raft::device_matrix_view<ValueType, IndexType, raft::col_major> V,
+             UType&& U = std::nullopt)
```
cjnolet, line 367: Can you add a quick test-case to the googletests to make sure the optional `U` input is working? (This will also help us ensure we don't hit regressions should anything change in the future).

cjnolet (changes requested):
Looks good. Just need a test to verify the nullopt case is working.

cjnolet (approved):

PR #1192, opened 2023-01-26
Expose raft::handle_t in the public header

vyasr:
#1140 migrated raft to using `device_resources` instead of `handle_t` everywhere internally, but it was not intended to be a breaking change since `handle_t` inherits from `device_resources`. However, `include/handle.hpp` was changed to include `device_resources.hpp` instead of `core/handle.hpp`, which is a breaking change since it means that downstream projects now need to include `core/handle.hpp` directly if they want access to `handle_t`. While that change may be desired eventually, I do not think that enforcing that requirement was an intended consequence of the change right now.


vyasr:
Both cuML and cuGraph CI is blocked on this change, as can be seen in some of the failures on https://github.com/rapidsai/cuml/pull/5165 (see [this build log](https://github.com/rapidsai/cuml/actions/runs/4016685951/jobs/6900089612)) and https://github.com/rapidsai/cugraph/pull/3189 (see [this build log](https://github.com/rapidsai/cugraph/actions/runs/4016644675/jobs/6901190612)).

vyasr:
OK, it looks like both the cuML and cuGraph failures are actually due to slightly different problems. cuML was relying on transitive includes in raft that were no longer applicable after the change, while cuGraph has an embedded copy of some raft Cython code that needed updating for the changes to raft. The change in this PR still seems helpful and necessary for other users of raft, but it no longer appears to be a blocker for the other RAPIDS libraries.

cjnolet:
@vyasr I do agree that this is a super useful change and I certainly don't mind merging it.

cjnolet (approved):

Issue #1187, opened 2023-01-26
[FEA] Possible improvements for sparse KNN


viclafargue:
Possible improvements for sparse KNN, some of which might be necessary to implement MNMG sparse KNN :
- ~Provide the possibility to specify different types for indices of sparse matrix inputs, and indices of dense nearest neighbors output~
- Allow multiple index inputs to be provided and the KNN results to be merged [in similar fashion to what is possible in the dense case](https://github.com/rapidsai/cuml/blob/72726933810b8b93d27e3f0c6fef5d3351858418/cpp/src/knn/knn_opg_common.cuh#L431-L434)
- Fix [issue for large batch sizes](https://github.com/rapidsai/raft/issues/1104)

cjnolet:
Hey @viclafargue these improvements look useful. Are you also planning to implement them as part of the sparse knn update work?

viclafargue:
Hey @cjnolet, I've already been working on making sparse KNN compile with `int64_t`, and also looked into ways to have different types for inputs and outputs indices. However, sparse KNN is quite complex and I feel like RAFT experts would be better suited to work on the two last points mentioned above. That said, if no one is available I can work on these.

cjnolet:
@viclafargue yes it would be very helpful if you have cycles to take these on. As you probably know, we are are short in resources across the board, but when issues like this arise in cuml it's somewhat expected that the corresponding raft pieces will be updated as well.

cjnolet:
@viclafargue, I'm reivewing your PR and I'm trying to think through why this is necessary:
```
Provide the possibility to specify different types for indices of sparse matrix inputs, and indices of dense nearest neighbors output
```

Can you explain the need for this?

viclafargue:
Sorry @cjnolet, I missed your question. This is actually something optional, I was thinking that we could systematically return indices as int64 as with the dense algorithm. But, when I think of it, compiling more templates is probably not the best option. The thing is, input indices could be int32 or int64, and it's getting complicated to create comms for MNMG KNN to send both datatypes. We should probably just systematically convert input indices to int64 in the MNGM KNN Python layer and only allow int64 inputs from the MNGM KNN C++ API.

viclafargue:
Closing this issue as it does not seem to be relevant anymore.

PR #1197, opened 2023-01-27
Sparse KNN improvements

viclafargue:
Answers #1187


cjnolet:
@viclafargue this is a breaking change for users downstream. Do we have a corresponding PR w/ cuML to test these changes?

cjnolet:
@viclafargue how are we feeling about these changes for 23.06?

viclafargue:
> @viclafargue how are we feeling about these changes for 23.06?

The PR would need a bit more work at the moment. I don't know if it could be part of 23.06.

cjnolet:
@viclafargue, sounds good, I'll push to 23.08.

PR #1198, opened 2023-01-27
[HELP-REQ] Expose KMeans `init_plus_plus` in pylibraft

### LINKED ISSUE #1166: [FEA] Expose KMeans cluster initialisation to Python
Expose kmeans++ cluster initialisation in pylibraft so that it can be called from Python. Maybe there is no need to have a random cluster init, `cupy.random.permutation()` might be enough.

The use case is to call the raft version of kmeans++ instead of using what is in scikit-learn. This means the arguments passed to https://github.com/scikit-learn/scikit-learn/blob/9b537396488a70de06e00de2330c4f00d7db1fc1/sklearn/cluster/_kmeans.py#L72 are available when calling the init function/having a similar function signature would be convenient.

**Additional context**
Recently there was a discussion about "k-means++", "greedy k-means++" and what is actually implemented (and what not) in https://github.com/scikit-learn/scikit-learn/issues/24973. Does the raft version match/can be configured to match "greedy k-means++"?

WDYT?

betatim:
Closes #1166

This adds `init_plus_plus` to pylibraft.

Three things to discuss:
1. can we come up with a better name?
2. what size should the `workspace` be? I couldn't quite work out if the size mattered or not.
3. I tried to add a `@property.setter` to `KMeansParams` so that I can set the `seed` after constructing it. But that leads to an error from cython(?). Does anyone know more about how to do this?


cjnolet (changes requested):
Thanks for opening this PR @betatim! Your impl looks quite clean and concise. Just a couple very small things.

python/pylibraft/pylibraft/cluster/kmeans.pyx
```diff
@@ -199,6 +200,52 @@ def compute_new_centroids(X,
         raise ValueError("dtype %s not supported" % x_dt)
 
 
+@auto_sync_handle
+@auto_convert_output
+def init_plus_plus(X, n_clusters, seed=None, handle=None):
```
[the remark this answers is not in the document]
betatim: Done. I made it so that it is an error to pass `n_clusters` and `centroids`.

python/pylibraft/pylibraft/cluster/kmeans.pyx
```diff
@@ -202,24 +202,33 @@ def compute_new_centroids(X,
 
 @auto_sync_handle
 @auto_convert_output
-def init_plus_plus(X, n_clusters, seed=None, handle=None):
+def init_plus_plus(X, n_clusters=None, seed=None, handle=None, centroids=None):
+    if n_clusters is not None and centroids is not None:
```
cjnolet, line 206: Rather than failing if they are both specified, could we just fail here when `n_clusters != len(centroids)`? That way as long they match, it's okay to specify both.

betatim: I'd be -1 on that. For me a caller that passes in both is signaling that they made a mistake or are confused (A bit like indicating left while being in a right turn lane with your car). The other thing I can't work out is when someone would use this possibility?

cpp/src/distance/cluster/kmeans_init_plus_plus_double.cu
```diff
@@ -0,0 +1,33 @@
...
+
+namespace raft::runtime::cluster::kmeans {
+
+void init_plus_plus(raft::device_resources const& handle,
+                    const raft::cluster::kmeans::KMeansParams& params,
+                    raft::device_matrix_view<const double, int> X,
+                    raft::device_matrix_view<double, int> centroids)
+{
+  // XXX what should the size of this be?
+  rmm::device_uvector<char> workspace(10, handle.get_stream());
```
betatim, line 29: What is a good size for this temporary workspace? Does it even matter or is it anyway resized to the size that is needed inside the function(s) that use it?

cjnolet: In general we have been working to remove the temporary workspace arguments from our public APIs in favor of using the "workspace_resource" in the device_resources instance. However, we can keep this one the way it is in the meantime until we scrape through and clean it up. The device_uvector is an RAII compliant (owns the memory, can resize it, and frees it when destructed) so you can just pass in a 0 size workspace and it'll get resized accordingly inside the init_plus_plus function.

python/pylibraft/pylibraft/cluster/kmeans.pyx
cjnolet: From a user experience perspective, I'd prefer to try to handle any accidental mistakes by providing the expected behavior (when we know for sure their intention) without raising an error when possible.

The `centroids` argument is specifically to allow users to pre-allocate their output buffer up front instead of having us do the allocation inside this function. Yes, we can infer the number of clusters from this output buffer, but if specified, the output buffer is the source of truth for the number of centroids.

What I'm suggesting is that in the case where they've done "a little extra" by specifying both centroids argument and n_clusters, there's no harm done if n_clusters does actually match the number of centroids they've passed in. All they've done is provided some redundant information. In this case I can't see a reason for failing outright when the information they've provided is still correct.

Now, if they've specified both arguments and n_clusters doesn't match the number of centroids, then that's a different story. In that case the entropy is high- they've clearly gotten confused or done something wrong inadvertently.

cjnolet (approved):
LGTM pending CI

PR #1200, opened 2023-01-27
Pin wheel dependencies to same RAPIDS release

sevagh:
Now that wheel CI is enabled, the nightly index will contain next release wheels, so we need to ensure those don't get installed while still building this release.

Also tests a new mechanism (`local-wheelhouse`) for cibuildwheel to prefer local wheels.


ci/release/update-version.sh
```diff
@@ -56,3 +57,15 @@ sed_runner "/^ucx_py_version:$/ {n;s/.*/  - \"${NEXT_UCX_PY_VERSION}\"/}" conda/
...
+CURRENT_SHORT_TAG_PEP440=$(python -c "from setuptools.extern import packaging; print(packaging.version.Version('${CURRENT_SHORT_TAG}')")
+NEXT_SHORT_TAG_PEP440=$(python -c "from setuptools.extern import packaging; print(packaging.version.Version('${NEXT_SHORT_TAG}')")
+CURRENT_UCX_PY_VERSION_PEP440=$(python -c "from setuptools.extern import packaging; print(packaging.version.Version('${CURRENT_UCX_PY_VERSION}')")
+NEXT_UCX_PY_VERSION_PEP440=$(python -c "from setuptools.extern import packaging; print(packaging.version.Version('${NEXT_UCX_PY_VERSION}')")
+
+# Wheel builds install intra-RAPIDS dependencies from same release
+sed_runner "s/${CURRENT_SHORT_TAG_PEP440}/${NEXT_SHORT_TAG_PEP440}/g" python/pylibraft/setup.py
+sed_runner "s/${CURRENT_SHORT_TAG_PEP440}/${NEXT_SHORT_TAG_PEP440}/g" python/pylibraft/_custom_build/backend.py
+sed_runner "s/${CURRENT_SHORT_TAG_PEP440}/${NEXT_SHORT_TAG_PEP440}/g" python/raft-dask/setup.py
+sed_runner "s/${CURRENT_UCX_PY_VERSION_PEP440}/${NEXT_UCX_PY_VERSION_PEP440}/g" python/raft-dask/setup.py
```
ajschmidt8, line 71: I think it's best to avoid using any kind of `CURRENT_SHORT_TAG` reference in the search term for these lines because if the source files become out of sync (which happens occasionally) then these `sed` commands will fail.

I would recommend making the search terms more specific.

ajschmidt8: BTW, you can test these lines with commands like: `bash ci/release/update-version.sh 23.04.00`.

I usually try to run a sequence of commands like this to ensure that everything works as expected:


```sh
bash ci/release/update-version.sh 23.06.00
bash ci/release/update-version.sh 22.06.00
bash ci/release/update-version.sh 23.02.00
```

ajschmidt8 (approved):

cjnolet (approved):

PR #1215, opened 2023-01-31
pre-commit: Update isort version to 5.12.0

wence-:
poetry version 1.5.0 broke installs of isort prior to 5.11.5 (see pycqa/isort#2077 and pycqa/isort#2078), so we need to upgrade.


wence-:
In line with the same changes across the rest of rapids. This fixes creation of new pre-commit environments.

benfred (approved):

PR #1226, opened 2023-02-02
Add innerproduct to the pairwise distance api

benfred:


mdoijade (changes requested):
I believe it would be ideal to use cuBLAS for inner product instead of pairwiseDistanceMatKernel perf wise.

benfred:
@mdoijade  I've updated to use cublas here -  I had to change the distance api to use a `raft::device_resources` instead of a just a cuda stream in order to pass in the cublas handle though, which makes this somewhat of a bigger change.  

I think in general though we should be passing `device_resources` instead of cuda streams - so I am hoping that this is ok. @cjnolet what do you think?

cpp/include/raft/distance/detail/kernels/gram_matrix.cuh
```diff
@@ -212,7 +212,7 @@ class GramMatrixBase {
                         int ld_out)
   {
     raft::distance::distance<raft::distance::DistanceType::L2Unexpanded, math_t, math_t, math_t>(
-      x1, x2, out, n1, n2, n_cols, stream, is_row_major);
+      raft::device_resources(stream), x1, x2, out, n1, n2, n_cols, is_row_major);
```
cjnolet, line 215: I think from the public API standpoint, we want to strive to accept `device_resources` everywhere (and eventually just `raft::resources`) but things inside of `detail` are okay to accept just the individual resources they need- so long as the public API is pulling resources like streams from the `raft::resources` speified from the user and not creating their own.

cjnolet: For example, I'm okay with this function accepting the stream directly since it's in the `detail` namespace. The `kernel gramm` API itself is in kind of weird state right now. We moved it over from cuml mostly because it's 1) primitives built directly from the pairwise distances API, and 2) it required just enough customization that it was hard to use the existing specializtions and thus cuml's compile time was increased. 

At some point, we do need to refactor the kernel gramm API into a set of flattened stateless public API functions but for now it's kind of been left in this state just so cuml doesn't break.

cpp/include/raft/distance/detail/distance.cuh
```diff
@@ -92,7 +93,8 @@ template <raft::distance::DistanceType distanceType,
           typename FinalLambda,
           typename Index_>
 struct DistanceImpl {
-  void run(const InType* x,
+  void run(raft::device_resources const& handle,
```
cjnolet, line 96: One thing you could do is even go a step further and just have this accept `raft::resources`. Then we can use the `raft::resources::get_stream(handle)` accessor function without having to import anything other than the stream.

cpp/include/raft/distance/detail/distance.cuh
```diff
@@ -186,12 +210,49 @@ struct DistanceImpl<raft::distance::DistanceType::CosineExpanded,
...
+           Index_ m,
+           Index_ n,
+           Index_ k,
+           void* workspace,
+           size_t worksize,
+           FinalLambda fin_op,
+           bool isRowMajor,
+           InType)
+  {
+    raft::linalg::gemm(handle,
```
cjnolet, line 245: Ah yes, I figured this is why we wanted to accept the handle in the distances API instead of jsut the stream- we ultimately need the cublas handle. 

I think this could be a little cleaner if we accept the `raft::resources` directly in the `distance()` function because then this file can include `raft/core/resource/cublas_handle.hpp` and none of the other files need to include it.

mdoijade:
> @mdoijade I've updated to use cublas here - I had to change the distance api to use a `raft::device_resources` instead of a just a cuda stream in order to pass in the cublas handle though, which makes this somewhat of a bigger change.
> 
> I think in general though we should be passing `device_resources` instead of cuda streams - so I am hoping that this is ok. @cjnolet what do you think?

Thanks @benfred  this looks good to me. Do you think it is worth adding test case for inner product on the c++ distance tests layer as well ?

cpp/include/raft/distance/detail/distance.cuh
benfred: done in the last commit

benfred:
> Thanks @benfred this looks good to me. Do you think it is worth adding test case for inner product on the c++ distance tests layer as well ?

@mdoijade All of the other distance metrics are only tested in the python layer - and using sklearn as a reference implementation to compare the results against.  I'm good with testing the IP distance here in the same way - especially since IP is so simple compared to the other metrics being tested.

cpp/include/raft/distance/detail/kernels/gram_matrix.cuh
benfred: yeah - I didn't want to break this here by changing the API (for the other changes, cuml is using the pairwise_distances api's which already take a raft handle instead of a stream afaict).

This is a little weird in that its upgrading a stream -> device_resources, but since the L2Unexpanded distance here only uses the stream, this won't do anything funky like allocate a cublas handle.

cpp/src/distance/distance/specializations/detail/inner_product.cu
```diff
@@ -0,0 +1,50 @@
...
+  float* dist,
+  int m,
+  int n,
+  int k,
+  void* workspace,
+  std::size_t worksize,
+  bool isRowMajor,
+  float metric_arg);
+
+template void distance<raft::distance::DistanceType::InnerProduct, double, double, double, int>(
```
cjnolet, line 35: We should separate these so they can be compiled in parallel.

python/pylibraft/pylibraft/test/test_distance.py
```diff
@@ -57,7 +58,10 @@ def test_distance(n_rows, n_cols, inplace, metric, order, dtype):
 
     output = np.zeros((n_rows, n_rows), dtype=dtype)
 
-    expected = cdist(input1, input1, metric)
+    if metric == "inner_product":
```
cjnolet, line 61: We mentioned offline, we should test this through both the C++ and python APIs [here](https://github.com/rapidsai/raft/tree/branch-23.04/cpp/test/distance). If you think it'd be helpful, I don't mind helping to add a quick google test case here.

cpp/include/raft/distance/distance.cuh
```diff
@@ -109,20 +108,20 @@ template <raft::distance::DistanceType distanceType,
           typename AccType,
           typename OutType,
           typename Index_ = int>
-void distance(const InType* x,
+void distance(raft::resources const& handle,
```
cjnolet, line 111: This might be a breaking change, though it might not? (e.g. [comment here](https://github.com/rapidsai/raft/pull/1226/files#r1099603821)). We probably want to build cuml w/ this change at least to check.

cpp/include/raft/distance/distance.cuh
```diff
@@ -202,21 +200,22 @@ template <raft::distance::DistanceType distanceType,
...
```
cjnolet, line 203: Same above (checking / updating cuml as needed).

cpp/include/raft/distance/distance.cuh
```diff
@@ -238,7 +237,7 @@ void distance(const InType* x,
  * @param metric_arg metric argument (used for Minkowski distance)
  */
 template <typename Type, typename Index_ = int>
-void pairwise_distance(raft::device_resources const& handle,
+void pairwise_distance(raft::resources const& handle,
```
cjnolet, line 240: I don't think this one is breaking

cjnolet (changes requested):
The changes look great overall! Mostly very minor things.

python/pylibraft/pylibraft/test/test_distance.py
benfred: thanks for the pointers! I can't believe I missed the c++ tests for the other metrics (didn't see them when I first looked for some , though its super obvious they are there). Added in the last commit -

cpp/src/distance/distance/specializations/detail/inner_product.cu
benfred: done

cpp/include/raft/distance/distance.cuh
benfred: testing out w/ cuml here https://github.com/rapidsai/cuml/pull/5230

benfred: needs https://github.com/rapidsai/cuml/pull/5230/commits/6f0d7fce048fd716a0e025882f79db241eb07d35 in cuml looks like

cpp/src/distance/distance/specializations/detail/canberra_double_double_double_int.cu
```diff
@@ -19,20 +19,8 @@
 namespace raft {
 namespace distance {
 namespace detail {
-template void distance<raft::distance::DistanceType::Canberra, float, float, float, int>(
-  const float* x,
```
cjnolet, line 23: Oh wow, so we were building the same canberra specialization multiple times? It makes sense now...

python/pylibraft/pylibraft/test/test_distance.py
cjnolet: Haha no problem. I just know a lot of things have been filling up people's plates lately so I didn't mind jumping in to help if needed.

cjnolet (approved):
LGTM!

PR #1230, opened 2023-02-02
Fix minkowki compile times

ahendriksen:


ahendriksen:
@cjnolet @benfred : This is definitely not passing label and style checker. Feel free to edit this PR.

cjnolet:
Im starting to wonder if it just makes sense to remove the uint32 indexed specialization altogether (and not use it anywhere). That seems to be the one bloating the build from what I can tell.

ahendriksen:
Closing in favor of https://github.com/rapidsai/raft/pull/1254

PR #1235, opened 2023-02-03
Remove uint32 distance idx specializations

cjnolet:
In addition to removing uint32_t idx specializations from pairwise distances, this splits a few of the remaining distance specializations into separate files so they can be compiled concurrently. Verified cuml is not using any of the uint32_t specializations.


cjnolet:
This PR takes the build of the distance/nn libs down to ~14min on a single arch:

Screenshot from 2023-02-03 10-34-40

cjnolet:
Unfortunately, the end-to-end build is still ~38min (largely due to ivf-flat googletest):

Screenshot from 2023-02-03 11-17-28

benfred (approved):
lgtm!

PR #1239, opened 2023-02-03
Un-reverting contractions tiling logic + masked l2 nn for 23.04

cjnolet:
This PR also includes changes from #1235 so it looks big, but it really only has a couple things. 

@ahendriksen I pulled these changes over from your other PRs and consolidated them. I don't want to step on your toes so if you would rather publish the commit or apply this patch (to have your name associated w/ the changes) then let me know and I can close this PR. I'm just hoping to get these merged back in early 23.04


cjnolet:
Just adding a note that we are still going to want to fix the unrolling for these changes before they go in

cpp/include/raft/distance/detail/pairwise_distance_base.cuh
```diff
@@ -116,96 +122,83 @@ struct PairwiseDistances : public BaseClass {
...
 
-  DI void prolog(IdxT gridStrideX, IdxT gridStrideY)
+  DI void reset_accumulator()
   {
-    if (gridStrideX == blockIdx.x * P::Nblk) { this->ldgXY(0); }
-
+    // Reset accumulator registers to zero.
 #pragma unroll
     for (int i = 0; i < P::AccRowsPerTh; ++i) {
 #pragma unroll
```
vyasr, line 197: Recording for posterity: changing these two unrolls to `pragma unroll 1` was sufficient to get the code to compile for me locally. Note that Github thinks these lines didn't change, but they have moved into a new function now which meaningfully changes where the unrolling ends up happening.

ajschmidt8 (approved):

ahendriksen:
I have no problem with the PR getting merged like this! 

> Just adding a note that we are still going to want to fix the unrolling for these changes before they go in
This is addressed by #1254

cjnolet:
Closing in favor of #1254

PR #1240, opened 2023-02-03
Determine conda build work paths

ajschmidt8:
This is just a test PR to determine the proper path to the `cpp/build` directory after a successful `conda build`.


ci/build_cpp.sh
```diff
@@ -1,11 +1,15 @@
 #!/bin/bash
-# Copyright (c) 2022, NVIDIA CORPORATION.
+# Copyright (c) 2022-2023, NVIDIA CORPORATION.
 
 set -euo pipefail
 
 source rapids-env-update
 
 export CMAKE_GENERATOR=Ninja
+BUILD_PATH=/opt/conda/conda-bld/*libraft-split*/cpp/build/
```
ajschmidt8, line 9:
```suggestion
BUILD_PATH=$(echo /opt/conda/conda-bld/*libraft-split*/cpp/build/)
```

I learned earlier in this PR that globs aren't expanded via bash when they're simply assigned to a variable.

Wrapping it like this seemed to resolve that. See my previous commit below for an example.


- https://github.com/rapidsai/raft/pull/1240/commits/3b2ceac292ea3157fe65fc14e31469690a5806be

ci/build_cpp.sh
```diff
@@ -14,3 +18,20 @@ rapids-logger "Begin cpp build"
 rapids-mamba-retry mambabuild conda/recipes/libraft
 
 rapids-upload-conda-to-s3 cpp
+
+
+rapids-logger "ls conda-bld"
+
+ls -al /opt/conda/conda-bld/
+
+rapids-logger "ls conda-bld/*libraft*/"
```
ajschmidt8, line 27:
```suggestion
rapids-logger "Upload ninja logs"
```

All this fluff can be removed in the future

ajschmidt8:
@cjnolet I think the latest issue we're seeing here is the result of conda moving around files like I mentioned on our call.

It looks like _during_ the build, the files might be located here: https://github.com/rapidsai/raft/actions/runs/4118626504/jobs/7111427444#step:6:2067

This might take some messing around with to get right.

I think it's safe to revert some of the changes back to my default `echo` recommendation.

PR #1252, opened 2023-02-06
Remove gpuCI scripts.

bdice:
Since migrating to GitHub Actions, the gpuCI scripts are no longer needed. This PR removes those outdated gpuCI scripts.


README.md
```diff
@@ -1,7 +1,5 @@
 # <div align="left"><img src="https://rapids.ai/assets/images/rapids_logo.png" width="90px"/>&nbsp;RAFT: Reusable Accelerated Functions and Tools</div>
 
-[![Build Status](https://gpuci.gpuopenanalytics.com/job/rapidsai/job/gpuci/job/raft/job/branches/job/raft-branch-pipeline/badge/icon)](https://gpuci.gpuopenanalytics.com/job/rapidsai/job/gpuci/job/raft/job/branches/job/raft-branch-pipeline/)
```
raydouglass, line 3: Should we consider adding a GHA badge? https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/adding-a-workflow-status-badge

ci/release/update-version.sh
```diff
@@ -52,8 +52,6 @@ for FILE in conda/environments/*.yaml dependencies.yaml; do
   sed_runner "s/ucx-py=.*/ucx-py=${NEXT_UCX_PY_VERSION}/g" ${FILE};
 done
 
-sed_runner "s/export UCX_PY_VERSION=.*/export UCX_PY_VERSION='${NEXT_UCX_PY_VERSION}'/g" ci/gpu/build.sh
```
raydouglass, line 55: I think this file is missing an update to `dependencies.yaml`.

This also begs the question: We update (manually) both `dependencies.yaml` and the resulting conda environment files. Should that be refactored to update only `dependencies.yaml` and then invoke `rapids-dependency-file-generator`? Technically, the manual updates should be fine, but do we want to be more consistent?

It's also worth pointing out that this script is typically run outside of CI not in a CI docker image. So the user (me) would need to ensure that `rapids-dependency-file-generator` is installed locally.

vyasr: I noticed this as well in some of the update-version.sh scripts. I think it would be nice to be consistent and only update dependencies.yaml with sed and then run dfg. IMO requiring local installation of dfg is not a major issue, and it's only going to become more important as we extend dfg to support more types of dependency files (especially since some of them will be easier to parse semantically rather than via simple textual regex-based updates).

README.md
bdice: I vote no. Our developers don't spend much time looking at the README, and it looks bad to users when our builds are failing.

ci/release/update-version.sh
bdice: Our CI requires that the files in `conda/environments` are consistent with the output of `rapids-dependency-file-generator`. Locally, I just invoke `rapids-dependency-file-generator` after updating `dependencies.yaml`.

The only reason I would keep the status quo is to avoid the requirement for whoever runs this script to have `rapids-dependency-file-generator` installed. If you are the one who runs this script and you prefer that approach, we can adjust these scripts.

bdice: As for "missing an update to `dependencies.yaml`, line 46 executes on `dependencies.yaml`. Is that what you were asking?

raydouglass: > As for "missing an update to `dependencies.yaml`", line 46 executes on `dependencies.yaml`. Is that what you were asking?

Whoops, yes. Just missed that. Thanks!

raydouglass (approved):

README.md
ajschmidt8: Agreed, but for a different reason.

The latest build status can already be seen on the repository's home page. And when you switch branches, it's automatically in-sync. So no `update-version.sh` hackery is necessary.

image

PR #1257, opened 2023-02-08
Bump pinned pip wheel deps to 23.4

sevagh:
We introduced a change to pin RAPIDS wheel dependencies to the same release version. However, branch 23.04 was created before that last PR was merged, so as of now raft's 23.4 wheels are installing 23.2 RAPIDS dependencies. This PR updates those pins to the current release.


sevagh:
The failing raft-dask-cu11 wheel test is because "dask-cuda==23.4.*" is not available yet. When does that typically get released?

cjnolet:
cc @jakirkham @quasiben

vyasr:
Ah I didn't see this PR earlier, my apologies. I superseded it in #1266.

The dask-cuda problem should be a nonissue because we install it from source. The issue was that the update-version.sh script had a bug that was preventing the branch pulled in CI from being updated, so the install from source was still installing 23.02. I have fixed that in #1266.

PR #1259, opened 2023-02-09
Reduce error handling verbosity in CI tests scripts

AjayThorve:
This PR adds a less verbose trap method, for error handling to help ensure that we capture all potential error codes in our test scripts, and works as follows:

- setting an environment variable, EXITCODE, with a default value of 0
- setting a trap statement triggered by ERR signals which will set EXITCODE=1 when any commands return a non-zero exit code

cc @ajschmidt8


ci/test_cpp.sh
```diff
@@ -32,6 +31,8 @@ rapids-mamba-retry install \
 rapids-logger "Check GPU usage"
 nvidia-smi
 
+EXITCODE=0
+trap "EXITCODE=1" ERR
```
ajschmidt8, line 35: I can't leave a suggestion on the particular line in GitHub, but it looks like there's a `TODO` a few lines down that can be removed here.

After that is removed, the whitespace can also be removed between the logger statement and subsequent `for` loop.

ajschmidt8 (changes requested):
Left a quick comment.

Might be worth pulling in the latest changes when you update this PR also... looks like you're on the cusp of the [Recently Updated](https://docs.rapids.ai/resources/recently-updated/) threshold.

ajschmidt8 (approved):
:fire: :fire:

PR #1264, opened 2023-02-09
Remove duplicate `librmm` runtime dependency

ajschmidt8:
`libraft-headers` has `librmm` listed as a `host` and `run` dependency in its recipe.

Since `librmm` has a `run_exports` value, this causes there to be two `librmm` runtime dependency entries in the resulting `libraft-headers` conda package, each with slightly version specifiers.

To prevent this from occurring, I've added `librmm` to the `ignore_run_exports_from` key in the `libraft-headers` recipe.

Note that another way to solve this would be to remove `librmm` from the `run` dependencies, but this is inconsistent with the rest of the RAPIDS recipes.

### DIFF
--- a/conda/recipes/libraft/meta.yaml
+++ b/conda/recipes/libraft/meta.yaml
@@ -37,7 +37,7 @@ outputs:
       string: cuda{{ cuda_major }}_{{ date_string }}_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
       ignore_run_exports_from:
         - {{ compiler('cuda') }}
-        - {{ librmm }}
+        - librmm
     requirements:
       build:
         - {{ compiler('c') }}
--- a/conda/recipes/libraft/meta.yaml
+++ b/conda/recipes/libraft/meta.yaml
@@ -37,6 +37,7 @@ outputs:
       string: cuda{{ cuda_major }}_{{ date_string }}_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
       ignore_run_exports_from:
         - {{ compiler('cuda') }}
+        - {{ librmm }}
     requirements:
       build:
         - {{ compiler('c') }}


ajschmidt8:
This picture of the `libraft-headers` metadata on its [conda page](https://anaconda.org/rapidsai-nightly/libraft-headers/files) demonstrates the current issue.


image

vyasr:
> Note that another way to solve this would be to remove librmm from the run dependencies, but this is inconsistent with the rest of the RAPIDS recipes.

Isn't this approach the better one, though? It lets librmm be the single source of truth dictating that linking to it also requires its presence as a runtime dependency. Maybe we merge this PR as is for consistency as you say, then go back through and refactor?

ajschmidt8:
> > Note that another way to solve this would be to remove librmm from the run dependencies, but this is inconsistent with the rest of the RAPIDS recipes.
> 
> Isn't this approach the better one, though? It lets librmm be the single source of truth dictating that linking to it also requires its presence as a runtime dependency. Maybe we merge this PR as is for consistency as you say, then go back through and refactor?

This part is complicated by the fact that we use conda recipes with multiple outputs.

If you take [the `libcudf` recipe](https://github.com/rapidsai/cudf/blob/branch-23.04/conda/recipes/libcudf/meta.yaml) for example...  we build the all of the `libcudf` packages (libcudf, libcudf-tests, libcudf_kafka, etc) using the top-level dependencies [here](https://github.com/rapidsai/cudf/blob/0cab19aca4c65453b9d1715b34999dbfc5d0a162/conda/recipes/libcudf/meta.yaml#L31-L48) (which, therefore, is where `librmm` needs to be listed).

Because `librmm` is _only_ listed in the `host` dependencies of this top-level `libcudf` package, the `run_exports` listed in its recipe do not apply to the rest of the `libcudf` `outputs` packages.

The individual `outputs` entires (like [libcudf](https://github.com/rapidsai/cudf/blob/0cab19aca4c65453b9d1715b34999dbfc5d0a162/conda/recipes/libcudf/meta.yaml#L51-L68)) only run `cmake --install cpp/build` and therefore _don't_ list `librmm` as a `host`/`build` dependency... because they don't need it. They only need `cmake` as a dependency since they're just installing the library that was already built.

As a result of this arrangement, we manually add `librmm` as a `run` dependency wherever else it's needed.

raydouglass (approved):

PR #1266, opened 2023-02-09
Bump pinned deps to 23.4

vyasr:
Some dependencies were not fully updated, probably because update-version.sh was run before the most recent updates to the script.


cjnolet (approved):

vyasr:
Ah this needs an ops review as well. Will post a request.

jolorunyomi (approved):

PR #1268, opened 2023-02-10
matrix::select_k specializations

achirkin:
Add specializations to `matrix::detail::select_k`, which has a few use-cases across raft.


cjnolet (approved):
LGTM!

Issue #1205, opened 2023-01-30
[FEA] Support reconstruction of original vectors from IVF-Flat / IVF-PQ Index


cjnolet:
FAISS currently supports two functions (on CPU) for reconstructing original vectors from the IVF-Flat/IVF-PQ indexes: `reconstruct_n` and `reconstruct_batch`. FAISS' existing GPU implementation does support `reconstruct_n` (a linear set of indices from a specific start to specific end index) but `reconstruct_batch` would be more efficient so that we can reconstruct a specific set of vectors given their indexes, which don't even need to be in order. It's been requested that we support this in our GPU implementations.

cjnolet:
Referencing the corresponding method in FAISS' CPU IVF index: https://github.com/facebookresearch/faiss/blob/main/faiss/IndexIVF.cpp#L918

achirkin:
Update on the reconstruction.

At this moment, we have two draft implementations of reconstruction. My [IVF-PQ implementation](https://github.com/rapidsai/raft/blob/1326ac707a3eb330120d7e122c3d1587de92ba71/cpp/include/raft/neighbors/ivf_pq.cuh#L392-L400) takes the (cluster_id, [in_cluster_indices]) instead of the user-defined indices, and the [IVF-Flat implementation](https://github.com/rapidsai/raft/blob/2e8bc92bb85838e8d3f794eb122a34743d103285/cpp/include/raft/neighbors/detail/ivf_flat_build.cuh#L481-L482) seems to be sub-optimal in that it does not support non-contiguous user-defined indices and iterates through the whole index multiple times.

The ANN indices are one-directional by design. Thus, supporting the search in the opposite direction (find vectors by indices) requires either an exhaustive search through the whole DB for every query, or introduction of a helper struct for a faster search (an attempt to make such is in ivf-flat PR). I suggest we do the latter, but decouple the helper struct from the model. Something like the following:
```C++
template <typename ListT>
struct index_mapping_t {
  device_vector<ListT::index_type> user_indices;
  device_vector<std::Tuple<uint32_t, uint32_t>> clusters_and_offsets;
  template <typename = ivf::enable_if_valid_list<ListT>>
  index_mapping_t(std::vector<std::shared_ptr<ListT>> lists) {
    ...
    cub::DeviceRadixSort::SortPairs(user_indices, clusters_and_offsets);
    ...
  }
};
template <typename ListT>
[...]

PR #1270, opened 2023-02-10
IVF-Flat reconstruction

viclafargue:
Answers https://github.com/rapidsai/raft/issues/1205


achirkin:
NB: ivf-pq reconstruction: https://github.com/rapidsai/raft/pull/1298

cpp/include/raft/neighbors/ivf_pq.cuh
```diff
@@ -117,6 +145,39 @@ auto extend(raft::device_resources const& handle,
     handle, orig_index, new_vectors, new_indices, n_rows);
 }
 
+/**
+ * @brief Build a new index containing the data of the original plus new extra vectors.
+ *
+ * Implementation note:
+ *    The new data is clustered according to existing kmeans clusters, then the cluster
```
tfeher, line 152:
```suggestion
 *    The new data is clustered according to existing kmeans clusters, the cluster
```

cpp/include/raft/neighbors/ivf_pq.cuh
```diff
@@ -198,6 +281,59 @@ void search(raft::device_resources const& handle,
...
+ * k]
+ * @param mr an optional memory resource to use across the searches (you can provide a large enough
+ *           memory pool here to avoid memory allocations within search).
+ */
+template <typename T, typename IdxT>
+inline void search(raft::device_resources const& handle,
+                   const search_params& params,
+                   const index<IdxT>& index,
+                   const raft::device_matrix_view<const T, IdxT, row_major>& queries,
+                   uint32_t k,
```
tfeher, line 317: Is there a use case when we want to pass `k` directly instead of getting it from the extent of `neighbors` or `distances`?

cpp/src/distance/neighbors/ivfpq_build.cu
```diff
@@ -21,43 +21,36 @@
...
-              const T* new_vectors,                                                          \
-              const IdxT* new_indices,                                                       \
-              IdxT n_rows)                                                                   \
-    ->raft::neighbors::ivf_pq::index<IdxT>                                                   \
-  {                                                                                          \
-    return raft::neighbors::ivf_pq::extend<T, IdxT>(                                         \
-      handle, orig_index, new_vectors, new_indices, n_rows);                                 \
-  }                                                                                          \
-                                                                                             \
-  void build(raft::device_resources const& handle,                                           \
```
tfeher, line 45: Do the python wrappers still work after removing this function definition?

tfeher (changes requested):
Thanks @viclafargue for this PR!  The details of the reconstruction function needs to updated after #1271 is merged, therefore I did not review that part yet. Below please find a few comments for the rest.

Could you update the PR destription to mention that it also adds the mdspan interface for ivf-pq?

viclafargue:
Thanks for your review @tfeher! Sorry for the confusion, the mdspan change is actually taking place in two other PRs, no need to review it on this PR (btw, your review is welcome in #1328). Finishing this PR took me a bit of time, as the reconstruction kept failing. It looked like there was an issue with the calculation of the offsets during the construction phase. Fortunately, the new list system in #1271 fixed the problem entirely.

cpp/include/raft/neighbors/ivf_flat.cuh
```diff
@@ -469,6 +469,31 @@ void search(raft::device_resources const& handle,
...
+ * @tparam T data element type
+ * @tparam IdxT type of the indices in the source dataset
+ *
+ * @param[in] handle
+ * @param[in] idx
+ * @param[in] vector_ids a mdspan array containing the ids of vectors to be reconstructed [n_rows]
+ * @param[out] vector_out a mdspan array to hold the vectors to be reconstructed [n_rows,
+ * index.dim()]
+ */
+template <typename T, typename IdxT>
```
lowener, line 484: Can `value_t` and `idx_t` be used for the template names so that all the new public mdspan API functions use the same type name.

achirkin: I'd suggest to **not** do this until we change the corresponding [.clang-tidy](https://github.com/rapidsai/raft/blob/76c828dd4da4dc922626ba2a440a46dea6ab03b9/cpp/.clang-tidy#L211). Note also that we use camel case in plenty of public/mdspan api, including the ivf-pq that mirrors the IVF-Flat interface.

cpp/include/raft/neighbors/detail/ivf_flat_build.cuh
```diff
@@ -415,4 +417,103 @@ inline void fill_refinement_index(raft::device_resources const& handle,
...
+
+template <typename T, typename IdxT>
+void reconstruct_batch(raft::device_resources const& handle,
+                       const index<T, IdxT>& index,
+                       device_mdspan<const IdxT, extent_1d<IdxT>, row_major> vector_ids,
+                       device_mdspan<T, extent_2d<IdxT>, row_major> vector_out)
+{
+  thrust::device_ptr<const IdxT> vector_ids_ptr =
+    thrust::device_pointer_cast(vector_ids.data_handle());
+  IdxT max_indice = *thrust::max_element(
```
achirkin, line 481: Just a reminder: although seeming like a good compromise, this would be wasting too much memory at best and crash at worst when the database contains non-contiguous (e.g. beyond a few billions) 64 bit user-space indices.
We've discussed this in the corresponding ivf-pq PR https://github.com/rapidsai/raft/pull/1298#discussion_r1127379995 and I suggested to create a special mapping struct (a hashmap or a pair of sorted arrays) for faster search of the user-space indices in https://github.com/rapidsai/raft/issues/1205 .

I think, we'd better reconstruct by cluster+offset pairs here and dedicate a separate PR for the mapping between user-space and DB indices, and put it into `ivf_list.cuh` to share among the ivf implementations.

achirkin (changes requested):

viclafargue:
> @viclafargue is this PR ready for another review yet?

Sorry for the delay, it's ready for a review.

cpp/include/raft/neighbors/detail/ivf_flat_build.cuh
```diff
@@ -416,4 +420,152 @@ inline void fill_refinement_index(raft::resources const& handle,
...
+                           const index<T, IdxT>& index,
+                           device_matrix_view<T, IdxT, row_major> out_vectors,
+                           uint32_t label,
+                           uint32_t offset)
+{
+  auto stream = raft::resource::get_cuda_stream(handle);
+
+  uint32_t len = out_vectors.extent(0);
+  const dim3 block_dim(256);
+  const dim3 grid_dim(raft::ceildiv<size_t>(len, block_dim.x));
```
achirkin, line 561: A nitpick: use the newer [raft::div_rounding_up_safe](https://github.com/rapidsai/raft/blob/f34cdbd740d23dae06ac88938094b9429cb81e4e/cpp/include/raft/util/integer_utils.hpp#L116C23-L116C43) here and above. Perhaps, we should deprecate `raft::ceildiv` in favor of `raft::div_rounding_up_unsafe`?...

cpp/include/raft/neighbors/ivf_flat-inl.cuh
```diff
@@ -464,6 +464,52 @@ void search(raft::resources const& handle,
...
+ * @param[in] handle
+ * @param[in] index ivf-flat constructed index
+ * @param[out] vector_out matrix with the vectors contained in the cluster
+ * @param[in] label cluster index
+ * @param[in] offset offset
+ */
+template <typename T, typename IdxT>
+void reconstruct_list_data(raft::resources const& handle,
+                           const index<T, IdxT>& index,
+                           device_matrix_view<T, IdxT, row_major> out_vectors,
```
achirkin, line 505: In `ivf_pq`, list and query dimensions are limited to `uint32_t` rather than `IdxT` (type of DB indices).

Here in the ivf_flat API you use  `device_matrix_view<T, IdxT, row_major>` rather than  `device_matrix_view<T, uint32_t, row_major>` , but you also set both label and offset arguments to `uint32_t`. Could you please check here, in the overload above, and down the detail namespace what are the appropriate indexing types and make it consistent?

cpp/include/raft/neighbors/detail/ivf_flat_build.cuh
```diff
@@ -416,4 +420,152 @@ inline void fill_refinement_index(raft::resources const& handle,
...
+                       const index<T, IdxT>& index,
+                       raft::device_vector_view<const IdxT, IdxT> vector_ids,
+                       raft::device_matrix_view<T, IdxT, row_major> vector_out)
+{
+  auto stream      = raft::resource::get_cuda_stream(handle);
+  auto exec_policy = raft::resource::get_thrust_policy(handle);
+
+  thrust::device_ptr<const IdxT> vector_ids_ptr =
+    thrust::device_pointer_cast(vector_ids.data_handle());
+  IdxT max_indice =
```
achirkin, line 487: I thought the consensus was to opt for a "inverted-index" module (hashmap/array/whatever) shared among ivf methods to convert user indices to (label, in-cluster-offset) pairs?
This code could break if the user adds non-contiguous range of large indices to the DB.

achirkin (changes requested):
Thanks for the update, here's couple comments on the api.

cpp/include/raft/neighbors/detail/ivf_flat_build.cuh
achirkin: I was actually thinking of decoupling the ivf-list hashmap struct api from the ivf-pq and ivf-flat methods. Hence a user would need to construct the hashmap explicitly once (costly operation) and then either:
a) search-by-user-ids in two calls (e.g. hashmap::get_lists_offsets + index::get_vectors)
b) pass the hashmap as an argument to the ivf index methods (index::get_vectors)

Not sure what we will decide in the end, but point is the api may change as we progress with the hashmap. Then, maybe we just can keep this function in the `detail` namespace for now? Hence the work won't be lost and if someone needs the functionality urgently, they can use the detail function, and we don't need to break the public api in the follow-up PR.

cpp/include/raft/neighbors/ivf_flat-inl.cuh
viclafargue: We are using `IdxT` for the IVF-Flat API. I corrected the code so that it uses the template type.

achirkin: Oh, interesting! Out of curiosity, could you add a test case for ivf-flat with 64-bit indexing type and try to call something with an offset slightly outside the 32-bit range? :)

achirkin: A quick browsing confirms that 64-bit indexing/extents are actually not supported https://github.com/rapidsai/raft/blob/cad88b38a1de065fed933e8a910002d543b8cfa2/cpp/include/raft/neighbors/ivf_flat-inl.cuh#L455-L464

viclafargue: The portion of code you are mentioning is indeed probably a mistake. At the moment the IVF-Flat API functions are templated with `IdxT` everywhere, and it is set to use `uint64_t` during compilation. Do you recommend that we update the IVF-Flat API to systematically use `uint32_t` instead? If so, this is an extensive change, and that would maybe require a separate PR, right?

achirkin: If ivf-flat can actually work with extents and input sizes `IdxT` larger than 32-bit, I don't see any problem using `IdxT` in the api. But I suspect this is not the case. Hence we should fix it either by limiting extent types to 32-bit or making sure it works with all possible `IdxT` (not only those we instantiate explicitly btw). I agree this work is worth a separate PR.
However, for the new public functions added in this PR, I think it makes sense to fix the api from the beginning, keeping in mind the future changes in the other parts of the code.

viclafargue:
Closing this as it is not relevant anymore.

Issue #1170, opened 2023-01-24
[FEA] Separate IVFs into individual memory chunks.


cjnolet:
This has been strongly requested for the FAISS integration and for true incremental add support. It also matches more closely to the current FAISS implementation and allows the `copyFrom()` implementation to be much more straightforward.

tfeher:
**Background**

Currently both IVF-Flat an IVF-PQ stores its index as a single array of all vectors. This has a drawback when we extend the index: we need to recreate the whole index to insert the new values.

Both method has the same coarse level clustering:

- we groups the points into n_lists clusters
- each cluster store all its vectors (interleaved and padded) in one chunk lets call this cluster data
- The cluster data has to be stored in one chunk, for efficient search.
- Currently the index is just a concatenation of cluster data for each cluster

**Tasks**

- Store IVF-Flat index in separate array for each cluster #1176
- Store IVF-PQ index in separate array for each cluster

**Amortize cost of growing the array**
- Discuss if it is necessary to preallocate larger space for list, that would allow growth without reallocating and moving to new list.

cjnolet:
I also want to note here that after we separate the lists into their own individual arrays, we will need to use the new [CodePacker](https://github.com/facebookresearch/faiss/pull/2625) API in FAISS to perform the transformation of the CPU/GPU indexes into the RAFT interleaved indexes.

achirkin:
Resolved via #1249 and #1271

PR #1271, opened 2023-02-10
IVF-Flat index splitting

lowener:
Refactor of `ivf_flat::index` to split the cluster data in separate buffers
- Addressing #1170.
- Following #1249 in the index structure implementation.
- Adding serialization API with stream and filename overloads
- Moving `raft::spatial::knn::ivf_flat` namespace to `raft::neighbors::ivf_flat`


cpp/include/raft/neighbors/ivf_flat_types.hpp
```diff
@@ -65,6 +77,37 @@ struct search_params : ann::search_params {
...
+      RAFT_FAIL(
+        "ivf-flat: failed to allocate a big enough index list to hold all data "
+        "(requested size: %zu records, selected capacity: %zu records). "
+        "Allocator exception: %s",
+        size_t(size),
+        size_t(capacity),
+        e.what());
+    }
+    // Fill the index buffer with a pre-defined marker for easier debugging
+    matrix::fill(res, indices.view(), ivf_flat::kInvalidRecord<IdxT>);
```
achirkin, line 107: NB: I had to switch back from raft primitive to thrust here, because of compilation errors on the cython side (no dependency on cuda code allowed) https://github.com/rapidsai/raft/pull/1249/commits/adf7872d9997eef5745787a26f692405aac3bf0d

achirkin: NB 2: the ivf-pq PR merged and the list data is in a separate module now to avoid code duplication between ivf-flat and ivf-pq.

cpp/test/neighbors/ann_ivf_flat.cuh
```diff
@@ -172,7 +180,7 @@ class AnnIVFFlatTest : public ::testing::TestWithParam<AnnIvfFlatInputs<IdxT>> {
         auto half_of_data_view = raft::make_device_matrix_view<const DataT, IdxT>(
           (const DataT*)database.data(), half_of_data, ps.dim);
 
-        auto index_2 = ivf_flat::extend(handle_, index, half_of_data_view);
+        ivf_flat::extend(handle_, &index, half_of_data_view);
```
achirkin, line 183: Not likely related to the current test failures, but the original intent of this was to test both variants of the `extent`  api (make new extended index or modify the given index). I think it makes sense to keep this logic, especially given that now these are actually two different codepaths in detail namespace.

cpp/include/raft/neighbors/ivf_list.hpp
```diff
@@ -67,9 +67,9 @@ list<SpecT, IdxT, SizeT>::list(raft::device_resources const& res,
  * Resize a list by the given id, so that it can contain the given number of records;
  * copy the data if necessary.
  */
-template <template <typename> typename SpecT, typename IdxT, typename SizeT>
+template <template <typename> typename SpecT, typename ValueT, typename IdxT, typename SizeT>
```
achirkin, line 70: Why is the need to add the `ValueT`? Is `SpecT<SizeT>::value_type` not enough for something?

cpp/include/raft/neighbors/ivf_flat_types.hpp
```diff
@@ -278,26 +345,29 @@ struct index : ann::index {
...
   raft::distance::DistanceType metric_;
   bool adaptive_centers_;
-  device_mdarray<T, extent_2d<IdxT>, row_major> data_;
-  device_mdarray<IdxT, extent_1d<IdxT>, row_major> indices_;
-  device_mdarray<uint32_t, extent_1d<uint32_t>, row_major> list_sizes_;
-  device_mdarray<IdxT, extent_1d<uint32_t>, row_major> list_offsets_;
-  device_mdarray<float, extent_2d<uint32_t>, row_major> centers_;
-  std::optional<device_mdarray<float, extent_1d<uint32_t>, row_major>> center_norms_;
+  bool conservative_memory_allocation_;
+  host_vector<std::shared_ptr<list_data<T, IdxT>>, uint32_t> lists_;
```
tfeher, line 349:
```suggestion
  std::vector<std::shared_ptr<list_data<T, IdxT>>, uint32_t> lists_;
```

see discussion https://github.com/rapidsai/raft/pull/1249#discussion_r1107267407

cpp/include/raft/neighbors/detail/refine.cuh
```diff
@@ -108,17 +108,17 @@ void refine_device(raft::device_resources const& handle,
     handle.get_thrust_policy(), fake_coarse_idx.data(), fake_coarse_idx.data() + n_queries);
 
   raft::neighbors::ivf_flat::index<data_t, idx_t> refinement_index(
-    handle, metric, n_queries, false, dim);
+    handle, metric, n_queries, false, false, dim);
```
tfeher, line 111: The index will be filled once, and then discarded. We can set conservative_memory_allocation = true.
```suggestion
    handle, metric, n_queries, false, true, dim);
```

tfeher (changes requested):
Thanks @lowener for the PR, it looks great overall! I have just a few comments to add.

cpp/include/raft/neighbors/ivf_list.hpp
lowener: `SpecT<SizeT>::value_type` is not enough, to change the value_type I would have to use a template on Spec, and I think it makes more sense to have the value type attached to the list than to the Spec.

lowener: I reverted this change to bring back `SpecT::value_type`. For `IdxT` I don't mind keeping it like that but if you would prefer to see it in `SpecT` I can make the change.

cpp/include/raft/neighbors/ivf_flat_types.hpp
```diff
@@ -223,53 +232,111 @@ struct index : ann::index {
...
-      indices_(make_device_mdarray<IdxT>(handle, make_extents<IdxT>(0))),
-      list_sizes_(make_device_mdarray<uint32_t>(handle, make_extents<uint32_t>(n_lists))),
-      list_offsets_(make_device_mdarray<IdxT>(handle, make_extents<uint32_t>(n_lists + 1))),
-      centers_(make_device_mdarray<float>(handle, make_extents<uint32_t>(n_lists, dim))),
-      center_norms_(std::nullopt)
+      conservative_memory_allocation_{conservative_memory_allocation},
+      centers_(make_device_matrix<float, uint32_t>(res, n_lists, dim)),
+      center_norms_(std::nullopt),
+      lists_{make_host_vector<std::shared_ptr<list_data<T, IdxT>>, uint32_t>(n_lists)},
+      list_sizes_{make_device_vector<uint32_t, uint32_t>(res, n_lists)},
```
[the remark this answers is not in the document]
lowener: As you said the `list_data` object can be augmented by an extended index so I added a note in the code that the actual `list_data` size might be more than the `list_sizes_`.
https://github.com/rapidsai/raft/pull/1271/files#diff-5717ee33acf80deef560fdfd5de6b0ce2faa71f12b3acfcfc05dd6c8658d687fR170
Another reason is that these sizes are needed in the search kernel, so this avoid recomputing the sizes for each search call.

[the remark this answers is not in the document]
tfeher: Great, thanks!

cpp/include/raft/neighbors/ivf_pq_types.hpp
```diff
@@ -161,9 +161,9 @@ constexpr static uint32_t kIndexGroupVecLen = 16;
 template <typename IdxT>
 constexpr static IdxT kOutOfBoundsRecord = std::numeric_limits<IdxT>::max();
 
-template <typename SizeT = uint32_t>
+template <typename ValueT = uint8_t, typename SizeT = uint32_t>
```
tfeher, line 164: Do we need to have `ValueT` as a template parameter, when in practice we always want it to be uint8? Unlike IVF-FLAT, where the value type is related to the type of the dataset, here in IVF-PQ we compress the dataset to a set of bits,  and the value_type is just used to access memory with 1 byte granularity).

tfeher (approved):
Thanks @lowener for addressing the issues! I have just one question left regarding list template params. I would actually ask @achirkin to have a final look at those changes. Apart from that the PR looks good to me, pre-approving.

cpp/include/raft/neighbors/ivf_pq_types.hpp
achirkin: Yes, I'd prefer to have this reverted along with the changes to ivf_list_types.cuh unless there is a very strong reason for this change.

cpp/include/raft/neighbors/ivf_list.hpp
```diff
@@ -135,7 +136,7 @@ void serialize_list(const raft::device_resources& handle,
 template <template <typename> typename SpecT, typename IdxT, typename SizeT>
 void serialize_list(const raft::device_resources& handle,
                     std::ostream& os,
-                    const std::shared_ptr<const list<SpecT, IdxT, SizeT>>& ld,
+                    const std::shared_ptr<list<SpecT, IdxT, SizeT>>& ld,
```
achirkin, line 139: Is there a specific reason you removed the const modifier here, or is it just an artifact of reverting the previous changes? If the latter, please revert the changes in this file (const modifier is a part of the contract that says the lists are the inputs).

ajschmidt8 (approved):

cpp/include/raft/neighbors/ivf_list.hpp
lowener: Yes, I removed the const modifier here because it is breaking the template type inference and it is not what the function is being called with https://github.com/rapidsai/raft/blob/e7f0268c9f5c17756c415ed47e1856030b3f235d/cpp/include/raft/neighbors/ivf_pq_types.hpp#L497
I'm using this fix because passing `list_spec_wrapper<T>::list_spec` as the `SpecT` template parameter is leading to the following compilation error ```error: argument list for class template "raft::neighbors::ivf_flat::list_spec_wrapper<ValueT>::list_spec" is missing``` even though IVF-PQ seems to have no problem using `list_spec` without the template argument...

achirkin: You're right, I see it's more troublesome to modify `index::lists()` and other calls compared to having a little compromise like this.
I've played a little bit with the templates and suggest to change the signature to make it more flexible/convenient for both ivf-flat and ivf-pq. Here it is https://github.com/lowener/raft/pull/1

achirkin (approved):

PR #1272, opened 2023-02-10
Remove no longer instantiated templates from list of extern template declarations

vyasr:
A number of template instantiations were removed in #1255, but the corresponding extern declarations appear to have persisted, resulting in any library including those headers ending up with undefined symbols.

### DIFF
--- a/cpp/include/raft/neighbors/specializations/detail/ivf_pq_search.cuh
+++ b/cpp/include/raft/neighbors/specializations/detail/ivf_pq_search.cuh
@@ -32,14 +32,9 @@ using fp8u_t = fp_8bit<5, false>;
   extern template struct ivfpq_compute_similarity<IdxT, OutT, LutT>::configured<false, true>; \
   extern template struct ivfpq_compute_similarity<IdxT, OutT, LutT>::configured<true, false>;
 
-#define RAFT_INST_ALL_IDX_T(OutT, LutT) \
-  RAFT_INST(uint64_t, OutT, LutT)       \
-  RAFT_INST(int64_t, OutT, LutT)        \
-  RAFT_INST(uint32_t, OutT, LutT)
-
 #define RAFT_INST_ALL_OUT_T(LutT)  \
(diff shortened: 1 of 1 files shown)


cjnolet:
@vyasr have you verified that this fix works for you to fix the symbol lookup error in cuml? I'm kind of surprised cuml is using these symbols at the moment and hasn't failed until now, but it makes sense that it would happen.

vyasr:
I'm testing it in [this cuml PR](https://github.com/rapidsai/cuml/pull/5217).

cjnolet:
Ah, okay. So I take it cuml's CI does fail without the change? (I thought someone had mentioned they didn't see the failures in cuml's CI)

vyasr:
I think I was updating my previous comment while you responded :joy: so to reiterate, we see these failures only for wheels, not conda. That's probably because as you suspect cuml isn't actually using these symbols, but they are problematic when raft is rebuilt in the wheels CI for cuml and then statically linked to as though the symbols should exist.

cjnolet (approved):

cjnolet:
@vyasr got it. Also making a note here that I believe @achirkin recently found some of the ivfpq specializations were no longer being used within RAFT, which was adding some unecessary build time. That should be fixed soon too :-)

achirkin:
> Also making a note here that I believe @achirkin recently found some of the ivfpq specializations were no longer being used within RAFT, which was adding some unecessary build time. That should be fixed soon too (and should help speed up the build a bit). :-)

I'm afraid, on the contrary: I've made the specializations (copmute_similarity kernel) be used again (by ivf_pq::search). This does improve the compile times though :)

cjnolet:
> I'm afraid, on the contrary: I've made the specializations (copmute_similarity kernel) be used again (by ivf_pq::search). This does improve the compile times though :)


This is what I meant- you found specializations which were no longer being used in the tests and benchmarks, causing unecessary compile times and you are fixing them by having them use the specializations. Sorry for being confusing there.

PR #1276, opened 2023-02-13
Suppress unused params warning on overloaded function

vyasr:
On CUDA 11.8 raft currently compiles fine, but with CUDA 11.5 I see unused function parameter warnings, which are currently treated as errors and therefore prevents compilation. This PR reenables compilation on 11.5 (and probably older) CUDA versions by ignoring these warnings.

### DIFF
--- a/cpp/include/raft/distance/detail/distance.cuh
+++ b/cpp/include/raft/distance/detail/distance.cuh
@@ -247,9 +247,11 @@ struct DistanceImpl<raft::distance::DistanceType::InnerProduct,
            Index_ m,
            Index_ n,
            Index_ k,
+#pragma nv_diag_suppress 177
            void* workspace,
            size_t worksize,
            FinalLambda fin_op,
+#pragma nv_diag_default 177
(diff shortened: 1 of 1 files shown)


cpp/include/raft/distance/detail/distance.cuh
```diff
@@ -247,9 +247,11 @@ struct DistanceImpl<raft::distance::DistanceType::InnerProduct,
            Index_ m,
            Index_ n,
            Index_ k,
+#pragma nv_diag_suppress 177
```
benfred, line 250: will this work for older versions of cuda (like cuda 11.2)? 

I'm wondering if we need to use `__NVCC_DIAG_PRAGMA_SUPPORT__` to figure out if there is support for this compiler flag (sorta like I attempted in https://github.com/rapidsai/raft/pull/941/files).

vyasr:
Closing in favor of #1277. @benfred probably worth copying some of the description over, but your PR builds just as well on my 11.5 env.

PR #1277, opened 2023-02-13
fix build on cuda 11.5

benfred:
On CUDA 11.8 raft currently compiles fine, but with CUDA 11.5 I see unused function parameter warnings, which are currently treated as errors and therefore prevents compilation. This PR reenables compilation on 11.5 CUDA versions by removing the unused parameters


cjnolet (approved):
LGTM!

PR #1279, opened 2023-02-14
New Sparse Matrix APIs

cjnolet:
Closes https://github.com/rapidsai/raft/issues/348. 

This design addresses some problems we've had in the past when modeling sparse data where our objects were not flexible nor composable enough, which led to APIs which were hard to maintain and state which was hard to track. 

This design starts by decomposing sparse formats into two components which are utlimately combined to compose the full sparse object:
1. a `structural` component manages the sparsity of the object , indexing, and data-specific metadata, such as total number of rows and columns.
2. a `valued` or `matrix` component combines with a structure and manages the nonzero elements.

Note that this design also affords the ability to model a sparse tensor in the future, as a new format `tensor` could allow for composing multiple structural and multiple `valued` components. This could enable our algorithms to support things decompositions of higher ordered structures and/or associated values. 

In addition to being flexible and composable, this design also needs to satisfy a couple different levels of immutability:
1. Read-only: immutable structure, immutable nonzero elements
2. Fixed-sparsity and value-mutable: immutable structure, mutable nonzero elements
3. Mutable-sparsity and value-mutable: mutable structure, mutable nonzero elements

Two concepts introduced in this design are pretty core to the 3 states above:
1. `structure-preserving` formats are views and require the sparsity to be known at creation time. The actual structural components may or may not be mutable. 
2. `structure-owning` formats house owning containers and don't require the sparsity to be known at creation time and provide a way to `initialize()` the sparsity once it is known. These formats will have mutable structure and nonzero elements. 

Both the `structure` and `matrix` formats can be `structure-preserving` or `structure-owning`. While this PR only includes `csr_matrix` and `coo_matrix` (I'm considering dropping the `r` from `csr` since it doesn't really matter if it's csr or csc), the design will further allow for other formats, such as `dcsr` and `bcsr`, in the future.

1. `csr_matrix_view` - this is a `structure-preserving` matrix view. Sparsity must be known up front and the underlying arrays may or may not be `const`. 
2. `csr_matrix` - this can be `structure-owning` or `structure-preserving` depending on whether its underlying structural component is `structure-preserving` (view) or `structure-owning`. Calling `view()` on this object produces the `csr_matrix_view` above.
3. `coo_matrix_view` - this is a `structure-preserving` matrix view. Sparsity must be known up front and the underlying arrays may or may not be `const`. 
4. `coo_matrix` - this can be `structure-owning` or `structure-preserving` depending on whether its underlying structural component is `structure-preserving` (view) or `structure-owning`. Calling `view()` on this object produces the `csr_matrix_view` above.

Similar to `mdarray` and `mdspan`, a bunch of factory functions are provided in `raft/core/device_sparse_matrix.hpp` to ease the construction process for users. The owning matrix types can be constructed either to own the underlying structure or a view of the structure. 

These new formats will allow us to model our sparse APIs so they are much more concise- a function can explicitly require a structure-owning matrix, which is a signal to the user that the function itself will compute the sparsity and at least fill in the initial structure. This will allow us to continue to provide an API which is easier to use, and ultimately feels more like our dense API, while still considering the design differences in sparse computations.

I also want to thank @divyegala for his help making the template metaprogramming layer flexible, reusable, and generally pleasant to use. 

- [x] more comprehensive googletests
- [x] split into different files (types, compressed, coordinate, etc...)
- [x] add host APIs


cjnolet:
cc @mhoemmen for thoughts

mhoemmen:
Hi @cjnolet !  I do like the idea of distinguishing between structure-preserving formats and structure-owning formats.  The latter show up, for example, when wrapping third-party libraries that ingest 3-array CSR and produce an optimized opaque format.

Sometimes those libraries give users a way to modify values or even structure.  Regardless, users still may want to take some structure-preserving or opaque format, and "dissolve" it to get a transparent, modifiable format.

I'll take a look at the PR; thanks for tagging me!

cjnolet:
@mhoemmen, thanks so much! I'm looking forward to your feedback. I tried to find a way to generalize the different states w/ nomenclature that we could apply directly to the different options.

ajschmidt8:

cpp/include/raft/core/device_coo_matrix.hpp
```diff
@@ -0,0 +1,236 @@
...
+ * @tparam RowType
+ * @tparam ColType
+ * @tparam NZType
+ * @param[in] handle raft handle for managing expensive device resources
+ * @param[in] structure_ a sparsity-preserving coordinate structural view
+ * @return a sparsity-preserving sparse matrix in coordinate (coo) format
+ */
+template <typename ElementType, typename RowType, typename ColType, typename NZType>
+auto make_device_coo_matrix(raft::device_resources const& handle,
+                            device_coordinate_structure_view<RowType, ColType, NZType> structure_)
```
divyegala, line 121: Should this already be a `shared_ptr`, and we produce views we return a `shared_ptr` to users?

cpp/include/raft/core/detail/device_mdarray.hpp
```diff
@@ -137,6 +137,8 @@ class device_uvector {
     return device_reference<T const>{thrust::device_ptr<T const>{data_.data() + i}, data_.stream()};
   }
 
+  void resize(size_type size, rmm::cuda_stream_view stream) { data_.resize(size, stream); }
```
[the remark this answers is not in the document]
cjnolet: Yep this is just for the container policy though, which we are thankfully able to reuse. We can expose through mdarray at some point too.

cpp/include/raft/core/device_coo_matrix.hpp
cjnolet: No, the views are trivially copyable by design so we don't want to have to burden the user to have to produce a `shared_ptr` here. The `shared_ptr` is more of an implementation detail from the perspective of the `coo_matrix` because it's really only being used to unify the preserving and owning variants underneath without having to introduce bloat.

cpp/include/raft/core/csr_matrix.hpp
```diff
@@ -0,0 +1,307 @@
...
+  static constexpr auto get_sparsity_type() { return sparsity_type; }
+  using sparse_matrix_type =
+    sparse_matrix<ElementType,
+                  structure_type,
+                  csr_matrix_view<ElementType, IndptrType, IndicesType, NZType, is_device>,
+                  is_device,
+                  ContainerPolicy>;
+  using container_type = typename ContainerPolicy<ElementType>::container_type;
+
+  template <typename = typename std::enable_if<sparsity_type == SparsityType::OWNING>>
```
divyegala, line 275:
```suggestion
  template <typename = std::enable_if_t<sparsity_type == SparsityType::OWNING>>
```

cpp/include/raft/core/csr_matrix.hpp
```diff
@@ -0,0 +1,307 @@
...
+  csr_matrix(raft::resources const& handle,
+             IndptrType n_rows,
+             IndicesType n_cols,
+             NZType nnz = 0) noexcept(std::is_nothrow_default_constructible_v<container_type>)
+    : sparse_matrix_type(handle, n_rows, n_cols, nnz){};
+
+  // Constructor that owns the data but not the structure
+
+  template <typename = typename std::enable_if<sparsity_type == SparsityType::PRESERVING>>
```
divyegala, line 284:
```suggestion
  template <typename = std::enable_if_t<sparsity_type == SparsityType::PRESERVING>>
```

cpp/include/raft/core/csr_matrix.hpp
```diff
@@ -0,0 +1,307 @@
...
+  csr_matrix(raft::resources const& handle, std::shared_ptr<structure_type> structure) noexcept(
+    std::is_nothrow_default_constructible_v<container_type>)
+    : sparse_matrix_type(handle, structure){};
+
+  /**
+   * Initialize the sparsity on this instance if it was not known upon construction
+   * Please note this will resize the underlying memory buffers
+   * @param nnz new sparsity to initialize.
+   */
+  template <typename = std::enable_if<sparsity_type == SparsityType::OWNING>>
```
divyegala, line 294:
```suggestion
  template <typename = std::enable_if_t<sparsity_type == SparsityType::OWNING>>
```

cpp/include/raft/core/csr_matrix.hpp
```diff
@@ -0,0 +1,307 @@
...
+  void initialize_sparsity(NZType nnz)
```
divyegala, line 295: Same question as before. This function is compile-time hidden if you have `SparsityType::PRESERVING`. However, if you call `csr_matrix<..., SparsityType::PRESERVING, ...>::structure_view().initialize_sparsity()` then it's a runtime error now

cjnolet:
Opened up docs issue as follow-on https://github.com/rapidsai/raft/issues/1342.

divyegala (approved):

PR #1281, opened 2023-02-16
Fix `RAFT_USE_FAISS_STATIC` default value

trxcllnt:
This PR makes `RAFT_USE_FAISS_STATIC` default to `OFF` when `RAFT_COMPILE_LIBRARIES` is `ON`.

Currently, `RAFT_USE_FAISS_STATIC` defaults to `ON` when `RAFT_COMPILE_LIBRARIES` is `ON`.

This means configuring raft with all default options leads to building and statically linking faiss.

These are the new default values of `RAFT_USE_FAISS_STATIC` in this PR:

```shell
$ cmake -S . -B build 2>&1 | grep RAFT_USE_FAISS_STATIC
  RAFT_USE_FAISS_STATIC=OFF
$ cmake -S . -B build -DRAFT_COMPILE_LIBRARIES=ON 2>&1 | grep RAFT_USE_FAISS_STATIC
  RAFT_USE_FAISS_STATIC=OFF
$ cmake -S . -B build -DRAFT_COMPILE_LIBRARIES=OFF 2>&1 | grep RAFT_USE_FAISS_STATIC
  RAFT_USE_FAISS_STATIC=ON
```

It may also be good to make `RAFT_USE_FAISS_STATIC` depend on the value of `BUILD_SHARED_LIBS`, but I haven't done that here.

The [`cmake_dependent_option`](https://cmake.org/cmake/help/v3.26/module/CMakeDependentOption.html) docs state:
> Makes `<option>` available to the user if the [semicolon-separated list](https://cmake.org/cmake/help/v3.26/manual/cmake-language.7.html#cmake-language-lists) of conditions in `<depends>` are all true. Otherwise, a local variable named `<option>` is set to `<force>`.
>
> When `<option>` is available, the ... initial `<value>` [is] used.

To put it another way, the order of the values passed to `cmake_dependent_option` is similar to python ternaries:
```python
raft_use_faiss_static=(False if raft_compile_libraries else True)
```


cjnolet (approved):

cjnolet:
@trxcllnt now that RAFT no longer depends on FAISS, can we close this PR?

PR #1283, opened 2023-02-17
CI: Remove specification of manual stage for check_style.sh script.

csadorf:
Do not explicitly specify to run the "manual" stage when running pre-commits as part of the ci/check_style.sh script.

### DIFF
--- a/ci/check_style.sh
+++ b/ci/check_style.sh
@@ -15,4 +15,4 @@ rapids-mamba-retry env create --force -f env.yaml -n checks
 conda activate checks
 
 # Run pre-commit checks
-pre-commit run --hook-stage manual --all-files --show-diff-on-failure
+pre-commit run --all-files --show-diff-on-failure


csadorf:
> @csadorf, I've got no objections to this change, but can you elaborate on why it's being made?

It is simply not necessary. @vyasr can you elaborate?

ajschmidt8 (approved):

vyasr:
It's a historical artifact. We were originally manually running all linters in CI. At some point, [we discussed the possibility of centralizing around pre-commit usage to manage all style checks in CI](https://github.com/rapidsai/cudf/issues/9309). When I implemented that in https://github.com/rapidsai/cudf/pull/9412, we were still using a special script to run local installations of `clang-format` i.e. we assumed that the user would have `clang-format` installed in their local environment. To avoid that breaking in cases where it was not installed, I configured the `clang-format` hook so that it would only run in certain pre-commit stages so that whether or not `clang-format` would be included in a `pre-commit run` was controllable by specifying `--hook-stage=...`. Since [clang-format wheels became available](https://github.com/ssciwr/clang-format-wheel), cudf [switched to using them to run fully pre-commit-managed hooks](https://github.com/rapidsai/cudf/pull/10659). However, the `--hook-stage=manual` wasn't removed from the CI scripts at the time. As other RAPIDS repos followed cudf's pattern and copied the CI files, the same command invocation spread to everywhere. At this point it can be removed from every repo since all of our local hooks are either managing dependencies via `additional_dependencies` or are simply shell scripts with no dependencies at all.

@csadorf it's probably worth doing a quick check to verify that none of the repos where you've removed this have a `stages: ...` key present in their `.pre-commit-config.yaml` files. I am not aware of any, but I may be forgetting one somewhere.

csadorf:
> @csadorf it's probably worth doing a quick check to verify that none of the repos where you've removed this have a `stages: ...` key present in their `.pre-commit-config.yaml` files. I am not aware of any, but I may be forgetting one somewhere.

Checked, none of them do.

vyasr:
Great, thanks!

PR #1287, opened 2023-02-18
support k up to 2048 in faiss select

benfred:
The faiss repo allows k values up to 2048, but we were limiting to 512 or 1024 instead. This seems to be because of compiler errors such as

```
ptxas error   : Entry function '_ZN4raft7spatial3knn6detail15select_k_kernelIdmLb1ELi2048ELi8ELi64EEEvPKT_PKT0_mmPS4_PS7_S4_S7_i' uses too much shared data (0x10000 bytes, 0xc000 max)
ptxas error   : Entry function '_ZN4raft7spatial3knn6detail15select_k_kernelIdmLb0ELi2048ELi8ELi64EEEvPKT_PKT0_mmPS4_PS7_S4_S7_i' uses too much shared data (0x10000 bytes, 0xc000 max)
ptxas error   : Entry function '_ZN4raft7spatial3knn6detail15select_k_kernelIdmLb1ELi1024ELi8ELi128EEEvPKT_PKT0_mmPS4_PS7_S4_S7_i' uses too much shared data (0x10000 bytes, 0xc000 max)
ptxas error   : Entry function '_ZN4raft7spatial3knn6detail15select_k_kernelIdmLb0ELi1024ELi8ELi128EEEvPKT_PKT0_mmPS4_PS7_S4_S7_i' uses too much shared data (0x10000 bytes, 0xc000 max)
```

when compiling with larger k values.

However, these only affect double precision with 64bit indices - and float32 works up to k=2048 (even with 64 bit indices).


cpp/include/raft/spatial/knn/detail/selection_faiss.cuh
```diff
@@ -27,10 +27,13 @@ namespace spatial {
...
 constexpr int kFaissMaxK()
 {
-  return (sizeof(key_t) + sizeof(payload_t) > 8) ? 512 : 1024;
+  if (sizeof(key_t) >= 8) {
+    return sizeof(payload_t) >= 8 ? 512: 1024;
+  }
+  return 2048;
 }
 
 template <typename key_t, typename payload_t, bool select_min, int warp_q, int thread_q, int tpb>
```
cjnolet, line 39: Just for consistency could we make this one start with payload_t as well?

cjnolet (approved):
Looks good, just a small nitpick but I'm pre-approving

cjnolet (approved):
Github seems to be in a weird state

ajschmidt8 (approved):

cjnolet (approved):

PR #1292, opened 2023-02-21
Python API for brute-force KNN

### LINKED ISSUE #1289: [FEA] Expose brute-force knn through pylibraft
This should be a very straightforward wrap and since we are already going to be specializing the C++ brute-force API for some common types, the runtime API should be very straightforward to implement.

cjnolet:
Closes #1289


cjnolet:
Is there a way we can try to reproduce this on the raft side? It would be really useful to isolate this issue in a google test with multiple partitions in the index so once its fixed we can guarantee it doesn't happen again in follow-on changes.

benfred (commented):
looks good! I'm excited to see this functionality go in

benfred (approved):

Issue #875, opened 2022-10-03
[BUG] New gemm/gemv APIs only support col-major and don't accept leading dimensions


cjnolet:
During the addition of the new gemm/gemv APIs that use mdspan, we removed the ability to use these APIs w/ row-major inputs, as well as the ability to specify additional tiling offsets w/ the leading dimensions arguments. There are good examples in cuml of these being used, such as during the [kernel gramm matrix construction in svm](https://github.com/rapidsai/cuml/blob/branch-22.10/cpp/src_prims/matrix/grammatrix.cuh#L149). We need to support these options in order for cuml to use the new mdspan functions.

PR #1296, opened 2023-02-22
Gram matrix support for sparse input

mfoerste4:
This PR adds sparse input support (CSR) for GramMatrix kernel computation. This is a requirement to enable SVM support for sparse input in [cuML issue 2197](https://github.com/rapidsai/cuml/issues/2197). 

It also adds row norm computation for CSR which is utilized for expanded L2 norm computation within RBF kernels.

Although this branch introduces a new API it is still backwards compatible with the old GramMatrix API (which is marked as deprecated).

CC @cjnolet @tfeher


cpp/test/sparse/gram.cu
```diff
@@ -0,0 +1,342 @@
...
+
+    raft::random::Rng r(42137ULL);
+    r.uniform(x1.data(), x1.size(), math_t(0), math_t(1), stream);
+    r.uniform(x2.data(), x2.size(), math_t(0), math_t(1), stream);
+  }
+
+  ~GramMatrixTest() override { RAFT_CUDA_TRY_NO_THROW(cudaStreamDestroy(stream)); }
+
+  // Calculate the Gram matrix on the host.
+  void naiveKernel()
```
tfeher, line 168: This is the same as for dense, right? Could we reuse?

cpp/include/raft/distance/detail/matrix/matrix.hpp
```diff
@@ -0,0 +1,99 @@
...
+  };
+
+  int n_rows;
+  int n_cols;
+};
+
+template <typename math_t>
+class DenseMatrix : public Matrix<math_t> {
+ public:
+  DenseMatrix(math_t* data, int rows, int cols, bool row_major = false, int ld_in = 0)
```
tfeher, line 73: Ideally we would like to use mdspan for representing the dense matrix. Could you list the pain points that would motivate keeping a pointer based API still around?

tfeher (changes requested):
Hi Malte, thanks for the PR! I have looked at the changes of the Gram matrices. Overall it looks nice, below you will find my comments.

I have added the "braking" label to remind us that we have code in cuML that is affected by the interface changes (even though it is the detail namespace, so technically it should not be breaking).

I will share my comments about the new norm method separately.

cpp/test/sparse/normalize.cu
```diff
@@ -0,0 +1,127 @@
+/*
```
tfeher, line 1: This is a copy of the old tests from norm.cu.

cpp/include/raft/sparse/linalg/norm.cuh
```diff
@@ -66,6 +67,37 @@ void csr_row_normalize_max(const int* ia,  // csr row ind array (sorted by row)
...
+ *
+ * Row-wise norm is useful while computing pairwise distance matrix, for
+ * example.
+ * This is used in many clustering algos like knn, kmeans, dbscan, etc...
+ *
+ * @tparam Type the data type
+ * @tparam Lambda device final lambda
+ * @tparam IdxType Integer type used to for addressing
+ * @param dots the output vector of row-wise dot products
+ * @param ia the input matrix row pointers
```
tfeher, line 81:
```suggestion
 * @param ia the input matrix row index array
```

cpp/include/raft/sparse/linalg/norm.cuh
```diff
@@ -66,6 +67,37 @@ void csr_row_normalize_max(const int* ia,  // csr row ind array (sorted by row)
...
+ * @param data the input matrix nnz data
+ * @param N number of rows of data
+ * @param type the type of norm to be applied
+ * @param stream cuda stream where to launch work
+ * @param fin_op the final lambda op
+ */
+template <typename Type, typename IdxType = int, typename Lambda = raft::identity_op>
+void rowNormCsr(Type* dots,
```
tfeher, line 89: - How does the performance relate to `csr_row_normalize_l1` (and `.._max`)?
- Any other differences (apart from implementation detail)?

tfeher (changes requested):
Thanks Malte for updating the PR! Here are my second batch of comments (related to csr norm).

We need to find a  way not to break cuML with this PR. Can we just keep the old implementation as on overload, and remove it once the corresponding cuML PR is accepted?

cpp/include/raft/distance/detail/matrix/matrix.hpp
tfeher: - I am fine keeping this version until the sparse wrapper PR #1279 is merged, and consolidate the interface afterwards.
- Note that we have helper functions to deal with mdspan on the Python side, see usage example here
https://github.com/rapidsai/raft/blob/03e26b520ea281d9b239938eda30dd5100559e24/python/pylibraft/pylibraft/neighbors/ivf_pq/ivf_pq.pyx#L523

tfeher (approved):
Thanks Malte for addressing the issues! The PR looks good to me, but let's wait for feedback from @cjnolet on the interface changes.

tfeher:
@mfoerste4 after restoring the old API, we do not contain braking changes anymore, right? Could you then remove the warning message from the PR description?

cpp/include/raft/distance/detail/matrix/matrix.hpp
cjnolet: I agree with @tfeher, rather than introducing more abstractions for this, we can unify the dense and sparse APIs using overloads of the existing abstractions (with templates if needed).

We didn’t have abstractions like the mdspan available when we were building cuml so we used pointers for everything. Now that we have the proper abstractions and infrastructure in place, we should use them.

please note that the sparse API has been merged. We should be able to add overloads for both dense and sparse without having an explicit class to unify them.

cjnolet (changes requested):

cpp/include/raft/distance/detail/matrix/matrix.hpp
cjnolet: Tagging @divyegala for his thoughts here regarding public APIs and template metaprogramming techniques that might help.

tfeher: > rather than introducing more abstractions for this, we can unify the dense and sparse APIs using overloads of the existing abstractions (with templates if needed).

@cjnolet Having overloads instead of on unified matrix type might make sense when we consider this PR only. When looking at how this is used from cuML, that would motivate introducing a unified sparse dense type.

Let us consider the whole call chain in https://github.com/rapidsai/cuml/pull/5273:
- Cython warppers -> SvcFit -> SmoSolver.Solve -> KernelCache.getTile -> GramMatrix.evaluate
- Cython wrappers -> SvcPredict -> KernelCache.getTile -> GramMatrix.evaluate

Having overloads on this whole call chain is not feasible. Having the matrix type as template parameter would unnecessary add extra compilation steps when all we would need to pass on some metadata about the array structures.

The simplest thing to do is to wrap the input into a unified struct as high in the call chain as possible, and unpack it only  where it is needed (only in GramMatrix). The QN solver in cuML does the same: [pack inputs into a unified sturcture at the entry point](https://github.com/rapidsai/cuml/blob/6976c8b893c57b4a8ae9a4929d06cb2dc0dc4b5e/cpp/src/glm/qn/qn.cuh#L161-L200).

If you are against a unified sparse / dense type in raft, then we could keep that in cuML, but it I think we should think about having such interface in raft.

divyegala: From my understanding, the code duplication in `GramMatrix` is only at the API layer but the actual logic for each input types are separate pieces anyway. Here's my suggestion, since it's trivial to overload the API layer as that doesn't have significant impact to compile time or binary size and I don't think we need any template methods here. For `GramMatrix`:
1. Keep the dense API
2. Keep the sparse API
3. Introduce unified API. This is just a facade, we immediately dispatch to the above two

This way, we solve all issues. The cuML call chain is simplified, and for users of RAFT that may only care about either the sparse API or the dense API can call that directly without us forcing them to add another type layer of `unified` before interacting with RAFT.

cjnolet: My apologies @tfeher and @mfoerste4, I meant to revisit this discussion myself and things have been pretty crazy the past couple of weeks. Please understand that my scrutiny over adding new types is only because 

1) I want to make sure that we're not exploding the number of types that a user needs work with, and 
2) making sure that we aren't exploding the number of new types each time we encounter a new pattern, unless said types can be at least moderately reusable across the codebase.

It sounds like the goal here is to have a unified base class. So far, the cython APIs in RAFT have gotten pretty tedious to code- and that's just for host/device mdspan. Adding sparse to the mid and accepting that on host and device is going to make things even more tedious. I definiely understand the reasoning here. 

Currently, the `SimpleMat` contains linear algebra implementations. If the goal here is just to have a unified base class that can accept read-only inputs, could we maybe just create a very simple class called something like `read_only_matrix_view` which we extend for `csr_read_only_matrix_view` and `dense_read_only_matrix_view`?

cjnolet: > As for cuML, in addition to the read-only views we also need a data/structure-owning matrix wrapper that allows for resizing internal data structures ([ResizableCsrMatrix](https://github.com/mfoerste4/cuml/blob/078ba5dc744477e1a8c57c0799fe8af981e4c3a9/cpp/src/svm/sparse_util.cuh#L34)).

It sounds like what you need there is indeed a `structure-owning` CSR matrix. Why not use [the one](https://github.com/rapidsai/raft/blob/branch-23.04/cpp/include/raft/core/device_csr_matrix.hpp#L156) that's already in RAFT? Again, it looks like we're duplicating types, which means increased maintenance burden.

cpp/include/raft/core/device_mdspan.hpp
```diff
@@ -259,6 +259,36 @@ auto make_device_matrix_view(ElementType* ptr, IndexType n_rows, IndexType n_col
...
+ * @tparam ElementType the data type of the matrix elements
+ * @tparam IndexType the index type of the extents
+ * @param[in] ptr on device to wrap
+ * @param[in] n_rows number of rows in pointer
+ * @param[in] n_cols number of columns in pointer
+ * @param[in] is_row_major whether the data is in row major format (column major otherwise)
+ * @param[in] ld leading dimension / stride of data
+ */
+template <typename ElementType, typename IndexType = std::uint32_t>
+auto make_device_matrix_view(
```
cjnolet, line 276: This feels a little misplaced. First- we should be accepting the `layout_type`, rather than a simple boolean, for row_major/col_major (and we an constrain to a contiguous/strided type inside the function itself using validation.). 

From a user's perspective, I'd most often expect that a pointer would be given the leading dimension offset added before creating an mdspan from it. I suggest we create a helper/utility function that accepts a a contiguous/strided mdspan, along w/ a leading dimension and returns a new mdspan. If this is only expected to be used for testing then maybe we could just put this utility in `gram_base` for now? If you expect this to be more broadly used, maybe we should put the utility function in `core/mdspan.hpp` so it's more broadly usable (the accessor template will carry along whether it's device/host accessible).

cpp/include/raft/distance/detail/kernels/gram_matrix.cuh
```diff
@@ -16,13 +16,25 @@
...
+#include <raft/distance/distance_types.hpp>
+#include <raft/sparse/detail/cusparse_wrappers.h>
+#include <raft/sparse/distance/distance.cuh>
 
 #include <raft/linalg/detail/cublas_wrappers.hpp>
 #include <raft/linalg/gemm.cuh>
 
 namespace raft::distance::kernels::detail {
 
+template <typename math_t>
```
cjnolet, line 31: We should also be exposing these publicly so users like cuml aren't using the `detail` namespace.

cpp/include/raft/sparse/linalg/norm.cuh
```diff
@@ -66,6 +67,37 @@ void csr_row_normalize_max(const int* ia,  // csr row ind array (sorted by row)
...
+ * @param norm the output vector of row-wise norm, size [N]
+ * @param ia the input matrix row index array
+ * @param data the input matrix nnz data
+ * @param N number of rows of data
+ * @param type the type of norm to be applied
+ * @param stream cuda stream where to launch work
+ * @param fin_op the final lambda op
+ */
+template <typename Type, typename IdxType = int, typename Lambda = raft::identity_op>
+void rowNormCsr(Type* norm,
```
cjnolet, line 89: Please accept `raft::device_resources` on public APIs and pass the `cudaStream_t` down to the `detail` namespace. Arguments should be ordered as follows:
1. `device_resources`
2. input pointers
3. output pointers
4. additional parameters

cpp/include/raft/distance/detail/kernels/gram_matrix.cuh
```diff
@@ -182,37 +312,257 @@ class GramMatrixBase {
     }
   }
 
-  /** Calculates the Gram matrix using Euclidean distance.
+ protected:
+  bool get_is_row_major(dense_output_matrix_view_t<math_t> matrix)
```
cjnolet, line 316: Do we already have utilities for this?

cpp/include/raft/distance/detail/kernels/gram_matrix.cuh
```diff
@@ -182,37 +312,257 @@ class GramMatrixBase {
...
+                                                order));
+
+    auto alg = order == CUSPARSE_ORDER_COL ? CUSPARSE_SPMM_CSR_ALG1 : CUSPARSE_SPMM_CSR_ALG2;
+
+    // compute X1*X2^T
+    auto opX1 = CUSPARSE_OPERATION_NON_TRANSPOSE;
+    auto opX2 = CUSPARSE_OPERATION_TRANSPOSE;
+
+    size_t bufferSize;
+    RAFT_CUSPARSE_TRY(raft::sparse::detail::cusparsespmm_bufferSize(handle.get_cusparse_handle(),
```
cjnolet, line 474: We should be exposing these operations directly through our public APIs that can accept our sparse types as arguments. Can you please move this logic to a `raft/sparse/linalg/spmm.cuh` primitive like we've done for `raft/linalg/gemm.cuh`? It could just accept pointers for now. It's on our roadmap to scrape through the sparse APIs and accept our new types everywhere.

cpp/include/raft/distance/detail/kernels/gram_matrix.cuh
```diff
@@ -182,37 +312,257 @@ class GramMatrixBase {
...
+    // extract major stride
+    int ld1    = is_row_major ? x1.stride(0) : x1.stride(1);
+    int ld2    = is_row_major ? x2.stride(0) : x2.stride(1);
+    int ld_out = is_row_major ? out.stride(0) : out.stride(1);
+
+    math_t alpha = 1.0;
+    math_t beta  = 0.0;
+    if (is_row_major) {
+      // #TODO: Call from public API when ready
+      RAFT_CUBLAS_TRY(raft::linalg::detail::cublasgemm(handle.get_cublas_handle(),
```
cjnolet, line 370: Please use `raft/linalg/gemm.cuh` (preferably the mdspan-based APIs) for this instead fo invoking cublas wrappers directly.

cjnolet (changes requested):
@mfoerste4 I finally got a chance to review these changes more thoroughly. Overall, I think the general design looks fine but there's a few other things here to consider a few things to expose through the public API (both for reuse and so users aren't having to use the `detail` APIs.)

cpp/include/raft/distance/detail/kernels/gram_matrix.cuh
cjnolet: Can you explain how this would break cuml? cuML is currently not using the detail namespace for these because we're exposing them through `raft::distance::kernels` public APIs.  Such as [here](https://github.com/rapidsai/cuml/blob/branch-23.04/cpp/src/svm/smosolver.cuh#L84)

cpp/include/raft/sparse/linalg/norm.cuh
cjnolet: All of the pointer-based (dense) public APIs are deprecated in RAFT in favor of the mdspan-based APIs. We're going to be scraping through the sparse public APIs and using the new sparse types everywhere. While I'd prefer that we no longer create pointer-based sparse APIs, I do understand that we don't yet have an alternative at the moment. Still, I'd prefer to at least keep the ordering of new APIs consistent w/ the agreed-upon ordering for consistency (the dense norm computation is both deprecated and didn't yet implement the ordering.). We should always accept `device_resources` in new APIs instead of `cudaStream_t` directly.

cpp/include/raft/distance/detail/kernels/gram_matrix.cuh
cjnolet: Ah, yes you are right. We still have [an issue](https://github.com/rapidsai/raft/issues/875) open for that. I'm okay using the pointer-based APIs of `raft::linalg::gemm` in the meantime. Can you reference the issue I just linked in a comment at the call site for tracking?

cpp/include/raft/sparse/linalg/detail/norm.cuh
```diff
@@ -170,6 +173,93 @@ void csr_row_normalize_max(const int* ia,  // csr row ind array (sorted by row)
...
+                     MainLambda main_op,
+                     ReduceLambda reduce_op,
+                     FinalLambda final_op)
+{
+  IdxType i = threadIdx.y + (Policy::RowsPerBlock * static_cast<IdxType>(blockIdx.x));
+  if (i >= N) return;
+
+  Type acc = init;
+  for (IdxType j = ia[i] + threadIdx.x; j < ia[i + 1]; j += Policy::LogicalWarpSize) {
+    acc = reduce_op(acc, main_op(data[j]));
```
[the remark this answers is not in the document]
cjnolet: Given that this is almost a 1:1 w/ the `csr_row_op`, I'd prefer that we use that underneath at least. It's a pretty trivial operation but the more we reuse and centralize, the more further optimizations will propagate.

cpp/include/raft/core/device_mdspan.hpp
cjnolet: I guess I'm thinking of something like this:
```suggestion
auto make_device_matrix_view(
  ElementType* ptr, IndexType n_rows, IndexType n_cols, bool is_row_major, IndexType ld)
{
  IndexType stride0 = is_row_major ? (ld > 0 ? ld : n_cols) : 1;
  IndexType stride1 = is_row_major ? 1 : (ld > 0 ? ld : n_rows);

  assert(is_row_major ? stride0 >= n_cols : stride1 >= n_rows);

  matrix_extent<IndexType> extents{n_rows, n_cols};
  std::array<IndexType, 2> strides{stride0, stride1};
  using mapping_type  = typename layout_stride::template mapping<matrix_extent<IndexType>>;
  mapping_type layout = {extents, strides};

  return device_matrix_view<ElementType, IndexType, layout_stride>{ptr, layout};
}
```
```
template<typename T, typename IdxT, typename LayoutPolicy = layout_c_contiguous>
auto make_strided_device_matrix_view(IdxT nrows, IdxT ncols, IdxT stride) {
    constexpr auto is_row_major = std::is_same_v<LayoutPolicy, layout_c_contiguous>;
    IndexType stride0 = is_row_major ? (ld > 0 ? ld : n_cols) : 1;
    IndexType stride1 = is_row_major ? 1 : (ld > 0 ? ld : n_rows);   

    assert(is_row_major ? stride0 >= n_cols : stride1 >= n_rows);
    matrix_extent<IndexType> extents{n_rows, n_cols};

    auto layout = make_strided_layout(extents, std::array<IndexType, 2>{stride0, stride1});
    return device_matrix_view<ElementType, IndexType, layout_stride>{ptr, layout};
}
```

cjnolet:
@mfoerste4 just a note- I bumped this to 23.06. I'll re-review so we can get this in quickly.

cpp/include/raft/distance/kernels/kernel_matrices.cuh
```diff
@@ -0,0 +1,741 @@
...
+/** Epiloge function for polynomial kernel without padding.
+ * Calculates output = (gain*in + offset)^exponent
+ * @param inout device vector in column major format, size [len]
+ * @param len array length
+ * @param exponent
+ * @param gain
+ * @param offset
+ */
+template <typename math_t, typename exp_t>
+__global__ void polynomial_kernel_nopad(
```
cjnolet, line 37: The reason these objects were in detail originally and only the public bits exposed via typedef (`using`) was because these kernels shouldn't be exposed to the public APIs. Also, this class-based (mostly-stateless) design for host-callable computation runs counter to all of the other APIs in RAFT. I think we've all agreed that we should flatten these APIs into a series of stateless functions that can be called and have their state passed into them, but we haven't yet prioritized that, thus I wanted to keep the code out of the docs and obscure it from being discoverable by users until we have the new "formal" API exposed. I suggest we move these back into the `detail` namespace and expose them with `using` into the public namespace but also expose the other bits we need from this API to cuml in the same way (with `using`) so that cuml isn't calling APIs in the detail namespace. 

I do apologize if my wording has been confusing in previous reviews.

cpp/include/raft/sparse/linalg/spmm.cuh
```diff
@@ -0,0 +1,78 @@
...
+ * @param[in] y input raft::device_matrix_view
+ * @param[in] beta scalar
+ * @param[out] z output raft::device_matrix_view
+ */
+template <typename ValueType,
+          typename IndexType,
+          typename NZType,
+          typename LayoutPolicyY,
+          typename LayoutPolicyZ>
+void spmm(raft::device_resources const& handle,
```
cjnolet, line 51: Tagging @divyegala for awareness thoughts since he's starting on exposing the sparse APIs w/ our new core vocabulary.

cjnolet (changes requested):
Thanks for the updates, @mfoerste4. The new `norm` and `sparse::spmm` APIs look great. I think this is almost there.

cjnolet (approved):
LGTM. Thanks Malte!

cjnolet:
@mfoerste4 a few recent updates to the sparse API broke your changes here. I can make these updates and push to your banch if it's easier (since I know what changed).

PR #1297, opened 2023-02-23
Adding util to get the device id for a pointer address

cjnolet:


cpp/include/raft/util/cudart_utils.hpp
```diff
@@ -25,6 +25,7 @@
 #pragma once
 
 #include <raft/core/error.hpp>
+#include <raft/core/logger.hpp>
```
benfred, line 28: whats the logger include for ?

benfred (approved):
lgtm!

cpp/include/raft/util/cudart_utils.hpp
cjnolet: Good catch. I was logging some things initially but decided to remove it from the function.

PR #1298, opened 2023-02-23
IVF-PQ: manipulating individual lists

achirkin:
Add public functions for reading and writing into individual ivf-pq lists (clusters), in the input space (reconstructed data) and in flat PQ codes.

Partially solves (IVF-PQ) https://github.com/rapidsai/raft/issues/1205


cpp/include/raft/neighbors/ivf_pq.cuh
```diff
@@ -75,6 +75,48 @@ auto build(raft::device_resources const& handle,
...
+ *   The id of the list (cluster) to decode.
+ * @param[in] n_skip
+ *   How many records in the list to skip.
+ */
+template <typename T, typename IdxT>
+void reconstruct_list_data(raft::device_resources const& res,
+                           const index<IdxT>& index,
+                           device_matrix_view<T, uint32_t, row_major> out_vectors,
+                           uint32_t label,
+                           uint32_t n_skip)
```
achirkin, line 115: @cjnolet , could you please check if this interface is suitable for integration with FAISS?

achirkin: Also don't you need something like this interface to be able to convert our GPU index to their CPU index type (more-or-less efficiently, converting one cluster at a time)?

achirkin: So, if I understand this correctly, the current proposed API here covers the CodePacker needs, right?

cjnolet: If this is intended to cover the code packer's needs, I'd like it if we adjusted the verbiage there, and maybe even use the verbiage from the code packer itself (e.g. `pack_single`, `unpack_all`) because the two use-cases are pretty different. Naming aside, I think this *could* work for the `unpack` side. We will still need the version that goes the other way, though.

achirkin: As for the second task, I'm not sure how we're supposed to accomplish it. As I said, we don't sort input indices in the database, so locating a single id in the database would entail a full linear scan. We can get some speedup thanks to batching, but that still sounds really slow. Do you know if it is the same in faiss?

cjnolet: > We can get some speedup thanks to batching, but that still sounds really slow. Do you know if it is the same in faiss?

Maybe I've misunderstood how we handle being given indexes today. What happens today when someone calls `ivf_pq::extend` for the following 3 indexes: `[500, 2, 50]`? Let's say these 3 are the only 3 indexes in the database.

achirkin: At the moment, we store the cluster data and indices in `ivf::list` structures for each cluster separately. That is, you can get a list of source indices for each cluster. In your example, with two clusters we may end up with two index lists `[[500], [2, 5]]`. To complicate things further,  `ivf_pq::extend`  writes vectors/indices in an undefined order for performance reasons (concurrent `atomicAdd` involved).

As a result we cannot rely in any way on the order in which the data is written. That being said, we can sort all indices by their source id in a "preparation" phase for a given index, so that the indices can be reconstructed faster afterwards.

Btw, the same issue applies to ivf-flat. Luckily, the interleaved layout does not affect the indices in both implementations, they are stored flat (per-cluster).

cjnolet: > At the moment, we store the cluster data and indices in ivf::list structures for each cluster separately.

Okay, so you are saying that the indices provided from the user are directly stored in the sparse structure representing the IVF lists? So, for example, the index 500 would still be associated w/ the first (and only) vector in the  first cluster (going from your example cluster assignments)? If this is the case, you are right that we would need to use a hash table or sort and use a binary search to look this up. At least that tells me it could be possible, nonetheless. 

If, on the other hand, we are normalizing the given index so that it's linear with all the other indexes then I think all bets are off.

achirkin: Yes, basically, we have to store the user indices, because this is what we return from the `ivf_*::search` along with the distances. Hence we can create a mapping {user index -> (cluster, incluster_offset)} and use it for faster retrieval by user index. The only problem is that the time complexity of creating it would be something like O(n log n) and it would require at least `(sizeof(IdxT) + 8) * n` extra storage for the struct itself, both of which can be a lot if we're talking about billion-record ivf-pq databases.

achirkin: I'd suggest I enhance the interface to accept (cluster, list of in-cluster indices/offsets) in addition to the current one in this PR. Then in a follow-up we can add this index mapping struct directly to `ivf::list` module so it can be shared between ivf-flat and ivf-pq.

cjnolet: Sorry, @achirkin I'm still kind of stuck on the CodePacker problem because that's the blocker at the moment the integration. Can you explain how this interface allows us to implement the CodePacker interface for RAFT's IVFPQ methods? Im having trouble figuring out where the information (like block size, etc...) is specified. Is that a constant? 

In the end, we are going to have to fill that in for RAFT in order to convert to/from FAISS cpu/gpu indexes. The FAISS folks have asked us to use the CodePacker API instead of the reconstruct_n API. It sounds like we are almost there but I think there might be a gap in my understanding of your proposed interface.

cjnolet: Yes, they are using the per-block flat vectors as an intermediate format for conversion. So, for example, we should be able to loop through each vector or block and write its codes in the flat form using one code packer implementation and then immediately unpack from flat to another code packer implementation- they provide APIs for doing 1, many, and all. If we do this right, it should be super easy to convert to/from FAISS existing gpu formats in addition to the cpu formats.

cpp/include/raft/neighbors/detail/ivf_pq_build.cuh
```diff
@@ -557,6 +557,178 @@ void train_per_cluster(raft::device_resources const& handle,
...
+{
+  const auto out_dim = out_vectors.extent(1);
+  using layout_t     = typename decltype(out_vectors)::layout_type;
+  using accessor_t   = typename decltype(out_vectors)::accessor_type;
+
+  const uint32_t pq_dim = out_dim / pq_centers.extent(1);
+  auto pq_extents =
+    list_spec<uint32_t>{PqBits, pq_dim, true}.make_list_extents(out_vectors.extent(0) + n_skip + 1);
+  auto pq_dataset =
+    make_mdspan<const uint8_t, uint32_t, row_major, false, true>(data_ptrs[cluster_ix], pq_extents);
```
tfeher, line 642: Can't we access the mdspan from the list struct, instead of recreating it here?

tfeher (commented):
Thanks Artem for the PR! As far as I understand we are still waiting for additional helpers to enable looking up vectors using the original index. Until that is ready, I have looked at the implementation details. It looks good so far, below you will find my comments.

cpp/include/raft/neighbors/ivf_pq.cuh
achirkin: Update.

I've implemented `unpack_list_data` so far with a small change in the logic. I always write one PQ code per byte, ignoring `pq_bits` (faiss does not support smaller code sizes). Hence the constants change a little bit:
```
nvec == list.values.extent(2) == kIndexGroupSize;
code_size == pq_dim;
block_size ==  list.values.extent(1) *  list.values.extent(2) * list.values.extent(3)
```

NB: it's not a problem to pack the flat codes using `pq_bits` bits per code, but I figure there's simply no use for it; if you think otherwise, we can make this change.

achirkin: @cjnolet I've updated the public API (implementation follows soon), could you have a look?

Note, although the public api provides now a way to read raw flat PQ codes, it's not suitable for the CodePacker, because it's not raw pointers. However, one of the corresponding functions in `detail` namespace hopefully should work for this purpose.

cpp/include/raft/neighbors/ivf_pq.cuh
```diff
@@ -230,11 +230,180 @@ auto build(raft::device_resources const& handle,
...
+ *   the destination buffer [n_take, index.pq_dim()].
+ *   The length `n_take` defines how many records to unpack,
+ *   it must be smaller than the list size.
+ * @param[in] label
+ *   The id of the list (cluster) to decode.
+ * @param[in] offset
+ *   How many records in the list to skip.
+ */
+template <typename IdxT>
+void unpack_list_data(raft::device_resources const& res,
```
cjnolet, line 265: Can we please start moving additional functions like this and the others in this PR into a different namespace like `raft::neighbors::ivf_pq::helpers`? I'm also suggesting that we do this with the `kmeans` API and other APIs that provide functions in addition to the standard operations. 

These new helper functions are extremely useful, but i also wonder if it's sustainable for us to keep adding these helper functions to the same public API headers- they run the risk of growing without bound and it makes them harder to navigate. What do you think about establishing a convention to put extra helpers like these in a separate header like `raft/neighbors/ivf_pq_helpers.cuh`? We could do the same w/ kmeans `raft/cluster/kmeans_helpers.cuh`. From what I've gathered, most of the helper functions should not need to call any of the non-helper public API functions so they *should* be pretty independent.

cpp/include/raft/neighbors/ivf_pq.cuh
```diff
@@ -230,11 +230,180 @@ auto build(raft::device_resources const& handle,
...
+{
+  return detail::unpack_list_data<IdxT>(res, index, out_codes, label, offset);
+}
+
+/**
+ * @brief Unpack a series of records of a single list (cluster) in the compressed index
+ * by their in-list offsets, one code per byte (independently of pq_bits).
+ *
+ * Usage example:
+ * @code{.cpp}
```
cjnolet, line 279: Thank you for these!

cpp/include/raft/neighbors/ivf_pq.cuh
```diff
@@ -296,6 +465,66 @@ void extend(raft::device_resources const& handle,
   detail::extend(handle, idx, new_vectors, new_indices, n_rows);
 }
 
+/**
+ * @brief Extend one list of the index in-place, by the list label, skipping the classification and
```
cjnolet, line 469: Can we get brief usage examples for these as well that a user could copy/paste into their code?

cjnolet (changes requested):
I think this generally looks good, though I'll need to go back over the codepacker API just to make sure we'll be able to fill it in with these functions. Just a few other (mostly minor) API-ish things from my side. 

> Note, although the public api provides now a way to read raw flat PQ codes, it's not suitable for the CodePacker, because it's not raw pointers. However, one of the corresponding functions in detail namespace hopefully should work for this purpose (detail::pack_list_data and detail::unpack_list_data).

I think we should use raft::span` here- especially since we know the `block*` will be of size `block_size` and the `flat_codes` will be size `nvec*code_size` and

achirkin:
I'm not sure, what do you mean by using `raft::span` for the codepacker? Probably, this my comment was a bit misleading. I think the problem with codepacker integration is that it takes the pointers to arbitrary locations within the index / lists as arguments. I don't think we would want to expose such a matching public function, because it goes against the hiding implementation details logic. Instead, I chose to reorganize the internal code, so that `detail::pack_list_data` and `detail::unpack_list_data` matched the codepacker interface and didn't take the index itself as an argument (nb: it's still mdspans, not raw pointers).

cpp/include/raft/neighbors/ivf_pq.cuh
cjnolet: The purpose of these pack and unpack APIs is so that we can call them through FAISS. We can't invoke anything below the public APIs external to RAFT, so we need to expose these publicly. Exposing them in a `helper` namespace further highlights to the user that these aren't required for normal operation but provide additional functionality that might be desired. 

In order to integrate w/ FAISS, we need to be able to provide a (stable) contract to them so that we can help them convert between different index types through these codepacker APIs. I agree that how these pack/unpack functions ultimately work "under the hood" should be hidden behind `detail` but the functions themselves to perform the pack/unpack, should be publicly accessable. That says to the world "We will not change these APIs, nor the way they function, without deprecation first". 

I'm only suggesting that we use `raft::span` in the case where we don't feel `mdspan` accurately represents a type. If an mdspan vector can represent this, that's great!

achirkin: > That says to the world "We will not change these APIs, nor the way they function, without deprecation first".

Oh, now I see what you mean. I agree, we should export `detail::pack_list_data` and `detail::unpack_list_data` into a public namespace, to make sure we don't change their signature and break faiss integration by accident. I'm also not opposed to the idea of moving the construct/reconstruct and other similar functions to the `ivf_pq::helpers` public namespace.
The only thing that bothers me, is that the overloads we need for the codepacker go really against the current design of the public api. Maybe, as an exception, we export these two into something like `ivf_pq::codepacker` to discourage other users from using them? Or even implement the whole codepacker api in raft in that namespace?

tfeher:
@cjnolet Since Artem is on leave, I am addressing the open issues. Will push update later today.

cpp/include/raft/neighbors/ivf_pq.cuh
tfeher: I have added a public API for pack_list_data.

cpp/test/neighbors/ann_ivf_pq.cuh
```diff
@@ -221,17 +247,136 @@ class ivf_pq_test : public ::testing::TestWithParam<ivf_pq_inputs> {
...
+    ivf_pq::helpers::extend_list_with_codes<IdxT>(
+      handle_, index, codes.view(), indices.view(), label);
+
+    auto& new_list = index->lists()[label];
+    ASSERT_NE(old_list.get(), new_list.get())
+      << "The old list should have been shared and retained after ivf_pq index has erased the "
+         "corresponding cluster.";
+
+    ASSERT_TRUE(devArrMatch(
+      old_list->data.data_handle(), new_list->data.data_handle(), n_rows, Compare<uint8_t>{}));
```
tfeher, line 340: @achirkin instead of `n_rows` shouldn't we compare something like `n_rows * pq_dim * pq_bits/8` elements?

tfeher (commented):
I have addressed the open issues.

cpp/include/raft/neighbors/ivf_pq.cuh
```diff
@@ -296,6 +465,66 @@ void extend(raft::device_resources const& handle,
...
+ * @param[inout] index
+ * @param[in] new_vectors data to encode [n_rows, index.dim()]
+ * @param[in] new_indices source indices [n_rows]
+ * @param[in] label the id of the target list (cluster).
+ *
+ */
+template <typename T, typename IdxT>
+void extend_list(raft::device_resources const& res,
+                 index<IdxT>* index,
+                 device_matrix_view<const T, uint32_t, row_major> new_vectors,
```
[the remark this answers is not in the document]
achirkin: Across `ivf_pq` modules, I always use `uint32_t` for the indices/lengths within clusters or number of clusters. `ivf_pq::search` is optimized under the assumption that both the number of clusters and the sizes of clusters are within 32-bit limits (it may potentially incur some bugs in the other functions as well if we break this assumption).

cpp/test/neighbors/ann_ivf_pq.cuh
achirkin: Good point, you're right! Probably better replace it with `n_rows * .extent(1) * .extent(2) * .extent(3)` though.

cpp/include/raft/neighbors/ivf_pq.cuh
achirkin: Note, for the codepacker, we'd probably need to expose this overload https://github.com/rapidsai/raft/blob/926f510aab3ec16844b9b18b4c04fcc51e8f1f7d/cpp/include/raft/neighbors/detail/ivf_pq_build.cuh#L575-L580

cpp/test/neighbors/ann_ivf_pq.cuh
tfeher: I have used `(n_rows / kIndexGroupSize) .extent(1) * .extent(2) * .extent(3)`, (where `extent(3) == kIndexGroupsize`). Ceildiv would not work because of uninitialized values in the last group.

cpp/include/raft/neighbors/ivf_pq.cuh
tfeher: I have missed this message before. I have  exposed the pack/unpack versions  that take list_data directly under `ivf_pq::helpers::codepacker` namespace.

In our implementation we still need to provide the `pq_bits` parameter. Otherwise it is very similar to the CodePacker API.

One question: what does `block` refer to in CodePacker? In our API it refers to a set of interleaved blocks, which can be the data for a whole cluster, or less.

cjnolet (approved):
LGTM. Thanks Tamas!

PR #1301, opened 2023-02-23
Stop using versioneer to manage versions

vyasr:
This PR replaces usage of versioneer with hard-coded version numbers in setup.py and init.py. Since raft needs to manage versions across a wide range of file types (CMake, C++, Sphinx and doxygen docs, etc), versioneer cannot be relied on as a single source of truth and therefore does not allow us to single-source our versioning to the Git repo as is intended. Additionally, since the primary means of installing raft is via conda packages (or now, pip packages), information from the package manager tends to be far more informative than the version strings for troubleshooting and debugging purposes. Conversely, the nonstandard version strings that it produces tend to be problematic for other tools, which at best will ignore such versions but at worst will simply fail.

This PR also replaces usage of an environment variable to set the package name for wheels in setup.py, instead moving the renaming logic into the same sed script used to update package versions. This change makes setup.py essentially static, paving the way for migration to pyproject.toml.


ajschmidt8 (approved):

cjnolet (approved):

benfred (approved):

PR #1304, opened 2023-02-24
Add end-to-end CUDA ann-benchmarks to raft

### LINKED ISSUE #1211: [FEA] Integrate cuda ANN benchmarks
This is just a tracker issue to integrate our cuda ANN benchmarks code once it's ready to be open-sourced. 

cc @tfeher @achirkin @teju85

cjnolet:
New `bench/ann` artifact for comparing (C++ APIs for) GPU-acclerated algorithms end-to-end. Working on this w/ @tfeher but had to squash the original commits into a single commit.

Things left to do:
- [x] Separate `benchmarks` executables for each different algorithm
- [x] Separate build targets for `ggnn` and `hnswlib`
- [x] Revise `bench/ann` docs
- [x] Break `factory.cuh` abd `benchmark.cu` / `benchmark.cpp` into individual files for each different algorithm to make it easier to plug in new algorithms. 
- [x] Separate into its own conda package

closes #1211


raydouglass (approved):

cpp/bench/ann/src/common/dataset.h
```diff
@@ -0,0 +1,381 @@
...
+  mutable T* mapped_base_set_ = nullptr;
+};
+
+template <typename T>
+Dataset<T>::~Dataset()
+{
+  delete[] base_set_;
+  delete[] query_set_;
+  if (d_base_set_) { RAFT_CUDA_TRY_NO_THROW(cudaFree(d_base_set_)); }
+  if (d_query_set_) { RAFT_CUDA_TRY_NO_THROW(cudaFree(d_query_set_)); }
```
benfred, line 274: Rather than managing our own memory - can we use things like rmm::device_uvector etc here (and likewise some kind of smart pointer for the host memory)?

cpp/bench/ann/src/common/util.cpp
```diff
@@ -0,0 +1,68 @@
...
+}
+
+bool file_exists(const std::string& filename)
+{
+  struct stat statbuf;
+  if (stat(filename.c_str(), &statbuf) != 0) { return false; }
+  return S_ISREG(statbuf.st_mode);
+}
+
+bool dir_exists(const std::string& dir)
```
benfred, line 44: Since we're using c++17 - how about using the version in the std library ? https://en.cppreference.com/w/cpp/filesystem/is_directory

cpp/bench/ann/src/common/util.cpp
```diff
@@ -0,0 +1,68 @@
...
+  std::vector<std::string> tokens;
+  std::string token;
+  std::istringstream iss(s);
+  while (getline(iss, token, delimiter)) {
+    if (!token.empty()) { tokens.push_back(token); }
+  }
+  return tokens;
+}
+
+bool file_exists(const std::string& filename)
```
benfred, line 37: https://en.cppreference.com/w/cpp/filesystem/is_regular_file

cpp/bench/ann/src/common/util.h
```diff
@@ -0,0 +1,79 @@
...
+};
+
+std::vector<std::string> split(const std::string& s, char delimiter);
+
+bool file_exists(const std::string& filename);
+bool dir_exists(const std::string& dir);
+bool create_dir(const std::string& dir);
+
+template <typename... Ts>
+void log_(const char* level, Ts... vs)
```
benfred, line 50: can we use the existing raft logging instead? (RAFT_LOG_DEBUG etc)

benfred (commented):
This looks great! I'm excited to get this functionality in here.

Can we also publish the output of these benchmarks somewhere in the source tree, as well as instructions on how to run yourself and get the datasets?

I've left some minor comments below.

cpp/bench/ann/src/raft/raft_ivf_pq.cu
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "raft_ivf_pq_wrapper.h"
+
+namespace raft::bench::ann {
+template class RaftIvfPQ<float, uint64_t>;
```
tfeher, line 19: When we are switching to int8 spcializations, then we should not forget to change these to to int8.

cpp/bench/ann/src/common/conf.h
```diff
@@ -0,0 +1,72 @@
...
+    int run_count;
+    std::vector<nlohmann::json> search_params;
+    std::string search_result_file;
+    float refine_ratio{0.0f};
+  };
+
+  struct DatasetConf {
+    std::string name;
+    std::string base_file;
+    // use only a subset of base_file,
```
tfeher, line 47: I recommend to add another config parameter to describe data type. This can be deduced from the file name.
```suggestion
    data type of datastet, possible values ["float", "int8", "uint8"]
    std::string dtype;
    // use only a subset of base_file,
```

cpp/bench/ann/src/common/conf.cpp
```diff
@@ -0,0 +1,136 @@
...
+  dataset_conf_.name       = conf.at("name");
+  dataset_conf_.base_file  = conf.at("base_file");
+  dataset_conf_.query_file = conf.at("query_file");
+  dataset_conf_.distance   = conf.at("distance");
+
+  if (conf.contains("subset_first_row")) {
+    dataset_conf_.subset_first_row = conf.at("subset_first_row");
+  }
+  if (conf.contains("subset_size")) { dataset_conf_.subset_size = conf.at("subset_size"); }
+}
```
tfeher, line 69:
```suggestion
  if (conf.contains("dtype")) {
    dataset_conf_.dtype = conf.at("dtype");
  }
  else {
    auto filename = dataset_conf_.base_file;
    if (!filename.compare(filename.size() - 4, 4, "fbin")) {
      dataset_conf_.dtype = "float";
    }
    else if (!filename.compare(filename.size() - 5, 5, "u8bin")) {
      dataset_conf_.dtype = "uint8";
    }
    else if (!filename.compare(filename.size() - 5, 5, "i8bin")) {
      dataset_conf_.dtype = "int8";
    }
    else {
     THROW("Could not determine data type of the dataset");
    }
}
```

tfeher (commented):
Just a few quick notes

cpp/bench/ann/conf/sift-128-euclidean.json
```diff
@@ -0,0 +1,1321 @@
+{
```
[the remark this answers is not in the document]
cjnolet: By `configs` are you referring to parameter sets? The more I've been working through integrating these benchmarks, the more I'm realizing we should probably be specifying the parameters as ranges/lists, rather than individual hard-coded configs everywhere so that we can actually explore the qps/recall space and learn which parameters actually performed the best on which datasets. If we only test a very small set of parameters, we're not going to have all that great a representation of which ones worked the best. Even worse, it doesn't really give us a great reference point for comparison against other algorithms. 

The typical way I've seen to compare these algorithms is by allowing each algorithm to specify a range of reasonable parameter settings and then sweeping through all of them, which will yield a curve- qps drops as recall increases, and pulling off the pareto frontier. While looking around the [big-ann-benchmarks](https://github.com/harsha-simhadri/big-ann-benchmarks) documentation, I also noticed someone has created a repository that integrated the scripts almost directly from Erik Bernardsen's [ann-benchmarks](https://github.com/erikbern/ann-benchmarks/). I think we should consider doing the same (using their run/plot scripts, config infrastructure, etc...). 

Of course, this then begs the question- since all of the libraries we are comparing against have python apis, including our own, should we instead focus our efforts on benchmarking at the Python API? It would still allow us to make changes to the C++ code and instantly benchmark the results.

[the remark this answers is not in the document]
tfeher: - I am mostly referring to index building parameters. Index construction for 100M+ scale datasets can be long, so we probably want to limit to the most relevant configuration.
- For search parameters, I agree with you. Currently a parameter set sweeps trough 1 parameter (typically n_probbes). Having a way to sweep through a combination of parameters them would be nice.

> should we instead focus our efforts on benchmarking at the level of the Python API?

For small batch size the Python wrappers can have non negligible overhead. Therefore, I think there is a value in having benchmarks both for the C++ and Python APIs. But they do not have to provide the same level of convenience for configuring the test and reporting the results.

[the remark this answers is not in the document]
cjnolet: > For small batch size the Python wrappers can have non negligible overhead. 

Yep, I agree- we should keep the c++ tests then. Benchmarking Dr the c++ layer also makes it much easier to benchmark the algorithms in the c++ layer. This is what Milvus does, for example. 

Maybe what I'll do is try to replicate a little bit more of the user experience from the Ann-benchmarks Python scripts, but only implement the options that make sense (for example, I don't think we need to worry about running docker containers, but it would be great it we could maybe specify the parameter ranges in a similar way.

It would be nice to use the plotting scripts too, and potentially the runner scripts. It could also be pretty easy to replicate their h5d file layout.

While all of the aforementioned things would be ideal the most ideal, I think we could probably merge this initial version after everyone's happy with it and then continue to iterate it on improving over the next couple releases. What do you think?

cjnolet:
Thanks for the reviews, guys! I've opened https://github.com/rapidsai/raft/issues/1367 to follow-up with some of the larger items uncovered in this PR so that we can get this in for 23.04. Milvus folks have expressed interest in using this. 

@tfeher, I'll make the necessary changes to specify the data types in the configs in this PR, as I believe those should be a priority- manually editing source files to change the data types is not ideal at all.

cpp/bench/ann/scripts/split_groundtruth.pl
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/perl
```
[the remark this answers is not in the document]
cjnolet: Created https://github.com/rapidsai/raft/issues/1367 to capture this issue. I'd ultimately like to use the Python scripts from ann-benchmarks (which have also been ported to big-ann-benchmarks).

cpp/bench/ann/src/common/dataset.h
cjnolet: Created https://github.com/rapidsai/raft/issues/1367 to do this. I'm just a little wary of doing it here because I don't want to break things. I'd like to do it in the follow-ups

cpp/bench/ann/src/faiss/faiss_wrapper.h
```diff
@@ -0,0 +1,317 @@
...
+  faiss::MetricType metric_type_;
+  int nlist_;
+  int device_;
+};
+
+template <typename T>
+FaissGpu<T>::FaissGpu(Metric metric, int dim, int nlist)
+  : ANN<T>(metric, dim), metric_type_(parse_metric_type(metric)), nlist_(nlist)
+{
+  static_assert(std::is_same_v<T, float>, "faiss support only float type");
```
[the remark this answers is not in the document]
cjnolet: That's a great point. I added it to the follow-ups, but i also think we shold start benchmarking against the fp16 as well. It's more fair.

cpp/bench/ann/src/raft/raft_ivf_pq.cu
cjnolet: I'm wondering if maybe we should just build all supported templates up front? If we're using the pre-compiled specializations properly then it shouldn't really add much time (or space) to the build.

benfred:
> Does [this file](https://github.com/rapidsai/raft/blob/ff724f3c8bb6595c6626b105c3edaeecd5094ab0/docs/source/cuda_ann_benchmarks.md), which we're adding to the docs, help alleviate this concern? Or are you wanting us to either duplicate or point right to to the table of datasets in the `big-ann-benchmarks`?

yup, that looks good! I must have missed it there

benfred (approved):
looks good!

ahendriksen:
Was it intended behavior that the benchmarks are not built anymore when you build with `./build.sh bench-prims`? This currently sets `-DBUILD_PRIMS_BENCH`, but in the bench CmakeLists.txt, it checks for `if(BUILD_BENCH)`.

ahendriksen:
In any case, I have attempted a fix in https://github.com/rapidsai/raft/pull/1389.

PR #1307, opened 2023-02-28
Reduce compile times of distance specializations 

ahendriksen:
Following the findings in https://github.com/ahendriksen/raft/tree/investigate-compile-time-reduction-strategies#investigation-of-compile-times, this PR reduces the compile times of the pairwise distance specializations.
This is achieved by:
1. Reducing the number of included files in the translation units where kernels are instantiated, specifically `spdlog` and `rmm` are avoided. 
2. Limiting loop unrolling in kernels with expensive operations in the inner loop.

Additional improvements geared towards iterative development:
1. The tests do not have to be recompiled when the internals of a pairwise distance kernel change. Before, a rebuilt was triggered due an include of `raft/distance/distance.cuh`.
2. Addition of a  fine tuning benchmark for the pairwise distance kernels that separates building the kernel from the benchmark code. This dramatically speeds up development. Compiling an empty benchmark takes roughly 18 seconds on my machine. Whereas recompiling a kernel takes ~3.8 seconds. Without this addition, a commit like 35a2ad437 would require substantially more time to make sure that performance is not degraded. 

image

```
Parallel build time before: 270 seconds (6 cores, SMT, 12 jobs)
Parallel build time before: 147 seconds (6 cores, SMT, 12 jobs)

Sum of compile times before: 3022.6 seconds
Sum of compile times after:  1816.2 seconds

Comparison of compile times between headers and compiled: 
path                                                                         before (s)     after (s)  change (s) change (%)
pairwise_test                                                                    None        0.486       None     None
ance/distance/specializations/detail/lp_unexpanded_double_double_double_int.cu.o  101.1          10.3       -90.8     -89.8%
src/distance/distance/specializations/detail/canberra_float_float_float_int.cu.o   52.9           6.3       -46.6     -88.0%
/distance/distance/specializations/detail/canberra_double_double_double_int.cu.o   48.5           6.4       -42.1     -86.8%
stance/distance/specializations/detail/jensen_shannon_float_float_float_int.cu.o   65.3          10.4       -55.0     -84.1%
istance/distance/specializations/detail/kl_divergence_float_float_float_int.cu.o   70.2          12.6       -57.6     -82.0%
stance/distance/specializations/detail/correlation_double_double_double_int.cu.o   46.7           8.9       -37.8     -80.9%
distance/specializations/detail/hellinger_expanded_double_double_double_int.cu.o   41.6           8.1       -33.5     -80.6%
nce/distance/specializations/detail/jensen_shannon_double_double_double_int.cu.o   74.6          15.1       -59.5     -79.7%
ir/src/distance/distance/specializations/detail/l1_double_double_double_int.cu.o   40.9           8.4       -32.5     -79.4%
ance/distance/specializations/detail/l2_unexpanded_double_double_double_int.cu.o   40.7           8.6       -32.1     -78.8%
distance/specializations/detail/hamming_unexpanded_double_double_double_int.cu.o   40.8           9.0       -31.7     -77.8%
istance/distance/specializations/detail/lp_unexpanded_float_float_float_int.cu.o   45.9          10.2       -35.7     -77.8%
src/distance/distance/specializations/detail/l_inf_double_double_double_int.cu.o   41.2           9.5       -31.8     -77.0%
istance/distance/specializations/detail/russel_rao_double_double_double_int.cu.o   29.5           7.2       -22.3     -75.6%
t.dir/src/distance/distance/specializations/detail/l1_float_float_float_int.cu.o   47.3          13.2       -34.1     -72.2%
ce/distance/specializations/detail/hamming_unexpanded_float_float_float_int.cu.o   47.0          13.3       -33.7     -71.6%
/distance/distance/specializations/detail/correlation_float_float_float_int.cu.o   49.4          14.1       -35.3     -71.5%
ce/distance/specializations/detail/hellinger_expanded_float_float_float_int.cu.o   43.6          12.5       -31.1     -71.4%
c/distance/distance/specializations/detail/russel_rao_float_float_float_int.cu.o   28.5           8.2       -20.3     -71.2%
ance/distance/specializations/detail/kl_divergence_double_double_double_int.cu.o   75.8          21.9       -53.9     -71.1%
istance/distance/specializations/detail/l2_unexpanded_float_float_float_int.cu.o   46.2          13.5       -32.7     -70.7%
ir/src/distance/distance/specializations/detail/l_inf_float_float_float_int.cu.o   43.1          12.7       -30.4     -70.6%
stance/distance/specializations/detail/l2_expanded_double_double_double_int.cu.o   52.3          24.9       -27.3     -52.3%
/distance/distance/specializations/detail/l2_expanded_float_float_float_int.cu.o   75.8          40.3       -35.5     -46.8%
rc/distance/distance/specializations/detail/cosine_double_double_double_int.cu.o   53.5          28.7       -24.8     -46.4%
r/src/distance/distance/specializations/detail/cosine_float_float_float_int.cu.o   83.9          50.1       -33.8     -40.3%
CMakeFiles/pairwise_test.dir/test/distance/fused_l2_nn.cu.o                        85.1          64.1       -21.1     -24.7%
wise_test.dir/src/distance/distance/specializations/fused_l2_nn_float_int64.cu.o   56.2          42.9       -13.3     -23.6%
irwise_test.dir/src/distance/distance/specializations/fused_l2_nn_float_int.cu.o   52.5          40.2       -12.3     -23.5%
CMakeFiles/pairwise_test.dir/test/distance/dist_lp_unexp.cu.o                      56.3          43.3       -13.0     -23.1%
CMakeFiles/pairwise_test.dir/test/distance/dist_russell_rao.cu.o                   55.7          44.0       -11.7     -21.0%
rwise_test.dir/src/distance/distance/specializations/fused_l2_nn_double_int.cu.o   45.3          36.4        -9.0     -19.8%
CMakeFiles/pairwise_test.dir/test/distance/dist_l2_unexp.cu.o                      54.6          44.1       -10.6     -19.3%
CMakeFiles/pairwise_test.dir/test/distance/dist_canberra.cu.o                      51.6          42.1        -9.6     -18.6%
CMakeFiles/pairwise_test.dir/test/distance/dist_l2_exp.cu.o                        53.1          43.4        -9.6     -18.2%
CMakeFiles/pairwise_test.dir/test/distance/dist_l_inf.cu.o                         53.2          43.9        -9.3     -17.5%
CMakeFiles/pairwise_test.dir/test/distance/dist_hellinger.cu.o                     53.1          44.0        -9.0     -17.0%
CMakeFiles/pairwise_test.dir/test/distance/dist_hamming.cu.o                       52.3          43.4        -8.9     -17.0%
CMakeFiles/pairwise_test.dir/test/distance/dist_l2_sqrt_exp.cu.o                   54.0          45.6        -8.4     -15.6%
CMakeFiles/pairwise_test.dir/test/distance/dist_l1.cu.o                            52.6          44.5        -8.1     -15.4%
CMakeFiles/pairwise_test.dir/test/distance/dist_kl_divergence.cu.o                 52.4          44.7        -7.7     -14.8%
ise_test.dir/src/distance/distance/specializations/fused_l2_nn_double_int64.cu.o   43.5          37.2        -6.4     -14.7%
CMakeFiles/pairwise_test.dir/test/distance/dist_cos.cu.o                           52.4          44.8        -7.6     -14.5%
CMakeFiles/pairwise_test.dir/test/distance/dist_jensen_shannon.cu.o                53.2          45.7        -7.6     -14.2%
CMakeFiles/pairwise_test.dir/test/distance/dist_inner_product.cu.o                 51.1          44.8        -6.3     -12.4%
istance/distance/specializations/detail/inner_product_float_float_float_int.cu.o   39.5          35.1        -4.5     -11.3%
CMakeFiles/pairwise_test.dir/test/distance/dist_correlation.cu.o                   51.7          46.8        -4.9     -9.5%
ance/distance/specializations/detail/inner_product_double_double_double_int.cu.o   37.1          33.9        -3.1     -8.5%
src/distance/distance/specializations/detail/kernels/gram_matrix_base_float.cu.o   45.3          41.7        -3.6     -8.0%
rc/distance/distance/specializations/detail/kernels/gram_matrix_base_double.cu.o   42.5          39.6        -2.9     -6.8%
stance/distance/specializations/detail/kernels/polynomial_kernel_double_int.cu.o   40.4          38.5        -1.9     -4.8%
CMakeFiles/pairwise_test.dir/test/distance/dist_adj.cu.o                          123.3         117.8        -5.4     -4.4%
CMakeFiles/pairwise_test.dir/test/distance/gram.cu.o                               55.3          53.4        -1.9     -3.5%
build.ninja                                                                         4.0           4.0        +0.0     +0.1%
istance/distance/specializations/detail/kernels/polynomial_kernel_float_int.cu.o   45.2          45.6        +0.4     +0.8%
.dir/src/distance/distance/specializations/detail/kernels/tanh_kernel_float.cu.o   45.2          46.0        +0.8     +1.7%
dir/src/distance/distance/specializations/detail/kernels/tanh_kernel_double.cu.o   39.0          39.8        +0.8     +2.1%
CMakeFiles/pairwise_test.dir/src/distance/distance/pairwise_distance.cu.o          39.6          50.1       +10.5     +26.6%
```


ahendriksen:
This commit is a follow up to #1295.

Gabriel-Gotti (approved):

ahendriksen:
This PR is a follow up to #1335.

cjnolet:
Linking this change so it gets done as part of this PR: https://github.com/rapidsai/raft/pull/1335#pullrequestreview-1342454117

cpp/include/raft/core/kvp.hpp
```diff
@@ -20,7 +20,7 @@
 
 #ifdef _RAFT_HAS_CUDA
 #include <cub/cub.cuh>
-#include <raft/util/cuda_utils.cuh>
+#include <raft/util/cuda_utils.cuh>  // raft::shfl_xor
```
divyegala, line 23: I believe RAFT's developer guide makes no recommendations to have comments for headers that are being included, and as such I'd like to avoid them completely since the rest of the codebase doesn't use them

cpp/include/raft/distance/detail/distance_ops/cosine.cuh
```diff
@@ -16,19 +16,34 @@
...
 
 namespace raft::distance::detail::ops {
 
+// Epilogue operator for CUTLASS based kernel
+template <typename DataT, typename AccT>
+struct cosine_cutlass_op {
+  __device__ cosine_cutlass_op() noexcept {}
+  __device__ AccT operator()(DataT& aNorm, const DataT& bNorm, DataT& accVal) const noexcept
+  {
+    return static_cast<AccT>(1.0) - (AccT)(accVal / (aNorm * bNorm));
```
divyegala, line 29: There's two different cast styles in the same line here
```suggestion
    return static_cast<AccT>(1.0 - accVal / (aNorm * bNorm));
```

cpp/include/raft/distance/detail/distance_ops/cosine.cuh
```diff
@@ -60,16 +75,8 @@ struct cosine_distance_op {
...
 
-template <typename DataT, typename AccT>
-struct cosine_cutlass_op {
-  __device__ cosine_cutlass_op() noexcept {}
-  __device__ AccT operator()(DataT& aNorm, const DataT& bNorm, DataT& accVal) const noexcept
-  {
-    return static_cast<AccT>(1.0) - (AccT)(accVal / (aNorm * bNorm));
-  }
-  __device__ AccT operator()(DataT aData) const noexcept { return aData; }
+  cosine_cutlass_op<DataT, AccT> get_cutlass_op() { return cosine_cutlass_op<DataT, AccT>(); }
```
divyegala, line 79:
```suggestion
  constexpr cosine_cutlass_op<DataT, AccT> get_cutlass_op() { return cosine_cutlass_op<DataT, AccT>(); }
```

cpp/include/raft/distance/detail/distance_ops/cutlass.cuh
```diff
@@ -0,0 +1,40 @@
...
+namespace raft::distance::detail::ops {
+
+// This file defines the named requirement "has_cutlass_op" that can be used to
+// determine if a distance operation has a CUTLASS op that can be used to pass
+// to CUTLASS. Examples of distance operations that satisfy this requirement are
+// cosine_distance_op and l2_exp_distance_op.
+
+// Primary template handles types that do not support CUTLASS.
+// This pattern is described in:
+// https://en.cppreference.com/w/cpp/types/void_t
```
divyegala, line 30: I like the idea of describing a specific method from the multitude of template metaprogramming methods that one can use, where relevant. Maybe we should do this more often.

cpp/include/raft/distance/detail/distance_ops/cutlass.cuh
```diff
@@ -0,0 +1,40 @@
...
+template <typename, typename = void>
+struct has_cutlass_op : std::false_type {
+};
+
+// Specialization recognizes types that do support CUTLASS
+template <typename T>
+struct has_cutlass_op<T, std::void_t<decltype(&T::get_cutlass_op)>> : std::true_type {
```
divyegala, line 37: What's the necessity for the ampersand in `&T::get_cutlass_op`?

cpp/include/raft/distance/detail/distance_ops/l2_exp.cuh
```diff
@@ -75,26 +101,8 @@ struct l2_exp_distance_op {
...
-    AccT outVal = aNorm + bNorm - DataT(2.0) * accVal;
-    // outVal could be negative due to numerical instability, especially when
-    // calculating self distance.
-    // clamp to 0 to avoid potential NaN in sqrt
-    outVal = outVal * (outVal > DataT(0.0));
-    return sqrt ? raft::sqrt(outVal) : outVal;
-  }
 
-  __device__ AccT operator()(DataT aData) const noexcept { return aData; }
+  l2_exp_cutlass_op<DataT, AccT> get_cutlass_op() { return l2_exp_cutlass_op<DataT, AccT>(sqrt); }
```
divyegala, line 105:
```suggestion
  constexpr l2_exp_cutlass_op<DataT, AccT> get_cutlass_op() { return l2_exp_cutlass_op<DataT, AccT>(sqrt); }
```

cpp/include/raft/distance/detail/pairwise_distance_cutlass_base.cuh
```diff
@@ -59,26 +62,29 @@ template <typename DataT,
...
-                           IdxT n,
-                           IdxT k,
-                           IdxT lda,
-                           IdxT ldb,
-                           IdxT ldd,
-                           OutT* dOutput,
-                           FinalLambda fin_op,
-                           DistanceFn dist_op,
-                           cudaStream_t stream)
+typename std::enable_if<ops::has_cutlass_op<OpT>::value>::type cutlassDistanceKernel(
```
divyegala, line 67:
```suggestion
std::enable_if_t<ops::has_cutlass_op<OpT>::value> cutlassDistanceKernel(
```

cpp/include/raft/distance/detail/pairwise_distance_cutlass_base.cuh
```diff
@@ -59,26 +62,29 @@ template <typename DataT,
...
+  IdxT ldd,
+  OutT* dOutput,
+  FinalLambda fin_op,
+  OpT distance_op,
+  cudaStream_t stream)
 {
   static_assert(!(std::is_same<OutT, bool>::value),
                 "OutType bool is not supported use uint8_t instead");
 
+  auto dist_op     = distance_op.get_cutlass_op();
```
divyegala, line 86:
```suggestion
  auto constexpr dist_op     = distance_op.get_cutlass_op();
```

cpp/include/raft/distance/detail/pairwise_matrix/dispatch_layout.cuh
```diff
@@ -0,0 +1,116 @@
...
+ *                   row_major and vec_len.
+ */
+template <typename F>
+auto dispatch_layout(bool row_major, int vec_len, F&& f)
+{
+  if (row_major) {
+    switch (vec_len) {
+      case 4: return f(std::bool_constant<true>(), vec_len_constant<4>());
+      case 2: return f(std::bool_constant<true>(), vec_len_constant<2>());
+      default: return f(std::bool_constant<true>(), vec_len_constant<1>());
```
divyegala, line 105:
```suggestion
      case 4: return f(std::bool_constant<true>(), vec_len_constant<4>());
      case 2: return f(std::bool_constant<true>(), vec_len_constant<2>());
      default: return f(std::bool_constant<true>(), vec_len_constant<1>());
```
```suggestion
      case 4: return f(std::true_type(), vec_len_constant<4>());
      case 2: return f(std::true_type(), vec_len_constant<2>());
      default: return f(std::true_type(), vec_len_constant<1>());
```

cpp/include/raft/distance/detail/pairwise_matrix/dispatch_layout.cuh
```diff
@@ -0,0 +1,116 @@
...
+    }
+  } else {
+    switch (vec_len) {
+      case 4: return f(std::bool_constant<false>(), vec_len_constant<4>());
+      case 2: return f(std::bool_constant<false>(), vec_len_constant<2>());
+      default: return f(std::bool_constant<false>(), vec_len_constant<1>());
```
divyegala, line 111:
```suggestion
      case 4: return f(std::false_type(), vec_len_constant<4>());
      case 2: return f(std::false_type(), vec_len_constant<2>());
      default: return f(std::false_type(), vec_len_constant<1>());
```

cpp/include/raft/distance/detail/pairwise_matrix/dispatch_sm60.cuh
```diff
@@ -0,0 +1,76 @@
...
+          typename FinOpT,
+          typename SM_compat_t>
+pairwise_matrix_sm60_wrapper<OpT, IdxT, DataT, OutT, FinOpT> pairwise_matrix_sm60_get_wrapper(
+  OpT distance_op,
+  pairwise_matrix_params<IdxT, DataT, OutT, FinOpT> params,
+  SM_compat_t sm_compat_range)
+{
+  int vec_len = determine_vec_len(params);
+
+  return dispatch_layout(params.is_row_major, vec_len, [&](auto row_major, auto vec_len_aligned) {
```
divyegala, line 38: Would you consider declaring the lambda before the return line? I find it a bit confusing to read it like this, when there's code extending after a return statement

cpp/include/raft/distance/detail/pairwise_matrix/dispatch_sm60.cuh
```diff
@@ -0,0 +1,76 @@
...
+    // To keep compile times in check, we only specialize on veclen > 1 when
+    // the inner loop is relatively cheap (< 5 flops).
+    constexpr int vec_len_op = distance_op.expensive_inner_loop ? 1 : vec_len_aligned();
+
+    // Prevent double, vec_len=4 combination (this is not supported)
+    constexpr int vec_len = std::min(vec_len_op, static_cast<int>(16 / sizeof(DataT)));
+
+    typedef typename raft::linalg::Policy4x4<DataT, vec_len>::Policy RowPolicy;
+    typedef typename raft::linalg::Policy4x4<DataT, vec_len>::ColPolicy ColPolicy;
+    typedef typename std::conditional<row_major(), RowPolicy, ColPolicy>::type Policy;
```
divyegala, line 51: This PR has generally used using instead of typedef, can we keep it the former?

cpp/include/raft/distance/detail/pairwise_matrix/dispatch_sm80.cuh
```diff
@@ -0,0 +1,62 @@
...
+          typename FinOpT,
+          typename SM_compat_t>
+void pairwise_matrix_sm80_dispatch(OpT distance_op,
+                                   pairwise_matrix_params<IdxT, DataT, OutT, FinOpT> params,
+                                   SM_compat_t sm_compat_range,
+                                   cudaStream_t stream)
+{
+  int vec_len = determine_vec_len(params);
+
+  dispatch_layout(params.is_row_major, vec_len, [&](auto row_major, auto vec_len_aligned) {
```
divyegala, line 37: Same suggestion on declaring the lambda before the caller site

cpp/include/raft/distance/specializations/detail/00_write_template.py
divyegala, line 1: This is fantastic

cpp/include/raft/distance/specializations/detail/correlation.cuh
```diff
@@ -18,36 +18,25 @@
...
-  int k,
-  void* workspace,
-  size_t worksize,
-  bool isRowMajor,
-  float metric_arg);
+  pairwise_matrix_instantiation_point<ops::correlation_distance_op<float, float, int>,
+                                      int,
+                                      float,
+                                      float,
+                                      decltype(raft::identity_op())>(
```
divyegala, line 28: Iff you want the type, can't it be used directly? If it can then this will apply in all other places
```suggestion
                                      raft::identity_op>(
```

cpp/test/distance/distance_base.cuh
```diff
@@ -523,9 +560,25 @@ class BigMatrixDistanceTest : public ::testing::Test {
...
+                           float* y,
+                           float* dists,
+                           int m,
+                           int n,
+                           int k,
+                           raft::distance::DistanceType metric,
+                           bool isRowMajor,
+                           float metric_arg);
+    bool row_major   = true;
+    float metric_arg = 0.0f;
```
divyegala, line 574:
```suggestion
    bool constexpr row_major   = true;
    float constexpr metric_arg = 0.0f;
```

cpp/include/raft/core/kvp.hpp
ahendriksen: I really needed these comments while developing this PR though. It is so easy to leave an unneeded header on a file. Even though I was actively trying to reduce the number of includes, I constantly found unneeded includes that I had inadvertently left in. With these comments, you can at least quickly search through the file if for instance `raft::shfl_xor` is still used. If not, it might be worth trying to remove the include. If it is, the check was quick and easy.

Would it be okay to go with the following policy: 
- leave the comments when they are already present
- update them when they are present
- not require them on new code/code that doesn't have these comments yet

I agree that consistency should be valued, but I also think that removing these comments in the entire distance/detail tree would really make it hard to keep the includes minimal (especially during review).

cpp/include/raft/distance/detail/distance_ops/cutlass.cuh
ahendriksen: Thanks! When I have to look something up to write the code, I try to make it a habit to to link a reference. It makes no sense to expect the reader to have a background in C++ that I didn't have before I wrote the code.

cpp/include/raft/distance/detail/distance_ops/cutlass.cuh
ahendriksen: Not necessary. I removed it :+1:

ahendriksen: It is actually necessary. Without the ampersand, you get the error `invalid use of non-static member function` ([godbolt](https://godbolt.org/z/59YTeYse8)). Not entirely sure why.

cpp/include/raft/distance/detail/pairwise_distance_cutlass_base.cuh
ahendriksen: This will not work: `distance_op` is a function parameter. In the case of L2 expanded, it is *not* a zero-sized struct (as it contains the runtime boolean `sqrt`). Therefore `distance_op` is not constexpr and the code will not compile. 

At first, this change did compile for me. The reason was that I had removed the ampersand in `struct has_cutlass_op<T, std::void_t<decltype(&T::get_cutlass_op)>> : std::true_type {};`. As a result, the compilation of the cutlass kernels was skipped. 

When I added back the ampersand, the cutlass kernels were compiled again and this change caused a compiler error.

cpp/include/raft/distance/detail/pairwise_matrix/dispatch_layout.cuh
ahendriksen: Thanks! That is indeed shorter.

cpp/include/raft/distance/detail/pairwise_distance_cutlass_base.cuh
ahendriksen: Great tip! Thanks! This is a lot less ugly.

cpp/include/raft/distance/detail/distance_ops/cutlass.cuh
divyegala: Ah I see, because `T` is a class and `get_cutlass_op` is a method. This is the correct way to add a temporary reference and check if a member function exists:
`std::declval<T>().get_cutlass_op()`

ahendriksen: Thanks! I will update the code.

ahendriksen:
> Linking this change so it gets done as part of this PR: [#1335 (review)](https://github.com/rapidsai/raft/pull/1335#pullrequestreview-1342454117)

This has been done.

cjnolet:
sorry @ahendriksen, we had a high priority to fix the RAFT packaging for 23.04 (in time for GTC keynote) but this caused a bunch of conflicts in this PR. They should be straightforward to fix, though- the `libraft` source files directory changed from `src/distance/.../` to `/src/.../` so most of it is just renaming.

cjnolet (approved):

cpp/include/raft/distance/detail/distance_ops/l2_exp.cuh
```diff
@@ -102,7 +102,10 @@ struct l2_exp_distance_op {
     }
   }
 
-  l2_exp_cutlass_op<DataT, AccT> get_cutlass_op() { return l2_exp_cutlass_op<DataT, AccT>(sqrt); }
+  constexpr l2_exp_cutlass_op<DataT, AccT> get_cutlass_op()
+  {
+    return l2_exp_cutlass_op<DataT, AccT>(sqrt);
```
divyegala, line 107: Is this path being tested? Where's the `sqrt` coming from?

cpp/include/raft/distance/detail/distance_ops/l2_exp.cuh
```diff
@@ -102,7 +102,10 @@ struct l2_exp_distance_op {
     }
   }
 
-  l2_exp_cutlass_op<DataT, AccT> get_cutlass_op() { return l2_exp_cutlass_op<DataT, AccT>(sqrt); }
+  constexpr l2_exp_cutlass_op<DataT, AccT> get_cutlass_op()
```
divyegala, line 105: On further thoughts, since this acts as a "create" function that doesn't rely on state of the object - it should be `static constexpr`. That will probably convert my question below to a compiler error.

cpp/include/raft/distance/detail/distance_ops/l2_exp.cuh
ahendriksen: Short story:
- This path is tested in `cpp/test/distance/dist_l2_sqrt_exp.cu`
- the sqrt is coming from the l2_expanded_distance op. It is an instance variable.

Long story:
The `DistanceType::L2SqrtExpanded` and `DistanceType::L2Expanded` distance types both map to an l2_exp_distance_op. The l2 expanded distance operation has an instance boolean variable `sqrt` which determines whether to calculate the square root. The distance op (and its member variable `sqrt`) gets passed to the kernel at runtime and the kernel decides at runtime whether to compute the square root or not.

cpp/include/raft/distance/detail/distance_ops/l2_exp.cuh
ahendriksen: It does rely on the state of the object: the `sqrt` comes from the object. Is there a way that I can document this better?

divyegala: Hmmm I understand it now. No, it's fine as is - my confusion came because I just looked at the diff but looking at the whole file is enough context. Thank you!

ahendriksen:
> sorry @ahendriksen, we had a high priority to fix the RAFT packaging for 23.04 (in time for GTC keynote) but this caused a bunch of conflicts in this PR. They should be straightforward to fix, though- the `libraft` source files directory changed from `src/distance/.../` to `/src/.../` so most of it is just renaming.

No problem! The merge was quite painless actually. Really looking forward to the GTC keynote and all the great progress we will make going forward!

cpp/include/raft/distance/detail/fused_l2_nn.cuh
```diff
@@ -300,59 +275,25 @@ void fusedL2NNImpl(OutT* min,
...
-                                                redOp,
-                                                pairRedOp,
-                                                core_lambda,
-                                                raft::identity_op{});
-  }
 
+  using AccT = DataT;
+  ops::l2_exp_distance_op<DataT, AccT, IdxT> distance_op{sqrt};
+
+  raft::identity_op fin_op{};
```
divyegala, line 282: This can be constexpr

cpp/include/raft/distance/detail/pairwise_matrix/kernel_sm60.cuh
```diff
@@ -36,43 +36,27 @@ __global__ __launch_bounds__(Policy::Nthreads, 2) void pairwise_matrix_kernel(
...
-                                            IdxT gridStrideX,
-                                            IdxT gridStrideY) {
-    // Use .template to disambiguate (See:
-    // https://en.cppreference.com/w/cpp/language/dependent_name)
-    distance_op.template epilog<Policy>(acc, regxn, regyn, gridStrideX, gridStrideY);
-  };
-
+  // The epilog is already provided by distance_op. Do not provide additional
+  // epilogs.
+  auto epilog_op = raft::void_op();
```
divyegala, line 48: Same as above

cpp/include/raft/distance/detail/fused_l2_nn.cuh
ahendriksen: I agree. I don't expect it to make a difference on the final binary though. Let's merge this PR first.

ahendriksen:
I think we have to wait for #1363 to clear so that the CI passes for this run.

cjnolet:
@ahendriksen yep, and CI hasn't been particularly friendly lately.... Jobs have been getting queued up for several hours at a time. If it helps, you can merge that branch into this one, then at least this one will be up to date and pass through CI so we can merge it right after.

PR #1309, opened 2023-02-28
Update to GCC 11

bdice:
This PR updates builds to use GCC 11.

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -59,12 +59,12 @@ dependencies:
           - matrix:
               arch: x86_64
             packages:
-              - gcc_linux-64=9
+              - gcc_linux-64=11.*
               - sysroot_linux-64==2.17
           - matrix:
               arch: aarch64
             packages:
-              - gcc_linux-aarch64=9
+              - gcc_linux-aarch64=11.*
               - sysroot_linux-aarch64==2.17
   checks:
     common:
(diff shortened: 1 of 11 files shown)


bdice:
This is ready for review, but please **_do not merge_** until all RAPIDS repos are ready to migrate. (I do not have privileges to add the DO NOT MERGE label to prohibit merging.)

jjacobelli (approved):

dependencies.yaml
```diff
@@ -59,12 +59,12 @@ dependencies:
...
               arch: x86_64
             packages:
-              - gcc_linux-64=9
+              - gcc_linux-64=11.*
               - sysroot_linux-64==2.17
           - matrix:
               arch: aarch64
             packages:
-              - gcc_linux-aarch64=9
+              - gcc_linux-aarch64=11.*
```
jjacobelli, line 67: Approved, but I'm curious, what's the difference between `=11` and = `11.*` here?

bdice: I'm aligning with the way that we specified this in other packages (cudf, cuspatial, cusignal, ...). I'm not sure which to prefer, I think both may be equivalent.

https://github.com/rapidsai/cudf/blob/b6d9fd1f26ccda9b3043a8967621bae234187c37/dependencies.yaml#L80

Ref: https://docs.conda.io/projects/conda-build/en/latest/resources/package-spec.html#examples

cjnolet (approved):
LGTM

Issue #1299, opened 2023-02-23
[FEA] mdspan/mdarray-based utility to copy a pointer to device iff it is not already on device


cjnolet:
This is a very useful utility in FAISS and something that we should support in raft so that we can have a unified means of handling both device and host inputs (and copying to temporary device buffers only if needed). 

The challenge here is that what we really seem to want is some type of object that can act like a view all the time, but which also owns the underlying memory only as needed. This is not something our current mdspan/mdarray infrastructure can support, so we will need to build something for this case. 

@mhoemmen has suggested we use something like the [kokkos::DualView](https://github.com/kokkos/kokkos/wiki/Kokkos::DualView) which can mirror device and host views with container(s) underneath to copy between temporary buffers as needed, keeping them up to date between host and device. I think this could work in our case, but as Mark points out, we would want to make it read-only on the host side so that the device side isn't copied back to the host.

mhoemmen:
@cjnolet wrote:
> mhoemmen has suggested we use something like the Kokkos::DualView....

Please please do not get the impression that I suggested using Kokkos::DualView.  DualView does not give a good example of how to solve this problem.

I'll discuss this offline with you; I've given talks about this and have slides ready to share.

PR #1313, opened 2023-03-01
Temporary buffer to view host or device memory in device

divyegala:
closes https://github.com/rapidsai/raft/issues/1299


cpp/test/core/mdarray.cu
```diff
@@ -947,4 +951,60 @@ void test_mdarray_unravel()
 
 TEST(MDArray, Unravel) { test_mdarray_unravel(); }
 
+TEST(DeviceBuffer, DevicePointer)
```
cjnolet, line 954: Just a small reminder to move these into their own test file

cpp/include/raft/core/device_buffer.hpp
```diff
@@ -0,0 +1,108 @@
...
+  using data_store           = std::variant<ElementType*, owning_device_buffer>;
+
+ public:
+  device_buffer(device_buffer const&) = delete;
+  device_buffer& operator=(device_buffer const&) = delete;
+
+  constexpr device_buffer(device_buffer&&) = default;
+  constexpr device_buffer& operator=(device_buffer&&) = default;
+
+  device_buffer(device_resources const& handle,
```
cjnolet, line 46: To keep this consistent with our other APIs, can we add a factory function to create this? Something like `make_device_buffer` or `make_readonly_device_buffer` and `make_writeback_device_buffer` (or  have the latter 2 call the former?)

cjnolet (changes requested):
Looks great so far. Just a couple small notes

cpp/test/core/temporary_device_buffer.cu
```diff
@@ -0,0 +1,87 @@
...
+TEST(TemporaryDeviceBuffer, HostPointerWithWriteBack)
+{
+  raft::device_resources handle;
+  auto exts  = raft::make_extents<int>(5);
+  auto array = raft::make_host_mdarray<int, int>(exts);
+  thrust::fill(array.data_handle(), array.data_handle() + array.extent(0), 1);
+  rmm::device_uvector<int> result(5, handle.get_stream());
+
+  {
+    // raft::device_buffer d_buf{handle, array.data_handle(), exts, true};
```
cjnolet, line 66: Should we remove these comments just for maintainability?

cpp/include/raft/core/temporary_device_buffer.hpp
```diff
@@ -0,0 +1,220 @@
...
+
+/**
+ * @brief Factory to create a `raft::temporary_device_buffer`
+ *
+ * @tparam ElementType type of the input
+ * @tparam IndexType index type of `raft::extents`
+ * @tparam LayoutPolicy layout of the input
+ * @tparam ContainerPolicy container to be used to own device memory if needed
+ * @tparam Extents variadic dimensions for `raft::extents`
+ * @param handle raft::device_resources
```
cjnolet, line 140: Can we add a brief usage example here? We'll also need to manually add these (ideally create a doxygengroup and link it in to a new `rst` file for `temporary_device_buffer).

divyegala: Should the usage example be for any particular method? Maybe something from `neighbors`?

cjnolet: Really, it could just be a copy/paste example of including the `device_resources.hpp`, defining a pointer on host, and then using the pointer to invoke the factory function. (Only the factory functions here  need usage examples).

divyegala: Done!

cpp/include/raft/core/temporary_device_buffer.hpp
```diff
@@ -0,0 +1,266 @@
...
+
+/**
+ * \defgroup TemporaryDeviceBuffer `raft::temporary_device_buffer` and associated factories
+ * @{
+ */
+
+/**
+ * @brief An object to have temporary representation of either a host/device pointer in device
+ * memory. This object provides a `view()` method that will provide a `raft::device_mdspan` that may
+ * be read-only depending on const-qualified nature of the input pointer.
```
wphicks, line 36:
```suggestion
 * @brief An object which provides temporary access on-device to memory from either a host or device pointer.
 * This object provides a `view()` method that will provide a `raft::device_mdspan` that may
 * be read-only depending on const-qualified nature of the input pointer.
```

wphicks (approved):
Looks great! Just one suggestion to clarify docs.

cpp/include/raft/core/temporary_device_buffer.hpp
divyegala: That reads much better, thanks!

cjnolet (approved):
The docs and usage examples look great. I think this is ready to go. Very nice to also see the container policy being reused here too.

PR #1317, opened 2023-03-07
Consolidate linter configuration into pyproject.toml

vyasr:
raft currently stores config information for linters across many different pyproject.toml and setup.cfg files. This PR moves all linter configuration out of setup.cfg files into pyproject.toml where possible. flake8 does not support pyproject.toml, so that data is placed in the flake8-specific .flake8 file. Old and unused .flake8 files have been removed. These changes are a precursor to build system migrations that should allow further consolidation of the data in these files.


benfred (approved):
this looks good to me - thanks for making these changes!

PR #1325, opened 2023-03-09
Using int64_t specializations for `ivf_pq` and `refine`

cjnolet:
Since FAISS and Milvus are both using `int64_t` everywhere and we haven't explicitly encountered places where `uint64_t` is being used for ann indices, we're opting to just use `int64_t` instead.


benfred (approved):

cjnolet:
This is fixing a bug in cuml and I don't believe any existing users will be negatively impacted by the changes to the specialized types so I'm going to go ahead and merge this over.

PR #1328, opened 2023-03-10
Update and standardize IVF indexes API

viclafargue:
Update and standardize IVF indexes API + edits on specializations


cpp/bench/neighbors/knn.cuh
```diff
@@ -179,9 +179,8 @@ struct ivf_pq_knn {
   {
     index_params.n_lists = 4096;
     index_params.metric  = raft::distance::DistanceType::L2Expanded;
-
-    auto data_view = raft::make_device_matrix_view<const ValT, IdxT>(data, ps.n_samples, ps.n_dims);
-    index.emplace(raft::neighbors::ivf_pq::build(handle, index_params, data_view));
+    index.emplace(raft::neighbors::ivf_pq::build(
```
cjnolet, line 182: We don't want to revert this back to the pointer based APIs. We should always be preferring the mdspan APIs now (eventually pointer APIs will be removed)

cpp/bench/neighbors/knn.cuh
```diff
@@ -190,13 +189,8 @@ struct ivf_pq_knn {
...
   {
     search_params.n_probes = 20;
-
-    auto queries_view =
-      raft::make_device_matrix_view<const ValT, IdxT>(search_items, ps.n_queries, ps.n_dims);
-    auto idxs_view  = raft::make_device_matrix_view<IdxT, IdxT>(out_idxs, ps.n_queries, ps.k);
-    auto dists_view = raft::make_device_matrix_view<dist_t, IdxT>(out_dists, ps.n_queries, ps.k);
     raft::neighbors::ivf_pq::search(
-      handle, search_params, *index, queries_view, ps.k, idxs_view, dists_view);
+      handle, search_params, *index, search_items, ps.n_queries, ps.k, out_idxs, out_dists);
```
cjnolet, line 193: Here too with using mdspan APIs.

cpp/include/raft/neighbors/ivf_flat.cuh
```diff
@@ -397,37 +397,35 @@ void search(raft::device_resources const& handle,
...
-template <typename value_t, typename idx_t, typename int_t>
+template <typename value_t, typename idx_t>
 void search(raft::device_resources const& handle,
             const index<value_t, idx_t>& index,
             raft::device_matrix_view<const value_t, idx_t, row_major> queries,
             raft::device_matrix_view<idx_t, idx_t, row_major> neighbors,
             raft::device_matrix_view<float, idx_t, row_major> distances,
-            const search_params& params,
-            int_t k)
+            const search_params& params)
```
cjnolet, line 407: I would prefer that we keep any params structs to be the second argument (after the handle) and then any additional single parameter arguments come last. 

This is the way it's done across the codebase. Kmeans is a good example of an API outside of neighbors that uses that design.

achirkin (changes requested):
Thanks for the PR! Would you mind moving the 'xxx_params' to a second argument in all apis instead of the last one? That will be "much less breaking" change, because the initial implementation, both public and `detail`, followed this pattern.

cjnolet (changes requested):
Thanks for fixing these changes. It's look good so far but want to make sure we are still being consistent with the other APIs in the codebase. 

For consistency, we use the following order of arguments always:

1. Handle 
2. Params structs
3. In 
4. Out 
5. Additional (non struct) parameters

cpp/include/raft/neighbors/ivf_flat.cuh
```diff
@@ -397,37 +397,35 @@ void search(raft::device_resources const& handle,
  * [n_queries, k]
  * @param[out] distances a device pointer to the distances to the selected neighbors [n_queries, k]
  * @param[in] params configure the search
- * @param[in] k the number of neighbors to find for each query.
  */
-template <typename value_t, typename idx_t, typename int_t>
+template <typename value_t, typename idx_t>
```
lowener, line 401: Remove the `@tparam` documentation for this as well

cpp/include/raft/neighbors/ivf_flat.cuh
```diff
@@ -270,38 +328,32 @@ void extend(raft::device_resources const& handle,
...
- * @tparam value_t data element type
- * @tparam idx_t type of the indices in the source dataset
- * @tparam int_t precision / type of integral arguments
- * @tparam matrix_idx_t matrix indexing type
+ * @tparam T data element type
+ * @tparam IdxT type of the indices in the source dataset
  *
- * @param[in] handle
- * @param[inout] index
+ * @param handle
```
cjnolet, line 339: handle should be `[in]`

cjnolet (changes requested):
PR is looking good. I think just some updates to the docs of the arguments to consistently use `[in]`, `[out]` and `[inout]` everywhere.

python/pylibraft/pylibraft/common/mdspan.pxd
```diff
@@ -37,3 +38,6 @@ cdef device_matrix_view[int8_t, int64_t, row_major] get_dmv_int8(
 
 cdef device_matrix_view[int64_t, int64_t, row_major] get_dmv_int64(
     array, check_shape) except *
+
+cdef optional[device_matrix_view[int64_t, int64_t, row_major]] create_optional(
```
cjnolet, line 42: I don't know if cython will allow automatic overloads of this based on type, will it? We might need to rename this something like `make_optional_view_int64` or something.

cjnolet (approved):
This looks good to me so I'm approving. I think we should visit the `create_optional` overload in an immediate follow-on though. What do you think?

achirkin (approved):

PR #1329, opened 2023-03-10
Generic linalg::map

achirkin:
Update the implementation behind `raft::linalg::map` and `raft::linalg::map_offset` to allow multiple inputs and optional index.

Originally, this is a part of the effort to reduce the latency of ivf-pq search. The new implementation replaces several helpers, which have been using thrust; at the moment, raft uses a thrust policy that occasionally inserts extra `cudaStreamSynchronize`, and this negatively affects the latency on small inputs.

The new implementation is generic enough to replace many raft's utility functions. It uses vectorized load/stores if possible, which improves performance.


achirkin:
Anecdotal evidence shows improved performance. I've tested it `ivf_pq::search` with small batch sizes, where the removal of unnecessary stream synchronization reduces the latency by up to 5-10% in some cases. However, the new primitive is going to be used everywhere across the codebase, so more benchmarking wouldn't hurt.

cjnolet:
I'm really happy to see more consolidation here! This is great!

cpp/include/raft/linalg/map.cuh
```diff
@@ -22,119 +22,74 @@
...
-          typename... Args,
-          typename = raft::enable_if_input_device_mdspan<InType>,
-          typename = raft::enable_if_output_device_mdspan<OutType>>
-void map(raft::device_resources const& handle, InType in, OutType out, MapOp map, Args... args)
+template <typename OutType,
+          typename Func,
+          typename... InTypes,
+          typename = raft::enable_if_output_device_mdspan<OutType>,
+          typename = raft::enable_if_input_device_mdspan<InTypes...>>
+void map(const raft::device_resources& res, OutType out, Func f, InTypes... ins)
```
cjnolet, line 58: I really like this new function a lot and I'm really happy to see so many functions getting consolidated and using the same guts underneath. 

The concern that I have about this new API is that the ordering of the arguments breaks the consistency that we have been using across the new mdspan APIs.

We have been trying to follow this ordering:
1. Handle
2. Params struct (if it exists)
3. Inputs
4. Outputs
5. Any additional single params

This new signature puts the inputs after the outputs. To keep consistency, we could always swap the outputs and inputs across all the APIs... but that would be quite a change to make and would probably require another deprecation round if we did it right.

I do like the simplicity of the variadic templates, which is what makes this so hard. Another option could be to accept a std::vector of inputs like we do in the knn API, but then that would make it much harder to allow the types to differ across inputs.

Any other options you can think of? Again, I'm really happy with this change but would also like to try and keep the APIs as consistent as possible. I suppose if we squint our eyes enough, another perspective could be to look at the inputs individually as "additional parameters", but that's also kind of awkward.

cpp/include/raft/linalg/map.cuh
```diff
@@ -22,119 +22,74 @@
...
 
 /**
- * @brief Perform an element-wise unary operation on the input offset into the output array
+ *  @brief Map a function over zero-based flat index (element offset) and zero or more inputs.
  *
  * Usage example:
  * @code{.cpp}
- *  #include <raft/core/device_mdarray.hpp>
- *  #include <raft/core/handle.hpp>
- *  #include <raft/core/operators.hpp>
```
cjnolet, line 108: Why remove these from the usage example? The goal with these examples is to make them copy/paste for the users so they don't have to go searching through headers to find what they need.

cpp/include/raft/linalg/map.cuh
```diff
@@ -22,119 +22,74 @@
...
- * @tparam MapOp the device-lambda performing the actual operation
- * @tparam TPB threads-per-block in the final kernel launched
- * @tparam OutType data-type of result of type raft::device_mdspan
- * @tparam Args additional parameters
- * @param[in] handle raft::device_resources
- * @param[in] in the input of type raft::device_mdspan
- * @param[out] out the output of the map operation of type raft::device_mdspan
- * @param[in] map the device-lambda
- * @param[in] args additional input arrays
+ * @brief Map a function over zero or more input mdspans.
```
cjnolet, line 34: Can you add a few more words just to describe what we how we expect the map to produce an output ? (For example, by applying a function to zero or more inputs element-wise, to produce an output value). 

Can you also mention that this supports vectorized loads? From a user's perspective, I think it's unclear which functions support that at this point.

cpp/include/raft/linalg/map.cuh
```diff
@@ -22,119 +22,74 @@
...
-  } else {
-    map_k<in_value_t, MapOp, std::uint64_t, TPB, out_value_t, Args...>(
-      out.data_handle(), out.size(), map, handle.get_stream(), in.data_handle(), args...);
-  }
+  return detail::map<false>(res, out, f, ins...);
 }
 
 /**
- * @brief Perform an element-wise unary operation on the input offset into the output array
+ *  @brief Map a function over zero-based flat index (element offset) and zero or more inputs.
```
cjnolet, line 64: Same here, just add a couple words to make it more clear that the function is applied to the offset value and zero or more inputs element-wise to produce a single output value

cpp/include/raft/linalg/map.cuh
```diff
@@ -22,119 +22,74 @@
...
- * @param[in] handle raft::device_resources
- * @param[in] in the input of type raft::device_mdspan
- * @param[out] out the output of the map operation of type raft::device_mdspan
- * @param[in] map the device-lambda
- * @param[in] args additional input arrays
+ * @brief Map a function over zero or more input mdspans.
+ *
+ * Usage example:
+ * @code{.cpp}
+ *  auto input = raft::make_device_vector<int>(res, n);
```
cjnolet, line 38: These examples should be copy/paste for users without them having to dig through a bunch of headers to get them to compile. We should add the headers in here too. 

Can we also mention that this supports vectorized loads?

cpp/include/raft/neighbors/detail/ivf_pq_search.cuh
```diff
@@ -1311,13 +1309,13 @@ void ivfpq_search_worker(raft::device_resources const& handle,
...
+      make_device_mdarray<uint32_t>(handle, mr, make_extents<uint32_t>(n_queries * n_probes));
     rmm::device_uvector<uint32_t> cluster_labels_out(n_queries * n_probes, stream, mr);
-    auto index_list   = index_list_buf.data();
+    auto index_list   = index_list_buf.data_handle();
     index_list_sorted = index_list_sorted_buf.data();
-    thrust::sequence(handle.get_thrust_policy(),
-                     thrust::device_pointer_cast(index_list),
-                     thrust::device_pointer_cast(index_list + n_queries * n_probes));
+
+    linalg::map_offset(handle, index_list_buf.view(), identity_op{});
```
cjnolet, line 1318: Yes!

cpp/include/raft/neighbors/detail/ivf_pq_search.cuh
```diff
@@ -1379,10 +1377,11 @@ void ivfpq_search_worker(raft::device_resources const& handle,
...
   if (manage_local_topk) {
     query_kths.resize(n_queries, stream);
-    thrust::fill_n(handle.get_thrust_policy(),
-                   query_kths.data(),
-                   n_queries,
-                   float(dummy_block_sort_t<ScoreT, IdxT>::queue_t::kDummy));
+    linalg::detail::map<false>(
+      stream,
+      query_kths.data(),
+      n_queries,
```
cjnolet, line 1383: I wonder if this is an opportunity to expose our own "fill_n" function. Or would that just be too redundant to map?

cjnolet (changes requested):
I'm really happy to have this change. Mostly I just have some minor comments on the docs but 

I'd also like to get your ideas (and maybe gather some ideas from others) about the ordering of arguments in the public API.

cpp/include/raft/linalg/map.cuh
achirkin: Thanks for the suggestions!

I don't think there is an easy way out of this situation. On the one hand, it makes sense to move the outputs to the beginning of the signatures, because it's very unlikely that we may need to use template parameter packs for them. On the other hand, I can imagine the parameter packs being used either for inputs or for extra single params (although less likely), hence changing the order to be always inputs-last does not guarantee to solve this forever.
Besides, there is at least one more such controversial yet useful function: [matrix/linewise_op](https://github.com/rapidsai/raft/blob/b12fc82be4c06cd157925ef14411fa2ee9cf0d07/cpp/include/raft/matrix/linewise_op.cuh#L65-L70), which has its mdspan inputs in the beginning as well as in the end and single parameters in the middle.

In my opinion, it's ok to have a little compromise to consistency for the few parameter-pack public functions we have but keep it strict where possible.

Perhaps, another option is to add a few more fixed overloads to map:
```
void map(const raft::device_resources& res, InType1 in1, OutType out, Func f);
void map(const raft::device_resources& res, InType1 in1, InType2 in2, OutType out, Func f);
void map(const raft::device_resources& res, InType1 in1, InType2 in2, InType2 in3, OutType out, Func f);
```
In theory, thanks to the lambda at the end the generic `map` version should not participate in instance resolution, and the type inference should work fine.

cpp/include/raft/neighbors/detail/ivf_pq_search.cuh
achirkin: Technically, we have it in [matrix/init.cuh](https://github.com/rapidsai/raft/blob/1c321f298a0d308efd0795512221fb1093904aad/cpp/include/raft/matrix/init.cuh#L64-L70). However, here use the `detail` version because the input comes in `rmm::uvector`. I cannot use the mdspan api here, because I resize it conditional on runtime arguments ("enable" the buffer only if it's used). 

Perhaps, this would be a good place to use `std::optional<mdarray>`, but it's hard to `emplace` the optional when we mostly use helpers to create mdarrays.

cjnolet: To your last point- would `std::make_optional` not work here? The `mdarray` is moveable as far as I know.

cpp/include/raft/linalg/map.cuh
cjnolet: Yeah, it's a hard problem for sure- I recall doing the mdspan for linewise_op and thinking something very similar (like squinting my eyes and trying to think of the additional vectors as params). 

I like the overloads too, and I think 3 of them could probably be enough to cover most cases.

As a middle ground, what would you think of doing something like this? 

```
map(handle, in, out, func, additional_in...)? 
```

I know it slightly complicates the internal logic, but in my perspective it stays true to our consistent arg ordering and it turns the `additional_in` into an additional params argument (without having to squint as much to justify it). Then in the case where the user wants zero inputs, we only need the one additional overload. I suppose we could do something similar here for linewise_op as well.

achirkin: > As a middle ground, what would you think of doing something like this?
NB: `map` can have zero input vectors (then it plays as `fill` or `range`)/

cpp/include/raft/neighbors/detail/ivf_pq_search.cuh
achirkin: I guess it could work (didn't try yet), but may be clumsy to force specify all template parameters.

cpp/include/raft/linalg/map.cuh
achirkin: (removed the previous remark, which I didn't notice on my phone was repeating what you've already said).

I don't know.. I don't like putting the different meanings into the first and the rest of the arguments, when they are semantically equal.

Can we just slightly modify the signature rule to state the layout as you declared it PLUS one exception that _if whatever group of arguments contains a parameter pack, the whole group goes last_? That is, if a part of the inputs is a parameter pack, all inputs pop to the end of the signature; if a part of the outputs is a parameter pack, then all outputs go after the single-parameter args, and so on.

cpp/include/raft/neighbors/detail/ivf_pq_search.cuh
achirkin: Update: turned out not so bad, although it didn't reduce the number of lines.

What do you think about backporting [`std::optional::transform`](https://en.cppreference.com/w/cpp/utility/optional/transform) from C++23 into raft? I miss this function so much for one-line constructs like this:
```C++
query_kths.transform([](auto x){return x.data_handle();}).value_or(std::nullptr);
```

cpp/include/raft/linalg/map.cuh
achirkin: Added the "pseudo-code" explanation and the performance note about vectorization.

cpp/include/raft/linalg/map.cuh
cjnolet: > Can we just slightly modify the signature rule to state the layout as you declared it PLUS one exception that if whatever group of arguments contains a parameter pack, the whole group goes last? That is, if a part of the inputs is a parameter pack, all inputs pop to the end of the signature; if a part of the outputs is a parameter pack, then all outputs go after the single-parameter args, and so on.

I'm okay with that, but I think I'm more okay with that if we do it in concert w/ this change as well:

```
void map(const raft::device_resources& res, InType1 in1, OutType out, Func f);
void map(const raft::device_resources& res, InType1 in1, InType2 in2, OutType out, Func f);
void map(const raft::device_resources& res, InType1 in1, InType2 in2, InType2 in3, OutType out, Func f);
```

cpp/include/raft/linalg/map.cuh
```diff
@@ -22,119 +22,267 @@
...
+ *  }
+ * @endcode
+ *
+ * _Performance note_: when possible, this function loads the argument arrays and stores the output
+ * array using vectorized cuda load/store instructions. The size of the vectorization depends on the
+ * size of the largest input/output element type and on the alignment of all pointers.
+ *
+ * Usage example:
+ * @code{.cpp}
+ *  #include <raft/core/device_mdarray.hpp>
```
cjnolet, line 49: Sorry, I missed this one. This should be `device_mdspan` for the view layer (in all other docs too)

cjnolet (changes requested):

cpp/include/raft/linalg/map.cuh
achirkin: But in this example I'm using only functions from `core/mdarray.hpp` (make_device_vector(), .view()), hence I imported that one instead of mdspan.

cjnolet: Ah, I missed that- I'm so used to seeing `make_device_vector_view` and `make_device_matrix_view` used in the examples (which I think is fine too). I'm not nitpicky about that and I think it's fine. I'll approve.

cjnolet (approved):

PR #1333, opened 2023-03-11
Consolidate pre-compiled specializations into single `libraft` binary

### LINKED ISSUE #824: [FEA] Rename libraft-distance to libraft
we had originally created two pre-compiled shared libs and had planned to create potentially more of them over time. However, we have started putting more APIs in libraft-distance than just distance things and we really only need one additional package as a result of the FAISS dependency for a lot of the nearest neighbors stuff.

We should rename libraft_distance.so to just libraft to be more consistent with other rapids projects. Once we are able to remove the FAISS dependency, we can also merge the source files from libraft-nn into libraft and just have a single package for runtime APIs and precompiled template specializations.

cjnolet:
Now that we no longer need to rely on the FAISS dependency, this PR:
1. Consolidates  the `libraft-distance` and `libraft-nn` targets into a single `libraft` artifact
2. Introduces a new `raft::compiled` target to link against the new `libraft.so` binary. Removes `raft::distance`and `raft::nn` targets.
3. Consolidates `RAFT_DISTANCE_COMPILED` and `RAFT_NN_COMPILED` pre-processor vars into a single `RAFT_COMPILED` (to match similar pattern implementated by `spdlog`)
4. Consolidates `specializations.cuh` headers
5. Updates all docs, scripts, and build infra

This change has been a long time coming and is intended to be a 23.04 feature. This is further going to require updates to several projects downstream. Here's a checklist to track that progress:
- [x] offline announcement
- [x] cuml https://github.com/rapidsai/cuml/pull/5272
- [x] cugraph https://github.com/rapidsai/cugraph/pull/3348
- [x] cuopt https://github.com/rapidsai/cuopt/pull/1023
- [x] cugraph-ops https://github.com/rapidsai/cugraph-ops/pull/429

This PR depended on https://github.com/rapidsai/raft/pull/1340 (removing FAISS from the build) and on https://github.com/rapidsai/raft/pull/1202 (replacing the FAISS bfknn w/ our own), both of which have been merged. 

Closes #824


cpp/CMakeLists.txt
```diff
@@ -168,13 +146,8 @@ include(cmake/modules/ConfigureCUDA.cmake)
...
 # * Requirements -------------------------------------------------------------
 
-if(RAFT_COMPILE_LIBRARIES)
-  set(RAFT_COMPILE_DIST_LIBRARY ON)
-  set(RAFT_COMPILE_NN_LIBRARY ON)
-endif()
-
-if(RAFT_COMPILE_DIST_LIBRARY OR distance IN_LIST raft_FIND_COMPONENTS)
-  set(RAFT_ENABLE_cuco_DEPENDENCY ON)
+if(RAFT_COMPILE_LIBRARY)
```
robertmaynard, line 149: This can be dropped as it is a no-op

vyasr:
Could we please separate the FAISS removal into a separate PR from the consolidation of libraries? I'm excited for all of the changes, but this is a massive PR in its current form and anything we can do to split it up would be a huge boon to reviewers.

cjnolet:
@vyasr I can do that if it helps, but it's such a small change (just removing `get_faiss.cmake` and subsequent uses of `faiss::faiss`) that it seemed more straightforward to merge w/ the library consolidation as well.

vyasr:
> @vyasr I opened #1340 to encapsulate the FAISS removal, but please note that PR also depends upon #1202 (so it doesn't actually include as many changes as it appears. same w/ this PR).
> 
> I've updated the description of this PR to mention that it includes the changes in the PRs above.

Thanks! #1202 is merged and #1340 is just waiting on an ops review. I can look through the CMake here once #1340 is merged too so that the diff here is minimized.

build.sh
```diff
@@ -18,7 +18,7 @@ ARGS=$*
 # script, and that this script resides in the repo dir!
 REPODIR=$(cd $(dirname $0); pwd)
 
-VALIDARGS="clean libraft pylibraft raft-dask docs tests bench clean --uninstall  -v -g -n --compile-libs --compile-nn --compile-dist --allgpuarch --no-nvtx --show_depr_warn -h --minimal-deps"
+VALIDARGS="clean libraft pylibraft raft-dask docs tests bench clean --uninstall  -v -g -n --compile-lib --allgpuarch --no-nvtx --show_depr_warn -h"
```
divyegala, line 21: So what does the `libraft` arg here do that is different from `--compile-lib`?

conda/recipes/libraft/meta.yaml
divyegala, line 1: How come we were able to drop all the requirements from here?

cpp/src/neighbors/specializations/ball_cover_knn_query.cu
```diff
@@ -14,18 +14,10 @@
  * limitations under the License.
  */
 
+#include <cstdint>
 #include <raft/neighbors/ball_cover.cuh>
 #include <raft/neighbors/ball_cover_types.hpp>
-
-// Ignore upstream specializations to avoid unnecessary recompiling
```
divyegala, line 20: Can you highlight what the issue here was and why it's not there anymore?

divyegala (commented):
Really minor questions. What a fantastic PR!

build.sh
cjnolet: Libraft installs the headers, `compile-lib` compiles the library. This keeps the build.sh consistent with other rapids projects. The cpp target is always libraft while the compiled shared lib is an option.

conda/recipes/libraft/meta.yaml
cjnolet: We no longer need the libraft-nn package at all. Is that what you mean?

cpp/src/neighbors/specializations/ball_cover_knn_query.cu
cjnolet: The libraft-nn package and libraft-distance were independent and didn't depend on each other. This was done by design so they could be installed individually. As a result, we needed to explicitly check if the distance lib was also compiled/installed when the nn lib was being compiled so we could benefit from skipping those specializations. Otherwise we had to re-compile the stuff individually for each library.

This is why I mentioned last week (offline) that consolidating these libraries is going to lower the size and compile faster- all of the specializations can always benefit from all of the other specializations (eg disrance specializations get compiled once and can always be used by brute force knn, fused l2 knn, rbc now that they are all together in the same package).

sevagh (approved):

divyegala (approved):

benfred (approved):

vyasr (approved):
I get the sense that I'm missing some context for the reorganization of the template specialization headers, probably because there were dependencies when using FAISS that have now changed without our usage of FAISS, or possibly because they're now shared between the former nn and dist lib files thanks to them being combined. Assuming all of that was done correctly, the build-related changes look correct to me. Some minor questions that aren't worth blocking over.

cpp/src/neighbors/specializations/detail/ball_cover_lowdim_pass_two_3d.cu
```diff
@@ -15,11 +15,9 @@
  */
 
 #include <cstdint>
+#include <raft/neighbors/specializations.cuh>
 #include <raft/spatial/knn/detail/ball_cover/registers.cuh>
 
-// TODO: Change this to proper specializations after FAISS is removed
```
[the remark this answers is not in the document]
cjnolet: Yeah, so @divyegala had a similar question and I can understand why it looks confusing here. Imagine the inital goals of RAFT's building individual (and individually optional) packages for `nn` and `distance`. We had users like cugraph who didn't want a FAISS dependency but users like cuml who didn't care. We also needed to be careful to keep the faiss dependency out of `pylibraft` for the same reasons. As a result, we made it such that `libraft-nn` had no dependency on `libraft-distance` and vice versa. A user *could* configure their build to compile and install both, in which case they could share specializations which were already compiled. 

However, if a user only built `libraft-nn` or `libraft-distance` individually, we needed to be careful that any APIs they were using that had specializations headers would not ignore compilingn specializations for the functions which they had not pre-compiled already. This meant we had to be very careful about keeping them separate. As a "standard practice" we want users to only have to worry about including `raft/<namespace>/specializations.cuh` for the namespaces they are using to ease their use of raft, but if we always included all of the namespaces in all of the spcecializations files, they'd get linking errors if they didn't always compile both libraries. 

So what we did was deprecate the `spatial/` directory and all of its contents. Since we now have our own brute-force knn not based on faiss, we moved those specializtions over to the libraft-distance package. The only two APIs that remained in `libraft-nn` were random ball cover and the fused l2 1-nn (which specifically dependened on the faiss k-selection device functions). We isolated those headers down in `spatial/knn` to separate them and didn't include them in `neighrbos/specializations.cuh`. That way special care could be taken only for those two APIs instead of having to split out the majority. Now that we're consolidating these apis and have removed the need to compile or link against the faiss library (though kept some choice headers in the meantime which are copied into the codebase), we can merge the two together, which also means merging the specializations. Underneath, @benfred has updated the `spatial/knn` headers to just use the `neighbors/` headers since those are being used now anyways. Next steps (which will be done in 23.06) will be to move all the needed `spatial/knn` headers over to `neighbors/` and finally remove those which are no longer needed.

python/pylibraft/pylibraft/neighbors/__init__.py
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+# from .brute_force import knn as brute_force_knn
```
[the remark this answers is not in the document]
cjnolet: No, this is an upcoming 23.04 feature that accidentally got merged into this PR but we're unlinking it from the build for now. It'll prbably be merged tomorrow or wednesday (it just needs pytests).

python/pylibraft/pylibraft/neighbors/brute_force.pyx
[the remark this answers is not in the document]
cjnolet: No, hese leaked in from https://github.com/rapidsai/raft/pull/1292 originally when we thought this was going to be a 23.06 feature. Since the feature is so close to being merged for 23.04, I'm thinking we should jsut keep it for now in this PR because it's a higher priority to get these libs consolidated before more users start looking at RAFT (from GTC talks/keynote).

cpp/include/raft/spatial/knn/specializations/knn.cuh
```diff
@@ -0,0 +1,43 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <raft/spatial/knn/knn.cuh>
+
+namespace raft::spatial::knn {
```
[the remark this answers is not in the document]
cjnolet: This particular one is just for backwards compatiblity. The new `raft/neighbors/` specializations are what user's should be using.

cpp/src/neighbors/specializations/detail/ball_cover_lowdim_pass_two_3d.cu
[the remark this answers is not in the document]
vyasr: Got it, thanks for that context. It helps a lot to understand the various moving pieces at play (including those in PRs that I haven't actively reviewed) to know why some of these changes are all happening in concert.

python/pylibraft/pylibraft/neighbors/brute_force.pyx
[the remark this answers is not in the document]
cjnolet: Just updating here for posterity that I removed these changes from the PR

PR #1336, opened 2023-03-13
Revert "Generic linalg::map (#1329)"

cjnolet:
The change to the public API was siginficant enough here that it's going to require updates on the cuml side which invoke `map_k` with very different arguments (e.g. mdspan, different order). For now, it's best we revert this commmit to unblock cuml and then we can proceed by keeping the old APIs and deprecating them until we change cuml.

### DIFF
--- a/cpp/include/raft/linalg/binary_op.cuh
+++ b/cpp/include/raft/linalg/binary_op.cuh
@@ -18,9 +18,12 @@
 
 #pragma once
 
+#include "detail/binary_op.cuh"
+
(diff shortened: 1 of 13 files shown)


benfred (approved):

cjnolet:
cc @achirkin Github doesn't seem to provide a way to re-open a merged PR. Can you open a new PR and add the original `map_k` API back into it? That way we can deprecate it in the PR and merge the changes over without breaking cuml at all.

PR #1337, opened 2023-03-14
Generic linalg::map

achirkin:
Update the implementation behind `raft::linalg::map` and `raft::linalg::map_offset` to allow multiple inputs and optional index.

Originally, this is a part of the effort to reduce the latency of ivf-pq search. The new implementation replaces several helpers, which have been using thrust; at the moment, raft uses a thrust policy that occasionally inserts extra `cudaStreamSynchronize`, and this negatively affects the latency on small inputs.

The new implementation is generic enough to replace many raft's utility functions. It uses vectorized load/stores if possible, which improves performance.

This is the second take on the PR #1329 that keeps the deprecated `map_k` function, which is used in cuml.


cjnolet (approved):

MatthiasKohl:
@cjnolet This PR just broke our builds in cugraph-ops because it uses a host constexpr function (`std::min`) in device code.
The relevant compiler flag is mentioned here: https://github.com/rapidsai/raft/blob/branch-23.04/docs/source/build.md#using-raft-in-downstream-projects
but I was just wondering: is there some list of compiler flags that's required to build RAFT?
It's possible that we're missing other flags like this one

MatthiasKohl:
@achirkin @cjnolet IIUC, this PR breaks at least the API semantics of `raft::linalg::writeOnlyUnaryOp` because the Lambda now takes as first input a reference to the output that should be written to the output pointer, whereas previously it took as first input the output pointer.
This was also documented clearly at the API level here : https://github.com/rapidsai/raft/blob/79e1ce854329bd655f2b02c86e6ec39165eebfc6/cpp/include/raft/linalg/unary_op.cuh#L60-L62

Semantically, the new function doesn't do the same thing at all, and it took me a lot of time to trace our issues in CI to this.

I know this PR is marked as breaking but it would be great that if we change the semantics of very generic functions like these, we notify others ahead of time so that we can check whether we make use of those semantics (that were guaranteed before, see the docs linked above !).

Thanks

cjnolet:
@MatthiasKohl I do apologize for this. I did a quick verification to make sure this wouldn't break any other repositories downstream but I admit I did not check cugraph-ops. 

Would it help if we went ahead and backed out the change for the write only unary op for now? Please note that we are planning to deprecate the writeOnlyUnaryOp function in favor of `map_offset`.

PR #1340, opened 2023-03-14
Removing FAISS from build

cjnolet:
This depends on https://github.com/rapidsai/raft/pull/1202 so those changes are also included here. This should also not be merged until that PR is merged.


bdice:
@cjnolet Thanks for splitting this, I agree this will be better and allow the faiss removal to merge faster. If #1202 is targeting 23.04, should the faiss removal also target 23.04?

cjnolet:
@bdice I'm a little concerned doing that large of a change at the very end of 23.04, which is why we're targeting this specifically for early 23.06. What might not be obvious from the `bfknn` PR is that there are some follow-on features needed (in 23.04) as well in order remove FAISS fully from cuML.

bdice:
> @bdice I'm a little concerned doing that large of a change at the very end of 23.04, which is why we're targeting this specifically for early 23.06. What might not be obvious from the `bfknn` PR is that there are some follow-on features needed (in 23.04) as well in order remove FAISS fully from cuML.

Gotcha. Sounds good, thanks for the context!

benfred (approved):
looks good!

vyasr (approved):
Looks great!

vyasr (changes requested):
Looks like faiss still needs to be removed from the conda recipes (libcudf/meta.yaml and conda_build_config.yaml). Also, I still see faiss_select files and namespaces in a number of places, do those need to be removed as well?

cjnolet:
> Looks like faiss still needs to be removed from the conda recipes (libcudf/meta.yaml and conda_build_config.yaml). Also, I still see faiss_select files and namespaces (coming from selection_faiss.cuh) in a number of places, do those need to be removed as well?

@vyasr done. And no, the `faiss_select/` directory has a couple minor headers we've copied out of FAISS in the meantime. We're going to be getting rid of those soon too but it allows us to make progress otherwise.

raydouglass (approved):

vyasr (approved):

cjnolet:
Thinking about this more, we would like to have RAFT integrated into FAISS by early 23.06, which means FAISS will likely want to be using 23.04, at least to start. Given the transition has been smoother than I had originally thought this far, and since it's still a couple weeks from burndown, I'm going to go ahead and merge this over now. If we encounter any problems leading not burndown, im going to revert the change and push to 23.06. Cc @bdice @jakirkham @vyasr

PR #1341, opened 2023-03-14
Pass minimum target compile options through `raft::raft`

cjnolet:
cc @robertmaynard @MatthiasKohl

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -223,6 +223,8 @@ target_link_libraries(
 )
 
 target_compile_features(raft INTERFACE cxx_std_17 $<BUILD_INTERFACE:cuda_std_17>)
+target_compile_options(raft PUBLIC $<$<COMPILE_LANGUAGE:CUDA>:--expt-extended-lambda --expt-relaxed-constexpr>)
+
 
 # Endian detection
 include(TestBigEndian)
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -224,7 +224,8 @@ target_link_libraries(
 
 target_compile_features(raft INTERFACE cxx_std_17 $<BUILD_INTERFACE:cuda_std_17>)
 target_compile_options(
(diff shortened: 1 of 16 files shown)


cpp/CMakeLists.txt
```diff
@@ -223,6 +223,8 @@ target_link_libraries(
 )
 
 target_compile_features(raft INTERFACE cxx_std_17 $<BUILD_INTERFACE:cuda_std_17>)
+target_compile_options(raft PUBLIC $<$<COMPILE_LANGUAGE:CUDA>:--expt-extended-lambda --expt-relaxed-constexpr>)
```
robertmaynard, line 226:
```suggestion
target_compile_options(raft PUBLIC $<$<COMPILE_LANGUAGE_AND_ID:CUDA,NVIDIA>:--expt-extended-lambda --expt-relaxed-constexpr>)
```

robertmaynard (changes requested):

cpp/CMakeLists.txt
robertmaynard: Since this is public we should be extra safe and only apply this to `nvcc` and not clang ( for the edgiest of edge cases )

cjnolet: Thanks Robert! Just committed that change

cjnolet:
@robertmaynard it looks like we are getting an error in CI related to the changes. Is there something else I need to set in order to make that work? 

```
  Expression did not evaluate to a known generator expression

```

cpp/CMakeLists.txt
```diff
@@ -223,6 +223,10 @@ target_link_libraries(
 )
 
 target_compile_features(raft INTERFACE cxx_std_17 $<BUILD_INTERFACE:cuda_std_17>)
+target_compile_options(
+  raft INTERFACE $<$<COMPILE_LANGUAGE_AND_ID:CUDA,NVIDIA>:--expt-extended-lambda
```
robertmaynard, line 227:
```suggestion
  raft INTERFACE $<$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>:--expt-extended-lambda
```
Here is the correct generator expression.

robertmaynard (changes requested):

Issue #1310, opened 2023-03-01
[FEA] Max workspace size


tfeher:
**Is your feature request related to a problem? Please describe.**

PR #1137 added a workspace resource, which can be used for temporary memory allocations. I would like an interface to
- set / get max workspace size, 
- query the currently allocated workspace.

There are several algorithms, where the workspace size that we actually use depends on an internal batch_size parameter:
 - ivf-pq: https://github.com/rapidsai/raft/blob/ae15be761ccdb456426226c938790680ab64fbde/cpp/include/raft/neighbors/detail/ivf_pq_search.cuh#L1608-L1611
 - radix-topk:  https://github.com/rapidsai/raft/blob/ae15be761ccdb456426226c938790680ab64fbde/cpp/include/raft/matrix/detail/select_radix.cuh#L598-L603
 - kmeans balanced: https://github.com/rapidsai/raft/blob/ae15be761ccdb456426226c938790680ab64fbde/cpp/include/raft/cluster/detail/kmeans_balanced.cuh#L385-L388
 - cuML svm predict https://github.com/rapidsai/cuml/blob/branch-23.04/cpp/src/svm/svc_impl.cuh#L113-L123

In  some cases above, we just hardcode a reasonable parameter. But the performance of an algorithm usually depends on the batch size, so we would like to allow the user to specify the workspace size, and the algorithms can adapt their batch size accordingly.

**Describe the solution you'd like**
- Add an extra `max_size `attribute to the workspace resource,
- track currently allocated size
- print warning if we try to allocate larger than max workspace size.

**Describe alternatives you've considered**
- Simply use a pool allocator, with initial_size = max_size = workspace_size.
- This allows to specify the max workspace size during initialization.
- We can query the pool size.
- We don't have access of the currently allocated workspace.

The last point is problematic if we have nested algorithms that need workspace. Concrete example is IVF-PQ search can use radix-top-k. 
 
**Additional context**
- While the difference between max_workspace_size - currently_allocated_size gives the theoretical free space, in practice it is not guaranteed that we can allocate that much memory when a pooling allocator is used (due to fragmentation of the pool).
- The workspace is always freed when an algorithm returns, therefore fragmentation might not be an issue. 
- Even if max_workspace_size - currrently_allocated_size is not accurate, it is an estimate that is useful for setting batch size.

tfeher:
cc @achirkin

achirkin:
[...]

PR #1356, opened 2023-03-20
Limiting workspace memory resource

achirkin:
Wrap the `rmm::mr::device_memory_resource` returned by `get_workspace_resource` into a `rmm::mr::limiting_resource_adaptor`. This allows to control how much memory is dedicated for the temporary allocations within raft and query this limit in the code (`mr->get_allocation_limit()` - `mr->get_allocated_bytes()`).

**Breaking change**: workspace resource passed in arguments everywhere via a shared_ptr instead of a raw pointer. This affects `set_workspace_resource` function and the constructors of `device_resources` and `handle_t`. Luckily, it's optional everywhere and is likely rarely used, so the impact should be low.

Resolves https://github.com/rapidsai/raft/issues/1310


achirkin:
Update: alternative option:
- Using pool memory resource by default.
- Set the default limit to be half of free gpu memory, to encourage the user to set the limit manually if they desire.

cpp/include/raft/core/resource/device_memory_resource.hpp
```diff
@@ -40,36 +65,109 @@ class device_memory_resource : public resource {
...
  */
-inline void set_workspace_resource(resources const& res, rmm::mr::device_memory_resource* mr)
+inline void set_workspace_resource(resources const& res,
+                                   rmm::mr::device_memory_resource* mr,
+                                   std::optional<std::size_t> allocation_limit = std::nullopt,
+                                   std::optional<std::size_t> alignment        = std::nullopt)
 {
-  res.add_resource_factory(std::make_shared<workspace_resource_factory>(mr));
+  // NB: to preserve the semantics of passing memory resource without transferring the ownership,
+  //     we create a shared pointer with a dummy deleter (void_op).
```
tfeher, line 164: Is there any reason to wrap it into a `shared_ptr `in this case?

cpp/include/raft/core/resource/device_memory_resource.hpp
```diff
@@ -40,36 +65,109 @@ class device_memory_resource : public resource {
...
+  std::optional<std::size_t> alignment_;
+  std::shared_ptr<rmm::mr::device_memory_resource> mr_;
+
+  // Create a pool memory resource by default
+  static inline auto default_memory_resource(std::size_t limit)
+    -> std::shared_ptr<rmm::mr::device_memory_resource>
+  {
+    constexpr std::size_t kOneGb = 1024lu * 1024lu * 1024lu;
+    auto min_size                = std::min<std::size_t>(kOneGb, limit / 2);
+    auto max_size                = limit * 3lu / 2lu;
```
tfeher, line 95: Why the 3/2 factor?

tfeher (changes requested):
Thanks @achirkin for the PR! Overall it looks good, this PR addresses the problems in #1310. 

The main question is what shall be the default size. 

@ahendriksen could you help addressing the issues for this PR while Artem is away?

cpp/include/raft/core/resource/device_memory_resource.hpp
achirkin: That's a totally random guess. The idea is that the maximum pool size should be big enough that the `limiting_resource_adaptor` limit is always reached before the `pool_memory_resource`  maximum size, accounting for the fragmentation of the allocated pool blocks.

I'd say the `min_size` has a bigger impact on the overall behavior (the current setting is also a totally random guess).

cpp/include/raft/core/resource/device_memory_resource.hpp
achirkin: Yes, I opted for the `shared_ptr` semantics in `limited_memory_resource` to make it easier for the end-user to pass custom memory resources (e.g. create a pool memory resource and not worry about managing its lifetime). Since the  `limited_memory_resource` accepts only shared pointers, we need to wrap the raw pointer.

There's only one exception to this whole story - `rmm::mr::get_current_device_resource()`, which returns a pointer to the memory resource that is manages elsewhere. The original workspace resource API was tailored for this use case; to keep the PR non-breaking, I decided to leave all signatures as they were, hence the shared_ptr with a dummy deleter.

cjnolet:
@achirkin can you explain why exactly this feature is needed? We don't do this type of thing anywhere else in RAPIDS and we leave it to the user to configure these types of resources.

tfeher:
> why exactly this feature is needed?

@cjnolet #1310 tries to give the motivation why do we need this. As listed there, currently several algorithms hardcode a limit for batch size. This tries to be a reasonable guess, but if we have more memory available, then we could improve performance by increasing the batch size. 

This PR would
- give users a knob to (indirectly) set the batch size by configuring the workspace memory resource,
- give raft developers a way to query the available workspace memory.

cpp/include/raft/core/resource/device_memory_resource.hpp
```diff
@@ -40,36 +65,109 @@ class device_memory_resource : public resource {
...
+      upstream, min_size, max_size);
+  }
+
+  // Allow a fraction of available memory by default.
+  static inline auto default_allocation_limit() -> std::size_t
+  {
+    std::size_t free_size{};
+    std::size_t total_size{};
+    RAFT_CUDA_TRY(cudaMemGetInfo(&free_size, &total_size));
+    return free_size / 2;
```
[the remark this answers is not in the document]
tfeher: Note that using `free_size` here is probably not what we want: if an upstream pooling allocator has already grabbed the space, that would be misleading. 
- One could use `total_size`, to set a practically "unlimited" workspace. This is consistent with the existing not limited workspace allocator.  But then we could not rely on the workspace size to calculate batch size without risk of running OOM. 
- That brings me back to the point of allowing a smaller limited workspace size, that is consistent with the algorithm hardcoded heuristics.
- If I see it correctly, the workspace resource is now used for all mdarrays https://github.com/rapidsai/raft/blob/e963f5a64371c25f1e2e09ab1e3ca15658293e2d/cpp/include/raft/core/device_container_policy.hpp#L167  Does this mean that an ivf_pq::index is allocating its device_matrix from the workspace? @cjnolet

tfeher (commented):
Few more notes.

cpp/include/raft/core/resource/device_memory_resource.hpp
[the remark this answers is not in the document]
achirkin: > the workspace resource is now used for all mdarrays

Oh. Thanks for pointing this out! I think, the meanings we put into the workspace resource start to diverge again. 

With the change to container policy you quoted, the workspace resource essentially becomes a raft substitute to `rmm::mr::current_device_resource`, because the policy is used for allocating all mdarrays, including the ones passed to the user inside structs like `ivf_*::index`. If we're to hold on to this approach, this PR certainly does not make any sense at all.

In this PR we assume the workspace resource is used only for the temporary allocations within the algorithms. If we are to proceed with the PR, we must revert the container policy change. I also believe the default pool allocator is justified in this case, because (1) it's good for performance, (2) it does not affect the allocations outside its explicit use and thus does not conflict with other (e.g. third-party) code, (3) we can enforce a conservative limit to avoid taking up all gpu resources.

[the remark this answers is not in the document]
cjnolet: Sorry Artem, I'm still catching up on this discussion a bit. From what I've read so far, I think your design is headed in the right direction but I think there's still a few things to polish off (like the defaults, for example).

cjnolet (changes requested):
Sorry I've taken so long to respond on this one, @achirkin. I've had some comments pending for quite awhile that I thought I submitted already.

I think this is headed in a good place but I think we need to discuss the defaults a little further. Overall, I'll be really happy to see us removing any and all custom memory resource creation from within RAFT and using this instead.

cjnolet (changes requested):
Sorry I've taken so long to respond on this one, @achirkin. I've had some comments pending for quite awhile that I thought I submitted already.

I think this is headed in a good place but I think we need to discuss the defaults a little further. Overall, I'll be really happy to see us removing any and all custom memory resource creation from within RAFT and using this instead.

cpp/include/raft/core/resource/device_memory_resource.hpp
```diff
@@ -41,36 +66,109 @@ class device_memory_resource : public resource {
...
+  }
 
  private:
-  rmm::mr::device_memory_resource* mr;
+  std::size_t allocation_limit_;
+  std::optional<std::size_t> alignment_;
+  std::shared_ptr<rmm::mr::device_memory_resource> mr_;
+
+  // Create a pool memory resource by default
+  static inline auto default_memory_resource(std::size_t limit)
```
cjnolet, line 91: I like the idea of using the limiting memory resource so we can better compute batch sizes across raft (and downstream) algorithms. 

But if the user doesn't specify an explicit workspace memory resource to use, we should just default to wrapping a limited memory resource around the current default rmm resource and perhaps set the memory limit to the max memory of the device.

This allows a user to opt-in to this feature and gain all the benefit while not forcing us to make (potentially very bad) assumptions on the initial limits.  

This provides a very simple configuration option, allowing a user to specify "I want to use 3Gb for workspaces" while minimizing surprises from poorly set "automagical" settings that the uninformed or common user might now know anything about.

I think we should document this option heavily for the users, perhaps in a "getting the most out of RAFT" tutorial in the docs.

cpp/include/raft/core/resource/device_memory_resource.hpp
cjnolet: I know we were accepting raw pointers in some places already but shared_ptr does provide a great mechanism for unified already owned elsewhere pointers vs one being owned by the shared pointer itself. What do you say we just accept shared_ptr everywhere for the memory resources to make the api experience more consistent?

cpp/include/raft/neighbors/detail/ivf_pq_search.cuh
```diff
@@ -717,8 +717,7 @@ inline void search(raft::device_resources const& handle,
                    uint32_t n_queries,
                    uint32_t k,
                    IdxT* neighbors,
-                   float* distances,
-                   rmm::mr::device_memory_resource* mr = nullptr)
+                   float* distances)
```
cjnolet, line 720: It's great to see this change!

cpp/include/raft/core/device_container_policy.hpp
```diff
@@ -183,4 +183,43 @@ class device_uvector_policy {
   [[nodiscard]] auto make_accessor_policy() const noexcept { return const_accessor_policy{}; }
 };
 
+/**
+ * @brief A container policy for device mdarray allocated in the temporary workspace.
+ */
+template <typename ElementType>
+class device_uvector_workspace_policy {
```
cjnolet, line 190: I like the idea of having a different container policy specifically for workspace resources but is there any reason we can't just subclass the current device container policy and override create()?

cpp/include/raft/core/resource/device_memory_resource.hpp
```diff
@@ -15,24 +15,49 @@
...
+#include <rmm/mr/device/limiting_resource_adaptor.hpp>
 #include <rmm/mr/device/per_device_resource.hpp>
+#include <rmm/mr/device/pool_memory_resource.hpp>
+
+#include <cstddef>
+#include <optional>
 
 namespace raft::resource {
-class device_memory_resource : public resource {
+class limited_memory_resource : public resource {
```
cjnolet, line 32: Small nitpick: the rmm memory resource is called a "limiting_memory_resource" but this class is called "limited_memory_resource". We should be consistent.

cjnolet (changes requested):
Thanks again for working on this, Artem! I finally got a chance to catch up on all the discussions and take a closer look at your implementation. 

I love this feature and I think it's going to be very powerful. @tarang-jain has been working on some optimizations to single linkage clustering and hdbscan primitives which require batch sizes and temporary workspace allocations which could very kick benefit from these changes.

My main concern is the default and I'd like to minimize the surprises to the common (non power-user) as much as possible and continue to prefer the default rmm memory resource when not otherwise specified. 

Otherwise, this really is a fantastic addition for raft, cuml, and all of our users. 

Also noting that we should document this feature clearly in our docs and maybe in a new tutorial to help the interested user get the most out of RAFT.

cpp/include/raft/core/device_container_policy.hpp
achirkin: I guess we can! The only problem we would still have to verbosely re-export all of the type aliases, right?

cpp/include/raft/core/resource/device_memory_resource.hpp
achirkin: I'm keeping the raw-pointer version here for two reasons:
 1. Just to keep the PR 'non-breaking'
 2. To allow a user to pass `rmm::mr::get_current_device_resource()` that is managed statically by rmm (though this can be workarounded by wrapping it into a shared_ptr with a dummy deleter).

@cjnolet let me know if you'd like to remove this, I don't mind.

cjnolet: I think we are so early in the process of using the workspace resource that it would be a good time to clean this up and make it consistent. I doubt anyone is using it downstream, really, so I'm okay changing it here.

cjnolet:
@achirkin on the topic of the default- what if we provided a helper function/setter in raft/core/resource/workspace_resource.hpp that allowed a user to easily switch the default to the pool allocated workspace resource? 

That way we get the best of everything- by default if a user has specified nothing else, the workspace resource is just a limiting_memory_resource around the rmm current memory resource. The user can, in a simple one-liner, call something like:

resource::set_workspace_pool_default() 

In order to enable the pool option.

cpp/include/raft/core/resource/device_memory_resource.hpp
achirkin: In general, I agree having the default memory resource makes the most sense.
The only issue is if we want to use this in ivf-flat / ivf-pq, it will hit performance, because there we rely on the pooling allocator when creating many intermediate buffers on each iteration/batch when batching.

What do you think about this: query `rmm::mr::get_current_device_resource()` and wrap it in the pooling resource only if the resource is a plain `rmm::mr::cuda_memory_resource` (default rmm resource)?

achirkin: Alternatively, I could keep the default workspace resource plain/non-pooling unconditionally, but add a check to k-neighbours, ivf-pq, and ivf-flat with a warning sent to logs if the default workspace resource is used (suggesting to switch to the pooling workspace resource).

cpp/include/raft/core/resource/device_memory_resource.hpp
cjnolet: I think it's probably best to keep the use of "limiting" here so that we are consistent with RMM's naming conventions as well.

cpp/include/raft/core/device_container_policy.hpp
achirkin: I've opted to use only one policy and pass a memory resource to it, as before. The reason is that the policy type is aliased within mdarray type; so the arrays created with different policies would have different types, which would require a too large refactoring and may be difficult to use.

achirkin:
@cjnolet , @tfeher I marked our scattered discussions as resolved to consolidate the main points here.

### Updates

  1. Workspace usage by mdarrays allocations:
    - set the default rmm memory resource as a default for mdarrays;
    - recovered the old behavior allowing to pass the memory resource to the mdarray helpers explicitly;
    - did __not__ add an extra set of helpers for creating mdarrays in the workspace memory resource (I think we already have a lot there).
  2. Changed the API to always set the memory resource using shared pointers instead of plain pointers to make managing their lifetime easier.
  3. Changed the default resource logic (see code comments)
  4. Slightly changed the default sizes (see code comments)
  5. Added a few helpers to query the workspace size and easily switch between pool/plain memory resources as Corey suggested in our discussion.

### TBD

Points 1, 2, 5 seem to reflect our consensus in the discussions. For points 3, 4 I need you input for sure.

cjnolet (changes requested):
Thanks for the updates, @achirkin! I think this is getting closer but still a coupld things here.

cpp/include/raft/neighbors/ivf_pq-inl.cuh
```diff
@@ -458,7 +458,16 @@ void search(raft::resources const& handle,
             float* distances,
             rmm::mr::device_memory_resource* mr = nullptr)
 {
-  detail::search(handle, params, idx, queries, n_queries, k, neighbors, distances, mr);
+  if (mr != nullptr) {
+    // Shallow copy of the resource with the automatic lifespan:
+    //                               change the workspace resource temporarily
+    raft::resources res_local(handle);
+    resource::set_workspace_resource(
```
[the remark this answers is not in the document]
achirkin: This code is here only for backwards compatibility - to keep the ivf-pq changes non-breaking. We need it to not change the original semantics. I'm happy to remove this as soon as we remove the function argument.

cpp/include/raft/neighbors/ivf_pq-inl.cuh
```diff
@@ -458,7 +458,16 @@ void search(raft::resources const& handle,
             float* distances,
             rmm::mr::device_memory_resource* mr = nullptr)
```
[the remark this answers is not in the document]
achirkin: Sure. I split the function into two overloads with/out `mr` argument and deprecated one of them.

cpp/include/raft/core/resource/device_memory_resource.hpp
```diff
@@ -41,36 +66,182 @@ class device_memory_resource : public resource {
...
+ * (`rmm::mr::get_current_device_resource()`.
+ *
+ * Note, the workspace resource is always limited; the limit here defines how much of the global
+ * memory resource can be consumed by the workspace allocations.
+ *
+ * @param res raft resources object for managing resources
+ * @param allocation_limit
+ *   the total amount of memory in bytes available to the temporary workspace resources.
+ */
+inline void set_workspace_to_global_resource(
```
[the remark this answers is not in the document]
achirkin: In case the user wants to set the workspace multiple times or change the allocation limit.

cpp/include/raft/core/resource/device_memory_resource.hpp
achirkin: Ok, done.

cpp/include/raft/core/resource/device_memory_resource.hpp
```diff
@@ -41,36 +66,161 @@ class device_memory_resource : public resource {
...
+    //   1) 'limit' is accurate - the user should be more likely to run into the limiting
+    //      resource adaptor bad_alloc error than into the pool bad_alloc error.
+    //   2) The pool doesn't grab too much memory on top of the 'limit'.
+    auto max_size = std::min<std::size_t>(limit + kOneGb / 2lu, limit * 3lu / 2lu);
+    auto upstream = rmm::mr::get_current_device_resource();
+    return std::make_shared<rmm::mr::pool_memory_resource<rmm::mr::device_memory_resource>>(
+      upstream, min_size, max_size);
+  }
+
+  /** Get the global memory resource wrapped into an unmanaged shared_ptr (with no deleter). */
```
tfeher, line 109: Please add a note why no deleter is used (we have discussed in [this thread](https://github.com/rapidsai/raft/pull/1356#discussion_r1150360407), but it is better to record this information in the code comments).

cpp/include/raft/core/resource/device_memory_resource.hpp
```diff
@@ -41,36 +66,161 @@ class device_memory_resource : public resource {
...
+  }
+
+  auto get_resource_type() -> resource_type override { return resource_type::WORKSPACE_RESOURCE; }
+  auto make_resource() -> resource* override
+  {
+    return new limiting_memory_resource(mr_, allocation_limit_, alignment_);
+  }
+
+  /** Construct a sensible default pool memory resource. */
+  static inline auto default_pool_resource(std::size_t limit)
```
tfeher, line 86: Instead of encoding certain assumptions here, I would just remove `default_pool_resource` together with `set_workspace_to_pool_resource`. I think it is sufficient to have `set_workspace_resource` API, the user could set a pool using that. An example block to `set_workspace_resource` docstring would be nice to demonstrate this.

tfeher (changes requested):
Thanks @achirkin for the updates. Attached please find my comments.

cpp/include/raft/core/device_container_policy.hpp
```diff
@@ -164,10 +164,19 @@ class device_uvector_policy {
  public:
   auto create(raft::resources const& res, size_t n) -> container_type
   {
-    return container_type(n, resource::get_cuda_stream(res), resource::get_workspace_resource(res));
+    if (mr_ == nullptr) {
```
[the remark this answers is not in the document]
achirkin: You're right, this extra field in container policy increases the footprint of all device mdarrays.

The main reason for having it as a member is to not change the mdarray api. This is, by the way, also the way we used to have it before the introduction of the workspace memory resources.

I think you raise a very good point here, and I would be up to do the necessary change. I think, the minimum required change is to add an extra constructor to mdarray that would take the memory resource. However, there are couple problems with this:
  - The dependent types, such as raft sparse matrices, would need to be modified as well to allow them to specify the memory resource.
  - This also breaks the encapsulation: normally an mdarray should rely only on the container policy type for managing allocations, but with the change, there would be an explicit memory-related argument passed to mdarray besides the container policy.
@cjnolet , what are your thoughts on this?

cpp/include/raft/neighbors/detail/ivf_pq_search.cuh
```diff
@@ -775,11 +776,7 @@ inline void search(raft::resources const& handle,
...
 
-  auto pool_guard = raft::get_pool_memory_resource(mr, n_queries * n_probes * k * 16);
-  if (pool_guard) {
-    RAFT_LOG_DEBUG("ivf_pq::search: using pool memory resource with initial size %zu bytes",
-                   n_queries * n_probes * k * 16ull);
-  }
+  auto mr = resource::get_workspace_resource(handle);
 
   // Maximum number of query vectors to search at the same time.
   const auto max_queries = std::min<uint32_t>(std::max<uint32_t>(n_queries, 1), 4096);
```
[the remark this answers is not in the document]
achirkin: Good point! I modified `get_max_batch_size` to use the workspace size.

cpp/include/raft/neighbors/detail/ivf_pq_search.cuh
```diff
@@ -704,11 +708,11 @@ inline auto get_max_batch_size(uint32_t k,
   auto ws_size = [k, n_probes, max_samples](uint32_t bs) -> uint64_t {
     return uint64_t(is_local_topk_feasible(k, n_probes, bs) ? k * n_probes : max_samples) * bs;
   };
-  constexpr uint64_t kMaxWsSize = 1024 * 1024 * 1024;
-  if (ws_size(max_batch_size) > kMaxWsSize) {
+  auto max_ws_size = resource::get_workspace_total_bytes(res);
```
tfeher, line 711: Why not `get_workspace_free_bytes`?

achirkin: Honestly I'm not sure which one is better here. On the one hand, `get_workspace_total_bytes` gives slightly more consistent experience (the batch sizes and thus performance don't differ depending on what code is running alongside the ivf-pq search and consuming the workspace memory); on the other hand, there could be a possible scenario where the search would throw exception due to trying to get more space than available.

Thus, I think both solutions are not perfect. Let me know if you prefer the `get_workspace_free_bytes` slightly more, I'm happy to replace it.

cpp/include/raft/core/device_container_policy.hpp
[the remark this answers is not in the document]
cjnolet: I don't think we want to change the mdarray API, especially since we're eventually planning to consolidate w/ CCCL's mdarray (which will already be painful, but the more we diverge from the reference/standard, the more painful it will become). 

One thing we can do is maybe split the current container policy into two strategies- the standard device_containter_policy that just uses the `get_default_resource` and perhaps one that explicitly uses the `get_workspace_resource`? What do you think? I think it's important that we couple the container policy to the `raft::resources` as much as possible so that we limit the amount of potential places where different state is initialized- if we need a container policy with a different memory resource, it should come from the `raft::resources` layer.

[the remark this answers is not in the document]
achirkin: The problem is that the container policy is a template parameter of mdarray and stored by value inside it (we cannot just inherit it). In raft we have many helper functions and wrapper types, which fix the type of the container policy. If we're to make several different policies now, that would be a major overhaul. Thus I suggest we keep it as it is for now, and discuss whether we want to do this in a follow-up PR.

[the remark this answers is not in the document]
cjnolet: Just to clarify my stance here- I'm not suggesting inheritance here- the relationship could be from a composition. I'm just saying that it might make more sense to provide different container policies for workspace and default memory resource variants. We could also provide new helpers for constructing the "workspace" variants if that would make it easier. That being said, I'm okay with pushing that one off until we have had more time to think about and discuss it.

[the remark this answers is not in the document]
achirkin: Independently of how we do it, the type of the policy is the part of the mdspan template. We have all those synonyms for device_mdspans and host_mdspans that hide this and fix the type of the policy. We would need to change all of those if we want to allow more than one policy.
Thanks, let's push this off for the next release.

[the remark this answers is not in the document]
tfeher: Thanks for the discussion. Agreed, we can proceed for now with the current implementation.

tfeher (commented):
Thanks Artem for the update and the discussion! Looks good, apart from a minor change request.

cpp/include/raft/core/resource/detail/device_memory_resource.hpp
```diff
@@ -0,0 +1,58 @@
...
+  // Detect if the plain cuda memory resource is used for the workspace
+  if (rmm::mr::cuda_memory_resource{}.is_equal(*get_workspace_resource(res)->get_upstream())) {
+    static std::set<std::string> notified_names{};
+    static std::mutex mutex{};
+    std::lock_guard<std::mutex> guard(mutex);
+    auto [it, inserted] = notified_names.insert(std::move(user_name));
+    if (inserted) {
+      RAFT_LOG_WARN(
+        "[%s] the default cuda resource is used for the raft workspace allocations. This may lead "
+        "to a significant slowdown for this algorithm. Consider using the default pool resource "
```
cjnolet, line 50: Looks great! Thanks for adding this

cpp/include/raft/neighbors/detail/ivf_flat_serialize.cuh
```diff
@@ -45,7 +45,7 @@ struct check_index_layout {
                 "paste in the new size and consider updating the serialization logic");
 };
 
-template struct check_index_layout<sizeof(index<double, std::uint64_t>), 296>;
+template struct check_index_layout<sizeof(index<double, std::uint64_t>), 328>;
```
cjnolet, line 48: @divyegala has mentioned this in another PR. I don't think we need to hold up this PR for it but I'd prefer if we started creating a constexpr for this value so it's easier to spot and manage.

cjnolet (approved):
LGTM @achirkin. Thanks so much for this change!

PR #1358, opened 2023-03-20
Fix ivf flat specialization header IdxT from uint64_t -> int64_t

ahendriksen:
The ivf_flat specialization header declarations used a wrong index type. 

The specializations for ivf flat are defined for int64_t. The raft_runtime interface also uses the int64_t instances. The ivf_flat specialization header, however, declared an interface using uint64_t.

This is fixed with this PR. Should also reduce compile times for `src/distance/neighbors/ivf_flat_search.cu.o`


cjnolet (approved):
Thanks for fixing this Allard!

divyegala (approved):
Ah, thanks for spotting this. Totally missed the hard-code there.

Issue #1300, opened 2023-02-23
[BUG] Consider removing spdlog dependency for substantial compile time improvements


ahendriksen:
**Describe the bug**

Including the `spdlog` headers is quite expensive. Just adding `#include <spdlog/spdlog.h>` to an empty file adds 2.8 seconds to the compilation time. For the pairwise distance kernels, removing the `spdlog` include can [reduce compile times by 50%](https://github.com/ahendriksen/raft/tree/investigate-compile-time-reduction-strategies#obtain-50-compile-time-reduction-by-not-including-rmm-and-spdlog).

**Steps/Code to reproduce bug**
``` sh
time nvcc -arch sm_70 -I/path/to/spdlog/include -x cu -c <(echo '// empty file')
real    0m1.042s

time nvcc -arch sm_70 -I/path/to/spdlog/include -x cu -c <(echo '#include <spdlog/spdlog.h> ')
real    0m3.840s
```

**Expected behavior**
A smaller increase in compile time. For context, including `<string>` adds on the order of 100ms to the compilation time: 
``` sh
time nvcc -arch sm_70 -I/path/to/spdlog/include -x cu -c <(echo '#include <string> ')
real    0m1.160s
```

**Additional context**

**RMM**
RMM also uses `spdlog`. In practice the compile time improvements will only be obtained when RMM also removes its spdlog dependency. 

**Reason**
The reason that compilation takes much longer is that `spdlog` instantiates a bunch of templates in every translation unit when used as a header only library. This happens in [pattern_formatter::handle_flag_](https://github.com/gabime/spdlog/blob/da14258533cb951ce85087ceb45556e0b8253660/include/spdlog/pattern_formatter-inl.h#L1105), which is instantiated [here](https://github.com/gabime/spdlog/blob/da14258533cb951ce85087ceb45556e0b8253660/include/spdlog/pattern_formatter-inl.h#L1410). Just adding back the `spdlog` header doubles the compile times of `cicc` (device side) and also `gcc` on the host side.

**Precompiled-library**
Another option is to not use `spdlog` as a header only library. The effect can be simulated by defining [SPDLOG_COMPILED_LIB](https://github.com/gabime/spdlog/blob/da14258533cb951ce85087ceb45556e0b8253660/include/spdlog/common.h#L28).  When this is defined,  `spdlog` adds only 0.5 seconds:
``` sh
[...]

PR #1360, opened 2023-03-21
[ENH] Add extern template for ivfflat_interleaved_scan 

ahendriksen:
This should cut compilation time for refine_d_int64_t_float.cu.o et al from ~900 seconds to 29 seconds.

The refine specialization contain >100 instances of the ivfflat_interleaved_scan kernel, even though these should be seperately compiled by the ivfflat_search specializations. 

The call to ivf_flat_interleaved_scan is [here](https://github.com/rapidsai/raft/blob/56ac43ad93a319a61073dce1b3b937f6f13ade63/cpp/include/raft/neighbors/detail/refine.cuh#L121). 

Depends on (so please merge after) PR #1307.


ahendriksen:
I think we should find a more sustainable way to manage the template instantiations in the specializations directories.. It is very easy to call a function that does not have an `extern template` definition and as a result recompile many duplicate kernels. Right now, these are only noticeable if you hawkishly check the ninja logs and compare the kernels contained in each translation unit (using cuobjdump). 

For now, I have added comments at each declaration, definition and call of `ivfflat_interleaved_scan` with a tag `greppable-id-specializations-ivf-flat-search`.

ahendriksen:
Marking this as ready for review since most of the builds and tests were succeeding in CI. Changes should be straightforward.

cpp/src/neighbors/specializations/ivfflat_search_float_int64_t.cu
```diff
@@ -18,12 +18,37 @@
...
+// greppable-id-specializations-ivf-flat-search: The ivfflat_interleaved_scan
+// function is used in both raft::neighbors::ivf_flat::search and
+// raft::neighbors::detail::refine_device. To prevent a duplicate instantiation
+// of this function (which defines ~270 kernels) in the refine specializations,
+// an extern template definition is provided. To make sure
+// ivfflat_interleaved_scan is actually compiled here, we explicitly instantiate
+// it below. Please check related function calls after editing template
+// definition below. Search for `greppable-id-specializations-ivf-flat-search`
+// to find them.
+#define RAFT_MAKE_INSTANCE(T, IdxT)                                          \
```
cjnolet, line 30: Can we pull this into a separate file so we can parallelize the compilation here? Both files can include the specializations headers so they won't be compiled multiple times.

ahendriksen: Good question. We could, but unfortunately build times will not improve. I checked by pulling the scan into a separate file and compiling ([diff](https://github.com/ahendriksen/raft/commit/9c2395776a725fa6ea034a4d42b072e4a9c580bd)). The sum of the split compile times is higher than the compile times of the current file.
```
# current:
$ ninja /path/to/raft/cpp/src/neighbors/specializations/ivfflat_search_{float,int8_t,uint8_t}_int64_t.cu^
# split:
$ ninja /path/to/raft/cpp/src/neighbors/specializations/ivfflat_{search,interleaved_scan}_{float,int8_t,uint8_t}_int64_t.cu^
```
Results: 
```
$ python ../scripts/analyze-ninja-log.py current.ninja_log 
CMakeFiles/raft_lib.dir/src/neighbors/specializations/ivfflat_search_float_int64_t.cu.o                                   518.9 seconds
CMakeFiles/raft_lib.dir/src/neighbors/specializations/ivfflat_search_int8_t_int64_t.cu.o                                  785.0 seconds
CMakeFiles/raft_lib.dir/src/neighbors/specializations/ivfflat_search_uint8_t_int64_t.cu.o                                 817.2 seconds

$ python ../scripts/analyze-ninja-log.py split.ninja_log 
CMakeFiles/raft_lib.dir/src/neighbors/specializations/ivfflat_search_float_int64_t.cu.o                                    54.0 seconds
CMakeFiles/raft_lib.dir/src/neighbors/specializations/ivfflat_search_int8_t_int64_t.cu.o                                   54.2 seconds
CMakeFiles/raft_lib.dir/src/neighbors/specializations/ivfflat_search_uint8_t_int64_t.cu.o                                  54.3 seconds
CMakeFiles/raft_lib.dir/src/neighbors/specializations/ivfflat_interleaved_scan_float_int64_t.cu.o                         545.4 seconds
CMakeFiles/raft_lib.dir/src/neighbors/specializations/ivfflat_interleaved_scan_int8_t_int64_t.cu.o                        802.8 seconds
CMakeFiles/raft_lib.dir/src/neighbors/specializations/ivfflat_interleaved_scan_uint8_t_int64_t.cu.o                       822.1 seconds
```
There was some background activity on my system, so it is not completely accurate. But qualitatively, the results above show that it is not worth splitting (at least, not to achieve lower compile times). 
I will upload the analyze-ninja-log.py file as a separate PR. Might be useful..

cjnolet: Your findings here create more questions than answers for me, I'm afraid. If the `interleaved_scan` is a core bottleneck in the `ivfflat_search` then building it once in a separate file and ignoring it when building the pieces that depend on it should be reducing the compile times by nearly as much as building the `interleaved_scan` piece, right? 

To be quite honest I'm not concerned about a small overhead in order to gain parallelism. I agree that the overhead you are showing above does not look particularly negligable but I think it's worthwhile to learn a little more about where it's coming from.

cjnolet (approved):
Thinking about this a little further, I'm okay merging this as-is, even though we have separated all of our other source files. 

I would like it if we would continue investigating why those overheads are so high, though.

cpp/src/neighbors/specializations/ivfflat_search_float_int64_t.cu
ahendriksen: > [...] then building it once in a separate file and ignoring it when building the pieces that depend on it should be reducing the compile times by nearly as much as building the interleaved_scan piece, right?

Yes. And indeed it *nearly* reduces the compile times by as much. Problem is that parallelizing incurs a cost due to (among others) [expensive includes ](https://github.com/rapidsai/raft/issues/1300#issuecomment-1460047077). Also, apart from the interleaved_scan_kernel, there are quite some other kernels that are being compiled and that seem to have some overlap. Especially the coalescedRedcution kernels stand out. I have opened a separate [issue](https://github.com/rapidsai/raft/issues/1372) for that. 
``` bash
$ cuobjdump -res-usage CMakeFiles/raft_lib.dir/src/neighbors/specializations/ivfflat_search_float_int64_t.cu.o | grep Function | cu++filt | sed 's/<.*//' | sort -h | uniq -c 
      2  Function raft::cache::
      2  Function void cub::CUB_101702_700_750_NS::EmptyKernel
     14  Function void raft::linalg::detail::coalescedReductionMediumKernel
     28  Function void raft::linalg::detail::coalescedReductionThickKernel
     78  Function void raft::linalg::detail::coalescedReductionThinKernel
     54  Function void raft::linalg::detail::map_kernel
      4  Function void raft::linalg::detail::stridedReductionKernel
      6  Function void raft::linalg::detail::stridedSummationKernel
     24  Function void raft::matrix::detail::matrixLinewiseVecColsMainKernel
      8  Function void raft::matrix::detail::matrixLinewiseVecColsTailKernel
     24  Function void raft::matrix::detail::matrixLinewiseVecRowsMainKernel
      8  Function void raft::matrix::detail::matrixLinewiseVecRowsTailKernel
     12  Function void raft::neighbors::ivf_pq::detail::calc_chunk_indices_kernel
      4  Function void raft::random::detail::fillKernel
      2  Function void raft::spatial::knn::detail::utils::outer_add_kernel
```

```
$ cuobjdump -res-usage CMakeFiles/raft_lib.dir/src/neighbors/specializations/ivfflat_interleaved_scan_float_int64_t.cu.o | grep Function | cu++filt | sed 's/<.*//' | sort -h | uniq -c 
      2  Function raft::cache::
      2  Function void cub::CUB_101702_700_750_NS::EmptyKernel
      8  Function void raft::linalg::detail::coalescedReductionMediumKernel
     16  Function void raft::linalg::detail::coalescedReductionThickKernel
     44  Function void raft::linalg::detail::coalescedReductionThinKernel
     42  Function void raft::linalg::detail::map_kernel
      2  Function void raft::linalg::detail::stridedReductionKernel
      6  Function void raft::linalg::detail::stridedSummationKernel
     24  Function void raft::matrix::detail::matrixLinewiseVecColsMainKernel
      8  Function void raft::matrix::detail::matrixLinewiseVecColsTailKernel
     24  Function void raft::matrix::detail::matrixLinewiseVecRowsMainKernel
      8  Function void raft::matrix::detail::matrixLinewiseVecRowsTailKernel
    324  Function void raft::neighbors::ivf_flat::detail::interleaved_scan_kernel
     12  Function void raft::neighbors::ivf_pq::detail::calc_chunk_indices_kernel
      4  Function void raft::random::detail::fillKernel
```

PR #1365, opened 2023-03-22
Remove usage of Dask's `get_worker`

pentschev:
In dask/distributed#7580 get_worker was modified to return the worker of a task, thus it cannot be used by client.run, and we must now use dask_worker as the first argument to client.run to obtain the worker.


ajschmidt8 (approved):

PR #1369, opened 2023-03-23
Add Options to Generate Build Metrics Report

divyegala:
This PR will also automatically generate an HTML report in each CI run

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -18,7 +18,7 @@ ARGS=$*
 # script, and that this script resides in the repo dir!
 REPODIR=$(cd $(dirname $0); pwd)
 
-VALIDARGS="clean libraft pylibraft raft-dask docs tests bench clean --uninstall  -v -g -n --compile-lib --allgpuarch --no-nvtx --show_depr_warn -h"
+VALIDARGS="clean libraft pylibraft raft-dask docs tests bench clean --uninstall  -v -g -n --compile-lib --allgpuarch --no-nvtx --show_depr_warn --build-metrics --incl_cache_stats -h"
 HELP="$0 [<target> ...] [<flag> ...] [--cmake-args=\"<args>\"] [--cache-tool=<tool>] [--limit-tests=<targets>] [--limit-bench=<targets>]
  where <target> is:
    clean            - remove all existing build artifacts and configuration (start over)
@@ -42,6 +42,8 @@ HELP="$0 [<target> ...] [<flag> ...] [--cmake-args=\"<args>\"] [--cache-tool=<to
    --allgpuarch                - build for all supported GPU architectures
    --no-nvtx                   - disable nvtx (profiling markers), but allow enabling it in downstream projects
    --show_depr_warn            - show cmake deprecation warnings
+   --build_metrics               - generate build metrics report for libcudf
+   --incl_cache_stats            - include cache statistics in build metrics report
    --cmake-args=\\\"<args>\\\" - pass arbitrary list of CMake configuration options (escape all quotes in argument)
    --cache-tool=<tool>         - pass the build cache tool (eg: ccache, sccache, distcc) that will be used
                                  to speedup the build process.
@@ -65,6 +67,8 @@ BUILD_TYPE=Release
 BUILD_BENCH=OFF
 COMPILE_LIBRARY=OFF
 INSTALL_TARGET=install
+BUILD_REPORT_METRICS=OFF
+BUILD_REPORT_INCL_CACHE_STATS=OFF
 
 TEST_TARGETS="CLUSTER_TEST;CORE_TEST;DISTANCE_TEST;LABEL_TEST;LINALG_TEST;MATRIX_TEST;RANDOM_TEST;SOLVERS_TEST;SPARSE_TEST;SPARSE_DIST_TEST;SPARSE_NEIGHBORS_TEST;NEIGHBORS_TEST;STATS_TEST;UTILS_TEST"
 BENCH_TARGETS="CLUSTER_BENCH;NEIGHBORS_BENCH;DISTANCE_BENCH;LINALG_BENCH;MATRIX_BENCH;SPARSE_BENCH;RANDOM_BENCH"
@@ -303,7 +307,13 @@ fi
 if hasArg clean; then
     CLEAN=1
 fi
+if hasArg --build_metrics; then
+    BUILD_REPORT_METRICS=ON
+fi
 
+if hasArg --incl_cache_stats; then
+    BUILD_REPORT_INCL_CACHE_STATS=ON
+fi
 
 
 if [[ ${CMAKE_TARGET} == "" ]]; then
@@ -363,6 +373,36 @@ if (( ${NUMARGS} == 0 )) || hasArg libraft || hasArg docs || hasArg tests || has
           -DRAFT_COMPILE_LIBRARY=${COMPILE_LIBRARY} \
           ${CACHE_ARGS} \
           ${EXTRA_CMAKE_ARGS}
+
+  if [[ "$BUILD_REPORT_METRICS" == "ON" && -f "${LIBRAFT_BUILD_DIR}/.ninja_log" ]]; then
+      if ! rapids-build-metrics-reporter.py 2> /dev/null && [ ! -f rapids-build-metrics-reporter.py ]; then
+          echo "Downloading rapids-build-metrics-reporter.py"
+          curl -sO https://raw.githubusercontent.com/rapidsai/build-metrics-reporter/v1/rapids-build-metrics-reporter.py
+      fi
+
+      echo "Formatting build metrics"
+      MSG=""
+      # get some sccache stats after the compile
+      if [[ "$BUILD_REPORT_INCL_CACHE_STATS" == "ON" && -x "$(command -v sccache)" ]]; then
+          COMPILE_REQUESTS=$(sccache -s | grep "Compile requests \+ [0-9]\+$" | awk '{ print $NF }')
+          CACHE_HITS=$(sccache -s | grep "Cache hits \+ [0-9]\+$" | awk '{ print $NF }')
+          HIT_RATE=$(echo - | awk "{printf \"%.2f\n\", $CACHE_HITS / $COMPILE_REQUESTS * 100}")
+          MSG="${MSG}<br/>cache hit rate ${HIT_RATE} %"
+      fi
+      MSG="${MSG}<br/>parallel setting: $PARALLEL_LEVEL"
+      MSG="${MSG}<br/>parallel build time: $compile_total seconds"
+      if [[ -f "${LIBRAFT_BUILD_DIR}/libraft.so" ]]; then
+          LIBRAFT_FS=$(ls -lh ${LIBRAFT_BUILD_DIR}/libraft.so | awk '{print $5}')
+          MSG="${MSG}<br/>libraft.so size: $LIBRAFT_FS"
+      fi
+      BMR_DIR=${RAPIDS_ARTIFACTS_DIR:-"${LIBRAFT_BUILD_DIR}"}
+      echo "Metrics output dir: [$BMR_DIR]"
+      mkdir -p ${BMR_DIR}
+      MSG_OUTFILE="$(mktemp)"
+      echo "$MSG" > "${MSG_OUTFILE}"
+      PATH=".:$PATH" python rapids-build-metrics-reporter.py ${LIBRAFT_BUILD_DIR}/.ninja_log --fmt html --msg "${MSG_OUTFILE}" > ${BMR_DIR}/ninja_log.html
+      cp ${LIBRAFT_BUILD_DIR}/.ninja_log ${BMR_DIR}/ninja.log
+  fi
 
   if [[ ${CMAKE_TARGET} != "" ]]; then
       echo "-- Compiling targets: ${CMAKE_TARGET}, verbose=${VERBOSE_FLAG}"
--- a/conda/recipes/libraft/build_libraft.sh
+++ b/conda/recipes/libraft/build_libraft.sh
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 # Copyright (c) 2022-2023, NVIDIA CORPORATION.
 
-./build.sh libraft --allgpuarch --compile-lib --no-nvtx
+./build.sh libraft --allgpuarch --compile-lib --build_metrics --incl_cache_stats --no-nvtx


cjnolet:
cc @ahendriksen for review

build.sh
```diff
@@ -42,6 +42,8 @@ HELP="$0 [<target> ...] [<flag> ...] [--cmake-args=\"<args>\"] [--cache-tool=<to
    --allgpuarch                - build for all supported GPU architectures
    --no-nvtx                   - disable nvtx (profiling markers), but allow enabling it in downstream projects
    --show_depr_warn            - show cmake deprecation warnings
+   --build_metrics               - generate build metrics report for libcudf
```
cjnolet, line 45: libcudf  eh?

build.sh
```diff
@@ -348,6 +358,12 @@ if (( ${NUMARGS} == 0 )) || hasArg libraft || hasArg docs || hasArg tests || has
         echo "Building for *ALL* supported GPU architectures..."
     fi
 
+    # get the current count before the compile starts
+    if [[ "$BUILD_REPORT_INCL_CACHE_STATS" == "ON" && -x "$(command -v sccache)" ]]; then
```
cjnolet, line 362: Can we raise some sort of warning if sccache is not available? Just thinking about a user using this who might not have it installed. We could even fail here just to make sure it's clear

build.sh
```diff
@@ -372,6 +389,38 @@ if (( ${NUMARGS} == 0 )) || hasArg libraft || hasArg docs || hasArg tests || has
...
+      if ! rapids-build-metrics-reporter.py 2> /dev/null && [ ! -f rapids-build-metrics-reporter.py ]; then
+          echo "Downloading rapids-build-metrics-reporter.py"
+          curl -sO https://raw.githubusercontent.com/rapidsai/build-metrics-reporter/v1/rapids-build-metrics-reporter.py
+      fi
+
+      echo "Formatting build metrics"
+      MSG=""
+      # get some sccache stats after the compile
+      if [[ "$BUILD_REPORT_INCL_CACHE_STATS" == "ON" && -x "$(command -v sccache)" ]]; then
+          COMPILE_REQUESTS=$(sccache -s | grep "Compile requests \+ [0-9]\+$" | awk '{ print $NF }')
```
cjnolet, line 405: Does it have to be sccache? Can't we get these metrics on ccache itself too? Maybe we could see which one is installed, fall back to ccache if sccache isn't available, and fail if neither is?

build.sh
```diff
@@ -372,6 +389,38 @@ if (( ${NUMARGS} == 0 )) || hasArg libraft || hasArg docs || hasArg tests || has
...
+          LIBRAFT_FS=$(ls -lh ${LIBRAFT_BUILD_DIR}/libraft.so | awk '{print $5}')
+          MSG="${MSG}<br/>libraft.so size: $LIBRAFT_FS"
+      fi
+      BMR_DIR=${RAPIDS_ARTIFACTS_DIR:-"${LIBRAFT_BUILD_DIR}"}
+      echo "Metrics output dir: [$BMR_DIR]"
+      mkdir -p ${BMR_DIR}
+      MSG_OUTFILE="$(mktemp)"
+      echo "$MSG" > "${MSG_OUTFILE}"
+      PATH=".:$PATH" python rapids-build-metrics-reporter.py ${LIBRAFT_BUILD_DIR}/.ninja_log --fmt html --msg "${MSG_OUTFILE}" > ${BMR_DIR}/ninja_log.html
+      cp ${LIBRAFT_BUILD_DIR}/.ninja_log ${BMR_DIR}/ninja.log
```
ahendriksen, line 422: Can you print a log message where I will be able to find the ninja log html report? Either on disk or the url to the location that CI uploads it to? I want to make sure the build metrics are somewhat discoverable (even if in the log messages in the middle of a conda build...)

ahendriksen: I just realized that it might also be worth just printing the stats to the console. This might be more readable when this PR to the build-metrics-reporter lands: https://github.com/rapidsai/build-metrics-reporter/pull/1/

Worth considering. Also okay to ignore.

build.sh
```diff
@@ -42,6 +42,8 @@ HELP="$0 [<target> ...] [<flag> ...] [--cmake-args=\"<args>\"] [--cache-tool=<to
    --allgpuarch                - build for all supported GPU architectures
    --no-nvtx                   - disable nvtx (profiling markers), but allow enabling it in downstream projects
    --show_depr_warn            - show cmake deprecation warnings
+   --build_metrics               - generate build metrics report for libcudf
+   --incl_cache_stats            - include cache statistics in build metrics report
```
ahendriksen, line 46: The right column does not line up with lines above and below in my terminal. This is true for both --build-metrics and --incl_cache_stats

ahendriksen (commented):
Looks good! Some small comments.

build.sh
divyegala: That's what this line does. If you don't have sccache it just goes ahead and collects the normal report without the cache hit ratio

build.sh
```diff
@@ -18,7 +18,7 @@ ARGS=$*
 # script, and that this script resides in the repo dir!
 REPODIR=$(cd $(dirname $0); pwd)
 
-VALIDARGS="clean libraft pylibraft raft-dask docs tests bench clean --uninstall  -v -g -n --compile-lib --allgpuarch --no-nvtx --show_depr_warn -h"
+VALIDARGS="clean libraft pylibraft raft-dask docs tests bench clean --uninstall  -v -g -n --compile-lib --allgpuarch --no-nvtx --show_depr_warn --build_metrics --incl_cache_stats -h"
```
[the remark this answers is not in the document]
divyegala: Can I also change `--show_depr_warn` then?

build.sh
ahendriksen: How about printing: `The HTML report can be found in ${BMR_DIR}/ninja_log.html. In CI, this report will also be uploaded to the appropriate subdirectory of https://downloads.rapids.ai/ci/raft/.`

Seems less roundabout than:
1. git blame on the line `echo "Metrics output dir: [$BMR_DIR]".`
2. Find it was part of this PR
3. Take the description "This PR will also automatically generate an HTML report in conda-cpp-build CI runs under the task Upload Additional Artifacts"
4. Go to "Upload additional artifacts" in the CI
5. For now: I cannot find that step in the CI for this PR?

I finally found the upload directory by searching through internal communications. Not everybody can do that.

divyegala: `The HTML report can be found in ${BMR_DIR}/ninja_log.html. In CI, this report will also be uploaded to the appropriate subdirectory of https://downloads.rapids.ai/ci/raft/, and the entire URL can be found in "conda-cpp-build" runs under the task "Upload additional artifacts"`

How about this?

ahendriksen: Haha okay. Now I get it. That link is actually quite easy to find once you know where it is... Yes, your new description helped me find it. I think it will help others as well!

build.sh
```diff
@@ -372,6 +396,50 @@ if (( ${NUMARGS} == 0 )) || hasArg libraft || hasArg docs || hasArg tests || has
...
+                  MSG="${MSG}<br/>cache hit rate ${HIT_RATE} %"
+              fi
+          fi
+      fi
+      MSG="${MSG}<br/>parallel setting: $PARALLEL_LEVEL"
+      MSG="${MSG}<br/>parallel build time: $compile_total seconds"
+      if [[ -f "${LIBRAFT_BUILD_DIR}/libraft.so" ]]; then
+          LIBRAFT_FS=$(ls -lh ${LIBRAFT_BUILD_DIR}/libraft.so | awk '{print $5}')
+          MSG="${MSG}<br/>libraft.so size: $LIBRAFT_FS"
+      fi
```
ajschmidt8, line 432: I'd like to include @vyasr and @davidwendt in a discussion here about how we can eliminate some of this boilerplate code.

Presumably some of this functionality will be needed by other repositories, so we should strive to make `rapids-build-metrics-reporter.py` handle as much of this as possible so that we have a single source of truth.

I have some thoughts and proposals here:

- I think `rapids-build-metrics-reporter.py` should handle `sccache -s`/`ccache -s` parsing. The lines of code in this PR are already duplicated from cudf's [build.sh](https://github.com/rapidsai/cudf/blob/4c4fdd2c86deef16bb68a13bc401193274c45266/build.sh#L305-L310), so this is a prime candidate for deduping. Perhaps we can expose a `--cache-tool`/`-c` flag that will let consumers indicate whether `ccache` or `sccache` should be used.
- Whenever possible, I'm in favor of avoiding the manual generation of HTML in bash variables (e.g. `<br/>`). In place of the `--msg` flag (which appears to take in an HTML string), I think we should create the following flags:
  - `--stat`/`-s` - this flag should be able to be specified multiple times and its value will be a key/value pair that can be added to a final summary table (e.g. `--stat="Parallel Setting=$PARALLEL_LEVEL"`, `--stat="Parallel Build Time=$compile_total seconds"`)
  - `--file-size`/`-f` - this flag should be able to be specified multiple times and its value will be a path to a file whose name and size will be added to a final summary table (e.g. `--file-size="${LIBRAFT_BUILD_DIR}/libraft.so"`)

The sum of these changes should allow us to dedupe quite a bit here and also eliminate the `MSG_OUTFILE` lines in this PR.

Once we get a consensus on all of this, I will open an issue to track the rollout of this new script to other repositories as well.

ajschmidt8 (commented):
Looks good.

I left some comments soliciting feedback from you and others.

Let's wait to make any of these changes until we get some consensus so that we can avoid wasteful iterations.

build.sh
davidwendt: I'd rather have the .py strictly handle the `.ninja_log` file. It currently supports other formats besides html including csv and xml (unittest format?). Both of these are handy at times and and there is no place for the stats in these outputs.
Rather than including the ccache stats in the html output, maybe we generate a separate file with sccache states into the artifacts directory. The reason is what incorporated in the html output originally was to just minimize the burden on the Jenkins job at the time. Just piping the `sccache -s > $RAPIDS_ARTIFACT_DIR/cache-stats.txt` seems a reasonable approach.

divyegala: So we entirely eliminate formating sccache/ccache statistics and just add it as a raw file for the user to parse?

divyegala: > Perhaps we can expose a --cache-tool/-c flag

This already exists in RAFT!

davidwendt: If you are calling the output of `sccache -s` raw output then yes.

ajschmidt8: > I'd rather have the .py strictly handle the `.ninja_log` file. It currently supports other formats besides html including csv and xml (unittest format?). Both of these are handy at times and and there is no place for the stats in these outputs.
> Rather than including the ccache stats in the html output, maybe we generate a separate file with sccache states into the artifacts directory. The reason is what incorporated in the html output originally was to just minimize the burden on the Jenkins job at the time. Just piping the `sccache -s > $RAPIDS_ARTIFACT_DIR/cache-stats.txt` seems a reasonable approach.

I see. I wasn't aware of these other formats.

I think there's value in having a single HTML file summary that includes all relevant build stats for CI as opposed to the stats being strewed across many files.

I wonder if we can make some changes to our Python script to accommodate both uses.

What if we rename `rapids-build-metrics-reporter.py` to `rapids-reporter.py` and implement subcommands for each report type?

For instance:

- `rapids-reporter.py summary` can produce a single HTML file and support all of the flags I mentioned above
- `rapids-reporter.py ninja` can support multiple output file types and support a separate set of flags

This should be doable with argparse. @davidwendt, what are your thoughts on this?

I realize this adds a bit of complexity, but I think the upfront investment is worth it. Plus we can write unit tests for the Python script to ensure it continues to work.

The end goal here is to strike the right balance between reusability, simplicity, and user experience for RAPIDS users/repos.

> > Perhaps we can expose a --cache-tool/-c flag
> 
> This already exists in RAFT!

I was referring to creating this option in the Python script. Presumably its value would be set to the value passed into `raft`'s `build.sh` script.

vyasr: I think centralizing more of the logic is the right approach as long as it is done in a way that the different pieces of the document (cache stats, ninja log, etc) are composable and extensible by packages that want to add new files etc. I would advocate for starting by building out the different pieces and then figuring out how to combine them afterwards. As long as there is a consistent HTML output the combining shouldn't be too difficult.

For building up the HTML in Python a reasonable low-level approach is building up the HTML tree with [xml.etree](https://docs.python.org/3/library/xml.etree.elementtree.html). A nicer approach might be to define a [Jinja template](https://jinja.palletsprojects.com/en/3.1.x/) that can be filled in, leaving some arbitrary extra sections that could be appended to so that the output format is extensible.

davidwendt: > Are you suggesting that we just move the existing logic to another shell script? That doesn't seem to solve the problem of reusability / keeping things DRY. All of the repositories will just have another rapids-reporter.sh script that contains all of the duplicated lines of code still.

The existing shell script logic does not necessarily have move to into a new shell script. I just don't want to move this part of the existing shell script into the `rapids-build-metrics-reporter.py` file. We could move the shell script logic into a new separate .py file if that helps. I'm suggesting there will be 2 files in this repository -- one to build/format the sccache/ccache results and one to build/format the ninja-log results. And maybe a 3rd file to combine them if necessary.

> can you point out where / how the other output formats that you mentioned will be used? As far as I can tell, build.sh only outputs a single format type. Are you trying to preserve the multiple output formats for users who may run this tool manually outside of build.sh?

The other output formats are not being used by CI currently. I'm hoping the xml can be used in the future to fail a build if there is a problem with specific source file compile times. The CSV format has been helpful for me locally so I can load the results into Excel and sort them in different ways as well as compare them with other runs.

build.sh
```diff
@@ -348,6 +358,19 @@ if (( ${NUMARGS} == 0 )) || hasArg libraft || hasArg docs || hasArg tests || has
...
+    if [[ -z "${CACHE_TOOL}" ]]; then
+        # default is sccache for CI
+        CACHE_COMMAND="sccache"
+    else
+        CACHE_COMMAND="${CACHE_TOOL}"
+    fi
+    if [[ "$BUILD_REPORT_INCL_CACHE_STATS" == "ON" && -x "$(command -v ${CACHE_COMMAND})" ]]; then
+        # zero the cache statistics
+        "${CACHE_COMMAND}" --zero-stats
+    fi
```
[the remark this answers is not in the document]
divyegala: Okay! I didn't know about default values in bash. But I think the correct expression here is:
```
    CACHE_COMMAND=""
    if [[ "$BUILD_REPORT_INCL_CACHE_STATS" == "ON" && -x "$(command -v ${CACHE_TOOL:-sccache})" ]]; then
        CACHE_COMMAND=$CACHE_TOOL
        "${CACHE_COMMAND}" --zero-stats
    fi
```

build.sh
divyegala: I'd like to merge what we have in RAFT at this moment, and when there is consensus on how to share these scripts and how much logic to pluck out we can go ahead and update it. Is everyone okay with that? There are some major changes happening in RAFT right now where it would be very valuable to have these metrics for hindsight.

build.sh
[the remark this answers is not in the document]
ajschmidt8: Good point. LGTM.

ajschmidt8 (approved):
looks like there are some errors that might need to be sorted out, but pre-approving in anticipation of those fixes.

cjnolet (approved):

PR #1371, opened 2023-03-24
add a distance epilogue function to the bfknn call

benfred:
Add the ability for a user to specify an epilogue function to run after the distance in the brute_force::knn call.

This lets us remove faiss from cuml, by updating the hdbscan reachability code (https://github.com/rapidsai/cuml/pull/5293)


cpp/include/raft/neighbors/brute_force.cuh
```diff
@@ -153,7 +155,8 @@ void knn(raft::device_resources const& handle,
          value_int k,
```
cjnolet, line 155: Can we remove k from here? We have been removing it from the Ann APIs too. It doesn't make any sense to have it with mdspan because we can infer it based on the output arrays.

cpp/include/raft/neighbors/brute_force.cuh
```diff
@@ -138,13 +138,15 @@ inline void knn_merge_parts(
...
+ * @param[in] distance_epilogue: optional epilogue function to run after computing distances
  */
 template <typename idx_t,
           typename value_t,
           typename value_int,
           typename matrix_idx,
           typename index_layout,
-          typename search_layout>
+          typename search_layout,
+          typename epilogue_op = raft::identity_op>
```
cjnolet, line 149: I guess in the current implementation (at least as far as I can tell from the diff), we are not fusing the epilogue to the pairwise distances but instead calling it as a transform after the fact. I think this will probably be the best for the hdbscan case, anyways, since there's some random access lookups that slow down the pw dist computation otherwise. 

Besides, if we did fuse the epilogue to the pw dists, hdbscan would propagate quite far down into the distances, which would mean the precompiled specializations would no longer going to be used. 


Tagging @ahendriksen @teju85 @tfeher for their thoughts here. This seems like the right approach to me- it means launching another kernel but the needed logic for hdbscan's reachability computation is not a trivial mapping.

cpp/include/raft/neighbors/detail/knn_brute_force.cuh
```diff
@@ -169,8 +172,22 @@ void tiled_brute_force_knn(const raft::device_resources& handle,
...
             val = val * (val >= 0.0001);
             if (metric == raft::distance::DistanceType::L2SqrtExpanded) { val = sqrt(val); }
+            val = distance_epilogue(val, row, col);
             return val;
           });
+      } else {
+        // if we're not l2 distance, and we have a distance epilogue - run it now
+        if constexpr (!std::is_same_v<DistanceEpilogue, raft::identity_op>) {
+          auto distances_ptr = temp_distances.data();
+          raft::linalg::map_offset(
```
cjnolet, line 182: Very nice!

cpp/include/raft/neighbors/detail/knn_brute_force.cuh
```diff
@@ -265,7 +285,8 @@ void brute_force_knn_impl(
   bool rowMajorQuery                  = true,
   std::vector<IdxType>* translations  = nullptr,
   raft::distance::DistanceType metric = raft::distance::DistanceType::L2Expanded,
-  float metricArg                     = 0)
+  float metricArg                     = 0,
```
cjnolet, line 288: Can we default this to 2.0 everywhere? That's usually what I see in scikit learn and other places as well. It's only used in Minkowski/lp currently, which would make it equivalent to euclidean.

cpp/include/raft/neighbors/detail/knn_brute_force.cuh
```diff
@@ -169,8 +172,22 @@ void tiled_brute_force_knn(const raft::device_resources& handle,
...
           });
+      } else {
+        // if we're not l2 distance, and we have a distance epilogue - run it now
+        if constexpr (!std::is_same_v<DistanceEpilogue, raft::identity_op>) {
+          auto distances_ptr = temp_distances.data();
+          raft::linalg::map_offset(
+            handle,
+            raft::make_device_vector_view(temp_distances.data(),
+                                          current_query_size * current_centroid_size),
+            [=] __device__(size_t i) {
```
cjnolet, line 186: I think it makes sense to do this as an epilogue on the pairwise distances rather than burdening the (already register constrained) k selection with it.

This formulation of the algorithm (tiled knn with epilogue after tiling) makes this portion equivalent to our gramm matrix API for constructing RKHS kernels (see raft::distance::kernel) and it makes me wonder if there's something to be gained by consolidating these eventually (eg brute force knn primitive becomes a composition of [tiled gramm + epilogue] + k-selection. That could also allow us to reuse more from the gram APIs.

cpp/include/raft/neighbors/brute_force.cuh
```diff
@@ -138,13 +138,15 @@ inline void knn_merge_parts(
  * 					 is ignored if the metric_type is not Minkowski.
  * @param[in] global_id_offset: optional starting global id mapping for the local partition
  *                              (assumes the index contains contiguous ids in the global id space)
+ * @param[in] distance_epilogue: optional epilogue function to run after computing distances
```
cjnolet, line 141: Can you add the expectation if the epilogue function's argument list here too? Just a small example prototype of the function definition will do.

cpp/include/raft/neighbors/detail/knn_brute_force.cuh
```diff
@@ -47,7 +47,9 @@ using namespace raft::spatial::knn;
  * Calculates brute force knn, using a fixed memory budget
  * by tiling over both the rows and columns of pairwise_distances
  */
-template <typename ElementType = float, typename IndexType = int64_t>
+template <typename ElementType      = float,
+          typename IndexType        = int64_t,
+          typename DistanceEpilogue = raft::identity_op>
 void tiled_brute_force_knn(const raft::device_resources& handle,
```
cjnolet, line 53: As mentioned above, id eventually love to find ways we can consolidate (and reuse) this with the gram matrix APIs. Not an immediate priority but the current gramm matrix API is a class that doesn't need to store state and we have a todo to convert it into flattened public API functions like the rest of RAFT.

cjnolet (approved):

cjnolet (changes requested):
I'm sorry! I meant to request changes, not approve. I think it looks great so far but just some minor things

cpp/include/raft/neighbors/brute_force.cuh
ahendriksen: > which would mean the precompiled specializations would no longer going to be used.
Indeed.

I need a bit more context to weigh in. The epilogue is K-selection? Or building the MST?

benfred: The epilogue for hdbscan is this code https://github.com/rapidsai/cuml/blob/bed5213e3a671f18dcfdf2bae912508a17bd24da/cpp/src/hdbscan/detail/reachability.cuh#L167-L176  

We're not fusing it to the pairwise distances, or the select_k op right now (though we could in the future) -

cpp/include/raft/neighbors/brute_force.cuh
```diff
@@ -153,7 +155,8 @@ void knn(raft::device_resources const& handle,
          value_int k,
          distance::DistanceType metric         = distance::DistanceType::L2Unexpanded,
          std::optional<float> metric_arg       = std::make_optional<float>(2.0f),
-         std::optional<idx_t> global_id_offset = std::nullopt)
+         std::optional<idx_t> global_id_offset = std::nullopt,
+         epilogue_op distance_epilogue         = raft::identity_op())
```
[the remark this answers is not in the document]
benfred: Doing it like this also lets us remove code at compile time - like we can check if the op is just an identity function using `if constexpr (std::is_same_v ...` and then exclude compiling it altogether for the default case.

cpp/include/raft/neighbors/brute_force.cuh
benfred: done w/ https://github.com/rapidsai/raft/pull/1371/commits/78f7b1f6b0ba133b0d50bc0340ecc5759d0ebcf8

cjnolet (approved):
LGTM!

PR #1375, opened 2023-03-27
CAGRA

tfeher:
This PR adds CAGRA, a graph based method for nearest neighbor search.


tfeher:
Todo:
- [ ] refactor search params
- [ ] memory allocation using RMM
- [ ] extent test coverage
- [ ] raft logging is missing from a few places
- [ ] revise multi GPU build/prune

tfeher:
Todo:
- [ ] refactor search params
- [ ] memory allocation using RMM
- [ ] extent test coverage
- [ ] raft logging is missing from a few places
- [ ] revise multi GPU build/prune

cpp/src/neighbors/cagra/make_search_cores.sh
tfeher, line 1: Note for reviewer: all the `search_core_<dtype>_<dim>_<team>.cu` files are generated by this script.

ahendriksen: Perhaps it is a good idea to add a notice to the generated files that states that the file is generated by this script. This avoids nasty surprises where the script and the source file start to diverge as new changes are being pushed. (Mostly saying this because I realized that I should have done this for my own generating script for the distance specializations).

tfeher:
**Warning** in its current form, this PR increases the `libraft.so`[ binary size](https://downloads.rapids.ai/ci/raft/pull-request/1375/f0d9210/cuda11_x86_64.ninja_log.html) to 1.7 GiB, which is 1.1 GiB larger than without this PR!
This is due to the many template specializations added.

I will restrict the dataset type to float, and the team_size search param to its default value, this will reduce binary size increase to 125 MiB

tfeher:
The latest commit limits the binary size increase to 125 MiB. [ninja_log](https://downloads.rapids.ai/ci/raft/pull-request/1375/c52dd33/cuda11_x86_64.ninja_log.html).

There is an issue with the specializations, they are not used for build / search, resulting in extra long compile time for these. I am considering to remove the high level specializations, and only leave for the lower level kernels. We can re-introduce these once we ensure that the RAFT_COMPILED flag is visible during compilation of raft_lib.

The ninja log is really very useful for keeping track of these issues!

cpp/include/raft/neighbors/cagra.cuh
```diff
@@ -0,0 +1,223 @@
...
+ * @param[in] dataset a matrix view (host or device) to a row-major matrix [n_rows, dim]
+ * @param[out] knn_graph a host matrix view to store the output knn graph
+ * @param[in] refine_rate refinement rate for ivf-pq search
+ * @param[in] build_params (optional) ivf_pq index building parameters for knn graph
+ * @param[in] search_params (optional) ivf_pq search parameters
+ */
+template <typename DataT, typename IdxT, typename accessor>
+void build_knn_graph(raft::device_resources const& res,
+                     mdspan<const DataT, matrix_extent<IdxT>, row_major, accessor> dataset,
+                     raft::host_matrix_view<IdxT, IdxT, row_major> knn_graph,
```
divyegala, line 76: Is this a sparse or dense data structure?

cpp/include/raft/neighbors/cagra.cuh
```diff
@@ -0,0 +1,223 @@
...
+                     const uint32_t refine_rate                         = 2,
```
divyegala, line 77: Should this also be an optional?

cpp/include/raft/neighbors/cagra_types.hpp
```diff
@@ -0,0 +1,197 @@
...
+        mdspan<IdxT, matrix_extent<IdxT>, row_major, graph_accessor> knn_graph)
+    : ann::index(),
+      metric_(metric),
+      dataset_(make_device_matrix<T, IdxT>(res, dataset.extent(0), dataset.extent(1))),
+      graph_(make_device_matrix<IdxT, IdxT>(res, knn_graph.extent(0), knn_graph.extent(1)))
+  {
+    RAFT_EXPECTS(dataset.extent(0) == knn_graph.extent(0),
+                 "Dataset and knn_graph must have equal number of rows");
+    raft::copy(dataset_.data_handle(), dataset.data_handle(), dataset.size(), res.get_stream());
+    raft::copy(graph_.data_handle(), knn_graph.data_handle(), knn_graph.size(), res.get_stream());
```
divyegala, line 185: Are these copies necessary if the dataset is already on device?

cpp/include/raft/util/cache_util.cuh
```diff
@@ -50,7 +50,7 @@ __global__ void get_vecs(
   if (tid < n_vec * n) {
     size_t out_col   = tid / n_vec;  // col idx
     size_t cache_col = cache_idx[out_col];
-    if (cache_idx[out_col] >= 0) {
+    if (!std::is_signed<idx_t>::value || cache_idx[out_col] >= 0) {
```
divyegala, line 53:
```suggestion
    if constexpr (!std::is_signed<idx_t>::value || cache_idx[out_col] >= 0) {
```

divyegala (commented):
No specializations for this version?

cpp/include/raft/neighbors/cagra_types.hpp
```diff
@@ -0,0 +1,197 @@
...
+  size_t num_parents = 1;
+  /** Lower limit of search iterations. */
+  size_t min_iterations = 0;
+
+  /** Bit length for reading the dataset vectors. 0, 64 or 128. Auto selection when 0. */
+  size_t load_bit_length = 0;
+  /** Thread block size. 0, 64, 128, 256, 512, 1024. Auto selection when 0. */
+  size_t thread_block_size = 0;
+  /** Hashmap type. "auto", "hash", or "small-hash". Auto selection when "auto". */
+  std::string hashmap_mode = "auto";
```
benfred, line 87: minor nitpick: can we make this an enum, instead of a string?

cpp/include/raft/neighbors/cagra.cuh
tfeher: A dense matrix that stores the neighbor indices, same number of neighbors for all points.

cpp/test/CMakeLists.txt
```diff
@@ -255,6 +255,7 @@ if(BUILD_TESTS)
     NAME
     NEIGHBORS_TEST
     PATH
+    test/neighbors/ann_cagra/test_float_uint32_t.cu
```
benfred, line 258: Do we need the other 3 files in `test/neighbors/ann_cagra` if we're only going to compile this one version?

benfred, line 258: do we need the other 3 files in `test/ann_cagra/` - if we're only compiling this one specialization?

cpp/include/raft/neighbors/detail/cagra/hashmap.hpp
```diff
@@ -0,0 +1,86 @@
+/*
```
benfred, line 1: Is it possible to use the hashmap from cuCollections for this https://github.com/NVIDIA/cuCollections - since we already have this as a dependency?

cpp/include/raft/neighbors/detail/cagra/bitonic.hpp
```diff
@@ -0,0 +1,226 @@
+/*
```
benfred, line 1: Is it possible to use the bitonic sort in `cpp/include/raft/util/bitonic_sort.cuh ` here?

benfred (commented):
I'm super excited about this change - but there is a *ton* of new code here, which makes this a bit hard to review.

Given that this is an experimental feature (and in an experimental namespace) - I think we can probably merge this in so other people can start trying this out.  @divyegala @cjnolet what do you think?

I left some minor comments on this below,

tfeher:
> No specializations for this version?

In an offline discussion @cjnolet recommended to leave out all the specializaitions. Even with limited float support the specializations would add 125 MiB (+20%) to `libraft.so`. Let's hope it won't timeout while compiling the test.

cpp/include/raft/neighbors/detail/cagra/topk_for_cagra/topk.h
```diff
@@ -0,0 +1,57 @@
...
+namespace raft::neighbors::experimental::cagra::detail {
+
+//
+size_t _cuann_find_topk_bufferSize(uint32_t topK,
+                                   uint32_t sizeBatch,
+                                   uint32_t numElements,
+                                   cudaDataType_t sampleDtype = CUDA_R_32F);
+
+//
+void _cuann_find_topk(uint32_t topK,
```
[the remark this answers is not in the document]
tfeher: Yes, I have missed that point earlier, just moved the code recently to `topk_for_cagra/topk_core.cuh`.

cpp/test/CMakeLists.txt
tfeher: I have removed them (for now).

tfeher (commented):
Thanks @divyegala and @benfred for the review! I have addressed the issues. 

Indeed there is a lot of code in the implementation details. That part will be improved in the follow up, to re-use already existing utilities from raft, improve structure and add developer docs.

cpp/include/raft/util/cache_util.cuh
tfeher: That does not compile because `cache_idx[out_col]` is not compile time constant.

benfred (approved):
thanks for the changes @tfeher!

cpp/include/raft/neighbors/cagra_types.hpp
divyegala: Yes, I think we can go about this a couple of ways:
1. Always copy, as is currently happening
2. Copy just the view, which would be like maintaining a reference to the underlying data
3. Construct the index on the fly if it's sufficiently cheap
4. Let the user pick which option to use from above. For example, if the user can guarantee that the data will stay in scope then I think they might go with 2

Could you please add this to your follow-on tasks as well? We can chat more later about how we want to go about this.

cpp/include/raft/util/cache_util.cuh
divyegala: Ah. You can write it as two if-conditions then, write? `if constexpr` as the outer if and then the inner condition which is a plain `if`. Anyway, this isn't blocking by any means. I will leave it up to you to decide what way you want to do it.

divyegala (approved):
Thanks @tfeher for this PR!

I agree with @benfred that we can merge this for now and do more iterations on reviews for implementation details as this feature makes it out of experimental. So for now, I just tried to review the public facing code.

cpp/include/raft/neighbors/cagra_types.hpp
tfeher: Thanks for the suggestions! I have added an item to to #1392

PR #1384, opened 2023-03-30
Fix IVF-PQ API to use `device_vector_view`

lowener:
This PR mainly intends to replace `device_matrix_view` for `ivf_pq::extend` to `device_vector_view`.
There are also a few updates to the documentation to reflect the current API.
The order of the arguments in the API is not touched.


viclafargue (approved):
Thanks, LGTM

cjnolet (approved):
LGTM. Thanks for fixing this @lowener! I think this is going to be a breaking change downstream, though passing explicit indices to `extend` is not an option that I would expect to be used a lot at the moment. I think we'll just need to be careful to verify the downstream projects we are aware of that are using these APIs, but we should be able to get this in 23.04.

cjnolet (approved):

cjnolet (approved):
LGTM!

ahendriksen:
Why are the docstrings of functions in `ivf_flat.cuh` changed in this PR? The function signatures themselves have not changed.

lowener:
The docstrings were not up to date with the order of the arguments in the function signatures.
So I updated the docstrings and kept the function signatures.

ahendriksen:
Ah my bad. I misread the function signatures. :+1:

PR #1386, opened 2023-03-30
Adding base header-only conda package without cuda math libs

cjnolet:
cc @MatthiasKohl  @bdice 

Making sure CI agrees w/ this change. @MatthiasKohl, if CI succeeds here let's try to plug the resulting conda packages into a cugraph-ops PR to make sure cugraph-ops CI is happy as well.


cjnolet:
cc @robertmaynard to make sure I did this properly. The idea is to make linking against the cuda math libraries optional for individual libraries after the `libraft-headers` conda package has been installed but pass them through `raft::raft` by default.

conda/recipes/libraft/meta.yaml
```diff
@@ -50,20 +50,25 @@ outputs:
...
-        - libcublas {{ libcublas_host_version }}
-        - libcublas-dev {{ libcublas_host_version }}
-        - libcurand {{ libcurand_host_version }}
-        - libcurand-dev {{ libcurand_host_version }}
-        - libcusolver {{ libcusolver_host_version }}
-        - libcusolver-dev {{ libcusolver_host_version }}
-        - libcusparse {{ libcusparse_host_version }}
-        - libcusparse-dev {{ libcusparse_host_version }}
         - librmm ={{ minor_version }}
+        - cudatoolkit {{ cuda_spec }}
```
bdice, line 54: We want to tightly constrain host requirements to ensure that we build with CUDA 11.8 (the `cuda_version` used at build time across all of RAPIDS). We use `cuda_spec` for run dependencies where any compatible 11.x version will suffice.
```suggestion
        - cudatoolkit {{ cuda_version }}
```

conda/recipes/libraft/meta.yaml
```diff
@@ -50,20 +50,25 @@ outputs:
...
+    about:
+      home: https://rapids.ai/
+      license: Apache-2.0
+      summary: libraft-headers-impl library
+  - name: libraft-headers
+    version: {{ version }}
+    build:
+      number: {{ GIT_DESCRIBE_NUMBER }}
+      string: cuda{{ cuda_major }}_{{ date_string }}_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
+      ignore_run_exports_from:
```
bdice, line 64: These ignores might be irrelevant because there’s no build or host sections for this output. I think you can delete them.

conda/recipes/libraft/meta.yaml
```diff
@@ -72,7 +77,7 @@ outputs:
         - libcusolver-dev {{ libcusolver_run_version }}
         - libcusparse {{ libcusparse_run_version }}
         - libcusparse-dev {{ libcusparse_run_version }}
-        - librmm ={{ minor_version }}
+
```
bdice, line 80: If we have blank lines in this file, they should separate the outputs and not be placed within a single output between its keys. This about section goes with libraft-headers, above, but it’s visually placed next to libraft below.

conda/recipes/libraft/meta.yaml
```diff
@@ -99,12 +104,6 @@ outputs:
         - cuda-profiler-api {{ cuda_profiler_api_host_version }}
         - libcublas {{ libcublas_host_version }}
         - libcublas-dev {{ libcublas_host_version }}
-        - libcurand {{ libcurand_host_version }}
```
bdice, line 102: To confirm, the compiled parts of the library don’t use libcurand, libcusolver, or libcusparse?

Note that because this package has a dependency on libraft-headers, the build will pull those math library packages anyway because they are run dependencies of libraft-headers — but they will be loosely pinned at 11.x “run versions” at build time rather than 11.8 “host versions”. It might make sense to keep these definitions so that you have tighter build-time pinnings, since the libs have to be fetched anyway.

conda/recipes/libraft/meta.yaml
```diff
@@ -170,19 +167,10 @@ outputs:
         - sysroot_{{ target_platform }} {{ sysroot_version }}
```
bdice, line 167: Unrelated comment: libcudf has a subpackage called libcudf-example. It does something similar to libraft-template. I think that the name “example” makes more sense than “template,” especially when template can be an overloaded word for header-only libraries (which feature C++ templates).

cpp/CMakeLists.txt
```diff
@@ -637,19 +600,47 @@ if(compiled IN_LIST raft_FIND_COMPONENTS)
...
+  code_string
+  [=[
+ option(RAFT_ENABLE_CTK_MATH_DEPENDENCIES "Enable CUDA math library dependencies (e.g. cublas, cusolver, cusparse, curand)" ON)
+mark_as_advanced(RAFT_ENABLE_CTK_MATH_DEPENDENCIES)
+message("RAFT: CUDA Toolkit Math dependencies enabled ${RAFT_ENABLE_CTK_MATH_DEPENDENCIES}")
+target_link_libraries(raft::raft INTERFACE $<$<BOOL:${RAFT_ENABLE_CTK_MATH_DEPENDENCIES}>:${RAFT_CTK_MATH_DEPENDENCIES}>)
+]=]
+)
 
 # Use `rapids_export` for 22.04 as it will have COMPONENT support
```
bdice, line 621: Comment seems outdated?
```suggestion
```

cpp/include/raft/linalg/detail/cublas_wrappers.hpp
```diff
@@ -965,7 +965,10 @@ inline cublasStatus_t cublasdot(cublasHandle_t handle,
                                 cudaStream_t stream)
 {
   RAFT_CUBLAS_TRY(cublasSetStream(handle, stream));
-  return cublasSdot(handle, n, x, incx, y, incy, result);
+  return cublasDotEx(
+    handle, n, x, CUDA_R_32F, incx, y, CUDA_R_32F, incy, result, CUDA_R_32F, CUDA_R_32F);
+
+  //  return cublasSdot(handle, n, x, incx, y, incy, result);
```
bdice, line 971: This commented line might be removable? Same below.

docs/source/build.md
```diff
@@ -156,6 +156,7 @@ RAFT's cmake has the following configurable flags available:.
 | BUILD_TESTS               | ON, OFF              | ON | Compile Googletests |
 | BUILD_BENCH               | ON, OFF              | OFF | Compile benchmarks |
 | raft_FIND_COMPONENTS      | compiled distributed | | Configures the optional components as a space-separated list |
+| RAFT_ENABLE_CTK_MATH_DEPENDENCIES | ON, OFF | ON | Link against CUDA Math Libraries in `raft::raft` |
```
bdice, line 159: We might want to note something like: “math libraries are required for (these headers/features) but optional for consumers of (these headers/features).”

bdice (commented):
Mostly commented on conda recipes. The CMake looks alright from my knowledge but I would like others to verify.

docs/source/build.md
cjnolet: Thanks for the suggestion here @bdice. We've tried that approach in the past and it made the headers (and the core library) very hard to consume. Most users don't (and shouldn't) care about these dependencies, which is why we're making this an "opt-out" feature. To try and list out all of the namespaces/headers which do and don't use these libraries would be an arduous task and nearly impossible to maintain.

bdice: That is totally reasonable!

MatthiasKohl:
@cjnolet looks good from my side, thank you !
I'm still unsure whether we can remove the "run" dependency of libraft-headers from cugraph-ops, since this was added by the cugraph team (cugraph-ops being a module necessary to build/run cugraph). Similarly, I would argue that cugraph-ops should not have libraft-headers as a "run" dependency, but the cugraph team will have to decide, since they may have to update their CMake files as well.

MatthiasKohl:
@cjnolet so IIUC, we will be able to make the `impl` package a host dependency on cugraph-ops, and remove the run dependency, right ?
This way, when building cugraph-ops, we should not download anything in cmake, but end users of cugraph-ops also don't need to download the additional dependencies from libraft-headers.
That would be ideal, thank you very much !

cjnolet:
> @cjnolet so IIUC, we will be able to make the impl package a host dependency on cugraph-ops, and remove the run dependency, right ?

You might be able to use `EXCLUDE_FROM_ALL` in `get_raft.cmake` to accomplish that, but it assumes that an end-user will not need raft at all in order to use `cugraph-ops` (e.g. you have not exposed any raft includes or apis on your own public api headers).

MatthiasKohl:
> You might be able to use EXCLUDE_FROM_ALL in get_raft.cmake to accomplish that, but it assumes that an end-user will not need raft at all in order to use cugraph-ops (e.g. you have not exposed any raft includes or apis on your own public api headers).

You're right, we do have public APIs with raft structures, so we'd need at least the impl package as a run dependency.

cjnolet:
@MatthiasKohl we have also deprecated our pointer-based APIs, which I notice you are using for rng. I'm going to prioritize changing over the rng APIs first so that they all accept raft::resources objects instead of raft::device_resources. raft::resources is a completely agnostic key/value store for resources (like cublas handles) and doesn't require any of the math libs to use. This way we can update cugraph-ops to use the  ww rng apis without polluting them with unecessary dependencies.

cpp/CMakeLists.txt
```diff
@@ -637,19 +604,59 @@ if(compiled IN_LIST raft_FIND_COMPONENTS)
 endif()
 ]=]
 )
+set(raft_components compiled distributed)
```
vyasr, line 610: I would recommend that we get #1154 merged first since it should simplify the component management in this PR.

cjnolet:
Making a note here that until all the APIs that are being compiled into `libraft` are using `raft::resources` instead of `raft::device_resources`, we will need to depend on all of the math libs.

cpp/CMakeLists.txt
```diff
@@ -457,7 +456,10 @@ if(RAFT_COMPILE_LIBRARY)
                INTERFACE_POSITION_INDEPENDENT_CODE ON
   )
 
-  target_link_libraries(raft_lib PUBLIC raft::raft $<TARGET_NAME_IF_EXISTS:OpenMP::OpenMP_CXX>)
+  target_link_libraries(
+    raft_lib PUBLIC raft::raft ${RAFT_CUBLAS_DEPENDENCY}
```
vyasr: Why does the raft lib require cublas but not the others? Why are the others only specified as interface dependencies below?

vyasr (commented):
>  It sounds like the request is to have a header package that supplies none of its dependencies if those can be acquired via CPM and/or via the system.

This doesn't quite align with my understanding, so could use some clarification. It's not so much that we want to avoid providing anything that can be acquired via CPM, it's that that we don't want to enforce any dependency requirements that are only needed if certain raft features are used since another package could leverage a limited subset of raft that doesn't have those requirements. Is that right?

> RAFT expects rmm and cudatoolkit at a bare minimum (though rmm will end up being downloaded automatically w/ cpm if they haven't been installed already), so, maybe we could remove rmm at least? The reality is that anyone building w/ the libraft headers should be using cmake, which is going to make sure these dependencies are there (everything other than cudatoolkit anyways), else it'll download them automatically. What are your thoughts here? Please guide...

Is there any way to use raft without rmm? If not, I would include rmm as a hard dependency. AIUI there is no benefit to removing a conda package dependency only to have it always be downloaded. The benefit is not including math libraries as a hard dependency if a user of raft might not actually need them.

cpp/CMakeLists.txt
cjnolet: @vyasr that's indeed the plan. I've merged that branch into this PR in the meantime because these changes won't work as intended otherwise.

cpp/CMakeLists.txt
cjnolet: Ideally we would only be linking against dependencies which are actually needed for each of the targets- cublas being the only dependency which is explicitly needed for the APIs compiled into `libraft`. The problem is that until we implement #1394, we have to depend on all the math libraries because `device_resources` (implicitly) requires all of the handles be available.

cjnolet:
@vyasr JFYI- @bdice and I did a huddle yesterday to iron out many of his initial comments (and to improve both of our understand- his from the cmake side and mine from the conda recipe / cudatoolkit packaging side).

MatthiasKohl:
> @MatthiasKohl we have also deprecated our pointer-based APIs, which I notice you are using for rng. I'm going to prioritize changing over the rng APIs first so that they all accept raft::resources objects instead of raft::device_resources. raft::resources is a completely agnostic key/value store for resources (like cublas handles) and doesn't require any of the math libs to use. This way we can update cugraph-ops to use the new rng apis without polluting them with unecessary dependencies.

Thanks, that makes sense, we should be able to do this during 23.06 without issues.

robertmaynard:
The whole discussion here and the requirements people are bringing up make me think we should offer `dev` packages. Where the `dev` provides all the required bits to compile against, and the base package offers the smallest set.

ajschmidt8 (approved):

cjnolet:
> The whole discussion here and the requirements people are bringing up make me think we should offer dev packages. Where the dev provides all the required bits to compile against, and the base package offers the smallest set.

@robertmaynard I think what you suggest here is a good state for us to get to (with the exception of rmm and cudatoolkit). I'm trying to minimize potential breakage downstream in the meantime with these changes so I've opted to keep the `libraft-headers` name for now. Given that any other users are going to need the ctk runtime and rmm at the very minimum, I don't foresee a problem w/ keeping these in both the cmake and conda.

divyegala (approved):

PR #1389, opened 2023-03-31
Fix primitives benchmarks

ahendriksen:
Due to some typos in https://github.com/rapidsai/raft/pull/1304, the benchmarks are not built any more. 

I have fixed this in this PR. I think it is important that this PR goes into 23.04, otherwise we will not have benchmarks for two months in the stable release.


cjnolet:
Thanks for fixing this @ahendriksen. I'm not sure what happened in the ann benchmarks PR but I had originally fixed this in the ann benchmarks PR (see [this file](https://github.com/cjnolet/raft/blob/122851983e78dc2d7d473a947172ddc64235aa4f/cpp/bench/prims/CMakeLists.txt)) and it looks like it was reverted. I think there was also a bad merge downstream at some point because the `bench/CMakeLists.txt` seems to have ended up back in the codebase.

ahendriksen:
> I'm not sure what happened

No problem. Happens to me all the time as well. Merge commits can be quite huge.. 

> Can you (1) merge the changes from bench/CMakeLists.txt in this PR as well and (2) remove that file please?

To (1): It looks like the changes from `bench/CmakeLists.txt` were properly merged into `bench/prims/CmakeLists.txt` (as in: the diff shows that the only difference is `/prims/` added in paths). 

To (2): This PR already deletes `bench/CmakeLists.txt`.

cjnolet:
> Will fix, but the tests should have caught this in the previous PR already right?

Unfortunately, I think those macros were removed after the file I linked above was reverted, which would explain why they weren't erroring in CI prior.

ahendriksen:
> Unfortunately, I think those macros were removed after the file I linked above was reverted, which would explain why they weren't erroring in CI prior. For some background- these macros were deprecated over a year ago now so they shouldn't have been used in the first place.

I overlooked. The compilation error was actually in a bench. In any case, it is fixed :)

cjnolet (approved):
LGTM. Thanks for fixing this!

PR #1412, opened 2023-04-13
Update clang-format to 16.0.1.

bdice:
This PR updates the clang-format version used by pre-commit.


bdice:
@cjnolet I think this just needs labels and should be good to merge. (Merging sooner means less likelihood of conflicts.)

raydouglass (approved):

cpp/scripts/run-clang-format.py
bdice, line 1: This script is not used by pre-commit any more. May we remove this file in favor of pre-commit's own logic for running clang-format? It would only matter if developers still want to use this script locally instead of running pre-commit locally (which I recommend, since it matches the CI outputs).

benfred: I'm good with removing this script - but we should also update the references in `docs/source/developer_guide.md` to mention pre-commit instead

cjnolet (approved):

cpp/scripts/run-clang-format.py
bdice: Updated in 83f1ab8 / c321744. I copied from the cuDF Contributing Guide. Does this look alright to you?

benfred: looks good - thanks!

benfred (approved):

PR #1418, opened 2023-04-14
Fix for function exposing KNN merge

viclafargue:


cjnolet (approved):
LGTM. Is this just popping up because Legate is enforcing const inputs or is this exposing a lack of testing on the RAFT side? I notice CI is tripping on this change but is this being tested?

viclafargue:
Yes, there were multiple issues during compilation. It doesn't appear to be tested anywhere. Do you want me to add some tests?

PR #1419, opened 2023-04-14
Fix is_min_close

benfred:
Correlation and Cosine distance both return (1 - similarity) in the pairwise distances apis, meaning that is_min_close is returning the wrong sort order for them. Fix.


cjnolet (approved):
LGTM!

PR #1422, opened 2023-04-15
Add python bindings for matrix::select_k

benfred:


python/pylibraft/pylibraft/matrix/select_k.pyx
```diff
@@ -0,0 +1,133 @@
...
+
+    >>> queries = cp.random.random_sample((n_rows, n_features),
+    ...                                   dtype=cp.float32)
+    >>> k = 40
+    >>> distances, ids = select_k(queries, k)
+    >>> distances = cp.asarray(distances)
+    >>> ids = cp.asarray(ids)
+    """
+
+    dataset_cai = cai_wrapper(dataset)
```
cjnolet, line 97: Wow- this whole function is quite elegant. For cython, anyways... but we've managed to isolate a lot of the boilerplate when working w/ mdspan/optional. It's great!

python/pylibraft/pylibraft/test/test_select_k.py
```diff
@@ -0,0 +1,54 @@
...
+import numpy as np
+import pytest
+
+from pylibraft.common import device_ndarray
+from pylibraft.matrix import select_k
+
+
+@pytest.mark.parametrize("n_rows", [32, 100])
+@pytest.mark.parametrize("n_cols", [40, 100])
+@pytest.mark.parametrize("k", [1, 5, 16])
```
cjnolet, line 25: Can we try a larger k here (some odd value like 35) or was that tripping the order of the tests?

cjnolet (approved):
LGTM! Really nice seeing all the hard work on the common pylibraft boilerplate paying off here. Even better seeing expansion of the Python APIs.

python/pylibraft/pylibraft/matrix/select_k.pyx
benfred: I think so too!  I like how we're getting to the point where the cython code is just a thin wrapper over c++. With the CAI wrappers, the `get_dmv_float` code, the decorators to convert the output etc - I think its really getting quite nice here

PR #1431, opened 2023-04-18
Use nvtx3 includes.

bdice:
This PR updates raft to use `#include <nvtx3/nvToolsExt.h>` instead of `#include <nvToolsExt.h>`. This ensures we fetch the header-only NVTX v3. See NVTX docs for more information: https://nvidia.github.io/NVTX/#c-and-c


benfred (approved):
lgtm - thanks!

Issue #1416, opened 2023-04-14
[DISCUSSION] A robust solution to precompiling function templates


ahendriksen:
This issue describes problems we currently have with using specialization headers to reduce compile times and proposes a fix.

This issue accompanies https://github.com/rapidsai/raft/pull/1415

I have added this issue to motivate the current design of the PR and solicit feedback on certain points.

# Problem

RAFT is a heavily templated library. Several core functions are very expensive to compile and we want to prevent duplicate compilation of this functionality. To limit build time, RAFT provides a precompiled library (`libraft.so`) where expensive function templates are instantiated for the most commonly used template parameters. This is not a complete solution, however, due to two problems (1) accidental reinstantiations and (2) unnecessary dependencies:

1. In practice, it often happens that a function template is **accidentally reinstantiated**, even when it is already provided by the library. This can have a big negative impact on compile times. It happens in consuming libraries, like cuML, but is also prevalent **within `libraft.so` itself** [[#1358](<https://github.com/rapidsai/raft/pull/1358>), [#1360](<https://github.com/rapidsai/raft/pull/1360>)].

2. A minor change in the core implementation of a function often causes the **recompilation of dependent** translation units. A particularly bad example is `neighbors/specializations.cuh`, which provides the extern template instantiations for many expensive kernels and is also included over 100 times in the RAFT project. Any change to any of these kernels triggers a rebuild of all 100 files that include the specializations header.

# Requirements

A solution to these two problems should satisfy the following requirements:

1. An accidental reinstantiation must be **impossible** when compiling `libraft.so`, i.e., it must result in an error.

2. In case of an error, the solution must result in useful and actionable error messages. Specifically, it should not introduce linker errors during development (they only fire at the very end of compilation and do not point to the source location that introduced the error).

3. It must be possible for downstream libraries to continue to use RAFT as a header-only library, as well as a precompiled library.

In addition, there are some nice-to-haves. That is, we obviously want to write as little boilerplate code as possible. Also, it would be nice if downstream code would not have to bother with the following if else preprocessor logic when including a header:
``` c++
    #if defined RAFT_COMPILED
    #include <raft_runtime/distance/pairwise_distance.hpp>
    #else
    #include <raft/distance/distance.cuh>
    #endif
```

# Background: C++ templates

I include some background on C++ function templates here, feel free to skip and refer back in case of questions.

[...]

Issue #1372, opened 2023-03-24
[ENH] Add specialization for coalescedReductionThinKernel


ahendriksen:
**Is your feature request related to a problem? Please describe.**
The `raft::linalg::detail::coalescedReductionThinKernel` is instantiated 2858 times in 84 objects that are inputs to libraft.so. 
Source: `python rapids-template-instantiation-reporter.py --build-dir /path/to/raft/cpp/build/  libraft.so` as described in https://github.com/rapidsai/build-metrics-reporter/pull/2

**Describe the solution you'd like**
It might make sense to add a specialization for the most commonly used instances. Using 
```
cuobjdump -res-usage /path/to/libraft.so | grep Function | cu++filt | sort -h | grep coalescedReductionThinKernel | uniq -c | sort -h
```
 we get the following kernels with instances in more than 80 objects:
```
     82  Function void raft::linalg::detail::coalescedReductionThinKernel<raft::linalg::detail::ReductionThinPolicy<(int)16, (int)8>, float, float, unsigned long, raft::abs_op, raft::add_op, raft::sqrt_op>(T3 *, const T2 *, T4, T4, T3, T5, T6, T7, bool):
     82  Function void raft::linalg::detail::coalescedReductionThinKernel<raft::linalg::detail::ReductionThinPolicy<(int)16, (int)8>, float, float, unsigned long, raft::abs_op, raft::max_op, raft::sqrt_op>(T3 *, const T2 *, T4, T4, T3, T5, T6, T7, bool):
     82  Function void raft::linalg::detail::coalescedReductionThinKernel<raft::linalg::detail::ReductionThinPolicy<(int)16, (int)8>, float, float, unsigned long, raft::identity_op, raft::add_op, raft::identity_op>(T3 *, const T2 *, T4, T4, T3, T5, T6, T7, boo>
     82  Function void raft::linalg::detail::coalescedReductionThinKernel<raft::linalg::detail::ReductionThinPolicy<(int)16, (int)8>, float, float, unsigned long, raft::sq_op, raft::add_op, raft::sqrt_op>(T3 *, const T2 *, T4, T4, T3, T5, T6, T7, bool):
     82  Function void raft::linalg::detail::coalescedReductionThinKernel<raft::linalg::detail::ReductionThinPolicy<(int)2, (int)64>, float, float, unsigned long, raft::abs_op, raft::add_op, raft::sqrt_op>(T3 *, const T2 *, T4, T4, T3, T5, T6, T7, bool):
[...]

PR #1437, opened 2023-04-20
[ENH] [1/5] Header structure: replace specializations

ahendriksen:
This PR replaces the specialization headers with a split header structure. It is a more reviewable version of #1415.

This is achieved by following the strategy outlined in issue https://github.com/rapidsai/raft/issues/1416.

The following steps are performed:

**Preparatory**
- **Move** files around. 
  - [d9801e8](https://github.com/rapidsai/raft/pull/1437/commits/d9801e87f266554c2c561490faf0fe5d76e66165) Move header files with expensive function templates from `header.cuh` to `header-inl.cuh` to preserve history and minimize the diff in later comits.
  - [50b374d](https://github.com/rapidsai/raft/pull/1437/commits/50b374da5c579d7d160c15ebc0d2b1db105798eb) Move raft_runtime source files into separate directory. They would otherwise clash with source files added in later commits. 
- **Fix** existing issues in RAFT unearthed by this PR:
  - [8974ae3](https://github.com/rapidsai/raft/pull/1437/commits/8974ae32255e702de9d707da34b353b61bedbc23) Fix several locations where files were depending on transitive includes
  - [95ef31b](https://github.com/rapidsai/raft/pull/1437/commits/95ef31bc974b6b2b4739cedd6d4d287e9e0266be) Fix distance::getWorkspaceSize. It did not compile previously.
- **Prepare** for changes that are coming:
  - [7edcb6a](https://github.com/rapidsai/raft/pull/1437/commits/7edcb6a59da62216d9521e15c3848ba676588f31) Create a functor for `rbf_fin_op` that can be used to instantiate templates later on.
  - [71de7bd](https://github.com/rapidsai/raft/pull/1437/commits/71de7bd4e33e788c17bf7c27893d764e3c410b06) Similar move for distance functors used in `registers.cuh`.

**Meat of PR**: Use header structure as outlined in the [new documentation](https://github.com/ahendriksen/raft/blob/enh-inl-headers-slow/docs/source/developer_guide.md#header-organization-of-expensive-function-templates). 
- [Change RAFT_COMPILED from INTERFACE to PUBLIC](https://github.com/rapidsai/raft/pull/1437/commits/541cabc385d193c908bd2faf39e6a9ce3c58dfc1)
- [Define RAFT_EXPLICIT and RAFT_EXPLICIT_INSTANTIATE_ONLY](https://github.com/rapidsai/raft/pull/1437/commits/d81b14ee5e25ac78e042671323765659f5f693af)
- [Update docs](https://github.com/rapidsai/raft/pull/1437/commits/48ea76957413b9f3238fbac2ce456fde893c60e4)
- [Replace specializations by split headers](https://github.com/rapidsai/raft/pull/1437/commits/e6bb5d5b957df2cf2b053be52b8359567374cd94). Big diff:
  - For each `header-inl.cuh`, add `header.cuh` and `header-ext.cuh`
  - Remove specialization source files. 
  - Add back source files with instances for each `-ext` header.
  - The only specializations that do not have a corresponding new instance are those in `src/distance/specializations/detail/kernels`. They were in fact not used in the tests, one did not compile, and adding instances would have caused conflicts with PR #1296. This was decided with @tfeher  [here](https://github.com/rapidsai/raft/pull/1415#discussion_r1169302147).
- [Deprecate specialization headers](https://github.com/rapidsai/raft/pull/1437/commits/ff79abf0860321985cee7228bef0afadaf4f3d5d): we do not remove the headers, but just add a `#pragma message`.

**Fixup**: Add some more template instances, remove unnecessary functions, fix a test. 
- [c9e7413](https://github.com/rapidsai/raft/pull/1437/commits/c9e741342b9382a60fe78fc1a9bc11c00c15cb09) Add instances for ivf_flat_interleaved_scan. 
- [Separate fused_l2_nn_helpers](https://github.com/rapidsai/raft/pull/1437/commits/0c889dc64a676287f43db60b884e26d882a2a278): This enables this PR to be **non-breaking**, as compared to PR #1415 . 
- [Remove pairwise_matrix_instantiation_point](https://github.com/rapidsai/raft/pull/1437/commits/f97b2a834a5cdc67894a2488c6850fd9cd09d2df): it is not used anymore. 
- [Rename specialization => instantiation](https://github.com/rapidsai/raft/pull/1437/commits/fb637f7a3cd0d78941f38f3afd67653ff2949b6c): Make usage of `instantiation` in code comments consistent with docs.
- [test/neighbors/selection.cu: Expose kFaissMaxK](https://github.com/rapidsai/raft/pull/1437/commits/7b065aff81a0b1976e2a9e2f3de6690361a1111b) Fix test for `selection.cu`. If it were built with `RAFT_EXPLICIT_INSTANTIATE_ONLY`, then `kFaissMax` would not be defined. By splitting it out into a separate header, it can be used regardless.


ahendriksen:
Review tip: I have moved the headers by adding the `-inl` suffix in the first commit. The diff is significantly smaller if you start from there.

cpp/src/distance/specializations/detail/canberra_double_double_double_int.cu
```diff
@@ -1,33 +0,0 @@
-/*
- * Copyright (c) 2021-2023, NVIDIA CORPORATION.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
```
cjnolet, line 6: Users downstream are still including these files in their code. Instead of removing all of these files, let's just remove their contents and add a deprecation warning that the files will be removed in 23.08 (or 23.10). The improvements you are making are already pretty massive, but I want to eliminate removing any files that are going to require synchronously updating all of the users before they get merged. In addition, we now have users outside of RAPIDS and our of courtesy we should be giving them time to update their code.

cpp/include/raft/neighbors/brute_force-ext.cuh
```diff
@@ -0,0 +1,109 @@
...
+      std::vector<raft::device_matrix_view<const value_t, matrix_idx, index_layout>> index, \
+      raft::device_matrix_view<const value_t, matrix_idx, search_layout> search,            \
+      raft::device_matrix_view<idx_t, matrix_idx, row_major> indices,                       \
+      raft::device_matrix_view<value_t, matrix_idx, row_major> distances,                   \
+      raft::distance::DistanceType metric,                                                  \
+      std::optional<float> metric_arg,                                                      \
+      std::optional<idx_t> global_id_offset,                                                \
+      epilogue_op distance_epilogue);
+
+instantiate_raft_neighbors_brute_force_knn(
```
ahendriksen, line 83: Related issue about these instances: https://github.com/rapidsai/raft/issues/1449

tfeher:
> Review tip: I have moved the headers by adding the -inl suffix in the first commit. The diff is significantly smaller if you start from there.

Indeed filtering the commits help tremendously. We have large mechanical changes in, the following commits:
- first commit: [MV: add -inl suffix to header paths](https://github.com/rapidsai/raft/pull/1437/commits/d9801e87f266554c2c561490faf0fe5d76e66165)
- 10th commit: [Replace specializations by split headers](https://github.com/rapidsai/raft/pull/1437/commits/e6bb5d5b957df2cf2b053be52b8359567374cd94)
- 11th commit [Deprecate specialization headers](https://github.com/rapidsai/raft/pull/1437/commits/ff79abf0860321985cee7228bef0afadaf4f3d5d)

It make sense to review these separately. The rest (commits 2-9, 12-16) can be then lumped together during review.

tfeher (approved):
Thanks Allard for breaking up the oiriginal PR (#1415) into multiple pieces! Also thanks for the detailed description of the changes in the PR description. The questions I had for #1415 are already answered here (or a tracking issue is created). I have one question left, but that could be answered in a follow up PR.

Overall the PR looks good to me, and I am excited about the compile time improvements that we can get once we go through all the steps outlined here and in he follow up PRs.

cpp/include/raft/distance/distance-ext.cuh
```diff
@@ -0,0 +1,1065 @@
...
+  raft::distance::DistanceType::RusselRaoExpanded, double, double, double, int);
+
+#undef instantiate_raft_distance_getWorkspaceSize
+
+#define instantiate_raft_distance_getWorkspaceSize(DistT, DataT, AccT, OutT, IdxT, layout)         \
+  extern template size_t raft::distance::getWorkspaceSize<DistT, DataT, AccT, OutT, IdxT, layout>( \
+    raft::device_matrix_view<DataT, IdxT, layout> const& x,                                        \
+    raft::device_matrix_view<DataT, IdxT, layout> const& y)
+
+// We could consider not taking template parameters for this function. The
```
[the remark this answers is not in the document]
ahendriksen: That is a good question. getWorkspaceSize is not an expensive template, so that's not the reason it is instantiated. The problem is that it is defined in the same header as the `distance` function template. To not instantiate it, we can do two things:
1. Move getworkspaceSize to a `distance_helpers.cuh` header and include it in `distance-ext.cuh` and `distance-inl.cuh`. This is what is done for the `fusedL2NN` `initialize` function.
2. Instead of defining `getWorkspaceSize` as `RAFT_EXPLICIT`, copy the definition from the `distance-inl.cuh` header. I think this is a disaster waiting to happen, as it will require a lot of attentiveness to make sure the two definitions do not diverge.

I believe the current concensus is to deprecate getWorkspaceSize anyway, so the long list of instances should be only a temporary issue.

benfred (commented):
These changes look great to me - I'm really looking forward to having this series of PR's merged, I think the improvements to build times will make RAFT development much easier.

I'd like to see the corresponding downstream PR's pass CI before getting this merged (like https://github.com/rapidsai/cuml/pull/5363) - do you know why its failing with cuda 11.2/11.4 but working with 11.8? The failure looks like a missing symbol w/ the RusselRao distance 

```python
In [3]: import cpp_demangle

In [4]: cpp_demangle.demangle("_ZN4raft8distance8distanceILNS0_12DistanceTypeE18EfffiEEvRKNS_9resourcesEPKT0_S8_PT2_T3_SB_SB_bS6_")
Out[4]: 'void raft::distance::distance<(raft::distance::DistanceType)18, float, float, float, int>(raft::resources const&, float const*, float const*, float*, int, int, int, bool, float)'
```

cjnolet:
> do you know why its failing with cuda 11.2/11.4 but working with 11.8? 

@vyasr  actually fixed this one today. It turns out some recent changes to the conda packages were causing the cuda toolkit to get pinned to a specific version (11.8) and it was causing cuml's CI to pull in an old version of raft packages for other cuda versions. 

Cuml's CI *should* run fine now.

benfred (approved):

tfeher:
Some of the headers was split under the `detail` folder, like (like `select_k.cuh`) while others were above the `detail` folder (public level).  @ahendriksen was it done this way to minimize the changes, or was there any other motivation?

The public API did not change in any of the cases, so functionally the current status is fine. But I think it is nicer if we split the files under the detail folder when possible. I would prefer if we do not hold up this PR for such cosmetic changes, but we should consider returning to this question.

ahendriksen:
>Some of the headers was split under the detail folder, like (like select_k.cuh) while others were above the detail folder (public level). @ahendriksen was it done this way to minimize the changes, or was there any other motivation?

Splitting at detail rather than at public level was done for three reasons: 
- The detail header was used internally: for instance the ivf_flat_interleaved_scan_kernel is called in the ivf_flat header as well as the refine header.
- Kind of accidentally: the coalesced reduction kernel was [duplicated many many times](https://github.com/rapidsai/raft/issues/1372) so I added an instance. In hindsight, it might be better to instantiate at a higher level (for instance `linalg::reduce`, `linalg::norm`, etc). To tease apart which `raft::linalg`  kernels are worth instantiating and how could be investigated in the future. The coalesced reduction instances show how it could be done.
- As part of design: the `raft::distance` kernels were optimized for compile time previously. Compiling the kernels without the public header includes can be substantially faster.



> But I think it is nicer if we split the files under the detail folder when possible. 

I think we should favor splitting at the public level. This is beneficial for two reasons: (1) to document which instances exist and (2) to allow downstream libraries to instantiate templates themselves. Instancing at public level allows downstream libraries to opt in to `RAFT_EXPLICIT_INSTANTIATE_ONLY`: when they use a template that has not been instantiated in `libraft.so`, they can instantiate it themselves. Here, it is better that these templates are part of the public API so they don't break as often.

tfeher:
> I think we should favor splitting at the public level. This is beneficial for two reasons: (1) to document which instances exist  ,,,,

Good points!

divyegala (approved):

cpp/include/raft/neighbors/detail/ivf_flat_search-inl.cuh
```diff
@@ -0,0 +1,1285 @@
...
+    }
+    // NB: this is the limitation of the warpsort structures that use a huge number of
+    //     registers (used in the main kernel here).
+    RAFT_EXPECTS(capacity == Capacity,
+                 "Capacity must be power-of-two not bigger than the maximum allowed size "
+                 "matrix::detail::select::warpsort::kMaxCapacity (%d).",
+                 matrix::detail::select::warpsort::kMaxCapacity);
+    RAFT_EXPECTS(
+      veclen == Veclen,
+      "Veclen must be power-of-two not bigger than the maximum allowed size for this data type.");
```
cjnolet, line 1021: Wow, we need to start breaking headers like this up. Maybe separating out the `kernels` into their own headers.

docs/source/developer_guide.md
```diff
@@ -260,6 +260,97 @@ Sometimes, we need to temporarily change the log pattern (eg: for reporting deci
...
+
+The `RAFT_EXPLICIT_INSTANTIATE_ONLY` macro is defined by `CMake` during compilation of `libraft.so` itself. When defined, it indicates that implicit instantiations of expensive function templates are forbidden (they result in a compiler error). In the RAFT project, we additionally define this macro during compilation of the tests and benchmarks. 
+
+Below, we summarize which combinations of `RAFT_COMPILED` and `RAFT_EXPLICIT_INSTANTIATE_ONLY` are used in practice and what the effect of the combination is. 
+
+| RAFT_COMPILED | RAFT_EXPLICIT_INSTANTIATE_ONLY | Which targets                                                                                        |
+|---------------|--------------------------------|------------------------------------------------------------------------------------------------------|
+| defined       | defined                        | `raft::compiled`, RAFT tests, RAFT benchmarks                                                        |
+| defined       |                                | Downstream libraries depending of `libraft` like cuML, cuGraph.                                      |
```
cjnolet, line 276:
```suggestion
| defined       |                                | Downstream libraries depending on `libraft` like cuML, cuGraph.                                      |
```

docs/source/developer_guide.md
```diff
@@ -260,6 +260,97 @@ Sometimes, we need to temporarily change the log pattern (eg: for reporting deci
...
+``` c++
+#include <raft/expensive-inl.cuh>
+
+template void raft::expensive<int>(int);
+template void raft::expensive<float>(float);
+```
+
+**Design considerations**: 
+
+1. In the `-ext.cuh` header, do not include implementation headers. Only include function parameter types and types that are used to instantiate the templates. If a primitive takes custom parameter types, define them in a separate header called `<primitive_name>_types.hpp`.
```
cjnolet, line 344:
```suggestion
1. In the `-ext.cuh` header, do not include implementation headers. Only include function parameter types and types that are used to instantiate the templates. If a primitive takes custom parameter types, define them in a separate header called `<primitive_name>_types.hpp` (see [Common Design Considerations](https://github.com/rapidsai/raft/blob/7b065aff81a0b1976e2a9e2f3de6690361a1111b/docs/source/developer_guide.md#common-design-considerations)).  
```

docs/source/developer_guide.md
```diff
@@ -260,6 +260,97 @@ Sometimes, we need to temporarily change the log pattern (eg: for reporting deci
...
+
+template void raft::expensive<int>(int);
+template void raft::expensive<float>(float);
+```
+
+**Design considerations**: 
+
+1. In the `-ext.cuh` header, do not include implementation headers. Only include function parameter types and types that are used to instantiate the templates. If a primitive takes custom parameter types, define them in a separate header called `<primitive_name>_types.hpp`. 
+
+2. Keep docstrings in the `-inl.cuh` header, as it is closer to the code. Remove docstrings from template definitions in the `-ext.cuh` header.
```
cjnolet, line 346:
```suggestion
2. Keep docstrings in the `-inl.cuh` header, as it is closer to the code. Remove docstrings from template definitions in the `-ext.cuh` header. Make sure to explicitly include public APIs in the RAFT API docs (e.g. that they point the user to the `expensive.cuh` header and not `expensive-inl.cuh`).
```

docs/source/using_libraft.md
```diff
@@ -1,59 +1,64 @@
...
 
-For most functions, this overhead is pretty minimal and not noticeable but some of RAFT's APIs consist of very complex hierarchies of function calls that ultimately end up dispatching to device code that's executed on the GPU. The compile times for these APIs may still be bearable when compiling for only a single compute architecture but could end up becoming extremely slow to compile for all of the supported architectures at once.
+For most functions, compile-time overhead is minimal but some of RAFT's APIs take a substantial time to compile. As a rule of thumb, most functionality in `raft::distance`, `raft::neighbors`, and `raft::spatial` is expensive to compile and most functionality in other namespaces has little compile-time overhead.
```
cjnolet, line 5:
```suggestion
For most functions, compile-time overhead is minimal but some of RAFT's APIs take a substantial time to compile. As a rule of thumb, most functionality in `raft::distance`, `raft::neighbors`, and `raft::cluster` is expensive to compile and most functionality in other namespaces has little compile-time overhead.
```

cjnolet: Please don't reference `raft::spatial` in the docs anymore.

docs/source/using_libraft.md
```diff
@@ -1,59 +1,64 @@
...
-As mentioned above, the pre-compiled template instantiations can save a lot of time if you are able to use the type combinations for the templates which are already specialized in the `libraft` binary. This will, of course, mean that you will need to add `libraft` to your link libraries.
+2. If you are able to use one of the template types that are already being
+   compiled into `libraft`, you can use the pre-compiled template
+   instantiations, which are described in more detail in the following section.
 
-At the top level of each namespace containing pre-compiled template specializations is a header file called `specializations.cuh`. This header file includes `extern template` directives for all the specializations which are compiled into libraft. As an example, including `raft/neighbors/specializations.cuh` in one of your source files will effectively tell the compiler to skip over any of the template specializations that are already compiled into the `libraft` binary.
+3. If you would like to use RAFT but either cannot or would prefer not to
+   compile any CUDA code yourself, you can simply add `libraft` to your link
+   libraries and use the growing set of runtime APIs.
```
cjnolet, line 21:
```suggestion
   libraries and use the growing set of `raft::runtime` APIs.
```

cjnolet: Can we link to the section on "Runtime APIs" here please?

cjnolet (changes requested):
Really looking forward to these changes, @ahendriksen! I've gone through all the files and I think everything looks good, with the exception of some trivial things (mostly docs related).

cpp/include/raft/distance/distance.cuh
```diff
@@ -13,470 +13,12 @@
...
-                    metric,
-                    rowmajor,
-                    metric_arg);
-}
-
-/** @} */
-
-};  // namespace distance
-};  // namespace raft
+#if !defined(RAFT_EXPLICIT_INSTANTIATE_ONLY)
```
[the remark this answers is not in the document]
ahendriksen: I misremembered a detail about `ifndef`. I thought it was standardized in C++ 23, but apparently it's `elifdef` and `elifndef` that are standardized in C++ 23. 

https://en.cppreference.com/w/cpp/preprocessor/conditional

cpp/include/raft/distance/distance-inl.cuh
```diff
@@ -0,0 +1,477 @@
...
+#include <rmm/device_uvector.hpp>
+#include <type_traits>
+
+#include <raft/core/device_mdspan.hpp>
+
+namespace raft {
+namespace distance {
+
+/**
+ * @defgroup pairwise_distance pointer-based pairwise distance prims
```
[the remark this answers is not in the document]
ahendriksen: Good point. I checked this previously, and when I build the docs locally, the docs point to the correct header. For instance: the documentation for `raft::neighbors::ivf_flat::extend` starts with `#include <raft/neighbors/ivf_flat.cuh>`.

I don't know why or how though. I couldn't find any documentation in Doxygen that it automatically points to a non `inl` header.

cpp/include/raft/neighbors/detail/ivf_flat_search-inl.cuh
ahendriksen: This header is broken up in #1440 :+1:

docs/source/developer_guide.md
ahendriksen: Do you happen to know how we can make the link always resolve to our latest branch? Now it is pointing to this specific commit. I could point it to the 23.06 branch, but that will be outdated in a few months too...

cpp/include/raft/distance/distance-inl.cuh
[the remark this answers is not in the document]
ahendriksen: I now see that we are manually writing the include header in the `.rst` files. So as long as we point users to the correct headers there, this should not be an issue. There seems to be no secret feature in doxygen/sphinx to automatically point to the correct header.

docs/source/using_libraft.md
ahendriksen: I couldn't find any documentation of the Runtime APIs. I have filed an [issue](https://github.com/rapidsai/raft/issues/1468) to keep track of it.

cpp/include/raft/distance/distance.cuh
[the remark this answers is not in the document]
ahendriksen: I have now replaced `# if !defined` by `#ifndef`.

PR #1438, opened 2023-04-20
 [ENH] [2/5] Header structure: remove specialization includes

ahendriksen:
Specializations are deprecated in PR #1437 . This PR removes includes of these deprecated headers.

On top of PR #1437, it adds a single commit:
- [d5b5673](https://github.com/rapidsai/raft/pull/1438/commits/d5b567397ca30c9e1f38e170adc52450626d97c5): Remove includes of specialization headers within RAFT.

PR #1439, opened 2023-04-20
 [ENH] [3/5] Header structure: force explicit instantiation in tests and benchmarks

ahendriksen:
Since PR #1437, expensive templates can be marked as `RAFT_EXPLICIT` to indicate they must be explicitly instantiated. This PR forces the tests and benchmarks to use these explicit instantiations. 

In addition, this PR adds tests to make sure that the header files that provide the expensive templates include all necessary files when compiled under different combinations of `RAFT_EXPLICIT` and `RAFT_EXPLICIT_INSTANTIATE_ONLY`. This is done in the following steps:

**1. Fix tests** so that they can be built with `RAFT_EXPLICIT_INSTANTIATE_ONLY` defined. 

Change the index type in the tests to match the instantiated index type:
- [test/matrix/select_k.cu: Change index type](https://github.com/rapidsai/raft/pull/1439/commits/e527efc56e407d7208ae6bfb4ae2fb7bbafed2f7)
- [test/neighbors/fused_l2_knn.cu: Change index type](https://github.com/rapidsai/raft/pull/1439/commits/17902e95d700426b355538ec408fc1da77f1451e)

Add an instance in the tests in a separate source file:
- [test/distance/dist_adj.cu: Add instance](https://github.com/rapidsai/raft/pull/1439/commits/94d8117dab13b4b04ad50ae8333a6d6c44667430)

Allow creating instances in the test itself: (filed issues #1447, #1448)
- [test/cluster/linkage.cu: Allow instance](https://github.com/rapidsai/raft/pull/1439/commits/361570b7e3a41f8fa7e00a09876ea121d21b31e1)
- [test/sparse/neighbors/connect_components.cu: Allow instance](https://github.com/rapidsai/raft/pull/1439/commits/5171de333cf4b406c3d572f6d24d00c3f50fa4cd)
- [test/neighbors/ann_ivf_pq/test_float_uint32_t.cu: Allow instance](https://github.com/rapidsai/raft/pull/1439/commits/4426c5001dda807224348608b20fc838f7e0a064)

**2. Force** explicit instantiation in the benchmarks and tests in CMake:

- [Force explicit instantiations in tests](https://github.com/rapidsai/raft/pull/1439/commits/976189b48447bb3c76ef12159d48e61c203c5770)
- [Force explicit instantiations in benchmarks](https://github.com/rapidsai/raft/pull/1439/commits/bdae61d76172bf53b2d2f94431076044fcd493d8)

**3. Add tests** to make sure that the headers work under all prevalent combinations of `RAFT_COMPILED` and `RAFT_EXPLICIT_INSTANTIATE_ONLY`:

- [Test that headers are free standing](https://github.com/rapidsai/raft/pull/1439/commits/b0b8fe5293964c185916cd783930464082e3c8d0)


ahendriksen:
Reviewer tip: this PR is a follow up to PR #1438. Only the last 9 commits are part of this PR. That should reduce the diff considerably.

cpp/test/sparse/neighbors/connect_components.cu
```diff
@@ -14,6 +14,15 @@
  * limitations under the License.
  */
 
+// XXX: We allow the instantiation of fused_l2_nn here:
+// raft::linkage::FixConnectivitiesRedOp<value_idx, value_t> red_op(colors.data(), params.n_row);
+// raft::linkage::connect_components<value_idx, value_t>(
+//   handle, out_edges, data.data(), colors.data(), params.n_row, params.n_col, red_op);
+//
+// TODO: consider adding this to libraft.so or creating an instance in a
```
ahendriksen, line 22: Filed issue #1447 for this.

cpp/test/neighbors/ann_ivf_pq/test_float_uint32_t.cu
```diff
@@ -14,6 +14,13 @@
  * limitations under the License.
  */
 
+// XXX: the uint32_t instance is not compiled in libraft.so. So we allow
+// instantiating the template here.
+//
+// TODO: consider removing this test or consider adding an instantiation to the
```
ahendriksen, line 20: Filed issue https://github.com/rapidsai/raft/issues/1448 for this.

cpp/test/cluster/linkage.cu
```diff
@@ -14,6 +14,15 @@
...
  */
 
+// XXX: We allow the instantiation of fused_l2_nn here:
+// raft::linkage::FixConnectivitiesRedOp<value_idx, value_t> red_op(colors.data(), params.n_row);
+// raft::linkage::connect_components<value_idx, value_t>(
+//   handle, out_edges, data.data(), colors.data(), params.n_row, params.n_col, red_op);
+//
+// TODO: consider adding this to libraft.so or creating an instance in a
+// separate translation unit for this test.
+#undef RAFT_EXPLICIT_INSTANTIATE_ONLY
```
ahendriksen, line 24: Filed issue #1447 for this.

cpp/test/CMakeLists.txt
```diff
@@ -140,7 +149,41 @@ if(BUILD_TESTS)
...
+    test/ext_headers/raft_neighbors_detail_selection_faiss.cu
+    test/ext_headers/raft_linalg_detail_coalesced_reduction.cu
+    test/ext_headers/raft_spatial_knn_detail_ball_cover_registers.cu
+  )
+
+  # Test that the split headers compile in isolation with:
+  #
+  # * EXT_HEADERS_TEST_COMPILED_EXPLICIT: RAFT_COMPILED, RAFT_EXPLICIT_INSTANTIATE_ONLY defined
+  # * EXT_HEADERS_TEST_COMPILED_IMPLICIT: RAFT_COMPILED defined
+  # * EXT_HEADERS_TEST_IMPLICIT:          no macros defined.
```
tfeher, line 178: Do I understand the following points correctly?
- These tests do not run any computation, just tests if all the valid flag combination result in valid code.
- No GPU code is generated for these test, therefore it is cheap to have them.

tfeher (commented):
Thanks @ahendriksen for the PR. Looks good overall, just have one question.

cpp/test/CMakeLists.txt
tfeher: Wow, that is surprising. Yes, I agree that it is better to keep the test and file an issue to investigate what is happening with these files.

ahendriksen: Filed: https://github.com/rapidsai/raft/issues/1465

tfeher (approved):
Reviewed the new commits in this PR. Issues are created for the few outstanding discussion points. Looks good to me. Thanks Allard for the clear description of the changes!

cpp/test/matrix/select_k.cu
```diff
@@ -228,9 +228,9 @@ struct SelectK  // NOLINT
     auto& in_dists   = ref.get_in_dists();
     auto compare_ids = [&in_ids, &in_dists](const IdxT& i, const IdxT& j) {
       if (i == j) return true;
-      auto ix_i = uint64_t(std::find(in_ids.begin(), in_ids.end(), i) - in_ids.begin());
-      auto ix_j = uint64_t(std::find(in_ids.begin(), in_ids.end(), j) - in_ids.begin());
-      if (ix_i >= in_ids.size() || ix_j >= in_ids.size()) return false;
+      auto ix_i = int64_t(std::find(in_ids.begin(), in_ids.end(), i) - in_ids.begin());
+      auto ix_j = int64_t(std::find(in_ids.begin(), in_ids.end(), j) - in_ids.begin());
+      if (size_t(ix_i) >= in_ids.size() || size_t(ix_j) >= in_ids.size()) return false;
```
divyegala, line 233:
```suggestion
      auto ix_i = static_cast<int64_t>(std::find(in_ids.begin(), in_ids.end(), i) - in_ids.begin());
      auto ix_j = static_cast<int64_t>(std::find(in_ids.begin(), in_ids.end(), j) - in_ids.begin());
      if (static_cast<size_t>(ix_i) >= in_ids.size() || static_cast<size_t>(ix_j) >= in_ids.size()) return false;
```

cpp/test/neighbors/fused_l2_knn.cu
```diff
@@ -77,9 +77,9 @@ class FusedL2KNNTest : public ::testing::TestWithParam<FusedL2KNNInputs> {
     rmm::device_uvector<T> temp_distances(num_db_vecs * num_queries, stream_);
     distance::pairwise_distance(
       handle_,
-      raft::make_device_matrix_view<T, int64_t>(search_queries.data(), num_queries, dim),
-      raft::make_device_matrix_view<T, int64_t>(database.data(), num_db_vecs, dim),
-      raft::make_device_matrix_view<T, int64_t>(temp_distances.data(), num_queries, num_db_vecs),
+      raft::make_device_matrix_view<T, int32_t>(search_queries.data(), num_queries, dim),
+      raft::make_device_matrix_view<T, int32_t>(database.data(), num_db_vecs, dim),
+      raft::make_device_matrix_view<T, int32_t>(temp_distances.data(), num_queries, num_db_vecs),
```
divyegala, line 82: Don't we specialize on `uint32_t`?

cpp/test/cluster/linkage.cu
```diff
@@ -14,6 +14,15 @@
  * limitations under the License.
  */
 
+// XXX: We allow the instantiation of fused_l2_nn here:
```
divyegala, line 17: Am I correct in understanding that we need to implicitly instantiate `fused_l2_nn` here because `connect_components` doesn't have an extern and explicit instantiation?

This makes sense. We implicitly instantiate `connect_components` here which I imagine has a cascading effect on every single function that is called within its function body. It's using the header-only version so `RAFT_EXPLICIT` error must be thrown for `fused_l2_nn`, and it's good to see that error working as expected.

cpp/test/CMakeLists.txt
```diff
@@ -17,7 +17,7 @@
 
 function(ConfigureTest)
 
-  set(options OPTIONAL LIB)
+  set(options OPTIONAL LIB EXPLICIT_INSTANTIATE_ONLY)
```
divyegala, line 20: Is `LIB` needed anymore now that `RAFT_COMPILED` is `PUBLIC`? We can just link to `raft::compiled` by default and we'll get that macro defined. This is out of scope for this PR though.

cjnolet (approved):
LGTM, pending changes from #1437 and successfull run of downstream CIs (cugraph, cugraph-ops, cuopt, cuml). I verified the `RAFT_COMPILED` was properly removed from all the tests and benchmarks.

cpp/test/CMakeLists.txt
ahendriksen: I think `LIB` is still needed and that its use is orthogonal to whether `RAFT_COMPILED` is `PUBLIC`.

My understanding is that `LIB` is used to optionally link to `raft::compiled`. This means that some tests (like `LINALG_TEST`) can be built without first building `libraft.so`. We specifically need the optional `LIB` flag now for `EXT_HEADERS_TEST_IMPLICIT`. 

In this PR, as well as before, `RAFT_COMPILED` would be defined automatically when linking `raft::compiled`.

cpp/test/neighbors/fused_l2_knn.cu
ahendriksen: `distance::pairwise_distance` only has instances for `int32_t`. There is an open issue to add 64 bit indexing though.

cpp/test/cluster/linkage.cu
ahendriksen: Yes, exactly :+1:

PR #1440, opened 2023-04-20
 [ENH] [4/5] Header structure: split search of ivf methods

ahendriksen:
This PR splits the ivf-pq and ivf-flat search methods. It is a follow up to PR #1439. 

This PR: 
- Halves the number of instances of interleaved_scan_kernel, by instantiating only for veclen=1 or veclen=16/sizeof(DataT). 
- Extracts the interleaved_scan_kernel code from the ivf_flat_search.cuh header
- Extracts the compute_similarity_kernel code from ivf_pq_search.cuh header
- Makes the template parameters of IVF-flat consistent


ahendriksen:
Review tip: the only difference compared to PR https://github.com/rapidsai/raft/pull/1439 is the last commit. The full diff wrt branch-23.06 includes PR https://github.com/rapidsai/raft/pull/1437, so it is a bit much.

cpp/include/raft/neighbors/detail/ivf_pq_compute_similarity-inl.cuh
```diff
@@ -0,0 +1,845 @@
...
+ * Maximum value of k for the fused calculate & select in ivfpq.
+ *
+ * If runtime value of k is larger than this, the main search operation
+ * is split into two kernels (per batch, first calculate distance, then select top-k).
+ */
+static constexpr int kMaxCapacity = 128;
+static_assert((kMaxCapacity >= 32) && !(kMaxCapacity & (kMaxCapacity - 1)),
+              "kMaxCapacity must be a power of two, not smaller than the WarpSize.");
+
+// using weak attribute here, because it may be compiled multiple times.
```
tfeher, line 41: Just curious: what are the reason for using `__attribute__((weak))` as opposed to `inline`?

cpp/include/raft/neighbors/detail/ivf_pq_compute_similarity-inl.cuh
```diff
@@ -0,0 +1,845 @@
...
+struct selected {
+  compute_similarity_kernel_t<OutT, LutT> kernel;
+  dim3 grid_dim;
+  dim3 block_dim;
+  size_t smem_size;
+  size_t device_lut_size;
+};
+
+template <typename OutT, typename LutT>
+void compute_similarity_run(selected<OutT, LutT> s,
```
tfeher, line 566: Most of the code in this file  is just moved from `ivf_pq_compute_similarity.cuh`. The main difference is this section, instead of keeping on `operator()` for the `selected` struct, the kernel launcher is factored out into this function. As described by Allard, this is necessary to ensure that we can throw error message for accidental instantiation. Looks good to me.

tfeher (approved):
Thanks Allard, I have just one question for clarification, otherwise the PR looks good to me.

cpp/include/raft/neighbors/detail/ivf_pq_compute_similarity-inl.cuh
ahendriksen: Inline implies that:
1. the function may be defined in multiple translation units
2. the function symbol may be elided from the symbol table of the object file

In this case, we want only behavior 1, not 2. The `is_local_topk_feasible` function is also declared in `-ext` so other translation units may depend on its existence for linking. 

The reason we cannot use `RAFT_INLINE_CONDITIONAL` is that we have multiple translation units that define `is_local_topk_feasible` (one for each instance of `ivf_pq_compute_similarity`). We would run into multiple definition errors during linking if we did not specify the symbol as weak.

cpp/include/raft/neighbors/detail/ivf_flat_search-ext.cuh
```diff
@@ -36,20 +35,6 @@ void search(raft::device_resources const& handle,
             float* distances,
             rmm::mr::device_memory_resource* mr = nullptr);
```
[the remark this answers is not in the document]
ahendriksen: No this is an oversight(!). I pushed a fix.

cpp/include/raft/neighbors/detail/ivf_pq_compute_similarity-inl.cuh
tfeher: Thanks for the explanation!

divyegala, line 41: Could you document this somewhere? I understand why this is needed from your explanation, but it looks like this is the way to emulate the behavior of `extern` and `explicit` template instantiation when the function in question is _not_ a function-template

divyegala (approved):
Just one comment about expanding documentation, but it doesn't have to be a part of this PR!

cpp/include/raft/neighbors/detail/ivf_pq_compute_similarity-inl.cuh
divyegala: Also, why is this function not just `inline`? Looks to me like a very small function which should ideally have little compile-time overhead

cjnolet (approved):
LGTM pending requested changes in #1437

cpp/include/raft/neighbors/detail/ivf_pq_compute_similarity-inl.cuh
ahendriksen: > Also, why is this function not just inline?

Copying from: https://github.com/rapidsai/raft/pull/1440#discussion_r1177615832 :
Inline implies that:
1. the function may be defined in multiple translation units
2. the function symbol may be elided from the symbol table of the object file

In this case, we want only behavior 1, not 2. The `is_local_topk_feasible` function is also declared in `-ext` so other translation units may depend on its existence for linking. 

The reason we cannot use `RAFT_INLINE_CONDITIONAL` is that we have multiple translation units that define `is_local_topk_feasible` (one for each instance of `ivf_pq_compute_similarity`). We would run into multiple definition errors during linking if we did not specify the symbol as weak. 

> Could you document this somewhere? I

I have pushed 6a0be4c3f that defines and documents `RAFT_WEAK_FUNCTION`.

PR #1444, opened 2023-04-20
Add RAPIDS cuDF as a library that supports cuda_array_interface

miguelusque:
Adding a mention and an example of RAPIDS cuDF compatibility with cuda_array_interface.


cjnolet (approved):
This is a great addition to the docs. Thanks @miguelusque!

PR #1455, opened 2023-04-21
Learn heuristic to pick fastest select_k algorithm

benfred:
This adds a script to generate a dataset of which select_k method is fastest for different numbers of rows/cols/k.

This dataset is then used to learn a heuristic to pick the fastest select_k variant based off the rows/ cols/k of the input. This heuristic is modelled as a DecisionTree, which is automatically exported in C++ code that is compiled into RAFT. 
This lets us learn a function to pick the fastest select_k method - which requires only a few `if` statements in C++ code, making it very cheap to evaluate.

Note: this requires https://github.com/rapidsai/raft/pull/1430 - and only the  last commit here is new code.


benfred:
You can see the learned heuristic has learned the fastest algorithm by looking at the plots here:
Screen Shot 2023-04-21 at 11 44 50 AM
Screen Shot 2023-04-21 at 11 44 41 AM

tfeher (changes requested):
This is a great initiative @benfred! 

I would recommend to extend the comparison to all the top-k implementations that we have internally: https://github.com/rapidsai/raft/blob/c0c4d52c5a72c494e070f42324fc80b3a7cda205/cpp/internal/raft_internal/matrix/select_k.cuh#L51-L60

Some context:
- `warpsort_filtered `was added while integrating IVF-Flat, and we still use it directly there: https://github.com/rapidsai/raft/blob/c0c4d52c5a72c494e070f42324fc80b3a7cda205/cpp/include/raft/neighbors/detail/ivf_flat_search.cuh#L704
- warp_sort_distributed was added for IVF-PQ https://github.com/rapidsai/raft/blob/c0c4d52c5a72c494e070f42324fc80b3a7cda205/cpp/include/raft/neighbors/detail/ivf_pq_search.cuh#L442-L453
- `warp_sort_distributed_ex` is the latest development from @yong-wang, and possibly the fastest. If your benchmark's confirm this, then we can deprecate the older versions, otherwise we could use the best from the existing methods.

Having this many variants of warpsort is part of the accumulated tech debt while integrating the ANN methods. It would be great to have a selection heuristic for the `warp_sort` methods. `warpsort::select_k` has an internal heuristic to choose between two variants of `warp_sort_`, but does not utilize the newer methods (`warp_sort_distributed`* variants).  

It should be considered whether we can have selection heuristic for the warp level classes that are are used in fused kernels (not necessarily in this PR).

The radix select kernels were recently updated in: https://github.com/rapidsai/raft/pull/1175. That pr added the [fused_last_filter](https://github.com/rapidsai/raft/blob/branch-23.06/cpp/include/raft/matrix/detail/select_radix.cuh#L1102) option, and it would be good to include that option in the decision process. This corresponds to the `kRadix11bits` vs `kRadix11bitsExtraPass` internal options above.

Last but not least, I would recommend to use our c++ benchmarks to generate the benchmark dataset for training the decision tree (see my comment below).

achirkin:
@benfred thanks for this work, I'm excited and looking forward to see the results of this! In particular, the top-k selection has a strong impact on latency of ivf-pq for small batch sizes. Thanks, @tfeher, for pointing to our google benchmarks and providing the comprehensive context.

The only thing need to add: if you have a chance, could you please check the launch configurations of the warpsort kernels? I clearly remember it was not optimal in some cases; at least, on small inputs, where it needs to decide whether to run the one-pass (one-block per line) or the two-pass version.

achirkin:
> Having this many variants of warpsort is part of the accumulated tech debt while integrating the ANN methods. It would be great to have a selection heuristic for the `warp_sort` methods. `warpsort::select_k` has an internal heuristic to choose between two variants of `warp_sort_`, but does not utilize the newer methods (`warp_sort_distributed`* variants).
> 
> It should be considered whether we can have selection heuristic for the warp level classes that are are used in fused kernels (not necessarily in this PR).

Also note, several warp_sort implementations differ in their usage of registers and shared memory. As a result, the slower ones may perform in some kernels better than the others.

benfred:
I've opened a new clean PR https://github.com/rapidsai/raft/pull/1523  here - rather than trying to resolve the conflicts on this branch (i could force push, but then we'd lose the discussion that has already happened).

PR #1456, opened 2023-04-22
fix ivf_pq n_probes

benfred:
The ivf-pq search code was including a guard like

```auto n_probes = std::min<uint32_t>(params.n_probes, index.n_lists());```

to check to make sure that we weren't selecting more values than are available. However,
this wasn't being used and instead just `params.n_probes` was being passed to functions
like `select_k`. This lead to asking select_k to select say 100 items, when there
were only 90 to choose from - and caused some issues downstream when trying to update
the select_k algorithm

Fix.


lowener (approved):

tfeher (approved):
Thanks @benfred, good catch! LGTM.

PR #1462, opened 2023-04-25
Use pin_compatible to ensure that lower CTKs can be used

vyasr:
The runtime requirement for the CTK is weaker than the build-time requirement. We allow installing raft into environments with an older CTK (any 11.x) than the build requirement (11.8)

### DIFF
--- a/conda/recipes/libraft/meta.yaml
+++ b/conda/recipes/libraft/meta.yaml
@@ -52,6 +52,9 @@ outputs:
       host:
         - librmm ={{ minor_version }}
         - cudatoolkit {{ cuda_version }}
+      run:
+        - {{ pin_compatible('cudatoolkit', max_pin='x', min_pin='x') }}
+        - librmm ={{ minor_version }}
     about:
       home: https://rapids.ai/
       license: Apache-2.0
@@ -67,7 +70,6 @@ outputs:
     requirements:
       run:
         - {{ pin_subpackage('libraft-headers-only', exact=True) }}
-        - cudatoolkit {{ cuda_spec }}
         - cuda-profiler-api {{ cuda_profiler_api_run_version }}
         - librmm ={{ minor_version }}
         - libcublas {{ libcublas_run_version }}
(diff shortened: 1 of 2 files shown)


conda/recipes/libraft/meta.yaml
```diff
@@ -66,9 +66,9 @@ outputs:
         - librmm
     requirements:
       run:
+        - {{ pin_compatible('cudatoolkit', max_pin='x', min_pin='x') }}
```
ajschmidt8, line 69: Since the `libraft-headers` package doesn't explicitly list `cudatoolkit` as a `build` or `host` dependency, I don't think `pin_compatible` will work as intended.

We can inspect the packages that result from the CI build to confirm.

There are docs on `pin_compatible` here: https://docs.conda.io/projects/conda-build/en/latest/resources/variants.html#extra-jinja2

To account for this limitation, you can use the `cuda_spec` variable that's at the top of the recipe if necessary.

ajschmidt8 (commented):
Good catch. Left one comment that we should investigate before I approve

conda/recipes/libraft/meta.yaml
vyasr: Yeah I noticed the same and was going to suggest just adding a host req but cuda_spec is cleaner.

vyasr:
Blocking while we discuss the appropriate dependency structure for the different packages.

cjnolet (approved):
LGTM

divyegala (approved):

ajschmidt8 (approved):

conda/recipes/libraft/meta.yaml
```diff
@@ -52,6 +52,9 @@ outputs:
       host:
         - librmm ={{ minor_version }}
         - cudatoolkit {{ cuda_version }}
+      run:
+        - {{ pin_compatible('cudatoolkit', max_pin='x', min_pin='x') }}
```
bdice, line 56: FYI, this is also what the cuda_spec variable is meant for (defined above). If you’re pinning cudatoolkit and don’t have a host pinning to pin_compatible with, you can use cuda_spec.

bdice (approved):

PR #1467, opened 2023-04-26
Remove include statements from inside namespace

robertmaynard:
Bringing in the below headers into the `raft::common::nvtx::detail` namespace breaks downstream users that need to use global symbols they provide. In the direct case I encountered the dlopen and dlclose functions became unusable.
```
#include <cstdint>
#include <cstdlib>
#include <mutex>
#include <nvtx3/nvToolsExt.h>
#include <string>
#include <type_traits>
#include <unordered_map>
```


vyasr (approved):
Thanks!

cjnolet (approved):

PR #1469, opened 2023-04-27
[ENH] [FINAL] Header structure: combine all PRs into one

### LINKED ISSUE #1416: [DISCUSSION] A robust solution to precompiling function templates

ahendriksen:
This is a rebase of all the commits in PRs:
- #1437 
- #1438 
- #1439 
- #1440 
- #1441 

The original PRs have not been rebased to preserve review comments. This PR is up to date with branch 23.06.

Closes #1416


divyegala (approved):

divyegala:
Is the rebase done correctly? CI is saying it's still 14 commits behind

ahendriksen:
I missed a `ifndef` in a previous version of commit 7af7711dd. Then I redid the rebase and naively thought it would be alright (which it wasn't). Now I have done the rebase from the beginning again. All commits should now be part of this PR.

cjnolet (approved):

PR #1471, opened 2023-04-28
CAGRA: Separate graph index sorting functionality from prune function

### LINKED ISSUE #1446: [FEA] Separate index sorting from CAGRA graph pruning
Hi, @tfeher 

The current implementation of CAGRA index pruning always sorts the input index before performing the pruning. 
However, it is unnecessary when CAGRA builds the input index because the CAGRA index-building function outputs a sorted index.
On the other hand, when the index is not built by CAGRA and not sorted, the sorting process is necessary to use CAGRA pruning.
Therefore, it would be useful to make sorting optional, depending on whether the index is built by CAGRA or not.

One of the solutions for this issue is to separate the sorting process from the CAGRA pruning function and create a separate sorting function.
This would involve modifying the current prune function as follows:

```cpp
template <typename IdxT = uint32_t,
          typename g_accessor =
            host_device_accessor<std::experimental::default_accessor<IdxT>, memory_type::host>>
void prune(raft::device_resources const& res,
           mdspan<IdxT, matrix_extent<IdxT>, row_major, g_accessor> knn_graph, //[in]
           raft::host_matrix_view<IdxT, IdxT, row_major> new_graph); //[out]

template <typename DataT, typename IdxT, typename accessor>
void sort_knn_graph(raft::device_resources const& res,
                     mdspan<const DataT, matrix_extent<IdxT>, row_major, accessor> dataset, // [in]
                     raft::host_matrix_view<IdxT, IdxT, row_major> knn_graph); // [in, out]
```

An alternative solution is to make the `dataset` argument `std::optional` and sort the index only if the argument is provided.

I prefer the approach of separating the sorting function because it can reduce the time required for compiling sorting operations during compilation when the sorting function is not used in the source code.

If there are no issues with this approach, I will modify the implementation accordingly.

enp1s0:
# Changes

This PR separates the graph index sorting functionality from the CAGRA pruning function and creates a new function. (Related issue: https://github.com/rapidsai/raft/issues/1446)

# Unit test

I have included a new unit test for the sorting function. The test utilizes a separate dataset from the one used in the CAGRA main test to avoid the effect of rounding errors during norm computation between two vectors in the dataset. More details are in the source code.
https://github.com/enp1s0/raft/blob/ea6c449c260895e9125a591a4848eed06f5b72c4/cpp/test/neighbors/ann_cagra.cuh#L93-L96

# Issue
Close #1446


enp1s0:
@tfeher Can you check the PR and let me know if there are any problems?

cpp/test/neighbors/ann_cagra.cuh
```diff
@@ -214,6 +295,79 @@ class AnnCagraTest : public ::testing::TestWithParam<AnnCagraInputs> {
...
+        raft::make_host_matrix<IdxT, IdxT>(ps.n_rows, index_params.intermediate_graph_degree);
+
+      if (ps.host_dataset) {
+        cagra::build_knn_graph<DataT, IdxT>(handle_, database_host_view, knn_graph.view());
+      } else {
+        cagra::build_knn_graph<DataT, IdxT>(handle_, database_view, knn_graph.view());
+      };
+
+      handle_.sync_stream(stream_);
+      handle_.sync_stream();
```
tfeher, line 331: Since` stream_ == handle.get_stream()`, this two statements are actually equivalent. (It is somewhat obscure but the we just [call this](https://github.com/rapidsai/raft/blob/a44ca96c5cddec7ca67b510f3c21163d3958bc7e/cpp/include/raft/core/resource/cuda_stream.hpp#L100) in the second case.
```suggestion
      handle_.sync_stream();
```

cpp/test/neighbors/ann_cagra.cuh
```diff
@@ -214,6 +295,79 @@ class AnnCagraTest : public ::testing::TestWithParam<AnnCagraInputs> {
...
+      ASSERT_TRUE(CheckOrder<DistanceT>(knn_graph.view(), database_host.view()));
+
+      RandomSuffle(knn_graph.view());
+
+      cagra::sort_knn_graph(handle_, database_view, knn_graph.view());
+      handle_.sync_stream(stream_);
+      handle_.sync_stream();
```
tfeher, line 338: Same as above.

tfeher (changes requested):
Thanks @enp1s0 for the PR! It looks good overall, below you will find a few small comments (and just as a reference, I have also linked two existing issues).

enp1s0:
@tfeher Thank you for your comments.
I have added a code example of `sort_knn_graph` and fixed the stream synchronization in the test code. I'll fix the memory allocation and multi-GPU support in another PR because there will be a lot of code changes.

cpp/test/neighbors/ann_cagra.cuh
```diff
@@ -190,18 +271,18 @@ class AnnCagraTest : public ::testing::TestWithParam<AnnCagraInputs> {
     std::cout << "Done.\nRuning rng" << std::endl;
     raft::random::Rng r(1234ULL);
     if constexpr (std::is_same<DataT, float>{}) {
-      r.uniform(database.data(), ps.n_rows * ps.dim, DataT(0.1), DataT(2.0), stream_);
-      r.uniform(search_queries.data(), ps.n_queries * ps.dim, DataT(0.1), DataT(2.0), stream_);
+      r.normal(database.data(), ps.n_rows * ps.dim, DataT(0.1), DataT(2.0), stream_);
```
tfeher, line 274: It will be probably more consistent to just remove `stream_` and use `handle.get_stream()` everywhere. I have added a commit into my follow-up PR: https://github.com/rapidsai/raft/pull/1494/commits/40ba2068fd9cd123d5449665e9e13dea98fac164

tfeher (approved):
Thanks @enp1s0  for addressing the issues. The PR looks good to me.

cpp/test/neighbors/ann_cagra.cuh
enp1s0: I have fixed the test code to use handle.get_stream() instead of stream_ in AnnCagraSortTest.

enp1s0:
I have changed the test case parameters to satisfy the following requirements in int8 and uint8 tests.

https://github.com/rapidsai/raft/blob/20141120f6ae2fd9ea1ea0648507af04594921ec/cpp/include/raft/neighbors/detail/cagra/cagra_build.cuh#LL52C45-L52C45

```cpp
RAFT_EXPECTS(
    dataset.extent(1) * sizeof(DataT) % 8 == 0,
    "Dataset rows are expected to have at least 8 bytes alignment. Try padding feature dims.");
```

PR #1482, opened 2023-05-02
Add missing ext declaration for log detail::format

tfeher:
The `format` function is used by [debug and trace loggers](https://github.com/rapidsai/raft/blob/a44ca96c5cddec7ca67b510f3c21163d3958bc7e/cpp/include/raft/core/logger-macros.hpp#L44-L75). While PR #1469 has restructured the logger headers it was forgotten to expose `detail::format` in case the `RAFT_EXPLICIT_INSTANTIATE_ONLY` is defined. This PR fixes that.


tfeher:
Tagging @ahendriksen for review. 

We should update the test so that such problems do not go undetected. The problem is, by default the debug and trace logging are disabled by RAFT_ACTIVE_LEVEL. Shall we just define adequate active level for the unit test?

cpp/include/raft/core/logger-ext.hpp
```diff
@@ -24,6 +24,9 @@ namespace raft {
 static const std::string RAFT_NAME = "raft";
 static const std::string default_log_pattern("[%L] [%H:%M:%S.%f] %v");
 
+namespace detail {
+std::string format(const char* fmt, ...);
```
ahendriksen, line 28: I would consider adding `RAFT_INLINE_CONDITIONAL` here as well just for consistency with the function definition in `logger-inl.hpp`.

ahendriksen (approved):
Good catch! I overlooked the detail namespace because I thought it would not get called from outside code..

ahendriksen:
> Shall we just define adequate active level for the unit test?

I agree it makes sense to `#define RAFT_ACTIVE_LEVEL=6` at the top of the logger test (i.e. before including `logger.hpp`). It would be even nicer if we could use `RAFT_LEVEL_TRACE` instead of `6`, but that would require including `logger.hpp`.

cpp/include/raft/core/logger-ext.hpp
tfeher: Added.

tfeher (commented):
Thanks Allard for the suggestion!

cjnolet (approved):

PR #1483, opened 2023-05-02
Fix get_pool_memory_resource return type

tfeher:
The return type of `get_pool_memory_resource` was changed in #1469 from `pool_memory_resource` to `device_memory_resource`. There are debug logs in the code ([example](https://github.com/rapidsai/raft/blob/a44ca96c5cddec7ca67b510f3c21163d3958bc7e/cpp/include/raft/neighbors/detail/ivf_pq_build.cuh#L1328-L1329)), which query the pool size, that would fail when debug logging is enabled. This PR restores the output type of `get_pool_memory_resource`.

### DIFF
--- a/cpp/include/raft/util/memory_pool-ext.hpp
+++ b/cpp/include/raft/util/memory_pool-ext.hpp
@@ -18,10 +18,11 @@
 #include <cstddef>                                   // size_t
 #include <memory>                                    // std::unique_ptr
 #include <rmm/mr/device/device_memory_resource.hpp>  // rmm::mr::device_memory_resource
+#include <rmm/mr/device/pool_memory_resource.hpp>    // rmm::mr::pool_memory_resource
 
 namespace raft {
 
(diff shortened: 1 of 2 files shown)


ahendriksen:
The return type of `get_pool_memory_resource` was changed specifically to avoid the inclusion of the `pool_memory_resource.hpp` header, which is [one of the few](https://github.com/rapidsai/rmm/pull/1241) RMM headers which unconditionally includes spdlog. Including `spdlog` slows down compile times considerably.

The `memory_pool.hpp` file is included from `cudart_utils.cuh`, so this PR in current state would undo all compile time gains from splitting out `memory_pool-inl.hpp`. 

As an alternative, would it be possible to dynamically cast to `pool_memory_resource` in those places where the debug logging takes place?

tfeher:
Thanks @ahendriksen for the explanation. I am closing this in favor of #1484.

Issue #1490, opened 2023-05-04
[BUG] linker complains about duplicate symbol get_cache_idx


alexanderguzhva:
**Describe the bug**
Many errors of the following kind:
```
ld.lld: error: duplicate symbol: raft::cache::(anonymous namespace)::get_cache_idx(int*, int, int*, int, int, int*, int*, bool*, int)
```

**Expected behavior**
No errors

**Environment details (please complete the following information):**
upstream hash is "a98295b516ef58bc855177077860bab2a2a76d77" (Apr 12 ?)
CUDA 11.4.2

**Additional context**
The error can be fixed in a hacky way by turning
```
__global__ void get_cache_idx(int* keys,
                              int n,
                              int* cached_keys,
                              int n_cache_sets,
                              int associativity,
                              int* cache_time,
                              int* cache_idx,
                              bool* is_cached,
                              int time)
```
into
```
template<typename T>
__global__ void get_cache_idx(T* keys,
                              T n,
                              T* cached_keys,
                              T n_cache_sets,
                              T associativity,
                              T* cache_time,
                              T* cache_idx,
                              bool* is_cached,
                              T time)
```
I haven't tried moving it to a source file.

cjnolet:
@alexanderguzhva this is indeed strange because the anonymous namespace should be protecting us from this compile error. Does adding `__inline__`  to the definition help overcome this error for you? If so, we can remove the anonymous namespace and just use that?

alexanderguzhva:
@cjnolet A kernel cannot be inlined, can it be?

cjnolet:
@alexanderguzhva oddly, I'm not able to reproduce that compile error when building FAISS. Is there a specific branch I can check out? What version of gcc are you using? 

I verified I'm able to build raft successfully by removing the offending kernel from the anonymous namespace and adding `__forceinline__` to the definition. If I create a RAFT PR for this, can you give it a try and see if the error goes away?

ahendriksen:
> A kernel cannot be inlined, can it be?

A kernel can be inlined, i.e, this works: `inline __global__ void kernel() {}`. It doesn't result in the kernel being literally "inlined" into a function, but it does change the linking behavior (symbol is defined as weak and may be omitted from symbol table by compiler).

In general, I think we should prefer inline over an anonymous namespace. With an anonymous namespace, a copy of the kernel is included in each translation unit and also in the final linked object. With inline, duplicate definitions should be deduplicated when linked.

[...]

PR #1492, opened 2023-05-05
Inline get_cache_idx

ahendriksen:
Related to issue #1490. 

The `get_cache_idx` kernel is currently defined in an anonymous namespace. In #1490, it was reported that this could lead to linker errors in CUDA 11.4 in combination with (presumably?) clang.

I tried to reproduce the linker error in commit 01cf4099b206b6b37dcdc84e852507306b018af5 but could not, as described in the commit message.

This PR attempts to fix the linker errors by marking `get_cache_idx` as inline and by removing the anonymous namespace.


cjnolet (approved):
LGTM pending verification from Meta side.

alexanderguzhva:
It seems to be working! Thanks.

PR #1494, opened 2023-05-08
Cagra index construction without copying device mdarrays

### LINKED ISSUE #1479: [FEA] CAGRA avoid unnecessary device memory copies of dataset / knn graph

When data / graph is in device memory, consider just storing a reference in the index for the data. See discussion https://github.com/rapidsai/raft/pull/1375#discussion_r1159230818

Tracker #1392

tfeher:
This PR aims to improve the workflow when dealing with large datasets. When experimenting with different versions of the knn-graph, we might want to construct indices with the same dataset (see https://github.com/rapidsai/raft/pull/1435 for further discussion).

If the dataset is already in device memory (and rows are properly aligned / padded), then we only store a reference to the dataset, therefore multiple indices can refer to the same dataset. Similarly, when `knn_graph` is a device array, then store only a reference. 

Additionally, this PR adds `update_dataset` and `update_graph` methods to the index.
Closes #1479

### DIFF
--- a/cpp/include/raft/neighbors/cagra_types.hpp
+++ b/cpp/include/raft/neighbors/cagra_types.hpp
@@ -34,6 +34,7 @@
 #include <thrust/fill.h>
 #include <type_traits>
 
+#include <raft/core/logger.hpp>
 namespace raft::neighbors::experimental::cagra {
 /**
  * @ingroup cagra
@@ -137,7 +138,7 @@ struct index : ann::index {
   /** Graph degree */
   [[nodiscard]] constexpr inline auto graph_degree() const noexcept -> uint32_t
   {
-    return graph_.extent(1);
+    return graph_view_.extent(1);
   }
 
   /** Dataset [size, dim] */
@@ -148,15 +149,12 @@ struct index : ann::index {
   }
 
   /** neighborhood graph [size, graph-degree] */
-  inline auto graph() noexcept -> device_matrix_view<IdxT, IdxT, row_major>
-  {
-    return graph_.view();
-  }
+  inline auto graph() noexcept -> device_matrix_view<IdxT, IdxT, row_major> { return graph_view_; }
 
   [[nodiscard]] inline auto graph() const noexcept
     -> device_matrix_view<const IdxT, IdxT, row_major>
   {
-    return graph_.view();
+    return graph_view_;
   }
 
   // Don't allow copying the index for performance reasons (try avoiding copying data)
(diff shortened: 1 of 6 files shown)


cpp/include/raft/neighbors/cagra_types.hpp
```diff
@@ -183,15 +185,36 @@ struct index : ann::index {
...
+    dataset_view_ = dataset_.view();
+    graph_view_   = graph_.view();
     res.sync_stream();
   }
 
+  /** Construct an index from dataset and knn_graph arrays */
+  index(raft::device_resources const& res,
+        raft::distance::DistanceType metric,
+        raft::device_matrix_view<const T, IdxT, row_major> dataset,
+        raft::device_matrix_view<const IdxT, IdxT, row_major> knn_graph)
```
tfeher, line 200: Right now, passing `device_matrix_views` to the constructor would select this specialization (assuming that `const` qualifiers are not forgotten). Shall we make it more explicit for the user to select which constructor to call (e.g. by an additional param / template param)?

tfeher:
Linking the original discussion: https://github.com/rapidsai/raft/pull/1375#discussion_r1159230818 and tagging @divyegala and @enp1s0 for review.

tfeher:
Currenty API change is non breaking. Behavior is changed: if device arrays go out of scope while the index is alive, then index would be invalid. Therefore labelled as breaking change.

cpp/include/raft/neighbors/cagra_types.hpp
divyegala: I think we should just add more explicit docs

cpp/include/raft/neighbors/detail/cagra/cagra_serialize.cuh
```diff
@@ -36,7 +36,7 @@ struct check_index_layout {
                 "paste in the new size and consider updating the serialization logic");
 };
 
-template struct check_index_layout<sizeof(index<double, std::uint64_t>), 136>;
+template struct check_index_layout<sizeof(index<double, std::uint64_t>), 184>;
```
divyegala, line 39: Can we put this "magic number" in a `constexpr` variable with a comment?

cpp/test/neighbors/ann_cagra.cuh
```diff
@@ -181,37 +194,72 @@ class AnnCagraTest : public ::testing::TestWithParam<AnnCagraInputs> {
...
+    search_queries.resize(0, handle_.get_stream());
   }
 
  private:
   raft::device_resources handle_;
-  rmm::cuda_stream_view stream_;
   AnnCagraInputs ps;
   rmm::device_uvector<DataT> database;
   rmm::device_uvector<DataT> search_queries;
+  raft::device_matrix_view<DataT, IdxT> cagra_graph_dev_;
```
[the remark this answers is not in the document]
enp1s0: Or change to `raft::device_matrix` and remove `auto` in the following line
https://github.com/rapidsai/raft/blob/40ba2068fd9cd123d5449665e9e13dea98fac164/cpp/test/neighbors/ann_cagra.cuh#L216

cpp/include/raft/neighbors/cagra_types.hpp
tfeher: In the context of this PR extending the docs sounds good, but in #1505, even `device_mdspan` input might be copied if the number of dimensions would not be compatible with vectorized loads. Do you think adding clear docs is still the best option? https://github.com/rapidsai/raft/pull/1505/files#diff-0512c89e79e0d6300fda20bf29340258cd48ce7c5b5b81f4fbad61e8eae3db5aR191-R206

divyegala: I do think so, yes, with a tiny code example in the associated docs. For the first constructor, perhaps we can use host_mdspan instead of vanilla mdspan? Then the overloads will be automatically selected based on the input, and if there's any instances where the code queries the data using driver/rt APIs to check where a pointer is located, we can remove that entirely. What are your thoughts?

cjnolet: We should probably explicitly add the constructors to the docs source that we intend for users to interact with.

Small thing I noticed while looking over this PR: device_resources in the constructor should be changed to raft::resources.

cpp/include/raft/neighbors/cagra_types.hpp
```diff
@@ -183,15 +185,36 @@ struct index : ann::index {
...
+                 "Dataset and knn_graph must have equal number of rows");
+    dataset_view_ = dataset;
+    graph_view_   = knn_graph;
+  }
+
  private:
   raft::distance::DistanceType metric_;
   raft::device_matrix<T, IdxT, row_major> dataset_;
   raft::device_matrix<IdxT, IdxT, row_major> graph_;
+  raft::device_matrix_view<const T, IdxT, row_major> dataset_view_;
```
cjnolet, line 216: We have some types to help us w/ the `sometimes owning` semantics that I think can be useful here. I'm okay merging this PR in for release the way it is, but I think we should consider using something like the `mdbuffer` or `TemporaryDeviceBuffer` in the future. Though I think the `TemporaryDeviceBuffer` will require some adjustments to make it useful for this specific case (tagging @divyegala).

cjnolet (changes requested):
Just a couple fairly minor things. It would help if we could capture an issue to revisit the `sometimes owning` semantics. I think it's another great case for the `mdbuffer` that @wphicks has been working on.

cpp/include/raft/neighbors/cagra_types.hpp
```diff
@@ -183,15 +185,36 @@ struct index : ann::index {
   {
     RAFT_EXPECTS(dataset.extent(0) == knn_graph.extent(0),
                  "Dataset and knn_graph must have equal number of rows");
+    RAFT_LOG_INFO("CAGRA constructor, copying arrays");
```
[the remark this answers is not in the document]
tfeher: Added more meaningful log messages.

tfeher (commented):
I have updated the PR according to the comments.

cpp/include/raft/neighbors/cagra_types.hpp
wphicks: +1 to what @cjnolet mentioned here. I think it will be pretty straightforward to transition to mdbuffer once it's available (hopefully very soon).

cjnolet (approved):
LGTM, thanks @tfeher!

PR #1495, opened 2023-05-08
Re-use memory pool between benchmark runs

benfred:
Don't recreate a new memory pool for each benchmark, and instead re-use the pool.

This significantly speeds up running the benchmarks that use a cuda memory pool. As an example running
`time ./cpp/build/MATRIX_BENCH --benchmark_filter=SelectK/float/uint32_t.*/0/` which runs benchmarks for 9 different selection algorithms - the time to run the benchmarks is reduced from`36.317s` on branch-23.06 to `10.038s` with this change.

### DIFF
--- a/cpp/bench/prims/common/benchmark.hpp
+++ b/cpp/bench/prims/common/benchmark.hpp
@@ -113,8 +113,19 @@ class fixture {
   raft::device_resources handle;
   rmm::cuda_stream_view stream;
 
-  fixture() : stream{handle.get_stream()}
+  fixture(bool use_pool_memory_resource = false) : stream{handle.get_stream()}
   {
+    // Cache memory pool between test runs, since it is expensive to create.
+    // This speeds up the time required to run the select_k bench by over 3x.
+    // This is part of the fixture class here so that the pool will get cleaned
+    // up, rather than outliving the benchmarks that require it.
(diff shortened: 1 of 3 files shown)


cpp/bench/prims/common/benchmark.hpp
```diff
@@ -113,8 +113,19 @@ class fixture {
...
   {
+    // Cache memory pool between test runs, since it is expensive to create.
+    // This speeds up the time required to run the select_k bench by over 3x.
+    // This is part of the fixture class here so that the pool will get cleaned
+    // up, rather than outliving the benchmarks that require it.
+    static std::unique_ptr<using_pool_memory_res> memory_pool;
+    if (use_pool_memory_resource) {
+      if (!memory_pool) { memory_pool.reset(new using_pool_memory_res()); }
+    } else if (memory_pool) {
+      memory_pool.reset();
```
divyegala, line 126: I'm not following what an empty `reset()` is doing here. I'm reading this line as "a new memory pool is created", but your analysis shows that's not the case. Could you please explain?

benfred: This frees the memory being allocated by the memory pool - like in this case `use_pool_memory_resource` is false, and the empty reset will free up any previously allocated memory pool.

divyegala (approved):

PR #1498, opened 2023-05-08
Remove raft/matrix/matrix.cuh includes

benfred:
The `raft/matrix/matrix.cuh` file has been marked as deprecated, and produces a compile warning when included. However it was still being referenced in a bunch of different spots within raft - making it hard to avoid these warnings.

Remove the includes, in favour of either the newer API's or in certain cases the detail API


cjnolet:
Oh man this makes me so happy. Thanks @benfred for taking this on!

cpp/include/raft/linalg/detail/eig.cuh
```diff
@@ -52,7 +52,7 @@ void eigDC_legacy(raft::device_resources const& handle,
   rmm::device_uvector<math_t> d_work(lwork, stream);
   rmm::device_scalar<int> d_dev_info(stream);
 
-  raft::matrix::copy(in, eig_vectors, n_rows, n_cols, stream);
+  raft::copy_async(eig_vectors, in, n_rows * n_cols, stream);
```
cjnolet, line 55: This is kind of a "quick fix" which is, in turn, coupling out detail APIs across namespaces, which is actually creating more tech debt. I'd prefer if instead of doing this, we would use the new matrix public APIs so that we can move closer to removing the deprecated APIs altogether.

If the goal here is really just to silence the deprecation warnings, there's other ways we can do that.

cjnolet (changes requested):

cpp/include/raft/linalg/detail/eig.cuh
benfred: thanks for reviewing this!  This is still a draft here (and there is still quite a few other things that need changed/fixed before we can merge this) - but the last couple commits remove most of the matrix::detail namespace hacks I had in here.

cpp/include/raft/matrix/copy.cuh
```diff
@@ -58,6 +58,24 @@ void copy_rows(raft::device_resources const& handle,
...
 }
 
+/**
+ * @brief copy matrix operation for row major matrices.
+ * @param[in] handle: raft handle
+ * @param[in] in: input matrix
+ * @param[out] out: output matrix
+ */
+template <typename m_t, typename matrix_idx_t>
+void copy(raft::device_resources const& handle,
```
cjnolet, line 68: wonderful!

cpp/include/raft/matrix/diagonal.cuh
```diff
@@ -34,7 +34,7 @@ namespace raft::matrix {
  * @param[out] matrix: matrix of size n_rows x n_cols
  */
 template <typename m_t, typename idx_t, typename layout>
-void set_diagonal(raft::device_resources const& handle,
+void set_diagonal(raft::resources const& handle,
```
cjnolet, line 37: Yes!

cjnolet (approved):
LGTM!

Issue #1457, opened 2023-04-24
CAGRA int8 / uint8 distance scaling


tfeher:
int8 / uint8 input data is scaled by factor 128 or 256, but the resulting distances are not scaled back. Possible solution: https://github.com/rapidsai/raft/pull/892/files#diff-bead3200ec8f9fec284357719c86805ce8ed546e5af25a015959e27c3879a954

PR #1499, opened 2023-05-09
Un-scale output distances

achirkin:
Reverts the scaling of input data and queries that normally happens for int8/uint8 types before distance computation.
Reuses the ivf-pq post-processing step for that.

Along the way, replaces the custom `fragment_scale` helper with the `utils::mapping` that is shared among several ANN methods.

Solves https://github.com/rapidsai/raft/issues/1457

### DIFF
--- a/cpp/include/raft/neighbors/detail/cagra/cagra_search.cuh
+++ b/cpp/include/raft/neighbors/detail/cagra/cagra_search.cuh
@@ -102,8 +102,8 @@ void search_main(raft::device_resources const& res,
                 "only float distances are supported at the moment");
   float* dist_out          = distances.data_handle();
   const DistanceT* dist_in = distances.data_handle();
-  // We're converting the data from T to DistanceT during distance computation;
-  // hence now we convert it back (DistanceT -> T).
+  // We're converting the data from T to DistanceT during distance computation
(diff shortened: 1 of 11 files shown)


tfeher (approved):
Thanks Artem for the fix, just a minor thing, otherwise it looks good to me.

Tagging @enp1s0 for visibility.

ahendriksen:
It looks like this PR increased the binary size of [raft_conda_cpp_cuda11_x86_64.tar.gz](https://downloads.rapids.ai/ci/raft/branch/branch-23.06/2014112/raft_conda_cpp_cuda11_x86_64.tar.gz) by 100MB (from 650MB to 750MB). 
See: 
- This PR: https://downloads.rapids.ai/ci/raft/branch/branch-23.06/2014112/
- Previous PR: https://downloads.rapids.ai/ci/raft/branch/branch-23.06/f60fb79/

Does anybody understand why? It is a significant enough change that might be worth investigating.

achirkin:
The archives contain tests, right? We add two more sets of instantiations to cagra in tests here, which is a lot. Hopefully, https://github.com/rapidsai/raft/pull/1428 will improve this?

ahendriksen:
Ah yes, that makes sense. I also don't see an increase in the wheels. Thanks for pointing out!

Issue #1459, opened 2023-04-24
Binary size too large. Note that rapids is compiled for sm 60, 70, 75, 80, 86, 90. Compiling for float only, with one team specilization per team adds 125 MiB to libraft.so. Four data types and multiple team size specialization goes over 1 GiB size increase. Find out what is essential for good perf, and what is acceptable size increase for RAFT.


tfeher:

tfeher:
See https://github.com/rapidsai/raft/pull/1428#issuecomment-1521538787 details in binary size (when compiled only for float input type).

If we could have block size as runtime param (instead of template) that would lead a significant reduction in binary size
https://github.com/rapidsai/raft/blob/branch-23.06/cpp/include/raft/neighbors/detail/cagra/search_single_cta.cuh#L827-L841

PR #1505, opened 2023-05-10
CAGRA pad dataset for 128bit vectorized load

### LINKED ISSUE #1458: CAGRA support arbitrary `dim` (number of features)

tfeher:
This PR adds padding to the dataset (if necessary) to make reading any of its rows compatible with 128bit vectorized loads. This change also enables handling arbitrary number of input features (before this PR each row had to be at least 64bit aligned, which constrained the acceptable number of input features).

Fixes #1458.

With this change, it is sufficient to keep a single "load type" specialization for the search kernels, which shall cut the binary size by half (#1459).

### DIFF
--- a/cpp/include/raft/neighbors/cagra_types.hpp
+++ b/cpp/include/raft/neighbors/cagra_types.hpp
@@ -26,6 +26,7 @@
 #include <raft/core/resources.hpp>
 #include <raft/distance/distance_types.hpp>
 #include <raft/util/integer_utils.hpp>
+#include <raft/util/pow2_utils.cuh>
 
 #include <memory>
 #include <optional>
@@ -82,8 +83,6 @@ struct search_params : ann::search_params {
   /** Lower limit of search iterations. */
   size_t min_iterations = 0;
 
-  /** Bit length for reading the dataset vectors. 0, 64 or 128. Auto selection when 0. */
-  size_t load_bit_length = 0;
   /** Thread block size. 0, 64, 128, 256, 512, 1024. Auto selection when 0. */
   size_t thread_block_size = 0;
   /** Hashmap type. Auto selection when AUTO. */
@@ -113,6 +112,7 @@ static_assert(std::is_aggregate_v<search_params>);
  */
 template <typename T, typename IdxT>
 struct index : ann::index {
+  using AlignDim = raft::Pow2<16 / sizeof(T)>;
   static_assert(!raft::is_narrowing_v<uint32_t, IdxT>,
                 "IdxT must be able to represent all values of uint32_t");
 
@@ -124,12 +124,15 @@ struct index : ann::index {
   }
 
   // /** Total length of the index. */
-  [[nodiscard]] constexpr inline auto size() const noexcept -> IdxT { return dataset_.extent(0); }
+  [[nodiscard]] constexpr inline auto size() const noexcept -> IdxT
+  {
+    return dataset_view_.extent(0);
+  }
 
   /** Dimensionality of the data. */
   [[nodiscard]] constexpr inline auto dim() const noexcept -> uint32_t
   {
-    return dataset_.extent(1);
+    return dataset_view_.extent(1);
   }
   /** Graph degree */
   [[nodiscard]] constexpr inline auto graph_degree() const noexcept -> uint32_t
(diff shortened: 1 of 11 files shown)


tfeher:
Marked as breaking change because it removes the `load_bit_length` search parameter.

tfeher:
There is still a bug with the new tests, otherwise it is ready for review.

enp1s0:
The solution of the compilation error is:

https://github.com/tfeher/raft/blob/fb175783a6cb6b9d0e56f0c18bd84c269fce9bb1/cpp/include/raft/neighbors/detail/cagra/search_multi_cta.cuh#L231
```diff
231:      dataset_size,
+         dataset_ld,
232:      result_buffer_size,
```

enp1s0:
It would be better to support a padded dataset in the argument of `cagra::sort_knn_graph`. What do you think?

enp1s0 (commented):
The changes look good to me.

One comment:
It would be better to support a padded dataset in the argument of `cagra::sort_knn_graph`. What do you think?

cjnolet:
@enp1s0 please fix the conflicts and we can merge this in for release 23.06.

tfeher:
I have fixed a bug in the serialization routine, and added more tests. @enp1s0 could you have a quick look again at the changes?

> It would be better to support a padded dataset in the argument of cagra::sort_knn_graph. What do you think?

I think it is a good suggestion, and it might be also useful to allow the constructor of index to accept padded dataset (i.e. strided mdspan). I would prefer to make these changes in a follow up PR.

enp1s0:
Thank you, @tfeher, for fixing it and adding tests. The code looks good to me.

enp1s0 (approved):
Thank you, @tfeher, for fixing it and adding tests. The code looks good to me.

cpp/include/raft/neighbors/detail/cagra/cagra_serialize.cuh
```diff
@@ -59,7 +60,19 @@ void serialize(raft::resources const& res, std::ostream& os, const index<T, IdxT
...
+  RAFT_CUDA_TRY(cudaMemcpy2DAsync(host_dataset.data_handle(),
+                                  sizeof(T) * host_dataset.extent(1),
+                                  dataset.data_handle(),
+                                  sizeof(T) * dataset.stride(0),
+                                  sizeof(T) * host_dataset.extent(1),
+                                  dataset.extent(0),
+                                  cudaMemcpyDefault,
+                                  resource::get_cuda_stream(res)));
+  resource::sync_stream(res);
+  serialize_mdspan(res, os, host_dataset.view());
```
cjnolet, line 75: Mostly a side question, but are we still planning to remove the dataset from the serialization in a future change?

cpp/include/raft/neighbors/cagra_types.hpp
```diff
@@ -179,15 +183,36 @@ struct index : ann::index {
...
-               dataset.size(),
-               resource::get_cuda_stream(res));
+    if (dataset_.extent(1) == dataset.extent(1)) {
+      raft::copy(dataset_.data_handle(),
+                 dataset.data_handle(),
+                 dataset.size(),
+                 resource::get_cuda_stream(res));
+    } else {
+      // copy with padding
+      RAFT_CUDA_TRY(cudaMemsetAsync(
```
cjnolet, line 199: If this is going to be a more common practice, I wonder if we should consider centralizing this somewhere eventually. Probably doesn't need to be done yet, or even in this PR, though.

cpp/include/raft/neighbors/detail/cagra/search_multi_cta.cuh
```diff
@@ -569,7 +546,7 @@ struct search : public search_plan_impl<DATA_T, INDEX_T, DISTANCE_T> {
   ~search() {}
 
   void operator()(raft::resources const& res,
-                  raft::device_matrix_view<const DATA_T, INDEX_T, row_major> dataset,
+                  raft::device_matrix_view<const DATA_T, INDEX_T, layout_stride> dataset,
```
cjnolet, line 549: Would there be any benefit to using a  [padded layout](https://github.com/rapidsai/raft/blob/567dfd74a5c092cec61de6b52ea239ea91b2ed46/cpp/include/raft/core/mdspan.hpp#L58) here or having an overload for it in the public API just to simplify the conversion?

cjnolet (approved):
LGTM. Just a couple questions and a suggestion for a future improvement overall. Since CAGRA is still experimental, we have some leeway for API changes.

PR #1507, opened 2023-05-10
Adding bfknn and ivf-pq python api to docs

### LINKED ISSUE #1506: [DOC] IVF-PQ Python docs broken and IVF-Flat needs to be added
@tfeher pointed out that the IVF-PQ python docs aren't rendering correctly so we should fix this. While we're in there, the ivf-flat python API needs to be added to the pylibraft.neighbors API.

cjnolet:
Closes #1506. The IVF-PQ docs seem to be rendering fine for me locally. I'm still unsure why they aren't rendering in the nightlies.


benfred (approved):

benfred (approved):

ajschmidt8 (approved):

PR #1508, opened 2023-05-10
Workaround for cuda 12 issue in cusparse

cjnolet:


divyegala (approved):

cpp/include/raft/spectral/detail/matrix_wrappers.hpp
```diff
@@ -241,6 +244,8 @@ struct sparse_matrix_t {
     RAFT_CUSPARSE_TRY(raft::sparse::detail::cusparsespmv(
       cusparse_h, trans, &alpha, matA, vecX, &beta, vecY, spmv_alg, external_buffer.raw(), stream));
 
+    // FIXME: This is a workaround for a cusparse issue being encountered in CUDA 12
```
benfred, line 247: Is there a link to a github issue for this problem that we can add here?

benfred (approved):

cjnolet:
There's not but I'd like to add that in a follow-on. This PR is just applying a patch from Paul Taylor that we had to use for the DLFW containers to workaround a gnarly cuda bug. It's currently holding up cugraph.

cjnolet:
Partially addresses #1509

Issue #1357, opened 2023-03-20
[BUG] Misleading signature of `ivf_pq::search`


achirkin:
The [raw pointer interface](https://github.com/rapidsai/raft/blob/81fd1e0f87f50f6501cdbd1e9b95baea35a1bb4b/cpp/include/raft/neighbors/ivf_pq.cuh#L342-L353) and internals of `ivf_pq::search` support only up to `uint32_t` number of queries, probes, `k`, and dimensionality. Yet, the new [mdspan api](https://github.com/rapidsai/raft/blob/81fd1e0f87f50f6501cdbd1e9b95baea35a1bb4b/cpp/include/raft/neighbors/ivf_pq.cuh#L160-L165) uses `IdxT` for the extents. I suggest to change that to `uint32_t` (the second (index) parameter of every mdspan).

PR #1513, opened 2023-05-14
Introduce sample filtering to IVFPQ index search

alexanderguzhva:
A prototype that introduces a per-sample filtering for IVFPQ search. Please feel free to use it as a foundation for the future change, if appropriate, because the code is functional, but is not super clean-and-neat.

The diff introduces a template parameter called `SampleFilterT`. An instance is expected 
* to be `SampleFilterT()` constructible (which was mostly needed to define a default behavior in the form of `SampleFilterT sample_filter=SampleFilterT()`, see below)
* to provide a `inline __device__ bool operator(...)`  that returns `true` is a given sample is valid for being used against a given query in IVFPQ search

The default filter (that I set as a default one in certain facilities in the form of `typename SampleFilterT = NoneSampleFilter` in order not to modify way to many files) allows all samples to be used:
```
struct NoneSampleFilter {
  inline __device__ __host__ bool operator()(
    // query index
    const uint32_t query_ix,
    // the current inverted list index
    const uint32_t cluster_ix,
    // the index of the current sample inside the current inverted list
    const uint32_t sample_ix
  ) const {
    return true;
  }
};
```
Here `__host__` is needed for a CPU-based testing only.
Also, I've provided an implementation of `BitMaskSampleFilter` that allows to filter samples based on a bit mask, as an example. The implementation was tested in the semi-production environment.

All the filter-related files were added to `cpp/include/raft/neighbors/detail/sample_filter.cuh`.

I did not change the default `ivf_pq_search()` method remains unchanged, but one more `ivf_pq_search_with_filtering()` method with an additional template argument `SampleFilterT` and one more input parameter was introduced.
```
template <typename T, typename IdxT, typename SampleFilterT>
void search_with_filtering(raft::device_resources const& handle,
            const raft::neighbors::ivf_pq::search_params& params,
            const index<IdxT>& idx,
            const T* queries,
            uint32_t n_queries,
            uint32_t k,
            IdxT* neighbors,
            float* distances,
            rmm::mr::device_memory_resource* mr = nullptr,
            SampleFilterT sample_filter = SampleFilterT());
```

All the current instantiations use `NoneSampleFilter` only.

I've used `SampleFilterT sample_filter` parameters passing instead of `const SampleFilterT sample_filter` in the function calls in order to be able to add some debugging facilities to a filter and with the hope that the compiler is smart enough to understand the de-facto constness if needed.

The filter does not take a computed distance score into account by design, thus the current implementation cannot have a distance threshold. This can be easily changed, if appropriate.

It is still questionable to me whether this filtering needs to be injected right inside the search kernel instead of doing post-processing, please let me know if you have any thoughts on the topic.

I'm happy to address the comments.

Thanks.


cpp/include/raft/neighbors/detail/ivf_pq_compute_similarity-inl.cuh
```diff
@@ -448,11 +456,18 @@ __global__ void compute_similarity_kernel(uint32_t n_rows,
       OutT score = kDummy;
       bool valid = i < n_samples;
       if (valid) {
```
achirkin, line 458: A neatpick: maybe put `valid && sample_filter(query_ix, label, i)` together to reduce the nesting level?

cpp/include/raft/neighbors/detail/ivf_pq_compute_similarity-inl.cuh
```diff
@@ -221,6 +222,11 @@ __device__ auto ivfpq_compute_score(uint32_t pq_dim,
  * @param index_list
  *   An optional device pointer to the enforced order of search [n_queries, n_probes].
  *   One can pass reordered indices here to try to improve data reading locality.
+ * @param query_kth
+ *   (TODO)
```
achirkin, line 226: Thanks for spotting this!

> query_kths keep the current state of the filtering - atomically updated distances to the k-th closest neighbors for each query [n_queries].

cpp/include/raft/neighbors/detail/ivf_pq_compute_similarity-inl.cuh
```diff
@@ -448,11 +456,18 @@ __global__ void compute_similarity_kernel(uint32_t n_rows,
...
-          pq_dim,
-          reinterpret_cast<const vec_t::io_t*>(pq_thread_data),
-          lut_scores,
-          early_stop_limit);
+        // Use the filter.
+        // The purpose of the filter is to decide whether a certain sample acceptable
+        // for a given query.
+        // Technically, a more sophisticated filter may be used that takes
+        // the computed score into account.
+        if (sample_filter(query_ix, label, i)) {
```
achirkin, line 464: Are you sure we must have `query_ix` as a parameter of the filter? We either need to remove this parameter or make sure we add the offset into the current batch  https://github.com/rapidsai/raft/blob/da7538c650d9f99495a83d961e845b033b7a20c4/cpp/include/raft/neighbors/detail/ivf_pq_search.cuh#LL833C23-L833C40  (offset_q + offset_b)

cpp/include/raft/neighbors/detail/sample_filter.cuh
```diff
@@ -0,0 +1,115 @@
+/*
+ * Copyright (c) 2022-2023, NVIDIA CORPORATION.
```
achirkin, line 2:
```suggestion
 * Copyright (c) 2023, NVIDIA CORPORATION.
```
(new file)

achirkin (changes requested):
Thanks for the PR! The change looks good and neat. Though I think a few things require further discussion.

cpp/include/raft/neighbors/detail/ivf_pq_compute_similarity-inl.cuh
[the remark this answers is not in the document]
alexanderguzhva: Tbh, I don't understand your comment about the security risk.

cpp/include/raft/neighbors/detail/ivf_pq_compute_similarity-inl.cuh
alexanderguzhva: I'll update the PR.

cpp/include/raft/neighbors/detail/ivf_pq_compute_similarity-inl.cuh
alexanderguzhva: I'll update the PR.

cpp/include/raft/neighbors/detail/ivf_pq_compute_similarity-inl.cuh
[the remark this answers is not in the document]
achirkin: I mean if filtering is used to suppress some sensitive data for the end user for a search query?..

cpp/include/raft/neighbors/detail/sample_filter.cuh
```diff
@@ -0,0 +1,115 @@
...
+      const uint32_t cluster_ix,
+      const uint32_t sample_ix) const {
+    const index_type database_idx = inds_ptr[cluster_ix][sample_ix];
+    const uint64_t bit_mask_element =
+        bit_mask_ptr[query_ix * bit_mask_stride_64 + database_idx / 64];
+    const uint64_t masked_bool =
+        bit_mask_element & (1ULL << (uint64_t)(database_idx % 64));
+    const bool is_bit_set = (masked_bool != 0);
+
+    return is_bit_set;
```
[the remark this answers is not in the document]
alexanderguzhva: I can remove this filter from the PR, because I've supplied it as a sort of demonstration, and we'll just reuse it on our end without littering. Or you may wish me to put this code into the comments together with `IndexSampleFilter` right above.  My purpose was to kinda confirm that this use case won't trigger some serious doubts.

In reality, we may wish to use Bloom filters instead of bit masks. but it needs to be analyzed for the performance first.

cpp/include/raft/neighbors/detail/ivf_pq_compute_similarity-inl.cuh
alexanderguzhva: Definitely, `offset_q + offset_b` needs to be propagated! I'll update the PR.

cpp/include/raft/neighbors/detail/sample_filter.cuh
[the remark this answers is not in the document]
alexanderguzhva: Moved `BitMaskSampleFilter` to comments

alexanderguzhva:
@achirkin I've made the changes to the PR, please feel free to take a look whenever you have time. Thanks!

achirkin:
Thanks, @alexanderguzhva. In the meanwhile, could you please elaborate a bit on the need for the `query_ix` parameter in the filtering? My concern is that the extra kernel parameter (query offset) increases the register pressure even if we don't do any filtering.

alexanderguzhva:
> Thanks, @alexanderguzhva. In the meanwhile, could you please elaborate a bit on the need for the `query_ix` parameter in the filtering? My concern is that the extra kernel parameter (query offset) increases the register pressure even if we don't do any filtering.

Well, the use case I'm interested in is when every query uses its own set of allowed samples. This is the requirement from the interested party.

alexanderguzhva:
Anything needs to be done on my end so far? Thanks

alexanderguzhva:
> /ok to test

@cjnolet , do you have any handy script for running a style check locally? Thanks!

cjnolet:
@alexanderguzhva we have pre-commit set up on the repository so it should do all the clang-formatting for you upon committing to git. Here's the instructions to install / configure it: https://docs.rapids.ai/api/raft/nightly/contributing/#python-pre-commit-hooks.

alexanderguzhva:
@cjnolet Thanks a lot!

cpp/include/raft/neighbors/detail/ivf_pq_search.cuh
```diff
@@ -712,16 +734,22 @@ inline auto get_max_batch_size(uint32_t k,
 }
 
 /** See raft::spatial::knn::ivf_pq::search docs */
+<<<<<<< HEAD
```
cjnolet, line 737: I think there might be some unmerged bits here.

alexanderguzhva: Yep, I'll fix that, thanks!

cpp/include/raft/neighbors/ivf_pq-inl.cuh
```diff
@@ -156,14 +156,16 @@ void extend(raft::resources const& handle,
...
-            raft::device_matrix_view<const T, IdxT, row_major> queries,
-            raft::device_matrix_view<IdxT, IdxT, row_major> neighbors,
-            raft::device_matrix_view<float, IdxT, row_major> distances)
+template <typename T, typename IdxT, typename SampleFilterT>
+void search_with_filtering(raft::resources const& handle,
+                           const search_params& params,
+                           const index<IdxT>& idx,
+                           raft::device_matrix_view<const T, IdxT, row_major> queries,
+                           raft::device_matrix_view<IdxT, IdxT, row_major> neighbors,
+                           raft::device_matrix_view<float, IdxT, row_major> distances,
```
achirkin, line 167: Could you please change the index type of the three arguments spans to `uint32_t` in the new api functions? We were meant to fix this for the existing api as well (https://github.com/rapidsai/raft/issues/1357) but haven't done this yet.

achirkin (approved):
I've compared the prim benchmarks of this PR (with `NoneSampleFilter`) against the current 23.06. There seem to be no performance hit whatsoever. The ncu stats, such as the number of registers stay the same as well, despite having more arguments in the signature. That's an LGTM from the performance side of things!

cpp/include/raft/neighbors/ivf_pq-inl.cuh
alexanderguzhva: @achirkin do you want me to change `template<typename T, typename IdxT, typename SampleFilterT> void search_with_filtering()` to something like `template<typename T, typename IdxT = uint32_t, typename SampleFilterT = detail::NoneSampleFilter> void search_with_filtering()`? I mean I can do it in theory, but it is a separate problem that needs to be addressed separately in a separate PR.
@cjnolet What do you think?

cjnolet: The `index_type` template parameter in the `device_matrix_view` is a little misleading because it only applies to the precision used to represent each extent in the underlying mdspan itself and shouldn't really be coupled to the indexing type in the resulting `indices` matrix. Given that this is already a problem that we're planning to address, I'm okay merging these changes as-is for now and cleaning this up when we clean up the others.

cjnolet (approved):

cjnolet (approved):

cjnolet:
Thanks again for contributing this feature @alexanderguzhva!

cpp/include/raft/neighbors/ivf_pq-inl.cuh
achirkin: @alexanderguzhva sorry I didn't clarify. No, as Corey noted I meant to just change the second template parameter in each of these three mdspans to `uint32_t`, because the underlying raw pointer api accepts only 32-bit dimension parameters (`n_queries`, `k`). Don't worry though, I'll add this shortly in a follow-up https://github.com/rapidsai/raft/pull/1539/files#diff-8b0311b97cc4f377506ed342bbe29dc3ac692304f8590feb3397926e8613a658R165-R167

PR #1514, opened 2023-05-15
Support uint64_t in CAGRA index data type

enp1s0:
This PR updates CAGRA::search to support `uint64_t` in the index data type. This update is required to implement the RAFT ANN benchmark for CAGRA.


enp1s0:
Currently, some unit tests fail because recalls of them are smaller than the threshold.

enp1s0:
TODO:

- hash table
- mottainai bit

enp1s0:
@tfeher Can you check the code and let me know if there are any problems?

tfeher:
Thanks @enp1s0 for this PR! I think it is very useful to update the code to work with other index type (even if we would mainly use with 32bit indices). I will have a look at the changes.

cpp/include/raft/neighbors/detail/cagra/search_multi_kernel.cuh
```diff
@@ -404,11 +406,14 @@ __global__ void remove_parent_bit_kernel(const std::uint32_t num_queries,
                                          INDEX_T* const topk_indices_ptr,  // [ld, num_queries]
                                          const std::uint32_t ld)
 {
+  constexpr INDEX_T index_msb_1_mask = static_cast<INDEX_T>(1)
```
tfeher, line 409: Instead of repeating this definition could reuse the definition, e.g. by introducing a templated struct with a constexpr mask.

cpp/include/raft/neighbors/detail/cagra/search_single_cta.cuh
```diff
@@ -51,6 +51,8 @@ __device__ void pickup_next_parents(std::uint32_t* const terminate_flag,
                                     const std::size_t dataset_size,
                                     const std::uint32_t num_parents)
 {
+  constexpr INDEX_T index_msb_1_mask = static_cast<INDEX_T>(1)
```
tfeher, line 54: Same as above, we should have a reusable expression.

cpp/test/CMakeLists.txt
```diff
@@ -316,6 +316,9 @@ if(BUILD_TESTS)
     test/neighbors/ann_cagra/test_float_uint32_t.cu
     test/neighbors/ann_cagra/test_int8_t_uint32_t.cu
     test/neighbors/ann_cagra/test_uint8_t_uint32_t.cu
+    test/neighbors/ann_cagra/test_float_uint64_t.cu
+    test/neighbors/ann_cagra/test_int8_t_uint64_t.cu
+    test/neighbors/ann_cagra/test_uint8_t_uint64_t.cu
```
tfeher, line 321: Adding these effectively doubles our [test binary size](https://downloads.rapids.ai/ci/raft/pull-request/1514/e790e82/cuda11_x86_64.tests_bench_prims.html):

File | Compile time | Size
-- | -- | --
test_float_uint32_t.cu.o | 43:55 min | 106.657 MB
test_float_uint64_t.cu.o | 42:24 min | 98.597 MB
test_uint8_t_uint32_t.cu.o | 41:31 min | 117.630 MB
test_int8_t_uint32_t.cu.o | 41:29 min | 116.698 MB
test_int8_t_uint64_t.cu.o | 40:04 min | 106.032 MB
test_uint8_t_uint64_t.cu.o | 39:56 min | 107.312 MB

See https://github.com/rapidsai/raft/pull/1499#issuecomment-1541583762 for potential impact on packages.

(Note that compile time is improved by #1428 by enabling parallel build for different template parameters, but binary size is not changed there.

cpp/test/neighbors/ann_cagra/test_float_uint64_t.cu
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+#include <gtest/gtest.h>
+
+#undef RAFT_EXPLICIT_INSTANTIATE_ONLY
+#include "../ann_cagra.cuh"
+
+namespace raft::neighbors::experimental::cagra {
+
+typedef AnnCagraTest<float, float, std::uint64_t> AnnCagraTestF_I64;
```
tfeher, line 24: Other ANN methods use `int64_t`, and it would be preferred to use the same here. Do we have any restrictions on signed vs non signed index types?

tfeher (changes requested):
Thanks @enp1s0 for this PR. It is great to have the freedom to customize the index type paramater.  I am somewhat concerned about the test size and compile time maybe it would be sufficient to test limited template type combinations in int64 mode? (In practice it is expected that `uint32_t` would be sufficient in most of the cases.)

cpp/include/raft/neighbors/detail/cagra/search_multi_kernel.cuh
enp1s0: I have modified it as you mentioned!

enp1s0:
> it would be sufficient to test limited template type combinations in int64 mode?

Yes, I agree with you. I'll remove the (`DataT`, `IdxT`) = (`int8`, `uint64`) and (`uint8`, `uint64`) tests.

enp1s0:
@tfeher I have a question about supporting signed 64-bit int (`int64_t`) as the index data type.
Currently, `int64_t` does not work as expected, with the recall becoming zero when using it. I'm not sure of the cause, but some parts of the current CAGRA kernels assume the index data type is unsigned. While it is true that the index is always positive, I understand that it is preferred to support signed data types to have freedom in the index data types.

While it is possible to modify the kernels to support int64_t and make them work as expected, I prefer an alternative approach: using kernel functions designed for unsigned index data types and using `reinterpret_cast` to treat the signed types as unsigned. The advantages of this implementation are as follows:

1. We can use the same kernel functions for both signed and unsigned index types, which helps to keep the binary size of the library from increasing.
1. We can avoid the need for additional out-of-range memory access checks for signed types.

This implementation should not have any disadvantages because, if the data type is signed, the index is always positive, and the range of representation for unsigned types is larger than that of signed types.

What do you think about it?

cpp/test/CMakeLists.txt
enp1s0: I have removed `test_int8_t_uint64_t` and `test_uint8_t_uint64_t`

cpp/test/neighbors/ann_cagra/test_float_uint64_t.cu
enp1s0: I have changed it to `int64_t`.

enp1s0:
@tfeher I have fixed the codes! Can you look at them again?

cpp/include/raft/neighbors/cagra.cuh
```diff
@@ -124,7 +134,20 @@ void sort_knn_graph(raft::resources const& res,
                     mdspan<const DataT, matrix_extent<IdxT>, row_major, d_accessor> dataset,
                     mdspan<IdxT, matrix_extent<IdxT>, row_major, g_accessor> knn_graph)
 {
-  detail::graph::sort_knn_graph(res, dataset, knn_graph);
+  using internal_IdxT = typename std::make_unsigned<IdxT>::type;
+
+  using g_accessor_internal =
+    host_device_accessor<std::experimental::default_accessor<internal_IdxT>, memory_type::host>;
```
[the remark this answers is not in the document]
enp1s0: Thanks @tfeher for the suggestion. I have adopted it.

Issue #1343, opened 2023-03-14
[BUG] IVF-FLAT OOM for large batch size


tfeher:
**Describe the bug**

IVF-flat can run out of memory if n_queries * n_probes is very large.

This is happens because we try to process all the query vectors at once, and we allocate a large workspace: 
https://github.com/rapidsai/raft/blob/667d873876fb1bb8669a2b9a5d1105b5ad67ca6e/cpp/include/raft/spatial/knn/detail/ivf_flat_search.cuh#L1290

**Additional context**
In contrast IVF-PQ processes the queries batch-wise, and therefore avoids OOM.
- Set workspace size to be max 1GiB
https://github.com/rapidsai/raft/blob/667d873876fb1bb8669a2b9a5d1105b5ad67ca6e/cpp/include/raft/neighbors/detail/ivf_pq_search.cuh#L1536-L1543
- process queries batch_wise https://github.com/rapidsai/raft/blob/667d873876fb1bb8669a2b9a5d1105b5ad67ca6e/cpp/include/raft/neighbors/detail/ivf_pq_search.cuh#L1655
- #1310

PR #1515, opened 2023-05-16
ivf-flat: limit the workspace size of the search via batching

achirkin:
closes https://github.com/rapidsai/raft/issues/1343

### DIFF
--- a/cpp/include/raft/neighbors/detail/ivf_flat_search-inl.cuh
+++ b/cpp/include/raft/neighbors/detail/ivf_flat_search-inl.cuh
@@ -213,22 +213,33 @@ inline void search(raft::device_resources const& handle,
                "n_probes (number of clusters to probe in the search) must be positive.");
   auto n_probes = std::min<uint32_t>(params.n_probes, index.n_lists());
 
-  auto pool_guard = raft::get_pool_memory_resource(mr, n_queries * n_probes * k * 16);
+  // a batch size heuristic: try to keep the workspace within the specified size
+  constexpr uint32_t kExpectedWsSize = 1024 * 1024 * 1024;
+  const uint32_t max_queries =
+    std::min<uint32_t>(n_queries,
+                       raft::div_rounding_up_safe<uint64_t>(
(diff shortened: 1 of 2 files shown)


cpp/include/raft/neighbors/detail/ivf_flat_search-inl.cuh
```diff
@@ -213,22 +213,33 @@ inline void search(raft::device_resources const& handle,
                "n_probes (number of clusters to probe in the search) must be positive.");
   auto n_probes = std::min<uint32_t>(params.n_probes, index.n_lists());
 
-  auto pool_guard = raft::get_pool_memory_resource(mr, n_queries * n_probes * k * 16);
+  // a batch size heuristic: try to keep the workspace within the specified size
+  constexpr uint32_t kExpectedWsSize = 1024 * 1024 * 1024;
+  const uint32_t max_queries =
+    std::min<uint32_t>(n_queries,
+                       raft::div_rounding_up_safe<uint64_t>(
+                         kExpectedWsSize, 16ull * uint64_t{n_probes} * k + 4ull * index.dim()));
```
tfeher, line 221: Why 16 and 4? Are these related to the date / index types?

tfeher (commented):
Thanks Artem for this fix! Just one question regarding the size heuristic.

cpp/include/raft/neighbors/detail/ivf_flat_search-inl.cuh
achirkin: Yes, that's very approximate heuristic for the allocated buffers inside search_impl<>. I figured there's no point in calculating these precise because (1) batching it does not seem to affect performance much (2) we plan to work on our workspace size management soon anyway.

tfeher: Sounds good, thanks!

tfeher (approved):
Thanks Artem for this fix! Just one question regarding the size heuristic.

PR #1516, opened 2023-05-16
Python API for IVF-Flat serialization

### LINKED ISSUE #752: [FEA] IVF-flat serialization

tfeher:
This PR adds Python API for IVF-Flat serialization.

closes #752


python/pylibraft/pylibraft/neighbors/ivf_flat/ivf_flat.pyx
```diff
@@ -708,3 +708,141 @@ def search(SearchParams search_params,
...
+        c_ivf_flat.serialize(deref(handle_), c_filename, deref(idx_int8.index))
+    elif index.active_index_type == "ubyte":
+        idx_uint8 = index
+        c_ivf_flat.serialize(deref(handle_), c_filename, deref(idx_uint8.index))
+    else:
+        raise ValueError("Index dtype %s not supported" % index.active_index_type)
+
+
+@auto_sync_handle
+def load(filename, dtype, handle=None):
```
tfeher, line 773: Note that IVF-Flat index has a dtype parameter. The default `serialize` / `deserialize` functions do not save this information, therefore we need an extra user arg. (Well, actually it is saved in the numpy header of the dataset mdspan, but it would be somewhat hacky to load that information).

Alternatively, we could modify `save` / load` `to serialize dtype information as well. Do we need such convenience function?

cjnolet: I think serializing this with the index makes the most sense. Then the user wouldn't have to know or even care what dtype was serialized- we can always give them back the correct index dtype.

tfeher: I have exposed a variant of the serialize function that uses a string as output. We save the dtype information + the serialized string on the Cython side.

python/pylibraft/pylibraft/neighbors/ivf_flat/ivf_flat.pyx
```diff
@@ -708,3 +708,157 @@ def search(SearchParams search_params,
...
+    elif index.active_index_type == "ubyte":
+        idx_uint8 = index
+        c_ivf_flat.serialize(
+            deref(handle_), c_string, deref(idx_uint8.index))
+    else:
+        raise ValueError(
+            "Index dtype %s not supported" % index.active_index_type)
+
+    dtype = np.dtype(index.active_index_type)
+    with open(filename, 'wb') as f:
```
cjnolet, line 777: Will this still work w/ the fully C++-based API as well? What happens if they save it in C++ but load it in Python (or vice versa?)

tfeher: No, that will not work. I have ignored that requirement. I see now that this is a problem. I have fixed that now.

cpp/src/raft_runtime/neighbors/ivf_flat_serialize.cu
```diff
@@ -0,0 +1,65 @@
...
+  };                                                                                   \
+                                                                                       \
+  void deserialize_file(raft::resources const& handle,                                 \
+                        const std::string& filename,                                   \
+                        raft::neighbors::ivf_flat::index<DTYPE, int64_t>* index)       \
+  {                                                                                    \
+    if (!index) { RAFT_FAIL("Invalid index pointer"); }                                \
+    *index = raft::neighbors::ivf_flat::deserialize<DTYPE, int64_t>(handle, filename); \
+  };                                                                                   \
+  void serialize(raft::resources const& handle,                                        \
```
tfeher, line 42: Note that this is not used by the current PR. But I think this can be used to fix https://github.com/rapidsai/cuml/issues/4743, that motivates to keep it in the PR.

cjnolet (approved):
LGTM!

PR #1522, opened 2023-05-17
[REVIEW] make raft::cache::Cache protected to allow overrides

mfoerste4:
We are planning to do some experiments with partial cache updates within cuml. In order to implement overrides for the given public functions and to allow re-usability for the cache internals we need to move the cache members to protected.

CC @cjnolet @tfeher


tfeher (approved):
Thanks Malte, LGTM.

PR #1523, opened 2023-05-17
Learn heuristic to pick fastest select_k algorithm

benfred:
This uses the select_k dataset from https://github.com/rapidsai/raft/pull/1497 to learn a heuristic of the fastest select_k variant based off the rows/ cols/k of the input. This heuristic is modelled as a DecisionTree, which is automatically exported in C++ code that is compiled into RAFT. This lets us learn a function to pick the fastest select_k method - which requires only a few if statements in C++ code, making it very cheap to evaluate.


cpp/include/raft/matrix/detail/select_k-inl.cuh
```diff
@@ -21,11 +21,101 @@
...
+
+/**
+ * Predict the fastest select_k algorithm based on the number of rows/cols/k
+ *
+ * The body of this method is automatically generated, using a DecisionTree
+ * to predict the fastest algorithm based off of thousands of trial runs
+ * on different values of rows/cols/k. The decision tree is converted to c++
+ * code, which is cut and paste below.
+ *
+ * The code to generate is in cpp/scripts/heuristics/select_k, running the
```
cjnolet, line 42: Just a tiny nitpick: 
```suggestion
 * NOTE: The code to generate is in cpp/scripts/heuristics/select_k, running the
```

cjnolet (approved):
LGTM. Just one tiny suggestion.

cpp/include/raft/matrix/detail/select_k-inl.cuh
```diff
@@ -21,11 +21,101 @@
...
+ * on different values of rows/cols/k. The decision tree is converted to c++
+ * code, which is cut and paste below.
+ *
+ * NOTE: The code to generate is in cpp/scripts/heuristics/select_k, running the
+ * 'generate_heuristic' notebook there will replace the body of this function
+ * with the latest learned heuristic
+ */
+inline Algo choose_select_k_algorithm(size_t rows, size_t cols, int k)
+{
+  if (k > 134) {
```
achirkin, line 48: I think, we'd better use log_2(k) instead of k when constructing the heuristic, so that all values of k go in powers of two. For all warp-based algorithms, performance for non-powers of two is equal to their rounded-up powers of two (queue capacity parameter).

PR #1525, opened 2023-05-18
ivf-flat: fix incorrect recomputed size of the index

achirkin:
Fix ivf-flat's `recompute_internal_state` incorrectly using the amortized list sizes to compute the index size.


enp1s0:
Hi @achirkin, the error in cpp-tests (TiledKNNTest) was caused by the issue mentioned in https://github.com/rapidsai/raft/issues/1526 . It has been fixed now, so please merge the latest branch-23.06 commits. Thanks!

cjnolet (approved):

PR #1536, opened 2023-05-19
Rename kernel arch finding function for dispatch

mdoijade:
-- as the kernel arch given by the cudaFuncAttribute ptxVersion depends on what archs the kernel was compiled for
we should renam kernel_runtime_arch() as kernel_virtual_arch().
-- accordingly update comments to reflect this.


cjnolet (approved):
LGTM!

PR #1537, opened 2023-05-19
Check python brute-force knn inputs

benfred:
The input validation code wasn't being triggered for the python bfknn api, causing invalid output when passed col-major inputs. Fix.


cjnolet (approved):
LGTM!

PR #1539, opened 2023-05-20
ivf-pq::search: fix the indexing type of the query-related mdspan arguments

achirkin:
closes https://github.com/rapidsai/raft/issues/1357

breaking change: the type of argument mdspans has slightly changed (second template parameter fixed to `uint32_t`)

### DIFF
--- a/cpp/bench/prims/neighbors/knn.cuh
+++ b/cpp/bench/prims/neighbors/knn.cuh
@@ -181,9 +181,10 @@ struct ivf_pq_knn {
   {
     search_params.n_probes = 20;
     auto queries_view =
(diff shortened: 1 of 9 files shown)


achirkin:
@cjnolet I'd like to squeeze this in 23.06 to have higher chances that nobody uses the original api. But this is a breaking change during burndown, so I surely don't mind moving it to 23.08 if you think it's more appropriate.

cjnolet:
@achirkin I'm okay putting this in 23.06. I'm sure it's fine but for completeness, can you please open a PR to test this change in cuml?

cjnolet (approved):

PR #1544, opened 2023-05-23
IVF-PQ tutorial notebook

achirkin:
Add a new folder `notebooks` and a notebook with an example of tweaking ivf-pq using a dataset from ann-benchmarks.com


tfeher:
Let's dust this off for the coming release. I think the only thing missing is the last section about the parameters for index building. Since we will have an IVF-Flat notebook that discusses clustering parameters (n-clusters, trainset ratio), we can focus here on additional parameters for the product quantization.

notebooks/tutorial_ivf_pq.ipynb
```diff
@@ -0,0 +1,1226 @@
...
+    "## Build the index\n",
+    "# This function takes a row-major either numpy or cupy (GPU) array.\n",
+    "# Generally, it's a bit faster with GPU inputs, but the CPU version may come in handy\n",
+    "# if the whole dataset cannot fit into GPU memory (or even CPU RAM -- use mmap for that).\n",
+    "index = ivf_pq.build(index_params, dataset, handle=resources)\n",
+    "# Sync the GPU to make sure we've got the timing right\n",
+    "resources.sync()\n",
+    "index"
+   ]
+  },
```
tfeher, line 253: After the index is built, please demonstrate how to save and load the index (reference: https://github.com/rapidsai/raft/blob/0abedc6661748bd64ff9828a86304dd29ae0ca05/python/pylibraft/pylibraft/test/test_ivf_pq.py#L518-L520)

notebooks/tutorial_ivf_pq.ipynb
```diff
@@ -0,0 +1,1226 @@
...
+     "output_type": "stream",
+     "text": [
+      "12.8 ms ± 1.51 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)\n",
+      "13.6 ms ± 1.73 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)\n",
+      "13.8 ms ± 926 ns per loop (mean ± std. dev. of 7 runs, 100 loops each)\n",
+      "14.1 ms ± 1.4 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)\n",
+      "14.8 ms ± 1.41 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)\n",
+      "18.1 ms ± 3.24 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)\n",
+      "19.9 ms ± 13.1 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)\n",
+      "30.6 ms ± 37.5 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)\n",
```
tfeher, line 437: Why this drop at k=128, is it an sisue of top-k kernel selection?

notebooks/tutorial_ivf_pq.ipynb
```diff
@@ -0,0 +1,1226 @@
...
+      "text/plain": [
+       "<Figure size 960x480 with 1 Axes>"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "bench_k = np.exp2(np.arange(10)).astype(np.int32)\n",
```
tfeher, line 454: I am not sure if we need a specific benchmark section here on k, It is good to point out in the text, that the number of neighbors are controlled by k, but actually benchmarking that could be postponed to the blog.

notebooks/tutorial_ivf_pq.ipynb
```diff
@@ -0,0 +1,1226 @@
...
+    "\n",
+    "#### n_lists\n",
+    "\n",
+    "`n_lists` is the first parameter to look at. It has a profound impact on overall performance during both training and search.\n",
+    "`n_lists` defines the number of clusters into which the index data is partitioned; you should keep this in mind when selecting the `n_probes` search parameter.\n",
+    "\n",
+    "The ratio `n_probes/n_lists` tells how large fraction of the dataset is compared to each query. If `n_lists == n_probes`, that is like a brute force search: we compare all dataset vectors to all query vectors. One would expect the recall is equal to `1` in such a case, but that does not take into account the PQ compression; in reality the recall is always lower unless you refine the search results.\n",
+    "\n",
+    "As `n_probes` approaches `n_lists`, IVF-PQ becomes slower than brute force because of all the extra work the algorithm does: dimension padding / transform, two-step search, extra PQ compute, etc. In practice searching around 1% of lists is enough for many datasets. But this depends on how well the input can be clustered. (e.g. for uniform random numbers as inputs, IVF methods don't work well).\n",
```
tfeher, line 853:
```suggestion
    "As `n_probes` approaches `n_lists`, IVF-PQ becomes slower than brute force because of all the extra work the algorithm does: dimension padding / transform, two-step search, extra PQ compute, etc. In practice searching around 0.1-1% of lists is enough for many datasets. But this depends on how well the input can be clustered. (e.g. for uniform random numbers as inputs, IVF methods don't work well).\n",
```

tfeher (commented):
Thanks Artem for the notebook! This looks great, a really detailed discussion of IVF-PQ search and training. While this content is great for the deep dive blog that we are planning, I am wondering whether this is the right level of  detail for the example notebook. I feel that there would be a value in making it more concise. Let's wait what the team says.

notebooks/tutorial_ivf_pq.ipynb
```diff
@@ -0,0 +1,1226 @@
+{
```
cjnolet, line 1: <p>Most notebooks that we publish for RAPIDS have longer descriptions before the first cell. I think this is a great lead-in sentence but can you please provide more details about what the notebook is going to do? Can you move the links and descriptions for the datasets up here as well so it's more immediately obvious where they came from and that the notebook is going to use them? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1544/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

cjnolet, line 1: <pre class="ql-syntax" spellcheck="false">Line #3.    import matplotlib.pyplot as plt
</pre><p>Please put all the includes together in this cell. This tells a user right away whether or not the notebook is going to work, rather than them getting to the 5th cell and realizing they don't have something installed, just to fail at 6th cell when they realize they didn't have something else. We are going to provide a separate conda environment yaml file for them to set up their environment for running notebooks as well, but having all the imports up front will immediately tell them whether the notebook is going to work. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1544/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

cjnolet, line 1: <pre class="ql-syntax" spellcheck="false">Line #2.    def show_properties(obj):
</pre><p>This is really nice. I almost wonder if this should eventually be added to pylibraft itself (and for the corresponding <code>__str__()</code> methods be overloaded to print it.  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1544/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

cjnolet, line 1: <p><code>RAFT</code> is an acronym. Please capitalize it.  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1544/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

cjnolet, line 1: <pre class="ql-syntax" spellcheck="false">Line #5.    # if the whole dataset cannot fit into GPU memory (or even CPU RAM -- use mmap for that).
</pre><p>It would help to either elaborate upon the <code>mmap</code> comment or remove it. I thin I understand what you are saying but I'm not expecting it to be immediately obvious to a user.   </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1544/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

cjnolet, line 1: <pre class="ql-syntax" spellcheck="false">Line #3.    # Sync the GPU to make sure we've got the timing right
</pre><p>I would propose maybe something more user friendly here like "This function is asychronous so we need to explicitly synchronize the GPU before we can measure the execution time". </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1544/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

cjnolet, line 1: <pre class="ql-syntax" spellcheck="false">Line #1.    ## Check the quality of the prediction (recall)
</pre><p>More user-friendly would be "We use recall to measure the quality of the prediction" (and then provide a link to wikipedia for recall). </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1544/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

cjnolet, line 1: <pre class="ql-syntax" spellcheck="false">Line #2.    def calc_recall(found_indices, ground_truth):
</pre><p>Side note: It would be nice to eventually see the recall computation from raft::stats exposed through pylibraft so we can just call that. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1544/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

cjnolet, line 1: <p>We should definitely be keeping the outputs in the checked in notebooks. We're going to potentially embed this into the docs so that users can still go through the notebooks even without running them. Can you also add a cell at the beginnign (maybe under the imports cell) to print <code>nvidia-smi</code>?This way the user has information about the gpu and architecture this was executed on. </p><p>  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1544/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

cjnolet, line 1: <p>I don't see QPS used anywhere above.. We should define it the first time we use it, just to be more straightforward for users. </p><p> </p><blockquote>Let's see how QPS (queries per second) depends on <code>k</code></blockquote><p><code>There's a couple typos in the original sentence too. </code> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1544/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

cjnolet, line 1: <blockquote>&nbsp;This is the core parameter control the QPS/recall trade-off</blockquote><p>Typo: This is the core parameter to control the QPS/recall trade-off"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1544/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='25'/>

cjnolet, line 1: <blockquote>Besides <code>n_probes</code>, <code>ivf_pq.SearchParams</code> contains couple more parameters, which affect internal workings of the algorithm. </blockquote><p>Besides <code>n_probes</code>, <code>ivf_pq.SearchParams</code> contains a couple more parameters, which affect the internal workings of the algorithm. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1544/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='29'/>

cjnolet, line 1: <p>Can we make this chart more readable please? Tou might be able to do it with some simple styling, but a couple of the data point labels are colliding and you can't read them. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1544/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='31'/>

cjnolet, line 1: <blockquote>Also, often you may see <code>16/16</code> version being faster than <code>16/8</code>. </blockquote><p>For user readability, can we please define what these ratios are depicting? You've got it in the code above but as a first-time reader of the notebook, it wasn't immediately obviously to me and it's simple enough to say something in the text like "You'll notice the datapoints are labeled as a/b, which corresponds to the number of bits from<code>internal_distance_dtype/lut_dtype</code> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1544/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='32'/>

cjnolet, line 1: <p>Again, these charts look pretty mangled with overlapping labels. Since the image is often encoded as base64, I'm assuming it'll look the same for the user as it does on ReviewNB. These should be styled a little bit or made more clear. Right now it just looks like a bunch of unreadable overlapping text. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1544/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='34'/>

cjnolet, line 1: <blockquote>we would pick three winning candidates along the Pareto curve:</blockquote><p>Please define <code>Parato curve</code>.This maybe obvious to us, but we're hoping to target users that may have no clue what terms like "QPS, recall, and Pareto frontier" mean.   </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1544/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='35'/>

cjnolet, line 1: <blockquote>Luckily training raft's ivf-pq does not take too much time thanks to using its fast implementation of Balanced Hierarchical K-Means clustering during build and an efficient encoding kernel. </blockquote><p>Maybe less sales pitchy to say something like:</p><p>"Since RAFT's IVF-PQ algorithm uses balanced-hierarchical k-means clustering and efficient logic for encoding, we find significantly improved index build times".</p><p> </p><blockquote> - <code>add_data_on_build</code> is a convenience flag that tells if you prefer to do both training (clustering) and filling the DB in one call to <code>ivf_pq.build</code> or start with an empty DB and add all data using <code>ivf_pq.extend</code> </blockquote><p>"when activated, automatically adds the training data to the index during <code>ivf_pq.build. </code>Otherwise, no data is added during <code>ivf_pq.build</code> and vectors need to be explicitly added to the index using  <code>ivf_pq.extend</code> "</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1544/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='38'/>

cjnolet, line 1: <blockquote>but that does not take into account the PQ compression; </blockquote><p>It would help to say "does not take into account the PQ compression, which is lossy."</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1544/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='39'/>

cjnolet, line 1: <blockquote>For the given data set (SIFT-128) and the selected parameters, one can see that 5000 clusters is clearly too much </blockquote><p>I think we should avoid assuming that the user reading this tutorial is familiar w/ reading qps@recall charts. Maybe instead of "one can see that 5000 cluster is clearly too much" we can say something "In this chart you can different queries per second for different levels of recall, calculated by training the index with different n_lists and querying with different numbers of probes. The charts demonstrate that 5000 lists is probably too many, since we see the qps for that curve doesn't reach as high as the others. We can also see that 100 lists is probably not enough, since we see its recall doesn't reach as high as the other curves". </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1544/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='42'/>

cjnolet, line 1: <blockquote>"This is useful when in conjugation with <code>add_data_on_build = True"</code></blockquote><p>I think "This can also be useful when used in conjunction with...". Or I would suggest just removing it all together. The add_data_on_build is surely a convenience, but just because it's set to false doesn't mean you can assume anything more about how many vectors are ultimately going to be added to the index. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1544/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='43'/>

cjnolet:
@achirkin ReviewNB didn't allow me to submit a message w/ the review. It's really a delight reading through the tutorial. It's very in-depth and I think it's going to be a great resource for users. I was able to make my way through most of it, but not all of it. I'm going to try and read through the rest next week.

notebooks/tutorial_ivf_pq.ipynb
achirkin: There seems to be a performance regression, which results in IVF-PQ in the range of `k = 65-128` being slower than for `k = 129...~400`. I observed this in internal benchmarks, and probably it's also manifested in https://github.com/rapidsai/raft/issues/1621

notebooks/tutorial_ivf_pq.ipynb
achirkin: <p>Thanks, I'm adding the <code>nvidia-smi</code> output; but I'm not sure what do you mean regarding keeping the outputs? Would you like me to print the neighbors/distances? </p>

achirkin: By this point in the notebook we've shown several QPS/Recall charts and gone rather deep into discussion of the performance implications of the search parameters. I think, it should be safe to assume that the reader, who is interested in such details, should already be familiar with the concept :) But I got the point, I'll try to put it a bit more verbose.

achirkin: This one is a funny typo :)
Regarding the usefulness: I meant to say that if `add_data_on_build = False`, then `kmeans_trainset_fraction` doesn't make much sense: a user can just pass a smaller dataset to the effect.
I've added a small note to clarify this.

achirkin:
@cjnolet, @tfeher many thanks for the reviews! While I still plan to do some minor changes, I think it's ready for another round.

notebooks/tutorial_ivf_pq.ipynb
cjnolet: <p>I just mean running the notebook and checking in the notebook w/ the outputs. Some libraries do a "clear outputs" before they check the notebooks in but I find having the outputs is useful to users, especially if they don't want to run the notebooks themselves. </p>

cjnolet: <p>I still don't see these defined anywhere. Can you expend a few words to explain what is the numerator and what is the denominator please?</p>

cjnolet: <p>It's definitely better. Thanks!</p>

notebooks/tutorial_ivf_pq.ipynb
```diff
@@ -0,0 +1,1345 @@
...
```
cjnolet, line 1: <p>Can we make the reference to <code>&nbsp;(see "Product quantization for nearest neighbor search" by Herve Jegou, Matthijs Douze, Cordelia Schmid). </code> into a link please? </p><p> </p><blockquote>), maps the vector y to the nearest cluster center. The number of&nbsp;clusters is n_lists. </blockquote><p>Please make the <code>n_lists</code> into an inline code sample for readability.  </p><p> </p><blockquote>  - a. "subspace" (C++ api: <code>codebook_gen::PER_SUBSPACE</code>): &nbsp;    creates <code>pq_dim</code> second-level quantizers - one for each slice of the data along features; &nbsp;  - b. "cluster" (C++ api: <code>codebook_gen::PER_CLUSTER</code>): &nbsp;    creates <code>n_lists</code> second-level quantizers - one for each first-level cluster. </blockquote><p>This isn't rendering as an enumerated list. Can you please explicitly add a line break?</p><p> </p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1544/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='49'/>

cjnolet:
@achirkin minor feedback from me at this point. End-to-end, this really is a great tutorial and I'm actually thinking we might want to consider keeping this as it is even after the blog is published (and maybe providing an additional notebook with more concise and simple usage examples for the blog?)

I think developers and users alike can benefit from this notebook as it gives a pretty thorough but readable overview of IVF-PQ. Thanks again for creating this!

notebooks/tutorial_ivf_pq.ipynb
achirkin: <p>I've added it directly to the plot, is that fine?</p>

cjnolet (approved):
Thanks Artem! I think this notebook is good to go for 23.08 and we can revisit some portions for 23.10. Overall I think it's a great deep dive into ivf pq.

PR #1546, opened 2023-05-23
Migrate wheel workflow scripts locally

divyegala:
Depends on https://github.com/rapidsai/shared-action-workflows/pull/97


ci/build_wheel.sh
```diff
@@ -0,0 +1,54 @@
...
+git diff
+
+cd "${package_dir}"
+
+# sccache configuration
+export SCCACHE_S3_KEY_PREFIX=gha-cibw
+export SCCACHE_REGION=us-east-2
+export SCCACHE_IDLE_TIMEOUT=32768
+export SCCACHE_BUCKET=rapids-sccache-east
+export SCCACHE_S3_USE_SSL=true
```
bdice, line 27: Use the env vars set upstream in the image. The only one you might need is `SCCACHE_S3_KEY_PREFIX`.

We use a different key prefix for each repo in conda -- e.g. `libraft-aarch64` rather than `gha-cibw`. Should we do the same here?

Upstream vars:
https://github.com/rapidsai/cibuildwheel-imgs/blob/fa583bdcb03693a1d8b77f96550b07cf5228dbd2/manylinux_v2_ubuntu/Dockerfile#L29-L37

conda just passes these through, only setting values for the key prefix:
https://github.com/rapidsai/raft/blob/42c9c180f2dcdb1437c45bfb0c753db49ec0a732/conda/recipes/libraft/meta.yaml#L33-L39

divyegala: This is a good point. I can set this up!

ci/build_wheel.sh
divyegala, line 1: Should this script be in `gha-tools` instead?

ci/build_wheel.sh
divyegala: Okay, I made this change. First run is running through now and it seems like we are not getting any `sccache` hits. Once I run again, we'll know if this worked.

vyasr: Removing these was the goal of https://github.com/rapidsai/cibuildwheel-imgs/pull/21 :slightly_smiling_face:

ci/build_wheel.sh
vyasr: It's worth considering whether we can standardize more than we currently do. The conda scripts are fairly duplicated across repos. I'd be more inclined to pull in a bunch of small tools rather than put this entire script up, though, especially since this script won't actually be common to all repos (e.g. pure wheel vs not).

ci/build_wheel.sh
```diff
@@ -0,0 +1,56 @@
...
+cd "${package_dir}"
+
+# sccache configuration
+if [ "${CI:-false}" = "false" ]; then
+  # Configure sccache for read-only mode since no credentials
+  # are available in local builds.
+  export SCCACHE_S3_NO_CREDENTIALS=true
+  export PARALLEL_LEVEL=${PARALLEL_LEVEL:-$(nproc)}
+  export RAPIDS_BUILD_TYPE=${RAPIDS_BUILD_TYPE:-"pull-request"}
+fi
```
vyasr, line 29: Maybe this could be a gha-tool?

ci/build_wheel.sh
```diff
@@ -0,0 +1,56 @@
...
+
+# Repair the wheel
+cd dist
+python -m auditwheel repair -w . ${package_name}*
+
+# Need to pick the final wheel out from all the dependencies and the
+# pre-repaired wheel.
+cd ..
+mkdir final_dist
+mv dist/${package_name}*manylinux* final_dist
```
vyasr, line 46: I believe I wrote this before I added the `--no-deps` to the pip wheel command above. With that, I think this may actually be much simpler now. Try replacing it with
```suggestion
mkdir -p final_dist
python -m auditwheel repair -w final_dist dist/*
```

vyasr: To be more precise you could still do `dist/${package_name}*` in the final line, but I don't think it'll be necessary anymore.

ci/build_wheel.sh
```diff
@@ -0,0 +1,56 @@
...
+
+# rapids-upload-wheels-to-s3 uses rapids-package-name which implicitly relies
+# on this variable being set
+export RAPIDS_PY_WHEEL_NAME="${package_name}_${RAPIDS_PY_CUDA_SUFFIX}"
+
+if [[ ! -d "/tmp/gha-tools" ]]; then
+  git clone https://github.com/divyegala/gha-tools.git -b wheel-local-runs /tmp/gha-tools
+fi
+
+/tmp/gha-tools/tools/rapids-upload-wheels-to-s3 final_dist
```
vyasr, line 56: Let's scope this a bit more tightly
```suggestion
if [[ ! -d "/tmp/gha-tools" ]]; then
  git clone https://github.com/divyegala/gha-tools.git -b wheel-local-runs /tmp/gha-tools
fi

RAPIDS_PY_WHEEL_NAME="${package_name}_${RAPIDS_PY_CUDA_SUFFIX}" /tmp/gha-tools/tools/rapids-upload-wheels-to-s3 final_dist
```

ci/build_wheel_pylibraft.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+# Copyright (c) 2023, NVIDIA CORPORATION.
+
+set -euo pipefail
+
+pyenv global ${RAPIDS_PY_VERSION}
```
vyasr, line 6: Now that we have one Python version per container I believe this is redundant.

ci/build_wheel_raft_dask.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+# Copyright (c) 2023, NVIDIA CORPORATION.
+
+set -euo pipefail
+
+pyenv global ${RAPIDS_PY_VERSION}
```
vyasr, line 6: Redundant?

ci/build_wheel.sh
```diff
@@ -0,0 +1,56 @@
...
+# Copyright (c) 2023, NVIDIA CORPORATION.
+
+set -euo pipefail
+
+package_name=$1
+package_dir=$2
+
+# Use gha-tools rapids-pip-wheel-version to generate wheel version then
+# update the necessary files
+RAPIDS_EPOCH_TIMESTAMP=$(date +%s)
```
vyasr, line 11: Hmm maybe we want this to be a bit more sophisticated. The question is, do we care if the different matrix builds for a given package produce slightly different versions? That's what will happen if we move the timestamp generation into the script. The purist in me says that's bad, although in practice it's probably not a huge deal either way.

ci/build_wheel.sh
```diff
@@ -0,0 +1,56 @@
...
+versioneer_override="$(rapids-pip-wheel-version ${RAPIDS_EPOCH_TIMESTAMP})"
```
vyasr, line 12: We don't use versioneer anymore. Let's take this opportunity to rename the variable to something like `version_override` instead.

ci/build_wheel.sh
```diff
@@ -0,0 +1,56 @@
...
+
+cd "${package_dir}"
+
+# sccache configuration
+if [ "${CI:-false}" = "false" ]; then
+  # Configure sccache for read-only mode since no credentials
+  # are available in local builds.
+  export SCCACHE_S3_NO_CREDENTIALS=true
+  export PARALLEL_LEVEL=${PARALLEL_LEVEL:-$(nproc)}
+  export RAPIDS_BUILD_TYPE=${RAPIDS_BUILD_TYPE:-"pull-request"}
```
vyasr, line 28: We may want to be a bit more flexible than this, see https://github.com/rapidsai/gha-tools/pull/62/

ci/build_wheel.sh
```diff
@@ -0,0 +1,56 @@
...
+fi
+export SCCACHE_S3_KEY_PREFIX="libraft-$(arch)"
+
+# Set up for pip installation of dependencies from the nightly index
+export PIP_EXTRA_INDEX_URL=https://pypi.k8s.rapids.ai/simple
```
vyasr, line 33: When we switch to the new index we should consider upstreaming this setting to the containers. Alternatively, we could upstream it now and just update the containers later. I'm fine with either approach. The latter is probably nicer for making this PR be an exemplar of future work. In either approach we may have a small hiccup during the switch, but I think that's fine.

vyasr (changes requested):
Looking good!

ci/build_wheel.sh
divyegala: I'm not able to understand how https://github.com/rapidsai/gha-tools/pull/62 will help here. That PR didn't set up anything for SCCACHE, did it?

ci/build_wheel.sh
divyegala: Yes, we can just upstream it now and export this variable to the environment. That way, when we do change the index we won't have to change every downstream repository.

ci/build_wheel.sh
divyegala: I can add this step back in the workflow, and then export this variable to the environment. Is that okay?

ci/build_wheel.sh
divyegala: `rapids-configure-local-sccache`?

ci/build_wheel.sh
divyegala: For containers actually running in CI, `cibuildhweel-imgs` or `ci-imgs` set up these variables to what we need them to be.

ci/build_wheel.sh
vyasr: I would probably name it `rapids-configure-sccache` since it'll run both locally and in CI, and it will internally flip the switch based on `${CI}`. Otherwise yes that sounds good.

ci/build_wheel.sh
vyasr: Yes that's fine with me.

ci/build_wheel.sh
vyasr: I think that makes sense, yes.

.github/workflows/build.yaml
```diff
@@ -66,15 +66,13 @@ jobs:
       run_script: "ci/build_docs.sh"
   wheel-build-pylibraft:
     secrets: inherit
-    uses: rapidsai/shared-action-workflows/.github/workflows/wheels-manylinux-build.yml@branch-23.08
+    uses: rapidsai/shared-action-workflows/.github/workflows/wheels-build.yaml@local-script
```
vyasr, line 69:
```suggestion
    uses: rapidsai/shared-action-workflows/.github/workflows/wheels-build.yaml@branch-23.08
```

.github/workflows/build.yaml
```diff
@@ -88,15 +86,13 @@ jobs:
   wheel-build-raft-dask:
     needs: wheel-publish-pylibraft
     secrets: inherit
-    uses: rapidsai/shared-action-workflows/.github/workflows/wheels-manylinux-build.yml@branch-23.08
+    uses: rapidsai/shared-action-workflows/.github/workflows/wheels-build.yaml@local-script
```
vyasr, line 89:
```suggestion
    uses: rapidsai/shared-action-workflows/.github/workflows/wheels-build.yaml@branch-23.08
```

.github/workflows/pr.yaml
```diff
@@ -67,12 +67,10 @@ jobs:
   wheel-build-pylibraft:
     needs: checks
     secrets: inherit
-    uses: rapidsai/shared-action-workflows/.github/workflows/wheels-manylinux-build.yml@branch-23.08
+    uses: rapidsai/shared-action-workflows/.github/workflows/wheels-build.yaml@local-script
```
vyasr, line 70:
```suggestion
    uses: rapidsai/shared-action-workflows/.github/workflows/wheels-build.yaml@branch-23.08
```

.github/workflows/pr.yaml
```diff
@@ -85,13 +83,10 @@ jobs:
   wheel-build-raft-dask:
     needs: wheel-tests-pylibraft
     secrets: inherit
-    uses: rapidsai/shared-action-workflows/.github/workflows/wheels-manylinux-build.yml@branch-23.08
+    uses: rapidsai/shared-action-workflows/.github/workflows/wheels-build.yaml@local-script
```
vyasr, line 86:
```suggestion
    uses: rapidsai/shared-action-workflows/.github/workflows/wheels-build.yaml@branch-23.08
```

vyasr (approved):
Just need to update the workflow names then this is good to go I think.

ci/build_wheel.sh
```diff
@@ -0,0 +1,30 @@
...
+source rapids-configure-sccache
+source rapids-date-string
+
+# Use gha-tools rapids-pip-wheel-version to generate wheel version then
+# update the necessary files
+version_override="$(rapids-pip-wheel-version ${RAPIDS_DATE_STRING})"
+
+RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"
+
+bash ci/release/apply_wheel_modifications.sh ${version_override} "-${RAPIDS_PY_CUDA_SUFFIX}"
```
ajschmidt8, line 18:
```suggestion
ci/release/apply_wheel_modifications.sh ${version_override} "-${RAPIDS_PY_CUDA_SUFFIX}"
```

as long as this file is executable (`chmod +x`) and contains a `#!/bin/bash` shebang at the top of its file, we can remove this `bash` keyword.

ci/build_wheel.sh
```diff
@@ -0,0 +1,30 @@
...
+
+cd "${package_dir}"
+
+# Hardcode the output dir
+python -m pip wheel . -w dist -vvv --no-deps --disable-pip-version-check
+
+mkdir -p final_dist
+python -m auditwheel repair -w final_dist dist/*
+
+RAPIDS_PY_WHEEL_NAME="${package_name}_${RAPIDS_PY_CUDA_SUFFIX}" rapids-upload-wheels-to-s3 final_dist
```
ajschmidt8, line 30:
```suggestion
export RAPIDS_PY_WHEEL_NAME="${package_name}_${RAPIDS_PY_CUDA_SUFFIX}"
rapids-upload-wheels-to-s3 final_dist
```

can we break up these long lines? if `rapids-upload-wheels-to-s3` needs the `RAPIDS_PY_WHEEL_NAME` var set, we can just export it on the line above it.

ci/build_wheel_raft_dask.sh
```diff
@@ -0,0 +1,13 @@
...
+# Copyright (c) 2023, NVIDIA CORPORATION.
+
+set -euo pipefail
+
+# Set up skbuild options. Enable sccache in skbuild config options
+export SKBUILD_CONFIGURE_OPTIONS="-DRAFT_BUILD_WHEELS=ON -DDETECT_CONDA_ENV=OFF -DFIND_RAFT_CPP=OFF"
+
+RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"
+
+RAPIDS_PY_WHEEL_NAME=pylibraft_${RAPIDS_PY_CUDA_SUFFIX} rapids-download-wheels-from-s3 ./local-pylibraft && python -m pip install --no-deps ./local-pylibraft/pylibraft*.whl
```
ajschmidt8, line 11:
```suggestion
export RAPIDS_PY_WHEEL_NAME=pylibraft_${RAPIDS_PY_CUDA_SUFFIX}
rapids-download-wheels-from-s3 ./local-pylibraft
python -m pip install --no-deps ./local-pylibraft/pylibraft*.whl
```

can we split these lines up too?

ci/build_wheel.sh
divyegala: Actually @vyasr pointed out that we only need `RAPIDS_PY_WHEEL_NAME` set for `rapids-upload-wheels-to-s3`, so we shouldn't pollute the environment

ci/build_wheel_raft_dask.sh
divyegala: Going by the above comment, would you be okay if we split up into two lines instead of 3?

ajschmidt8: That's fine with me

ci/build_wheel.sh
```diff
@@ -0,0 +1,30 @@
...
+RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"
+
+ci/release/apply_wheel_modifications.sh ${version_override} "-${RAPIDS_PY_CUDA_SUFFIX}"
+echo "The package name and/or version was modified in the package source. The git diff is:"
+git diff
+
+cd "${package_dir}"
+
+# Hardcode the output dir
+python -m pip wheel . -w dist -vvv --no-deps --disable-pip-version-check
```
ajschmidt8, line 25:
```suggestion
python -m pip wheel . -w dist --no-deps --disable-pip-version-check
```

Is the verbosity here necessary?

In the past, we've intentionally turned off verbosity for things like `pytest` because it can make the GHA logs difficult to read.

ci/test_wheel_pylibraft.sh
```diff
@@ -0,0 +1,19 @@
...
+# Copyright (c) 2023, NVIDIA CORPORATION.
+
+set -eoxu pipefail
+
+mkdir -p ./dist
+RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"
+RAPIDS_PY_WHEEL_NAME="pylibraft_${RAPIDS_PY_CUDA_SUFFIX}" rapids-download-wheels-from-s3 ./dist
+
+# echo to expand wildcard before adding `[extra]` requires for pip
+python -m pip install -v $(echo ./dist/pylibraft*.whl)[test]
```
ajschmidt8, line 11:
```suggestion
python -m pip install $(echo ./dist/pylibraft*.whl)[test]
```

Same question about verbosity here.

ci/test_wheel_pylibraft.sh
```diff
@@ -0,0 +1,19 @@
...
+
+# Run smoke tests for aarch64 pull requests
+arch=$(uname -m)
+if [[ "${arch}" == "aarch64" && ${RAPIDS_BUILD_TYPE} == "pull-request" ]]; then
```
ajschmidt8, line 15:
```suggestion
if [[ "$(arch)" == "aarch64" && "${RAPIDS_BUILD_TYPE}" == "pull-request" ]]; then
```

`arch` and `uname -m` will return the same value, so this can be made a little more succinct.

Also, it's a good practice to quote variables. If we had `shellcheck` enabled for this script, it would recommend the double quotations around `RAPIDS_BUILD_TYPE`.

ci/test_wheel_raft_dask.sh
```diff
@@ -0,0 +1,26 @@
...
+
+# Always install latest dask for testing
+python -m pip install git+https://github.com/dask/dask.git@main git+https://github.com/dask/distributed.git@main git+https://github.com/rapidsai/dask-cuda.git@branch-23.08
+
+# echo to expand wildcard before adding `[extra]` requires for pip
+python -m pip install -v $(echo ./dist/raft_dask*.whl)[test]
+
+# Run smoke tests for aarch64 pull requests
+arch=$(uname -m)
+if [[ "${arch}" == "aarch64" && ${RAPIDS_BUILD_TYPE} == "pull-request" ]]; then
```
ajschmidt8, line 22:
```suggestion
if [[ "$(arch)" == "aarch64" && "${RAPIDS_BUILD_TYPE}" == "pull-request" ]]; then
```

same suggestions here

ci/test_wheel_raft_dask.sh
```diff
@@ -0,0 +1,26 @@
...
+
+# Download the pylibraft built in the previous step
+RAPIDS_PY_WHEEL_NAME="pylibraft_${RAPIDS_PY_CUDA_SUFFIX}" rapids-download-wheels-from-s3 ./local-pylibraft-dep
+python -m pip install --no-deps ./local-pylibraft-dep/pylibraft*.whl
+
+# Always install latest dask for testing
+python -m pip install git+https://github.com/dask/dask.git@main git+https://github.com/dask/distributed.git@main git+https://github.com/rapidsai/dask-cuda.git@branch-23.08
+
+# echo to expand wildcard before adding `[extra]` requires for pip
+python -m pip install -v $(echo ./dist/raft_dask*.whl)[test]
```
ajschmidt8, line 18:
```suggestion
python -m pip install $(echo ./dist/raft_dask*.whl)[test]
```

same question about verbosity here

ci/build_wheel.sh
divyegala: I will let @vyasr comment on the verbosity questions

vyasr: Yes I was going to raise this point towards the end. I say let's turn off the verbosity for tests, but keep it for builds (i.e. `pip wheel -v`) That matches what we do for conda and I think that's generally helpful. Test failures tend to be self-contained, so it's enough to see that a single test failure to know what's going on, whereas with builds the real cause can often be hidden (e.g. pip does something wrong in dependency installation which results in compile-time errors).

ajschmidt8 (approved):

PR #1503, opened 2023-05-10
Use rmm allocator in CAGRA prune

enp1s0:
This PR modifies the CAGRA prune function to use rmm allocator instead of a custom memory allocator. And also removes the multi-GPU functionality that is not suitable for the [RAFT policy](https://docs.rapids.ai/api/raft/stable/developer_guide/#multi-gpu).

rel: #1461

PR #1550, opened 2023-05-29
Update CAGRA knn_graph_sort to use Raft::bitonic_sort

enp1s0:
This PR changes CAGRA `knn_graph_sort` function to use `raft::util::bitnic_sort` instead of a custom sorting function.

Rel: https://github.com/rapidsai/raft/pull/1503#discussion_r1190439804


enp1s0:
@tfeher Can you check the code and let me know if there are any problems?

tfeher (approved):
Thanks @enp1s0 for these changes, the PR looks good to me. It is great to see more reuse of existing raft kernels.

PR #1552, opened 2023-05-29
Add RAFT ANN benchmark for CAGRA

enp1s0:
This PR adds the RAFT ANN benchmark for CAGRA


cpp/bench/ann/src/raft/raft_benchmark.cu
```diff
@@ -117,28 +123,25 @@ void parse_search_param(const nlohmann::json& conf,
...
+template <typename T, typename IdxT>
+void parse_search_param(const nlohmann::json& conf,
+                        typename raft::bench::ann::RaftCagra<T, IdxT>::SearchParam& param)
 {
-  typename Algo<T>::BuildParam param;
-  parse_build_param<T>(conf, param);
-
-  (void)dev_list;
-  return std::make_unique<Algo<T>>(metric, dim, param);
+  param.search_params.max_queries = conf.at("max_batch");
```
tfeher, line 141: Just for my understanding: naively I would think that we want to use `max_batch` == `query_set_size` for best perf. In that case setting a lower `max_batch` would only make sense to limit the workspace memory usage. Is this true, or are there any cases where we would want to use a smaller value for `max_batch`?

cpp/bench/ann/src/raft/raft_cagra_wrapper.h
```diff
@@ -0,0 +1,192 @@
...
+                                 const IdxT* const src_ptr,
+                                 const size_t len,
+                                 cudaStream_t cuda_stream)
+{
+  const size_t block_size = 256;
+  const size_t grid_size  = (len + block_size - 1) / block_size;
+  convert_neighbor_index_type_kernel<IdxT>
+    <<<grid_size, block_size, 0, cuda_stream>>>(dst_ptr, src_ptr, len);
+}
+}  // anonymous namespace
```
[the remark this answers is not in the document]
enp1s0: I didn't know that there was such a function... Thank you for the information.

enp1s0:
TODO:

- [ ] Add support for deep-100M and sift-128-euclidean, that use L2-norm for similarity

cpp/bench/ann/src/raft/raft_benchmark.cu
enp1s0: I noticed "search_basic_param.batch_size" is in the conf json files, so I'll remove the "max_batch" setting from CAGRA conf.

> we want to use max_batch == query_set_size for best perf.

Yes, that's true.
However, there are instances where we need to process small batches or even single queries.

tfeher (approved):
Thanks @enp1s0 for addressing the issues, the PR looks good to me!

enp1s0:
Thanks @tfeher for checking the code.

Issue #1542, opened 2023-05-22
[ENH] Add `__half` overloads to math API wrappers


Nyrio:
The wrappers in https://github.com/rapidsai/raft/blob/branch-23.06/cpp/include/raft/core/math.hpp delegate work to the appropriate CUDA intrinsic. However, the CUDA intrinsics for square root, trigonometry functions, etc, have different names for `__half` (e.g `hsqrt` instead of `sqrt`).

To facilitate templated code, we could add overloads of `sqrt` calling `hsqrt` on the device and similarly for trigonometry functions.
On host, we can either not define those, or use the fp32 functions.

PR #1554, opened 2023-05-30
Add support to __half and nv_bfloat16 to most math functions

Kh4ster:
This PR adds support to __half and nb_bfloat16 to most math functions:

- cos
- sin
- exp
- log
- max
- min
- sqrt


cpp/include/raft/core/detail/macros.hpp
```diff
@@ -22,6 +22,12 @@
 #endif
 #endif
 
+#if defined(_RAFT_HAS_CUDA)
+#define CUDA_ONLY_CONDITION(condition) condition
+#else
+#define CUDA_ONLY_CONDITION(condition) true
+#endif
```
Nyrio, line 29: Do you mind future-proofing that with:
```c++
#if defined(_RAFT_HAS_CUDA)
#define CUDA_CONDITION_ELSE_TRUE(condition) condition
#define CUDA_CONDITION_ELSE_FALSE(condition) condition
#else
#define CUDA_CONDITION_ELSE_TRUE(condition) true
#define CUDA_CONDITION_ELSE_FALSE(condition) false
#endif
```

and using `CUDA_CONDITION_ELSE_TRUE`?

cpp/include/raft/core/math.hpp
```diff
@@ -91,7 +96,14 @@ RAFT_INLINE_FUNCTION auto atanh(T x)
 
 /** Cosine */
 template <typename T>
-RAFT_INLINE_FUNCTION auto cos(T x)
+RAFT_INLINE_FUNCTION
+#if _RAFT_HAS_CUDA
+  typename std::enable_if<!std::is_same<T, __half>::value && !std::is_same<T, nv_bfloat16>::value,
+                          T>::type
```
Nyrio, line 102: Nitpick: can you use C++17 constructs for readability? `enable_if_t` and `is_same_v` (everywhere, not just this line)

cpp/include/raft/core/math.hpp
```diff
@@ -91,7 +96,14 @@ RAFT_INLINE_FUNCTION auto atanh(T x)
...
+#else
+  auto
+#endif
```
Nyrio, line 105: Can you use `CUDA_ONLY_CONDITION` / `CUDA_CONDITION_ELSE_TRUE` when you have this kind of host guard?

cpp/include/raft/core/math.hpp
```diff
@@ -100,9 +112,49 @@ RAFT_INLINE_FUNCTION auto cos(T x)
...
+{
+#if (__CUDA_ARCH__ >= 530)
+  return ::hcos(x);
+#else
+  // static_assert(false) would be evaluated during host compilation stage while __CUDA_ARCH__ is at
+  // device compilation stage Using this sizeof(T) != sizeof(T) makes it work as it's only triggered
+  // during template instantiation and thus at device compilation stage
+  static_assert(sizeof(T) != sizeof(T), "__half is only supported on __CUDA_ARCH__ >= 530");
+  return T{};
+#endif
```
Nyrio, line 128: Question for @cjnolet : iirc RAPIDS is Pascal+, shall we have a centralized check of the minimum arch somewhere and remove these checks here, or would this have too many consequences on repos using raft?

cpp/include/raft/core/math.hpp
```diff
@@ -151,16 +243,89 @@ RAFT_INLINE_FUNCTION auto exp(T x)
...
+template <typename T>
+RAFT_DEVICE_INLINE_FUNCTION typename std::enable_if<std::is_same<T, __half>::value, __half>::type
+exp(T x)
+{
+#if (__CUDA_ARCH__ >= 530)
+  return ::hexp(x);
+#else
+  // static_assert(false) would be evaluated during host compilation stage while __CUDA_ARCH__ is at
+  // device compilation stage Using this sizeof(T) != sizeof(T) makes it work as it's only triggered
+  // during template instantiation and thus at device compilation stage
```
Nyrio, line 256: This comment appears in many places, can you please change it for a one-liner? Such as:
```c++
// Fail during template instantiation if the compute capability doesn't support this operation
```

or remove it altogether

Nyrio (commented):
Thanks for the PR. A few nitpicks

cpp/include/raft/core/math.hpp
```diff
@@ -100,8 +108,40 @@ RAFT_INLINE_FUNCTION auto cos(T x)
 #endif
 }
 
-/** Sine */
+#if _RAFT_HAS_CUDA
 template <typename T>
+RAFT_DEVICE_INLINE_FUNCTION typename std::enable_if<std::is_same_v<T, __half>, __half>::type cos(
```
Nyrio, line 113: -> `enable_if_t`

_note: 18 occurences_

Nyrio (changes requested):

python/pylibraft/pyproject.toml
```diff
@@ -19,7 +19,7 @@ requires = [
     "cuda-python>=11.7.1,<12.0",
     "cython>=0.29,<0.30",
     "ninja",
-    "rmm==23.8.*",
+    "rmm==23.6.*",
```
Nyrio, line 22: Why is this required?

Nyrio:
Note: this PR solves #1542

python/pylibraft/pyproject.toml
Nyrio: @bdice Yes, I think the `dependencies.yaml` still has 23.6, so the right fix is to update that to 23.8

bdice: Fix is filed here: https://github.com/rapidsai/raft/pull/1556

Nyrio (approved):

cjnolet (approved):
LGTM. Thanks for the PR!

Issue #1560, opened 2023-05-31
[FEA] host functionality for random generators


hello-11:
I wish I could use RAFT to generate random values on the CPU. 
Besides, I wish the host functionality of random generator should emulate device functionality when given the same seeds/base-subsequences..

PR #1568, opened 2023-06-01
[FEA] Expose RNG device API for host code

MatthiasKohl:
This PR fixes issues #1560.
It exposes RNG device functionality for `__host__` code in order to allow easier testing of kernels using the device code fused with other functionality.
The PR also adds a few comments: one question left is whether we should deprecate the constructors based on subsequence and offset (since they shouldn't really be used in other projects IMO, the main device interface should go through `DeviceState`).
What do you think @vinaydes @cjnolet  ?


MatthiasKohl:
@cjnolet I assume that your update version script doesn't include the toml files in `python` folder, this is mainly what the style checks were complaining about previously in this PR

MatthiasKohl:
Had a quick look and indeed, the RMM version isn't updated.
I can try to add this here so it's done

MatthiasKohl:
actually, I don't really understand, it should have been updated anyway by the script already, and I branched off of branch-23.08. Sorry for the confusion here, I'm not exactly sure what went wrong

MatthiasKohl:
@cjnolet I'm unsure whether the failed C++ test is related to this PR. It is this one: ReductionTest/ReductionTestInt.REDUCTIONS/0.
Everything else seems to be passing.
Can you help me a little to fix this?
I'm unfortunately very busy with other projects atm.

cjnolet (approved):
LGTM

vinaydes:
The change looks fine to me. About keeping only the `DeviceState` based constructor: Right now `DeviceState` does not contain `offset` as a member, which would be required for a complete initialization. Do we expect user to call `advance` method to accommodate `offset`?

MatthiasKohl:
@vinaydes to me there's two things:
1. the `offset`-based constructor shouldn't really be used at all anymore IMO: I believe we just kept this to keep things compatible, but I could probably mark that constructor deprecated now. Also to keep things in line with `curand` Philox which uses the `subsequence` and `offset` separately. In theory, other RNGs may at some point also have both a `subsequence` and `offset`.
2. the `subsequence` isn't contained in the device state either, but that would require us to create another class which contains the "host-to-device-state" (current `DeviceState` struct) plus the actual subsequence or additional info (in case we bring back something like `offset` later on). This potential new class would be created within a kernel and would then contain all info required to generate a new number.

Instead, I think that creating the RNGs based on `DeviceState` as defined now along with the `subsequence` is good enough for now, and this should allow things to work on host as well.

Finally, a different thing: do you happen to know why the `ReductionTest/ReductionTestInt.REDUCTIONS/0` C++ test fails here and everything else passes?
The only functional change in this PR is to add the `subsequence` to the `offset` in the `offset`-based constructor of PCG. This should not fail any test unless the test was bogus to begin with

MatthiasKohl:
@vinaydes were you able to make progress on this?

vinaydes:
I have added PCG host API with this PR #1767. May be we can close this.

MatthiasKohl:
Indeed, your PR superseeds this. Closing

PR #1571, opened 2023-06-02
[HOTFIX] Fix  distance metrics L2/cosine/correlation when X & Y are same buffer but with different shape and add unit test for such case.

mdoijade:
-- This is how tiled_brute_force_knn may use pairwise distance API hence assuming when X == Y the buffer has same shape is incorrect.

### DIFF
--- a/cpp/include/raft/distance/detail/distance.cuh
+++ b/cpp/include/raft/distance/detail/distance.cuh
@@ -126,9 +126,7 @@ void distance_impl(raft::resources const& handle,
                    bool is_row_major,
                    DataT)  // unused
 {
-  ASSERT(!((((x != y) || ((x == y) && (m != n))) && (worksize < 2 * (m + n) * sizeof(AccT))) ||
-           (worksize < 2 * m * sizeof(AccT))),
-         "workspace size error");
+  ASSERT(!(worksize < 2 * (m + n) * sizeof(AccT)), "workspace size error");
   ASSERT(workspace != nullptr, "workspace is null");
 
   cudaStream_t stream = raft::resource::get_cuda_stream(handle);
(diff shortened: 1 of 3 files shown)


mdoijade:
@cjnolet , @tfeher  for help with review.

cpp/include/raft/distance/detail/distance.cuh
```diff
@@ -137,7 +137,7 @@ void distance_impl(raft::resources const& handle,
   AccT* y_norm    = workspace;
   AccT* sq_x_norm = workspace;
   AccT* sq_y_norm = workspace;
-  if (x != y) {
+  if ((x != y) || ((x == y) && (m != n))) {
```
cjnolet, line 140: This conditional is getting a little complicated and I see it in multiple places in this PR. Can we create a helper function that can give this a name and be reused throughout the code? It would make it a lot easier to read and maintain.

cjnolet (changes requested):
Change looks good, just hoping to consolidate the conditional to make it easier to read.

Thanks for finding and fixing this issue. It's a super subtle detail.

cpp/include/raft/distance/detail/distance.cuh
```diff
@@ -210,15 +210,16 @@ void distance_impl(raft::resources const& handle,
...
+           (worksize < m * sizeof(AccT))),
          "workspace size error");
   ASSERT(workspace != nullptr, "workspace is null");
 
   cudaStream_t stream = raft::resource::get_cuda_stream(handle);
 
   DataT* x_norm = workspace;
   DataT* y_norm = workspace;
-  if (x != y) {
+  if ((x != y) || ((x == y) && (m != n))) {
```
tfeher, line 222: This solves the problem. But what do you think about an alternative solution?

```c++
if (x != y) {
// ... keep original code 
} else {
   // use max(m,n)
    raft::linalg::rowNorm(
      x_norm, x, k, std::max(m,n), raft::linalg::L2Norm, is_row_major, stream, raft::sqrt_op{});
  }
```
During workspace size calculation we also need to use `max(m,n)`

cjnolet:
@mdoijade do you still think we can get this fix into 23.06? It looks like there's some c++ test failures

mdoijade:
> @mdoijade do you still think we can get this fix into 23.06? It looks like there's some c++ test failures

I believe the failure is related to fp arithmetic accuracy issues in L2 distances for X == Y input cases. I don't see a bug as such I will fix the tolerance accordingly and push it soon.

cpp/include/raft/distance/detail/distance.cuh
mdoijade: as discussed offline, this is good solution but somehow for column major inputs there is perhaps fp arithmetic accuracy loss due to which I've kept `max(m,n)` only for row major inputs until we resolve it.

cpp/include/raft/distance/detail/distance.cuh
mdoijade: have removed this conditional.

cjnolet (approved):

PR #1573, opened 2023-06-02
Assigning Deterministic rank to Dask Workers Based on CUDA_VISIBLE_DEVICES

VibhuJawa:
Previously, `dask-raft` non-deterministically maps a process to a GPU. 

In this PR, we assign a deterministic order to each worker based on the CUDA_VISIBLE_DEVICES environment variable.
as NCCL>1.11 expects a process with `rank r` to be mapped to `r % num_gpus_per_node` . 

This  fixes https://github.com/rapidsai/cugraph/issues/3478 
and this raft-test in MNMG setting  https://github.com/rapidsai/raft/blob/c1a7b7c0e33b11d2e7ff3bc5014e3b410a2edd0d/python/raft-dask/raft_dask/test/test_comms.py#L82-L84

CC: @jnke2016

### DIFF
--- a/python/raft-dask/raft_dask/common/comms.py
+++ b/python/raft-dask/raft_dask/common/comms.py
@@ -14,13 +14,11 @@
 #
 
 import logging
+import os
 import time
 import uuid
 import warnings
-from collections import OrderedDict, Counter
-import os
-
-
+from collections import Counter, OrderedDict
 
 from dask.distributed import default_client
 from dask_cuda.utils import nvml_device_index
@@ -692,10 +690,12 @@ def _func_ucp_ports(client, workers):
 
 def _func_worker_ranks(client):
     """
-    For each worker connected to the client, compute a global rank which is the sum
+    For each worker connected to the client,
+    compute a global rank which is the sum
     of the NVML device index and the worker rank offset.
 
-    Args:
+    Parameters
+    ----------
         client (object): Dask client object.
     """
     ranks = client.run(_get_nvml_device_index)
@@ -706,7 +706,8 @@ def _func_worker_ranks(client):
 
 def _get_nvml_device_index():
     """
-    Return NVML device index based on environment variable 'CUDA_VISIBLE_DEVICES'.
+    Return NVML device index based on environment variable
+    'CUDA_VISIBLE_DEVICES'.
     """
     CUDA_VISIBLE_DEVICES = os.getenv("CUDA_VISIBLE_DEVICES")
     return nvml_device_index(0, CUDA_VISIBLE_DEVICES)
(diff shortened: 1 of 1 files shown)


python/raft-dask/raft_dask/common/comms.py
```diff
@@ -686,8 +688,75 @@ def _func_ucp_ports(client, workers):
...
+    return _append_rank_offset(ranks, worker_ip_offset_dict)
+
+
+def _get_nvml_device_index():
+    """
+    Return NVML device index based on environment variable
+    'CUDA_VISIBLE_DEVICES'.
+    """
+    CUDA_VISIBLE_DEVICES = os.getenv("CUDA_VISIBLE_DEVICES")
+    return nvml_device_index(0, CUDA_VISIBLE_DEVICES)
```
rlratzel, line 713: Can we guarantee `CUDA_VISIBLE_DEVICES` is always set? Otherwise, I think this will break if it isn't.
(untested, needs formatting and another import, but you get the idea)
```suggestion
    CUDA_VISIBLE_DEVICES = os.getenv("CUDA_VISIBLE_DEVICES", ",".join((str(d) for d in range(dask_cuda.utils.get_gpu_count()))))
    return nvml_device_index(0, CUDA_VISIBLE_DEVICES)
```

VibhuJawa: I think `CUDA_VISIBLE_DEVICES` is always set by `dask-cuda` 

https://github.com/rapidsai/dask-cuda/blob/bba3d3f2a1edf5b42160d8f081b676cd4e9b9a41/dask_cuda/cuda_worker.py#L207


https://github.com/rapidsai/dask-cuda/blob/bba3d3f2a1edf5b42160d8f081b676cd4e9b9a41/dask_cuda/cuda_worker.py#L233

jnke2016:
@VibhuJawa  thanks a lot for this PR. It looks good. However, it seems to uncover an issue with batch edge betweenness centrality as it is the only one failing. @ChuckHastings we were thinking about merging this PR as it fixes the `nccl` failure that we have been debugging and later, find a fix for `batch edge betweenness centrality` unless you have something else in mind. In the mean time, I have been working on bindings for edge betweenness centrality(with the hope that this failure doesn't occur in the new implementation) as you are implementing it but we have to make a call for this PR as raft is entering code freeze soon.

ChuckHastings:
> @VibhuJawa thanks a lot for this PR. It looks good. However, it seems to uncover an issue with batch edge betweenness centrality as it is the only one failing. @ChuckHastings we were thinking about merging this PR as it fixes the `nccl` failure that we have been debugging and later, find a fix for `batch edge betweenness centrality` unless you have something else in mind. In the mean time, I have been working on bindings for edge betweenness centrality(with the hope that this failure doesn't occur in the new implementation) as you are implementing it but we have to make a call for this PR as raft is entering code freeze soon.

Definitely merge this to fix the `nccl` failure.  The `batch edge betweenness centrality failure` is clearly a separate issue.  We would need to resolve that separately anyway.

cjnolet (approved):

cjnolet:
@VibhuJawa can you open a quick cuml PR that pins your branch in `get_raft.cmake` just so we can make sure this doesn't somehow break cuml?

VibhuJawa:
> @VibhuJawa can you open a quick cuml PR that pins your branch in `get_raft.cmake` just so we can make sure this doesn't somehow break cuml?

Started PR : https://github.com/rapidsai/cuml/pull/5462/files

PR #1577, opened 2023-06-06
Fixing merge_labels test

cjnolet:
The offending test was using an `rmm::device_scalar` for some memory that should have been a vector. Not sure how this didn't fail in the past but these changes fix it.


benfred (approved):
lgtm!

PR #1580, opened 2023-06-07
Fix raft::bitonic_sort small usage example

enp1s0:
The template argument `Size` of `raft::util::bitnic_sort` must be power-of-two but is set to three in the small usage example in the source code. This PR is only for documentation improvement and does not change any behavior.


achirkin (approved):
Indeed you're right, thanks for spotting and fixing this!

cjnolet (approved):

PR #1584, opened 2023-06-09
[IMP] move core CUDA RT macros to cuda_rt_essentials.hpp

MatthiasKohl:
The reasoning behind this PR is as follows:
for now, anyone wanting to use `RAFT_CUDA_TRY_NO_THROW` still needs to include `cudart_utils.hpp` which can be costly (compilation) due to the include of `memory_pool.hpp`.
By moving the macros to the essentials, we should not break anything for anyone, but allow anyone to improve compilation times by including the essentials only. At the same time, it should add minimal overhead to the compilation time of the essentials file since the pre-processor is (usually) fast compared to the rest of the compilation pipeline.

### DIFF
--- a/cpp/include/raft/util/cuda_rt_essentials.hpp
+++ b/cpp/include/raft/util/cuda_rt_essentials.hpp
@@ -23,6 +23,8 @@
 #include <cuda_runtime.h>
 #include <raft/core/error.hpp>
 
+#include <cstdio>
+
 namespace raft {
 
 /**
--- a/cpp/include/raft/util/cuda_rt_essentials.hpp
+++ b/cpp/include/raft/util/cuda_rt_essentials.hpp
@@ -58,3 +58,38 @@ struct cuda_error : public raft::exception {
       throw raft::cuda_error(msg);                 \
     }                                              \
   } while (0)
+
+/**
+ * @brief Debug macro to check for CUDA errors
+ *
+ * In a non-release build, this macro will synchronize the specified stream
+ * before error checking. In both release and non-release builds, this macro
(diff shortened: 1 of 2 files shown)


cpp/include/raft/util/cuda_rt_essentials.hpp
```diff
@@ -58,3 +58,38 @@ struct cuda_error : public raft::exception {
...
+
+// /**
+//  * @brief check for cuda runtime API errors but log error instead of raising
+//  *        exception.
+//  */
+#define RAFT_CUDA_TRY_NO_THROW(call)                               \
+  do {                                                             \
+    cudaError_t const status = call;                               \
+    if (cudaSuccess != status) {                                   \
+      printf("CUDA call='%s' at file=%s line=%d failed with %s\n", \
```
ahendriksen, line 89: One change request related to the header being self-standing. Can you add an include of `cstdio` to the top of `cuda_rt_essentials.hpp`? 

This should have been done before, but probably escaped review.

ahendriksen (changes requested):
Looks good to me :+1:.

cpp/include/raft/util/cuda_rt_essentials.hpp
MatthiasKohl: actually, nothing was directly using `cstdio` in this file and `cstdio` was only included transitively through `core/error.hpp` for things like `SET_ERROR_MSG`.

So all was actually good before (IMO), and I fixed the include here now since we now use `cstdio` directly in this file.

Thanks for pointing it out !

MatthiasKohl:
This should be good to be merged

cjnolet (approved):

PR #1586, opened 2023-06-09
Fix launchconfig y-gridsize too large in epilogue kernel

mfoerste4:
This fixes the launch config for epilogue kernels in the GramMatrix computation for large number of columns in the resulting kernel matrix.
This code-path is triggered when predicting scores with a support vector space larger than 262140 which would result in a y-grid-dimension larger than 65535. 

Although this is not a regression this code path might get hit more often now that we allow sparse input data, so we might want to fix it in 23.06 as well.

CC @tfeher .


tfeher (approved):
Thanks @mfoerste4 for the PR, LGTM!

PR #1587, opened 2023-06-09
Assigning Deterministic rank to Dask Workers Based on CUDA_VISIBLE_DEVICES (branch-23.06)

VibhuJawa:
This PR is same as https://github.com/rapidsai/raft/pull/1573 but targetted for branch-23.06 as a hotfix CC: @rlratzel 

Previously, `dask-raft` non-deterministically maps a process to a GPU. 

In this PR, we assign a deterministic order to each worker based on the CUDA_VISIBLE_DEVICES environment variable.
as NCCL>1.11 expects a process with `rank r` to be mapped to `r % num_gpus_per_node` . 

This  fixes https://github.com/rapidsai/cugraph/issues/3478 
and this raft-test in MNMG setting  https://github.com/rapidsai/raft/blob/c1a7b7c0e33b11d2e7ff3bc5014e3b410a2edd0d/python/raft-dask/raft_dask/test/test_comms.py#L82-L84

### DIFF
--- a/python/raft-dask/raft_dask/common/comms.py
+++ b/python/raft-dask/raft_dask/common/comms.py
@@ -14,12 +14,14 @@
 #
 
 import logging
+import os
 import time
 import uuid
 import warnings
-from collections import OrderedDict
+from collections import Counter, OrderedDict
 
 from dask.distributed import default_client
+from dask_cuda.utils import nvml_device_index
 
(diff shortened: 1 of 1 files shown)


rlratzel (approved):
I believe this PR was originally [here](https://github.com/rapidsai/raft/pull/1573) but created again to target `23.06` instead.

PR #1588, opened 2023-06-12
Rename the CAGRA prune function to optimize

enp1s0:
This PR renames the `cagra::prune` function to `cagra::optimize` since it adds reverse edges other than pruning unimportant edges.


cjnolet (approved):
LGTM

PR #1591, opened 2023-06-12
Add support for row-major slice

### LINKED ISSUE #1583: [FEA] `raft::matrix::slice` to support row-major mdspans
This should be pretty straightforward to implement. It's possible to perform a `slice` on existing row-major data by constructing a column-major mdspan and then flipping the rows/cols in the slice coords but this is not particularly straightforward to the user and confusing to the user. 

We should have a first-class row-major overload for it.

lowener:
Closes #1583


cpp/include/raft/matrix/slice.cuh
```diff
@@ -45,17 +46,18 @@ struct slice_coordinates {
...
-template <typename m_t, typename idx_t>
+template <typename m_t, typename idx_t, typename layout_t>
 void slice(raft::resources const& handle,
-           raft::device_matrix_view<const m_t, idx_t, col_major> in,
-           raft::device_matrix_view<m_t, idx_t, col_major> out,
+           raft::device_matrix_view<const m_t, idx_t, layout_t> in,
+           raft::device_matrix_view<m_t, idx_t, layout_t> out,
            slice_coordinates<idx_t> coords)
 {
+  RAFT_EXPECTS(raft::is_row_or_column_major(in), "Unsupported matrix layout");
```
cjnolet, line 60: In addition to the assertion, can we provide a little hint to the user to help them fix this? Maybe something like "Matrix layout must be row- or column-major".

cjnolet (commented):
Great to see more APIs beind expanded for multiple layouts. Just one little comment, otherwise this LGTM!

cjnolet (approved):

PR #1594, opened 2023-06-19
Use exceptions instead of `exit(-1)`

benfred:
Rather than exit the process immediately, use the `RAFT_FAIL` macro to throw an exception inside the cagra code when certain conditions aren't met


tfeher (approved):
Thanks @benfred far the changes, LGTM.

PR #1597, opened 2023-06-22
pass correct stream to cutlass kernel launch of L2/cosine pairwise distance kernels

mdoijade:
-- This fixes the runtime failure seen in the pytest on arm-A100 config in CI.


ahendriksen (approved):
Looks good to me. Very subtle issue..

tfeher (approved):
Thanks Mahesh for the fix, LGTM!

PR #1613, opened 2023-06-23
CAGRA max_queries auto configuration

enp1s0:
This PR changes the behavior of max_queries in CAGRA.

In the search operation of CAGRA, a batch of queries is divided into several `max_queries` queries sub-batches and operated; by increasing `max_queries,` higher throughput can be obtained. However, if one forgets to set `max_queries,` the default value, 1, will be used, resulting in lower throughput. This is not user-friendly.

This PR modifies CAGRA to automatically set `max_queries` as the batch size by default.

An alternative way is to remove `max_queries`.


tfeher (approved):
Thanks @enp1s0 for the PR, it looks good to me.

PR #1615, opened 2023-06-23
Adding sort option to matrix::select_k api

cjnolet:
The current `raft::matrix::select_k` API doesn't always return sorted neighborhoods, which can easily cause issues when the sorting is expected. This PR provides a new argument to the select_k API to guarantee the output is always sorted.


cpp/include/raft/matrix/detail/select_k-inl.cuh
```diff
@@ -116,6 +120,83 @@ inline Algo choose_select_k_algorithm(size_t rows, size_t cols, int k)
...
+    cub::DeviceSegmentedRadixSort::SortPairs(d_temp_storage,
+                                             temp_storage_bytes,
+                                             distances,
+                                             out_dists.data_handle(),
+                                             indices,
+                                             out_inds.data_handle(),
+                                             n_rows * k,
+                                             n_rows,
+                                             offsets.data_handle(),
+                                             offsets.data_handle() + 1);
```
benfred, line 180: should we pass in the stream here - like in https://github.com/rapidsai/raft/blob/185e933e123c043c488d252513f1360c40487af4/cpp/include/raft/matrix/detail/columnWiseSort.cuh#L292-L294 ?

cpp/include/raft/matrix/detail/select_k-inl.cuh
```diff
@@ -116,6 +120,83 @@ inline Algo choose_select_k_algorithm(size_t rows, size_t cols, int k)
...
+                                                       out_dists.data_handle(),
+                                                       indices,
+                                                       out_inds.data_handle(),
+                                                       n_rows * k,
+                                                       n_rows,
+                                                       offsets.data_handle(),
+                                                       offsets.data_handle() + 1);
+  }
+  // Allocate temporary storage
+  cudaMalloc(&d_temp_storage, temp_storage_bytes);
```
benfred, line 167: I believe this will synchronize the device - can we use a `rmm::device_uvector<char>` here for temporary storage (which should also pick up a pool allocator if set) ?

cpp/include/raft/matrix/detail/select_radix.cuh
```diff
@@ -1170,6 +1170,8 @@ void select_k(const T* in,
                                                         stream,
                                                         mr);
     }
+
+    // TODO SORT
```
benfred, line 1174: maybe remove this comment ?

benfred (commented):
looks great! thanks for this - 

I think for fixing faiss we will need to enable this option for the select_k call in `cpp/include/raft/neighbor/detail/knn_brute_force.cuh`.   If we do that, we might also be able to revert https://github.com/rapidsai/raft/pull/1533

cpp/include/raft/matrix/detail/select_k-inl.cuh
cjnolet: > I believe this will synchronize the device - can we use a rmm::device_uvector<char> here for temporary storage (which should also pick up a pool allocator if set) ?

Ah good catch! Yes this wasn't intentional. And I've been preferring the `mdarray` over directly using `rmm` in new code so that we can add a layer between the memory manager and the actual containers (e.g. container policy).

benfred (approved):
looks good!

cpp/internal/raft_internal/matrix/select_k.cuh
```diff
@@ -99,12 +99,21 @@ void select_k_impl(const resources& handle,
...
+
         // NB: std::nullopt prevents automatic inference of the template parameters.
         return matrix::select_k<T, IdxT>(
-          handle, in_span, std::nullopt, out_span, out_idx_span, select_min);
+          handle, in_span, std::nullopt, out_span, out_idx_span, select_min, true);
       } else {
-        return matrix::select_k(
-          handle, in_span, std::make_optional(in_idx_span), out_span, out_idx_span, select_min);
+        printf("RUnning sorted select_k2\n");
+
```
benfred, line 109:
```suggestion
```

cpp/internal/raft_internal/matrix/select_k.cuh
```diff
@@ -99,12 +99,21 @@ void select_k_impl(const resources& handle,
       auto out_span     = make_mdspan<T, int64_t, row_major, false, true>(out, out_extent);
       auto out_idx_span = make_mdspan<IdxT, int64_t, row_major, false, true>(out_idx, out_extent);
       if (in_idx == nullptr) {
+        printf("RUnning sorted select_k\n");
```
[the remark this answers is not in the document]
cjnolet: oh darn, I thought I had removed all the printfs. THanks for noticing this!

PR #1623, opened 2023-06-29
Use rapids-cmake new parallel testing feature

robertmaynard:
Corrects multiple test issues uncovered by running tests in parallel. Splits the matrix and neighbor tests into multiple executables to improve throughput

Previous performance on 1 GPU: 830sec
New performance on 1 GPU: 692sec
New performance on 2 GPU: 418sec


cpp/test/distance/gram.cu
```diff
@@ -127,6 +129,9 @@ class GramMatrixTest : public ::testing::TestWithParam<GramMatrixInputs> {
 
     (*kernel)(handle, x1_span, x2_span, out_span);
 
+    // Something in gram is executing not on the 'stream' and therefore
+    // a full device sync is required
+    RAFT_CUDA_TRY(cudaDeviceSynchronize());
```
vyasr, line 134: Can we open an issue to track this problem?

CC @cjnolet

cpp/test/matrix/select_k.cuh
vyasr, line 1: I've assumed that this entire file is just the subset of select_k.cu needed to allow separating the large test into select_k_large.cu and have not reviewed beyond that.

cpp/test/neighbors/selection.cu
```diff
@@ -441,7 +441,7 @@ auto inputs_random_largesize = testing::Values(SelectTestSpec{100, 100000, 1, tr
                                                SelectTestSpec{1, 100000000, 256, false, false});
 
 auto inputs_random_largek = testing::Values(SelectTestSpec{100, 100000, 1000, true},
-                                            SelectTestSpec{100, 100000, 2000, true},
+                                            SelectTestSpec{100, 100000, 2000, false},
```
vyasr, line 444: Why did this change?

cpp/test/neighbors/selection.cu
```diff
@@ -482,6 +482,11 @@ INSTANTIATE_TEST_CASE_P(SelectionTest,
  *  SelectionTest/ReferencedRandomFloatSizeT.LargeK/0
  *  Indicices do not match! ref[91628] = 131.359 != res[36504] = 158.438
  *  Actual: false (actual=36504 != expected=91628 @38999;
+ *
+ *  SelectionTest/ReferencedRandomFloatSizeT.LargeK/1
+ *  ERROR: ref[57977] = 58.9079 != res[21973] = 54.9354
+ *  Actual: false (actual=21973 != expected=57977 @107999;
+ *
```
vyasr, line 489: Should we be tracking these in issues?

cpp/test/util/device_atomics.cu
```diff
@@ -51,8 +51,8 @@ TEST(Raft, AtomicIncWarp)
   // Write all 1M thread indices to a unique location in `out_device`
   test_atomic_inc_warp_kernel<<<num_blocks, threads_per_block, 0, s>>>(counter.data(),
                                                                        out_device.data());
-
   // Copy data to host
+  RAFT_CUDA_TRY(cudaStreamSynchronize(s)); //ensure stream is done before copy
```
vyasr, line 55: Could we use a cudaMemcpyAsync below instead?

vyasr (changes requested):

cpp/test/neighbors/selection.cu
robertmaynard: Switcing this to false stopped the failure I was seeing, which I added to the todo comment about `SelectionTest/ReferencedRandomFloatSizeT.LargeK/1`.

cpp/test/util/device_atomics.cu
robertmaynard: Yes we can, and done.

cpp/test/matrix/select_k.cuh
robertmaynard: That is correct

cpp/test/neighbors/selection.cu
robertmaynard: I don't know. @cjnolet do we have an existing issue for the selection test failures?

cpp/test/distance/gram.cu
robertmaynard: I have updated the test to remove the usage of multiple streams which caused the sync issues.

cpp/test/neighbors/selection.cu
vyasr: This doesn't have to block this PR. I'll open an issue and we can proceed.

vyasr: https://github.com/rapidsai/raft/issues/1643

vyasr (approved):

cpp/test/util/device_atomics.cu
cjnolet: Please use raft::copy and raft::update_host instead of calling cuda directly.

cpp/test/distance/gram.cu
cjnolet: Please synchronize the stream in the handle or use interruptible to sync the stream instead of calling cuda APIs directly

robertmaynard: PR has been updated to fix the underlying bug that caused the `cudaDeviceSynchronize` to be needed.

cpp/test/util/device_atomics.cu
robertmaynard: There are no existing raft calls ( or handles ) in this test. The original version was calling `cudaMemcpy` on memory being written to by a kernel on a different stream.

cjnolet (approved):
LGTM!

cjnolet (approved):
LGTM!

raydouglass (approved):

PR #1626, opened 2023-06-29
[WIP] Test  CI issues 

VibhuJawa:


cjnolet:
Looks like it's still skipping the tests. Darn. Tomorrow I'll attempt removing the conditional testing skipping based on the exception catching and see if it might be a dependency issue. Perhaps its also possible a dependency changed recently and maybe it was just coincidence it was around the time your changes were merged.

VibhuJawa:
@cjnolet , I think the issue was due to `__init__.py` and we are running into below.    I am still trying to think what a long term good fix will be and why we are running into this problem. 

```
    from raft_dask.common import (
common/__init__.py:16: in <module>
    from .comms import Comms, local_handle
common/comms.py:28: in <module>
    from .comms_utils import (
E   ModuleNotFoundError: No module named 'raft_dask.common.comms_utils'
```

cjnolet (approved):
Thanks again for finding/fixing this, @VibhuJawa!

Issue #1575, opened 2023-06-05
CAGRA multi-cta solver accuracy


tfeher:
CAGRA's multi_cta solver gives incorrect results when max_queries > 1.

tfeher:
This was solved by #1628

PR #1628, opened 2023-06-30
Fix cagra multi CTA bug

enp1s0:
rel: #1575

The main causes of the bug are the wrong memory allocation size and the wrong number of threads to be launched.


enp1s0:
@tfeher Can you review the code?

cjnolet (approved):
LGTM, thanks @enp1s0! Once conflicts are resolved, I can merge this.

PR #1630, opened 2023-06-30
Fixing ann bench cmake (and docs)

cjnolet:
I'm removing the build for the raft bfknn ann benchmark target for now since it doesn't seem to be complete anyways. I'm going to create a separate github issue to add it back it. For now, the cmake isn't even producing what the docs are saying it's producing so we need to fix that ASAP.


cjnolet:
Linking https://github.com/rapidsai/raft/issues/1631

dantegd (approved):

PR #1641, opened 2023-07-07
[TEST-ONLY] Build `raft` with the latest `cuco`

PointKernel:
DO NOT MERGE

This PR is to make sure the latest cuco in `rapids-cmake` won't break `raft`. Needed by https://github.com/rapidsai/rapids-cmake/pull/435


PointKernel:
Close this as all tests passed.

PR #1644, opened 2023-07-12
IVF-PQ: Add a (faster) direct conversion fp8->half

achirkin:
Add a missing direct conversion from the custom ivf-pq `fp_8bit` type to `half`. This conversion is used in a tight ALU-bound loop that computes the distances between the query and the encoded cluster vectors.

This change improves QPS by 0-20% on the `deep-1B` dataset when `internal_distance_dtype = CUDA_R_16F` and `lut_dtype = CUDA_R_8U`. Note, however, on this dataset the <CUDA_R_16F, CUDA_R_8U> combination is often still slower than the <CUDA_R_16F, CUDA_R_16F> combination due to the couple extra ALU instructions. `lut_dtype = CUDA_R_8U` is the most beneficial when there's not enough shared memory for the lookup table, which is not the case for the best-performing parameter combinations on the `deep-1B` dataset.

### DIFF
--- a/cpp/include/raft/neighbors/detail/ivf_pq_fp_8bit.cuh
+++ b/cpp/include/raft/neighbors/detail/ivf_pq_fp_8bit.cuh
@@ -71,7 +71,7 @@ struct fp_8bit {
     return *this;
   }
   HDI explicit operator float() const { return fp_8bit2float(*this); }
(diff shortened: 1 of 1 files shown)


tfeher (approved):
Thanks @achirkin for the PR, it looks good to me.

PR #1649, opened 2023-07-14
Modify comm_split to avoid ucp

ChuckHastings:
During testing of a new feature in cugraph I discovered that the method required either MPI comms or UCP.  I have an application that has neither.

This PR modifies the `comm_split` implementation to continue using `allgather` when performing the split instead of using `allgather` followed by UCP comms.


cpp/include/raft/comms/detail/std_comms.hpp
```diff
@@ -140,48 +142,37 @@ class std_comms : public comms_iface {
...
+      std::distance(thrust::make_zip_iterator(h_colors.begin(), h_keys.begin()),
+                    std::find_if(thrust::make_zip_iterator(h_colors.begin(), h_keys.begin()),
+                                 thrust::make_zip_iterator(h_colors.end(), h_keys.end()),
+                                 [color](auto tuple) { return thrust::get<0>(tuple) == color; }));
+
+    auto subcomm_size = std::count(h_colors.begin(), h_colors.end(), color);
+
+    update_host(&id, d_nccl_ids.data() + offset, 1, stream_);
+
+    RAFT_CUDA_TRY(cudaStreamSynchronize(stream_));
```
cjnolet, line 171: We should probably use `sync_stream()` here so that this is more tolerant to failed ranks.

cpp/include/raft/comms/detail/std_comms.hpp
```diff
@@ -140,48 +142,37 @@ class std_comms : public comms_iface {
 
     RAFT_CUDA_TRY(cudaStreamSynchronize(stream_));
```
cjnolet, line 143: Now that we are using nccl instead of ucp, we should probably avoid any potential deadlocks from dying nccl ranks by using `sync_stream` here instead of the standard cuda synchronization.

cjnolet (changes requested):
This is definitely a welcome change! Looks good, just two very minor things.

cpp/include/raft/comms/detail/std_comms.hpp
ChuckHastings: Updated in latest push

cpp/include/raft/comms/detail/std_comms.hpp
ChuckHastings: Updated in latest push

cjnolet (approved):
LGTM, thanks Chuck!

PR #1651, opened 2023-07-18
ANN: Optimize host-side refine

achirkin:
Prior to this change, raft's host-side implementation of `raft::neighbors::refine` operation uses non-optimal OpenMP thread config by default, spawning as many threads as there are available cores, even if only one thread is used (per-query parallelism with batch size one).
This change fixes that and adds a few optimizations alongside:
  - Use the number of threads = min(cores, queries)
  - Use templates to push the metric-type condition outside the distance computation loop (this should also make it easier to implement new metrics later)
  - Force `tree-optimize` compilation flag in the hopes compiles does the vectorization
  - Split out the host implementation in separate files to be able to compile it without nvcc.


cpp/CMakeLists.txt
```diff
@@ -434,6 +437,10 @@ if(RAFT_COMPILE_LIBRARY)
   # ensure CUDA symbols aren't relocated to the middle of the debug build binaries
   target_link_options(raft_lib PRIVATE "${CMAKE_CURRENT_BINARY_DIR}/fatbin.ld")
 
+  # Optimization: extra compile flags for individual compilation units
+  file(GLOB REFINE_HOST_SRC_FILES "src/neighbors/detail/refine_host_*.cpp")
+  set_property(SOURCE ${REFINE_HOST_SRC_FILES} PROPERTY COMPILE_FLAGS "-ftree-vectorize")
+
```
achirkin, line 443: Note, I added these lines to cmake to do extra optimization and enable vectorization during my experiments. At some point, we could add here some flags to enable x86 avx or arm sve.  I'd like to hear any thoughts of @rapidsai/raft-cmake-codeowners, whether this is considered a good practice. Otherwise, I'd remove this as not critical.

robertmaynard: Specfying `-ftree-vectorize` isn't needed as it is part of the set of flags that `-O3` sets which is the optimization level we set when compiling in Release mode.

As far as setting SVE I am extremely hesistant as lots of arm nodes don't support that ( e.g. only AWS Graviton3 supports it from AWS ).

tfeher (approved):
Thanks Artem for the PR, it looks good to me (apart from the issue with the unnecessary extra compile flag)!

cpp/CMakeLists.txt
achirkin: AVX2 is availabe on virtually all processors used by the raft users. SVE is supported by Grace, so it may be worth adding a separate knob for that.
In the RelWithDebugInfo mode only O2 is used, which does not include 'tree-vectorize'. This is very handy during development and experimentation. Besides, there's no harm in specifying a flag multiple times, I've put it there more for humans to read and being aware that he vectorizarion is important for that bit of code.

Apart from the problems with specific flags, is a construct like this a bad style for a cmake project?

achirkin:
On the performance note: the new code gives some ridiculous speed up of up to x500-x1000 in the case of small batches (n_query = 1), I'm not sure why. Also one cannot rely on the perf reporting in the current `prims` bench, because it uses cuda events to measure the time, while the host-only refine obviously does not use cuda streams. I will bring up the necessary updates to the benchmarks in a follow-up PR.

cpp/CMakeLists.txt
robertmaynard: > Apart from the problems with specific flags, is a construct like this a bad style for a cmake project?
Yes, all sources should be listed explicitly.

> AVX2 is availabe on virtually all processors used by the raft users. SVE is supported by Grace, so it may be worth adding a separate knob for that

These very specific platform flags should be set in a toolchain file that best describes the platform. This toolchains can be selected by the build-infra when it needed. That way we don't publish packages that don't execute on a subset of supported hardware due to a misconfigured default.

We shouldn't conditionally add optimizations like these to a subset of files without first showing why these optimizations fail for the entire project. If somebody wants AVX2 or SVE I see no reason why they wouldn't want that for all of RAFT.

achirkin: Makes sense, thanks for the clarification. My particular use was was to isolate the optimization during the debugging to isolate the bug I faced. I guess this is too specific of a use case to keep it in CMakeLists indefinitely.

cpp/CMakeLists.txt
```diff
@@ -458,6 +461,10 @@ if(RAFT_COMPILE_LIBRARY)
...
```
achirkin, line 467:
```suggestion
```

PR #1654, opened 2023-07-19
Hotfix: wrong constant in IVF-PQ fp_8bit2half

achirkin:
#1644 introduced a direct conversion from the custom ivf-pq fp8 type to half and a bug alongside. The exponent bias value wrong by one bit. The result is the loss of precision in some datasets.
This hotfix just changes this constant.


tfeher (approved):
LGTM.

cjnolet (approved):

PR #1655, opened 2023-07-19
Switch to local testing script

vyasr:
Moves the wheel testing logic out of the workflow into the repo. This matches conda tests more closely and allows each repo to manage its own wheel tests.


divyegala (approved):

vyasr:
These changes were incorporated into https://github.com/rapidsai/raft/pull/1546.

Issue #1367, opened 2023-03-23
[FEA] Follow-up items for benchmarks PR


cjnolet:
The initial version of our ANN benchmarks leaves some things to be desired. I'm capturing them here as a tracker so we can quickly follow-up on them (likely in 23.06 when things calm down a bit):

- [ ] Convert pearl script to Python
- [ ] Use RMM (and corresponding RAFT utilities) throughout benchmarking code
- [ ] Port over Python client scripts (like `run.py`, `install.py`, and `plot.py` from big-ann-benchmarks.
- [ ] Use file/directory management functions in `std` library and remove mannual file management functions
- [ ] Replace logging statements w/ RAFT logger

PR #1657, opened 2023-07-20
ANN-benchmarks: avoid using the dataset during search when possible

achirkin:
This PR changes the behavior of ANN benchmark `dataset.h` to defer reading the data until it is definitely needed. This allows to avoid copying/reading huge datasets in search benchmarks when the database (index) is already built.

### DIFF
--- a/cpp/bench/ann/src/common/dataset.h
+++ b/cpp/bench/ann/src/common/dataset.h
@@ -14,20 +14,26 @@
  * limitations under the License.
  */
 #pragma once
+
+#include <cstdint>
+
+#ifndef CPU_ONLY
 #include <cuda_fp16.h>
+#include <raft/util/cudart_utils.hpp>
+#else
+typedef uint16_t half;
+#endif
+
 #include <errno.h>
 #include <sys/mman.h>
 #include <sys/stat.h>
 
 #include <cassert>
-#include <cstdint>
 #include <cstdio>
 #include <stdexcept>
 #include <string>
 #include <type_traits>
 #include <vector>
-
-#include <raft/util/cudart_utils.hpp>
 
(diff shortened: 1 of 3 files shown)


cpp/bench/ann/src/common/dataset.h
```diff
@@ -270,31 +295,37 @@ Dataset<T>::~Dataset()
...
 }
 
 template <typename T>
 const T* Dataset<T>::base_set_on_gpu() const
 {
+#ifndef CPU_ONLY
   if (!d_base_set_) {
     base_set();
-    RAFT_CUDA_TRY(cudaMalloc((void**)&d_base_set_, base_set_size_ * dim_ * sizeof(T)));
+    RAFT_CUDA_TRY(cudaMalloc((void**)&d_base_set_, base_set_size() * dim() * sizeof(T)));
```
tfeher, line 310: I wonder why we are not using RMM allocator here? (I know it was not changed in this PR, and should be a separate task to fix that).

tfeher (approved):
Thanks @achirkin for the PR. This is useful for performing benchmarks with already saved indices. LGTM.

cpp/bench/ann/src/common/dataset.h
achirkin: I think that was a straight file copy from the separate ann benchmarks project.

cjnolet: Yep, this has been on the TODO but we rushed this out in 23.04 and haven't gotten back to this.

cjnolet: Relevant issue here: https://github.com/rapidsai/raft/issues/1367. Item number 2

PR #1662, opened 2023-07-21
Add notebook for Vector Search - Question Retrieval

lowener:


notebooks/VectorSearch_QuestionRetrieval.ipynb
```diff
@@ -0,0 +1,254 @@
+{
```
cjnolet, line 1: <p><code>RAFT</code> is an acronym- for consistency can we use RAFT instead <code>Raft</code> please?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1662/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

cjnolet, line 1: <p>I wonder if we shold be creating a separate section in <code>depencies.yaml</code> that will output a new conda environment yaml which we can simply tell the user "please create a new conda environment using this yaml file and install raft to run this notebook". That would make it much easier for a user to go from zero to notebook example.</p><p> </p><p>Can we separate the imports from the loading of the data? I've found it to be a good practice as a sanity check so users can quickly see if they have all the needed packages. Another reason this will be useful is that we can document the actual data loading cell as such, while still keeping the general introduction and description separate. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1662/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

cjnolet, line 1: <p>Nitpick, can we use back-ticks around <code>pylibraft</code>?  <span class="ql-cursor">﻿</span></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1662/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

cjnolet, line 1: <pre class="ql-syntax" spellcheck="false">Line #1.    pq_index_mem = pq_index.pq_dim * pq_index.size * pq_index.pq_bits
</pre><p>Just for readability, can we put a space after the first <code>print</code> as well? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1662/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

cjnolet, line 1: <p>We're going to be referencing these notebooks in the docs and likely "embedding" them so that users can read through them without having to run them. As a result, I think it's worth checking in the executed notebook w/ the output text filled in.</p><p> </p><p>And since we'll be checking in the notebook with the output cells filled in, I also think we should run an <code>nvidia-smi</code> in the first or second cells so the user can see the GPU architecture the notebook was executed on. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1662/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

lowener: I think we should go to the simplest option. How about adding a cell with
`!pip install sentence_transformers torch pylibraft-cu11 --extra-index-url=https://pypi.nvidia.com`?
This will minimize user interaction and it will work out of the box.

cjnolet: I think to get this (and the IVF-PQ notebook from @achirkin) into 23.08, we should go ahead and do a pip install for now (let's just point to the raft documentation for raft install and have the cell do a `pip install sentence_transformers torch` for now). We can follow-up with this at a later time.

cjnolet (approved):
Thanks so much for these changes @lowener! And thanks again for adding the cagra example and fixing the build docs @dantegd! LGTM.

cjnolet (approved):

cjnolet (approved):

PR #1665, opened 2023-07-24
CAGRA Python wrappers

dantegd:
First verstion of a CAGRA API in pylibraft. 

Todos:

- [x] C++ raft_runtime instantiations and void overloads
- [x] Cython API
- [x] Solve issue of `cagra_types.hpp` including `#include <raft/util/pow2_utils.cuh>` that makes it need nvcc, blocking a clean C++ only cython build
- [x] Check in pytests
- [x] Add examples to docstrings 
- [x] Accommodate for parameter rename of #1676
- [x] Accomodate changes of #1664 
- [x] Move out of experimental namespace


python/pylibraft/pylibraft/neighbors/cagra/cagra.pyx
```diff
@@ -0,0 +1,1148 @@
...
+
+    def __cinit__(self):
+        self.trained = False
+        self.active_index_type = None
+
+
+cdef class IndexFloat(Index):
+    cdef c_cagra.index[float, uint32_t] * index
+
+    def __cinit__(self, handle=None):
```
tfeher, line 160: Not necessarily in the current release, but as a follow up work:

I would recommend that we have a constructor that takes two arrays (dataset, and graph) and just wraps them as an index object, like here: https://github.com/rapidsai/raft/blob/4574e9aaf68e8cb547d417f60359502d6bcbd94b/cpp/include/raft/neighbors/cagra.cuh#L268

This way we can experiment better with different index configurations. Here is an example workflow

```python
dataset = load_dataset(filename) # loads dataset on host
# it is efficient to build graph using host input array
knn_graph = build_knn_graph(dataset, ...) 

# One could use any method to bulid knn_graph.
# It might be needed to sort the neighbors, if it is not yet done.
# sort_knn_graph(knn_graph) # this is unnecessary if build_knn_graph was used

opt_graph = np.empty((n_rows, graph_degree)
opt_graph = optimize(knn_graph, opt_graph)

# if dataset is correctly aligned, then it makes sense to already move to GPU
# the index will only store a reference to it, and we can keep multiple indices in memory (assuming graphs fit)
dataset = cp.asarray(dataset) 
index = Index(dataset, opt_graph)
distances, neighbors = search(index, queries, k)

# Try another variation, using the same knn_graph, but different optimization options
opt_graph2 = np.empty(n_rows, other_degree)
opt_graph2 = optimize(knn_graph, opt_graph2)

index2 = Index(dataset, opt_graph2)

# search with index2, compare qps / recall to index
```

tfeher (commented):
Started to look at this, overall it looks nice, but had only a superficial view. From the current state of the API, it is not clear what is the use of the seprate `build_knn_graph` and `optimize` methods. They make sense if we can construct an index using their output. I have left a note at the index construction along this line.

python/pylibraft/pylibraft/neighbors/cagra/cagra.pyx
dantegd: Sounds good, I think we can do a follow up PR to add that and testing for it, so for this one will remove the functions so that we keep the released API consistent and simple to use

python/pylibraft/pylibraft/neighbors/cagra/cagra.pyx
```diff
@@ -0,0 +1,852 @@
...
+        max_queries: int, default = 0
+            Maximum number of queries to search at the same time (batch size).
+            Auto select when 0.
+        itopk_size: int, default = 64
+            Number of intermediate search results retained during the search.
+            This is the main knob to adjust trade off between accuracy and
+            search speed. Higher values improve the search accuracy.
+        max_iterations: int, default = 0
+            Upper limit of search iterations. Auto select when 0.*/
+        algo: search_algo, default = c_cagra.search_algo.AUTO
```
cjnolet, line 461: We should probalby list the possible values for this as strings.

python/pylibraft/pylibraft/neighbors/cagra/cagra.pyx
```diff
@@ -0,0 +1,852 @@
...
+            or 32.
+        search_width: int, default = 1
+            Number of graph nodes to select as the starting point for the
+            search in each iteration.
+        min_iterations: int, default = 0
+            Lower limit of search iterations.
+        thread_block_size: int, default = 0
+            Thread block size. 0, 64, 128, 256, 512, 1024.
+            Auto selection when 0.
+        hashmap_mode: hash_mode, default = c_cagra.hash_mode.AUTO
```
cjnolet, line 474: Should we list the possible values for this as strings?

python/pylibraft/pylibraft/neighbors/cagra/cagra.pyx
```diff
@@ -0,0 +1,852 @@
...
+    ...                                   dtype=cp.float32)
+
+    >>> handle = DeviceResources()
+    >>> build_params = cagra.IndexParams(metric="squeclidean",
+    ...     handle=handle
+    ... )
+
+    >>> index = cagra.build(build_params, dataset, handle=handle)
+
+    >>> distances, neighbors = ivf_flat.search(ivf_flat.SearchParams(),
```
cjnolet, line 331: Copy-paste typo

python/pylibraft/pylibraft/neighbors/cagra/cagra.pyx
```diff
@@ -0,0 +1,852 @@
...
+            del self.index
+
+
+@auto_sync_handle
+@auto_convert_output
+def build(IndexParams index_params, dataset, handle=None):
+    """
+    Build the CAGRA index from the dataset for efficient search.
+
+    The build consist of two steps: build an intermediate knn-graph, and
```
cjnolet, line 283:
```suggestion
    The build performs two different steps- first an intermediate knn-graph is constructed, then
```

python/pylibraft/pylibraft/neighbors/cagra/cagra.pyx
```diff
@@ -0,0 +1,852 @@
...
+    optimize it to create the final graph. The index_params object controls
```
cjnolet, line 284:
```suggestion
    it's optimized to create the final graph. The index_params object controls
```

python/pylibraft/pylibraft/neighbors/cagra/cagra.pyx
```diff
@@ -0,0 +1,852 @@
...
+    the node degree of these graphs.
+
+    It is required that dataset and the optimized graph fit the GPU memory.
```
cjnolet, line 287:
```suggestion
    It is required that both the dataset and the optimized graph fit the GPU memory.
```

python/pylibraft/pylibraft/neighbors/cagra/cagra.pyx
```diff
@@ -0,0 +1,852 @@
...
+
+    To customize the parameters for knn-graph building and pruning, and to
```
cjnolet, line 289: Probaby want to remove the remaining portion of this section (everything above the distance metric anyways).

python/pylibraft/pylibraft/neighbors/cagra/cagra.pyx
```diff
@@ -0,0 +1,852 @@
...
+    using pylibraft.cagra.build_knn_graph and
+    pylibraft.cagra.optimize.
+
+     The following distance metrics are supported:
+     - L2
+
+    Parameters
+    ----------
+    index_params : IndexParams object
+    dataset : array interface compliant matrix shape (n_samples, dim)
```
cjnolet, line 300: Can the dataset also exist on the GPU? Chould it support both cuda array interface AND array interface?

python/pylibraft/pylibraft/neighbors/cagra/cagra.pyx
```diff
@@ -0,0 +1,852 @@
...
+    >>> # Using a pooling allocator reduces overhead of temporary array
+    >>> # creation during search. This is useful if multiple searches
+    >>> # are performad with same query size.
+    >>> import rmm
+    >>> mr = rmm.mr.PoolMemoryResource(
+    ...     rmm.mr.CudaMemoryResource(),
+    ...     initial_pool_size=2**29,
+    ...     maximum_pool_size=2**31
+    ... )
+    >>> distances, neighbors = cagra.search(search_params, index, queries,
```
cjnolet, line 643: I don't think we expose the memory resource, do we? (We probaby shouldn't be anyways ... that should be eventually be getting set through the device_resources).

python/pylibraft/pylibraft/neighbors/cagra/cagra.pyx
```diff
@@ -0,0 +1,852 @@
...
+    >>> search_params = cagra.SearchParams(
+    ...     max_queries=100,
+    ...     itopk_size=64,
+    ... )
+
+    >>> # Using a pooling allocator reduces overhead of temporary array
+    >>> # creation during search. This is useful if multiple searches
+    >>> # are performad with same query size.
+    >>> import rmm
+    >>> mr = rmm.mr.PoolMemoryResource(
```
cjnolet, line 638: We should just remove this. I don't think this argument is supported on cagra anyways.

python/pylibraft/pylibraft/neighbors/cagra/cagra.pyx
```diff
@@ -0,0 +1,852 @@
...
+cdef class IndexParams:
+    cdef c_cagra.index_params params
+
+    def __init__(self, *,
+                 metric="sqeuclidean",
+                 intermediate_graph_degree=128,
+                 graph_degree=64,
+                 add_data_on_build=True):
+        """"
+        Parameters to build index for IVF-PQ nearest neighbor search
```
cjnolet, line 103: Copy/paste typo

python/pylibraft/pylibraft/neighbors/cagra/cagra.pyx
```diff
@@ -0,0 +1,852 @@
...
+    get_hmv_float,
+    get_hmv_int8,
+    get_hmv_int64,
+    get_hmv_uint8,
+    get_hmv_uint32,
+    make_optional_view_int64,
+)
+from pylibraft.neighbors.common cimport _get_metric_string
+from pylibraft.neighbors.ivf_pq cimport ivf_pq
+from pylibraft.neighbors.ivf_pq.cpp.c_ivf_pq cimport (
```
cjnolet, line 88: Do we need IVF PQ stuff in here?

python/pylibraft/pylibraft/neighbors/cagra/cagra.pyx
```diff
@@ -0,0 +1,852 @@
...
+            - sqeuclidean is the euclidean distance without the square root
+              operation, i.e.: distance(a,b) = \\sum_i (a_i - b_i)^2,
+            - euclidean is the euclidean distance
+            - inner product distance is defined as
+              distance(a, b) = \\sum_i a_i * b_i.
+        intermediate_graph_degree : int, default = 128
+
+        graph_degree : int, default = 64
+
+        add_data_on_build : bool, default = True
```
cjnolet, line 119: I don't think this parameter is part of cagra is it? This was something that was added to the IVF methods so the user could bypass extend if they wanted.

python/pylibraft/pylibraft/neighbors/cagra/cagra.pyx
```diff
@@ -0,0 +1,852 @@
...
+        ----------
+        max_queries: int, default = 0
+            Maximum number of queries to search at the same time (batch size).
+            Auto select when 0.
+        itopk_size: int, default = 64
+            Number of intermediate search results retained during the search.
+            This is the main knob to adjust trade off between accuracy and
+            search speed. Higher values improve the search accuracy.
+        max_iterations: int, default = 0
+            Upper limit of search iterations. Auto select when 0.*/
```
cjnolet, line 460: Is the training `*/` intentional?

cjnolet (changes requested):
Thanks so much for whipping up this feature at the last minute! I think most of the things are topical (doc related) from what I can tell.

cjnolet (approved):
LGTM, but since I made some changes I'd prefer @dantegd to look over them before we merge.

dantegd (commented):
The changes lgtm @cjnolet , thanks for applying them!

PR #1671, opened 2023-07-25
Update readme to make vector similariity use-case more clear

cjnolet:


img/raft_tech_stack.png
bdice, line 1: @cjnolet Is redis shown on this chart twice?

PR #1676, opened 2023-07-26
Rename CAGRA parameter num_parents to search_width

tfeher:
The name `search_width` is more expressive, therefore this PR renames `num_parents` to `search_width`.

The original CAGRA implementation used both names internally, but externally it used `search_width`.


cjnolet (approved):
LGTM!

PR #1680, opened 2023-07-26
Fix template types for create_descriptor function.

csadorf:
Use correct types for indices pointers and indices when creating the cusparse
descriptor from a device_csr_matrix_view.


cjnolet:
@csadorf is this one ready to be merged?

cjnolet (approved):

PR #1682, opened 2023-07-27
Fix missing parameter for select_k

ucassjy:
The parameter `rmm::mr::device_memory_resource* mr` is missing while passing down.


benfred (approved):
Lgtm-thanks !

PR #1687, opened 2023-07-27
Add algo parameter for CAGRA ANN bench

tfeher:
To achieve high QPS with small batch size, it is recommended to use the `multi_cta` algorithm variant. CAGRA has an input parameter to specify which algorithm to use, but until now the benchmarks only used the default algorithm.

This PR adds configuration option for changing the algorithm parameter for the CAGRA ANN benchmark wrappers. This will be useful while benchmarking various configurations with CAGRA.


cjnolet (approved):
LGTM. Thanks @tfeher!

PR #1693, opened 2023-07-28
Remove push condition on docs-build

raydouglass:
Nightly builds are skipping docs builds due to this condition: https://github.com/rapidsai/raft/actions/runs/5688559993/job/15419505390

Removing it should allow docs to build when triggered by the nightly pipeline.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -59,11 +59,14 @@ jobs:
     secrets: inherit
     uses: rapidsai/shared-action-workflows/.github/workflows/custom-job.yaml@branch-23.08
     with:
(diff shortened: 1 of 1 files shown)


ajschmidt8 (commented):
the rest of this job should look more like `cudf`'s block below

- https://github.com/rapidsai/cudf/blob/7746af421c88c59e2261d5edfcfe61b0fc7835fd/.github/workflows/build.yaml#L56-L69

ajschmidt8 (approved):

raydouglass:
Admin merging since these changes aren't tested in CI

PR #1716, opened 2023-08-03
[FEA] Provide device_resources_manager for easy generation of device_resources

wphicks:
Edit: I have renamed the object to raft::device_resources_manager on the (excellent) suggestion of @cjnolet. Substitute the name accordingly in the description below.

### Summary
This PR introduces a new utility (`raft::resource_manager`) to RAFT that is used to help downstream applications correctly generate the `raft::device_resources` they need to interact with the RAFT API.

### Purpose
As more vector search applications have begun integrating RAFT, it has become apparent that correctly managing CUDA resources like streams, stream pools, and device memory can be a challenge in a codebase that has previously focused exclusively on CPU execution. As a specific example, these applications are generally highly multi-threaded on the host. As they begin to use RAFT, they typically make use of the default `device_resources` constructor to generate the requisite `device_resources` object for the API. Because this default constructor makes use of the default stream per thread, the application uses as many streams as there are host threads. This behavior can lead to exhaustion of device resources because all of those host threads simultaneously launch work on independent streams.

In a CUDA-aware codebase, we might expect the application to manage its own limited pool of streams, but requiring this creates and unnecessary barrier for RAFT adoption. Instead, the `resource_manager` provides a straightforward method for limiting streams and other CUDA resources to sensible values in a highly multi-threaded application.

### Usage
To use the `resource_manager`, the host application will make use of setters which can provide control over various device resources. For instance, to limit the total number of streams used by the application to 16 per device, the application would call the following during its startup:
```
raft::resource_manager::set_streams_per_device(16);
```
After startup, if the application wishes to make a RAFT API call using a `raft::device_resources` object, it may call the following:
```
auto res = raft::resource_manager::get_device_resources();
some_raft_call(res);
```
If the same host thread calls `get_device_resources()` again in another function, it will retrieve a `device_resources` object based on the exact same stream it got with the previous call. This is similar in spirit to the way that the default CUDA stream per thread is used, but it draws from a limited pool of streams. This also means that while each host thread is associated with one CUDA stream, that stream may be associated with multiple host threads.

In addition to drawing the `device_resources` primary stream from a limited pool, we can share a pool of stream pools among host threads:
```
// Share 4 stream pools among host threads, with the same pool always assigned to any given thread
raft::resource_manager::set_stream_pools_per_device(4);
```

Besides streams and stream pools, the resource manager optionally allows initialization of an RMM memory pool for device allocations:
```
// Start the pool with only 2048 bytes
raft::resource_manager::set_init_mem_pool_size(2048);
// Allow the pool to use all available device memory
raft::resource_manager::set_max_mem_pool_size(std::nullopt);
```
For downstream consumers who know they want a memory pool but are not sure what sizes to pick, the following convenience function just sets up the memory pool with RMM defaults:
```
raft::resource_manager::set_mem_pool();
```
If no memory pool related options are set or if the maximum memory pool size is set to 0, no memory pool will be created or used. Furthermore, if the type of the current device memory resource is non-default, no memory pool will be created or used, and a warning will be emitted. We assume that if the application has already set a non-default memory resource, this was done intentionally and should not be overwritten.

### Design
This object is designed with the following priorities:

- Ease of use
- Thread safety
- Performance
- (Lastly) Access to as many `device_resources` options as possible

If a downstream application needs complete control over `device_resources` creation and memory resource initialization, that codebase is probably already CUDA-aware and will not benefit from the resource manager. Therefore, we do not insist that every possible configuration of resources be available through the manager. Nevertheless, codebases may grow more CUDA-aware with time, so we provide access to as many options as possible (with sensible defaults) to provide an on-ramp to more sophisticated manipulation of device resources.

In terms of performance, the goal is to make retrieval of `device_resources` as fast as possible and avoid blocking other host threads. To that end, the design of `resource_manager` includes a layer of indirection that ensures that each host thread needs to acquire a lock only on its first `get_device_resources` call for any given device. The `resource_manager` singleton maintains an internal configuration object that may be updated using the setter methods until the first call to `get_device_resources` on any thread. After that, a warning will be emitted and no changes will be made on any subsequent setter calls.

Within the `get_device_resources` call, each thread keeps track of which devices it has already retrieved resources for. If it has not yet retrieved resources for that device, it acquires a lock. It then marks the configuration as finalized and checks to see if any thread has initialized the shared resources for that device. If no other thread has, it initializes those resources itself. It then updates its own thread_local list of devices it has retrieved resources for so that it does not need to reacquire the lock on subsequent calls.

### Questions
~This PR is still a WIP. It is being posted here now primarily to gather feedback on its public API. A thorough review of the implementation can happen once tests have been added and it is moved out of WIP.~ Edit: This PR is now ready for review. Implementation feedback welcome!

One question I have is about two convenience functions I added right at the end: `synchronize_work_from_this_thread` and `synchronize_work_from_all_threads`. The idea behind these functions is that for the target audience of this feature, it may be helpful to provide synchronization helpers that clearly indicate what their execution mean in relation to the more familiar synchronization requirements of the host code. The first is intended to communicate that it is guaranteed to block host execution on this thread until all work submitted to the device from that thread is completed. The hope is that this can increase confidence around questions of synchronicity that developers unfamiliar with CUDA streams sometimes have. Because this helper is not strictly required for the core functionality of `resource_manager`, however, it would be useful to have feedback on whether others think it is worthwhile.

I am even less certain about `synchronize_work_from_all_threads`. The idea behind this is that it provides a way to block host thread execution until all work submitted on streams under the `resource_manager`'s control have completed. A natural question is why we would not just perform a device synchronization at that point. My justification for that is that the application may also be integrating other libraries which have their own stream management and synchronization infrastructure. In that case, it may be desirable _not_ to synchronize work submitted to the device by calls to the other library. It would be useful to hear if others think this might be useful for the target audience of this PR or if we should just push them toward a device synchronization to avoid unpleasant edge cases.


cjnolet:
I took a cursory look through the code. The intent is pretty clear and I think the implementation looks good. I'll take a closer look tomorrow and read through the description more thoroughly. 

I do suggest we name this `device_resource_manager` or event `device_resources_manager` so that it's consistent with all the other naming conventions in RAFT- it becomes blatantly clear just through the naming that it's specific to managing device resources.

wphicks:
Looks like my auto-signing configuration for git is not working. I'll rebase to add signatures to all commits before final review.

cpp/include/raft/core/device_resources_manager.hpp
```diff
@@ -0,0 +1,578 @@
...
+  {
+    static std::atomic<std::size_t> thread_counter{};
+    thread_local std::size_t id = ++thread_counter;
+    return id;
+  }
+
+  // This struct holds the various parameters used to control
+  // construction of the underlying resources shared by all
+  // `device_resources` objects returned by `get_device_resources`
+  struct resource_params {
```
divyegala, line 104: Every device will have the exact same set of `params` if I understand correctly?

divyegala (commented):
Thanks for this PR! It's shaping really well and I am very happy with your descriptions as well how well-commented the code is.

cpp/include/raft/core/device_resources_manager.hpp
```diff
@@ -0,0 +1,578 @@
...
+      return workspace_allocation_limit_;
+    }
+    void synchronize() const
+    {
+      for (auto i = std::size_t{}; i < stream_count(); ++i) {
+        interruptible::synchronize(streams_->get_stream(i));
+      }
+      for (auto i = std::size_t{}; i < pool_count(); ++i) {
+        for (auto j = std::size_t{}; j < pools_[i]->get_pool_size(); ++j) {
+          interruptible::synchronize(pools_[i]->get_stream(j));
```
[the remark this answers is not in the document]
wphicks: My own opinion is no, the RAFT API should at least leave the door open to overlapping usage of the stream pool with other calls and then synchronizing later, but I certainly understand why this is not as clear cut as the question of the primary stream. In general, given that RAFT provides primitive building blocks, I think the principle that "you can build safe on top of fast but not fast on top of safe" is the determining factor here. If a downstream application is sophisticated enough to do that overlapping of pool usages, I don't think we should preclude them from doing so.

cpp/include/raft/core/device_resources_manager.hpp
wphicks: That is correct. My feeling on this was that if the application requires fine-grained control of these sorts of parameters per device, it probably is CUDA-aware enough to build its own `device_resources`. This still leaves the application in control of whether or not it uses a specific device, just not e.g. different number of total streams per device.

cpp/include/raft/core/device_resources_manager.hpp
```diff
@@ -0,0 +1,580 @@
...
+   * @param int device_id If provided, the device for which resources should
+   * be returned. Defaults to active CUDA device.
+   * @param workspace_mr If provided, a separate memory resource to be used
+   * for allocating temporary workspaces in RAFT calls.
+   */
+  static auto get_device_resources(int device_id = device_setter::get_current_device(),
+                                   std::shared_ptr<rmm::mr::device_memory_resource> workspace_mr = {
+                                     nullptr})
+  {
+    auto const& components = get_manager().get_device_components(device_id);
```
cjnolet, line 405: From a larger overarching design standpoint, I'd have expected to see this manager actually manage a series of device resources objects, which might each have their own varying state, rather than as a "super factory" that just produces new instances. For one- this is assuming that the creation of the instances themselves is cheap enough that it's okay to do this often. Another problem with this is that it sidesteps the individual device resources from being able to maintain their own independent state. I don't mind that the `resources_manager` is the curator for consolidating the initial settings, but I'd prefer to have the resources manager actually manage a collection of `device_resources` instances, rather than just producing new instances each time they are needed. 

For example, FAISS does something very similar to what I'm suggesting, and I think it better defines the relationship and separation of concerns between the manager of instances and the actual instances.

cpp/include/raft/core/device_resources_manager.hpp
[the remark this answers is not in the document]
cjnolet: I agree w/ Divye here- this feels like a sign of a somewhat bad smell in the design- the device resources should be managing their streams and the device_resources_manager should be managing some set of device_resources, and then calling `res.sync_stream()` or `res.sync_stream_pool()` on those instead (which should ultimately be getting propagated down to `raft::resources::sync_stream()` to choose the implementation of syncer (interruptible in this case). 

If we are using a stream pool in the `device_resources_manager`, I think it's the responsibility of the manager to map the individual streams to their associated `device_resources` instances, but after it does that mapping, it should be allowing the `device_resources` to manage their own internal state (e.g. by calling `res.sync_stream()` at the higher layer).

cpp/include/raft/core/device_resources_manager.hpp
```diff
@@ -0,0 +1,580 @@
...
+        streams_{[&params, this]() {
+          auto scoped_device = device_setter{device_id_};
+          auto result        = std::unique_ptr<rmm::cuda_stream_pool>{nullptr};
+          if (params.stream_count) {
+            result = std::make_unique<rmm::cuda_stream_pool>(*params.stream_count);
+          }
+          return result;
+        }()},
+        pools_{[&params, this]() {
+          auto scoped_device = device_setter{device_id_};
```
cjnolet, line 144: I think the device id can and should be an instance variable and state on each individual `device_resources` instances and once "set", the corresponding device_resources should be the "source of truth" for the device id it's mapped to. Again- all of the state for each individual device_resources instances should be managed down at that layer- the device_resources_manager should just be managing its own device_resources instances.

cpp/include/raft/core/device_resources_manager.hpp
```diff
@@ -0,0 +1,580 @@
...
+      iter = std::find_if(
+        std::begin(per_device_components_),
+        std::end(per_device_components_),
+        [device_id](auto&& components) { return components.get_device_id() == device_id; });
+    }
+    return *iter;
+  }
+
+  // Thread-safe setter for the number of streams
+  void set_streams_per_device_(std::optional<std::size_t> num_streams)
```
cjnolet, line 300: How is this context configured? Is this object being mutated often from different threads?

cjnolet (changes requested):
I finally got a chance to review this more thoroghly. Mostly, I'd like to separate the concerns a little better between individual `device_resources` and the `device_resources_manager`. From the RAFT perspective, the `device_resources` is not just an API for invoking RAFT's APIs. It's supposed to be responsible for maintaining the state assigned to an individual device. In many ways, this `device_resources_manager` implementation is side-stepping the individual resource management when we could be using this as an opportunity to reuse the existing centralized resource management infrastructure. 

Other than that, I think the design is great RAFT has been in need of something that can manage multiple devices (e.g. instances of `device_resources` for different devices) across threads.

cpp/include/raft/core/device_resources_manager.hpp
wphicks: If we want to cache the `device_resources`, that's fine. I had considered doing so but did not like that it automatically extended the lifetime of the resources object beyond what might actually be used by the caller. I figured that we would prefer this to _just_ generate the `device_resources` object and then have the consumer use that object as it is normally used by consumers of the RAFT API. It feels like caching is adding an additional complication to do something that is already handled by the `device_resources` object. I'll add it for now and we can revert if we don't like it.

cpp/include/raft/core/device_resources_manager.hpp
wphicks: We still need to generate the streams to be shared by the `device_resources` objects, and those need to be associated with a device id. If we're caching the `device_resources` objects, then yes, we can query those objects in the `find` call to retrieve the `device_resources` associated with the requested device.

wphicks:
> In many ways, this device_resources_manager implementation is side-stepping the individual resource management when we could be using this as an opportunity to reuse the existing centralized resource management infrastructure.

Could you be more specific? The intention of this object is to ensure that `device_resources` objects are constructed from a fixed pool of resources. E.g. Multiple `device_resources` objects may share a single stream in order to control the total number of streams used by the application per device. If you are saying that the `device_resources` objects themselves should be solely responsible for ownership of the underlying streams, there is no way to achieve that. If you are suggesting that we should create initial `device_resources` objects to distribute their resources among the final `device_resources` objects returned, this seems unnecessary when all we really want is a collection of streams (adequately encapsulated by `rmm::cuda_stream_pool`) and a collection of pools (adequately encapsulated by any container of `rmm::cuda_stream_pool`).

cpp/include/raft/core/device_resources_manager.hpp
cjnolet: > I'm not entirely sure what you're asking.

I'm asking specifically from within the context of the the existing integrations (Redis and Milvus). Just trying to get a feel for the larger picture and how the threads interact there. I was under the impression there was likely some type of setup or initialization step, which would likely be done at the very beginning in a single thread and that the various threads are just being used to create/use their local `device_resources` instances on whatever device theyve been mapped to (or I guess even depending on whatever device happens to be available in the case of load balancing). I understand that it's better to be safe here and I'm not against that at all (especially since we are already locking in `raft::resources`).

wphicks:
Refactored to cache the generated `device_resources`. Still to finalize:

- [ ] Fix setting of workspace memory resource
- [ ] Fix docs
- [ ] Move the log_level_setter implementation

cpp/include/raft/core/device_resources_manager.hpp
```diff
@@ -0,0 +1,609 @@
...
+ * requested) stream pool.
+ *
+ * Typical usage might look something like the following:
+ * @code
+ * void initialize_application() {
+ *   raft::device_resources_manager::set_streams_per_device(16);
+ * }
+ *
+ * void foo_called_from_multiple_threads() {
+ *   auto res = raft::device_resources_manager::get_device_resources();
```
cjnolet, line 63: Does this usage need to change now from recent updates?

cpp/include/raft/core/device_resources_manager.hpp
```diff
@@ -0,0 +1,609 @@
+/*
+ * Copyright (c) 2023, NVIDIA CORPORATION.
```
cjnolet, line 2: The implementation looks good. Can you also explicit add this to the docs (might be easiest to create a doxygen group and reference the group in the docs).

wphicks:
Updated everything requested except the docs. I'll finish those up tomorrow.

cpp/include/raft/core/logger-inl.hpp
```diff
@@ -152,4 +152,16 @@ RAFT_INLINE_CONDITIONAL void logger::flush() { pimpl->spdlogger->flush(); }
...
+  explicit log_level_setter(int level)
+  {
+    prev_level_ = logger::get(RAFT_NAME).get_level();
+    logger::get(RAFT_NAME).set_level(level);
+  }
+  ~log_level_setter() { logger::get(RAFT_NAME).set_level(prev_level_); }
+
+ private:
+  int prev_level_;
+};
```
ahendriksen, line 165: Sorry, I should have given this a bit more thought in my previous comment. People who use the raft::compiled library will not be able to access `log_level_setter` (because they only include logger-ext.hpp).

The "perfect" solution would be to split up log_level_setter into a declaration (logger-ext.cuh) and a definition (logger-inl.cuh), like was done with logger.

The practical solution is to put log_level_setter in logger-ext.cuh. Then consumers of both `raft::compiled` and `raft::header_only` (?) will see log_level_setter. The compile time overhead of this struct is close to zero, so that should be fine. 

Background: the reason this header was split is that the logger-inl.cuh includes spdlog, which is a beast to compile.

ahendriksen (changes requested):
One more move of log_level_setter requested. Sorry, should have given this a bit more thought in the previous comment.

cpp/include/raft/core/logger-inl.hpp
wphicks: No worries! Thanks for catching it before this merged. Updating accordingly

ahendriksen (approved):
Thanks for moving `log_level_setter` to logger-ext.cuh. LGTM!

cjnolet (approved):
LGTM. Thanks @wphicks!

PR #1717, opened 2023-08-04
Update RAFT documentation

lowener:
- Various documentation updates on C++ and Python doc, mainly for raft::neighbors
- Add QPS vs Recall plot


cjnolet (commented):
Overall I think the changes look great. I'm holding off on approving only because of the format changes for the code examples in the python docs. We had originally shifted these examples to a different format and I don't know if we want to revert them. 

cc @benfred @galipremsagar for their thoughts here.

python/pylibraft/pylibraft/cluster/kmeans.pyx
```diff
@@ -84,37 +84,40 @@ def compute_new_centroids(X,
     Examples
     --------
 
-    >>> import cupy as cp
+    .. code-block:: python
```
galipremsagar, line 87: @lowener Any reason why we are adding python code-blocks? AFAIK sphinx does a good job at parsing the example code section of documentation.

lowener: Currently the code examples on Sphinx are split into many small blocks, I thought that it was unintentional and I didn't know that it was a shift to a different format.
If this small block format has already been decided then I should revert my changes.
Current
image
Proposed
image

python/pylibraft/pylibraft/cluster/kmeans.pyx
```diff
@@ -84,37 +84,40 @@ def compute_new_centroids(X,
...
 
-    >>> from pylibraft.common import Handle
-    >>> from pylibraft.cluster.kmeans import compute_new_centroids
+        >>> import cupy as cp
```
galipremsagar, line 90:
```suggestion
```
@lowener To slove the issue you are seeing in the docs, you just have to delete these empty lines instead of creating a python code-block.. Good catch!

python/pylibraft/pylibraft/cluster/kmeans.pyx
galipremsagar: Got it, I have a simpler solution for it. See my comment below: https://github.com/rapidsai/raft/pull/1717/files#r1284444539

python/pylibraft/pylibraft/cluster/kmeans.pyx
lowener: One last argument that I have in favor of the code-block is that it improves readability. In the previous example there is small breaks between the includes, the variable declaration, and the RAFT code. I think it can be appreciated by our Python users. What do you think?

cjnolet: IIUC, the .. code block is a sphinx-specific documentation style while the other format is more universal across python. Is that true?

lowener: From what I understand as well yes. When running `?cagra.build` in IPython for details about this function users would see written literally `.. code-block::python`.

galipremsagar: Yeah, that's right. Lets drop `code-block` usages in that case.

galipremsagar (approved):

cjnolet (approved):
Changes LGTM! If we are choosing to document simple param structs at the class level, we should be as consistent as possible and make sure we do it everywhere. 

Looks like the doctest pytests are failing. We should be able to merge after we fix those.

PR #1728, opened 2023-08-08
Fix `__repr__` on CAGRA python index

benfred:
Running `cagra.build(params, dataset)` in a jupyter notebook would throw an exception (`TypeError: can only concatenate str (not "list") to str`). This was happening in the `__repr__` method on the index

Fix an add a basic unittest


benfred:
no longer necessary w/ the fix in https://github.com/rapidsai/raft/pull/1717/files - closing

PR #1730, opened 2023-08-09
[FEA] Use CAGRA in C++ template

lowener:
Proposal to change the C++ template from a distance computation to a vector search application using CAGRA.


cpp/template/CMakeLists.txt
```diff
@@ -28,11 +28,13 @@ rapids_cuda_init_architectures(test_raft)
 
 project(test_raft LANGUAGES CXX CUDA)
 
+find_package(OpenMP)
```
cjnolet, line 31: Why are we adding openmp to the dependencies here?

cjnolet (changes requested):
Looks good- just one question about dependencies.

cpp/template/CMakeLists.txt
lowener: CAGRA is using openmp and the link from `raft::compiled` to openmp is not automatic.
I've found that openmp needs to be found in the dependencies of the consumer`s project for the link to happend: 
https://github.com/rapidsai/raft/blob/b7d2a9aba4c38c575c4b397a7260d139b2bca427/cpp/CMakeLists.txt#L445-L451

cjnolet: @robertmaynard do you have any ideas of why this might be happening? Are we missing something in the `raft_lib` side?

robertmaynard: There is no exporting of the need for OpenMP to consuming projects. https://github.com/rapidsai/raft/blob/branch-23.10/cpp/CMakeLists.txt#L130 would need to change to `rapids_find_package` with the proper export sets to remove the need for consumers to explicitly find OpenMP

cjnolet (approved):
LGTM!

PR #1734, opened 2023-08-10
Improvements to ANN Benchmark Python scripts and docs

divyegala:


conda/environments/bench_ann_cuda-118_arch-x86_64.yaml
```diff
@@ -34,6 +34,7 @@ dependencies:
 - nccl>=2.9.9
 - ninja
 - nlohmann_json>=3.11.2
+- pyyaml
```
cjnolet, line 37: Matplotlib doesn't seem to be getting installed as a dependency of the `libraft-ann-bench` conda package.

cjnolet (changes requested):
Couple minor things but these are huge improvements. Thanks!

bench/ann/run.py
```diff
@@ -121,7 +125,10 @@ def main():
         search = args.search
 
     # Read configuration file associated to dataset
-    conf_filepath = args.configuration
+    if args.configuration:
+        conf_filepath = args.configuration
+    else:
+        conf_filepath = os.path.join(scripts_path, "conf", f"{args.dataset}.json")
```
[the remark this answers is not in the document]
divyegala: Since dataset has a default, that enforces that one of them is specified. Should I remove the default and throw an error instead if both are unspecified?

conda/environments/bench_ann_cuda-118_arch-x86_64.yaml
divyegala: Yes, it's part of the bench-ann environment file. It should not be a dependency of the libraft-ann-bench conda package because that package is just the executables

cjnolet (approved):

raydouglass (approved):

PR #1736, opened 2023-08-14
Replace GEMM backend: cublas.gemm -> cublaslt.matmul

achirkin:
This PR replaces the current cublas `gemm` backend of `raft::linalg::gemm`  with cublasLt `matmul`. The latter is more flexible and allows to decouple selection of the algorithm heuristics from its execution.
Thanks to this change, this PR adds memoization of the matmul heuristics and the other arguments (matrix layouts and the matmul descriptor).

#### Performance on specific workloads
IVF-PQ performs two gemm operations during pre-processing on small work sizes. The preprocessing consists of a few kernel launches and a rather heavy logic on CPU side (which results in gaps between the kernel launches).
This PR **roughly halves the gemm kernel launch latency** (approx 10us -> 5us, as measured by NVTX from entering `matmul` wrapper on the host to the launch of the kernel).
As a motivation example: this PR improves QPS of IVF-PQ by ~5-15% on small batches (tested on SIFT-128, n_queries = 1, n_probes = 20 and 200) .

#### Synthetic benchmarks: no significant difference
Running all 4K+ benchmarks across RAFT does not bring significant difference in CPU/GPU exec time.
  - Overall, the average exec time reduction of ~0.5%
  - 100+ benchmarks show 5-10% time reduction
  - 9 benchmarks show 5-10% time increase (none of them use GEMM)

Only a small fraction of RAFT benchmarks actually use GEMM, so most of the stronger deviations are likely due to pure chance.  Having no gain across all benchmarks is not surprising, because we've designed most of them for somewhat larger work sizes, which hides the gemm latency.


achirkin:
### A note on caching
Since cublas 12, `cublasLtMatmulAlgoGetHeuristic` actually does the [heuristics caching](https://docs.nvidia.com/cuda/cublas/index.html#heuristics-cache) on their own (can be controlled by CUBLASLT_HEURISTICS_CACHE_CAPACITY env variable).
Yet I think, raft caching still makes sense for these reasons:
  - It caches not only heuristics, but also the descriptors saving a little bit more work (my NVTX profiling shows the a small reduction in the total launch latency ~ 6-7us -> 5us).
  - raft implementation restricts the space of possibly layouts, thus slightly increases the chance of the cache hit and the size of the cache

#### Cache scope
In raft, I see three ways to implement caching:
  1. [currently selected] `static thread_local` variable inside the function: create a new cache per-thread and per input types combination (float/half/etc)
  2. `static` variable inside the function: compared to (1) would allow to reuse the cache between threads, but would require mutexes/guards
  3. New `raft::resource`. Same as (2) but more flexible and explicit. However, this would erase input types and thus increase the size of the cache keys (may hurt performance?..).

cpp/include/raft/linalg/detail/gemm.hpp
```diff
@@ -13,231 +13,398 @@
...
-                             ldb,
-                             beta,
-                             C,
-                             ldc,
-                             stream));
+  common::nvtx::range<common::nvtx::domain::raft> batch_scope(
+    "linalg::matmul(m = %d, n = %d, k = %d)", m, n, k);
+  std::shared_ptr<matmul_desc> mm_desc{nullptr};
+  matmul_key_t mm_key{m, n, k, lda, ldb, ldc, trans_a, trans_b};
+  static thread_local cache::
```
cjnolet, line 262: cc @robertmaynard for thoughts on implications to static link library (and potential conflics between shared state across applications).

robertmaynard: The issue around caching is due to these line from cublas docs:
```
A cuBLAS library context is tightly coupled with the CUDA context that is current at the time of the cublasCreate() call. An application that uses multiple CUDA contexts is required to create a cuBLAS context per CUDA context and make sure the former never outlives the latter.
```

So in our case we need all functions that call cublas or cache cublas types to be hidden symbols so that we maintain separate versions in different DSO as those will have separate CUDA contexts. The current version would need to have everything marked with `visibility("hidden")`

achirkin: Regarding the visibility of the functions: this PR does not change anything in that regard, so it's orthogonal to any solution we come up with.

Regarding the visibility of the static/thread_local variable declared *inside the function*. Am I right that it's not visible by default, so no action is needed here? I assume if RAFT is statically/dynamically linked multiple times, the variables like this won't collide. Also, since it is a locally defined variable, it is instantiated only upon the first call to the function.

achirkin: Resolved: avoiding the issue by keeping the cache inside the raft resources handle.

cpp/include/raft/core/resource/user_resource.hpp
```diff
@@ -0,0 +1,71 @@
...
+ * @param[in] res the raft resources object
+ * @return a pointer to the user-defined resource.
+ */
+template <typename Store>
+auto get_user_resource(resources const& res) -> Store*
+{
+  if (!res.has_resource_factory(resource_type::USER_DEFINED)) {
+    res.add_resource_factory(std::make_shared<user_resource_factory>());
+  }
+  return res.get_resource<user_resource>(resource_type::USER_DEFINED)->load<Store>();
```
achirkin, line 68: Would be nice to reuse the resources' lock here rather than having two different locks for getting the resource and loading the store from the resource. This could be done e.g. by passing an optional lambda to the `resource::get_resource` to  apply post-processing while holding the lock.
I haven't added this to not make the PR overly invasive.

cpp/include/raft/linalg/gemm.cuh
```diff
@@ -27,8 +28,65 @@
...
+ * @param [in] ldc leading dimension of C
+ */
+template <bool DevicePointerMode = false, typename S, typename A, typename B, typename C>
+void matmul(raft::resources const& res,
+            bool trans_a,
+            bool trans_b,
+            uint64_t m,
+            uint64_t n,
+            uint64_t k,
+            const S* alpha,
```
cjnolet, line 65: Pointer-based APIs are deprecated. Please use mdspan (or at least add an mdspan variant to the group that will be included in the docs below and test it). Eventually this one will be removed with the other pointer-based APIs.

cpp/include/raft/linalg/gemm.cuh
```diff
@@ -27,8 +28,65 @@
 #include <raft/core/resources.hpp>
 #include <raft/util/input_validation.hpp>
 
-namespace raft {
-namespace linalg {
+namespace raft::linalg {
```
cjnolet, line 31: I wonder if instead of gemm.hpp, we should be putting the new matmul function in matmul.hpp?

cpp/include/raft/core/resource/cublaslt_handle.hpp
```diff
@@ -0,0 +1,68 @@
...
+#include <cublasLt.h>
+#include <raft/core/cublas_macros.hpp>
+#include <raft/core/resource/resource_types.hpp>
+#include <raft/core/resources.hpp>
+
+#include <memory>
+
+namespace raft::resource {
+
+class cublaslt_resource : public resource {
```
cjnolet, line 27: Please make sure this gets added to the docs like the others.

cpp/include/raft/core/resource/user_resource.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+#include <raft/core/resource/resource_types.hpp>
+#include <raft/core/resources.hpp>
+
+#include <memory>
+#include <typeindex>
+
+namespace raft::resource {
+
+class user_resource : public resource {
```
cjnolet, line 26: Please make sure this gets added to the docs (and is well documented)

cjnolet (changes requested):
I think the user defined resource is a great concept and it should solve some of the problems we have with extensibility from using the fixed-length array for the other resources.

Of course this does come with the cost of hashmap lookups, but I think the benefits will outweigh the drawbacks here.

cpp/include/raft/linalg/gemm.cuh
achirkin: resoling: I've removed this altogether, because it would mirror the mdspan api of the gemm that is already defined in this file.

cpp/include/raft/core/resource/user_resource.hpp
achirkin: Thanks for the reminder!

cjnolet (changes requested):

cpp/include/raft/linalg/detail/matmul.hpp
```diff
@@ -0,0 +1,293 @@
+/*
```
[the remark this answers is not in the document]
achirkin: In my opinion, this is a little different compared to the other wrapper code. In `detail/cublas_wrappers.hpp`, we have thin wrappers for unifying cuBLAS functions with different types and setting up the worker stream. For cuBLASLt `matmul`, we don't need it.
I think, the code in `detail/matmul.hpp` is more akin to `detail/gemm.hpp`, which introduces new functions that make use of the lower-level cuBLAS functions. It has different arguments and some logic to combine multiple cuBLASLt functions, caching, and some extra structures, etc. Hence I've put it in a dedicated `detail/matmul.hpp` to distance it a little bit from cuBLASLt.

That being said, I don't mind renaming `matmul.hpp` into `cublaslt_wrappers.hpp` and to host more cuBLASLt-related functions there later. Would you like me to proceed with the change?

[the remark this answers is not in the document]
cjnolet: we aren't only using the wrappers for the types. We're using them to centralize and add a layer of indirection between RAFT and the cuda math libs. As an example, we shouldn't making any calls to the cublasLt libraries directly outside of those wrappers. There are, unfortunately, a couple remnants of legacy code in places where we're still calling cusparse directly, but that's the exception and not the rule here. This header file is doing the same thing as the other cuda math libs wrappers are doing (albeit a tad differently) and I'd like to keep the wrapper designs consistent so we aren't directly making any cublaslt calls outside of the wrapper itself. I'm even okay using raft::resources in the `cublastlt_wrappers.h` instead of passing the handle itself.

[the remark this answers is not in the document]
achirkin: Ok, got it! I moved the `matmul.hpp` to `cublaslt_wrappers.hpp`.

cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp
```diff
@@ -0,0 +1,294 @@
...
+            uint64_t k,
+            const S* alpha,
+            const A* a_ptr,
+            uint64_t lda,
+            const B* b_ptr,
+            uint64_t ldb,
+            const S* beta,
+            C* c_ptr,
+            uint64_t ldc,
+            cudaStream_t stream)
```
cjnolet, line 264: This should either accept cublaslt handle + stream or raft::resources but not raft::resources + stream. This is too error-prone and confusing for potential users.

cpp/include/raft/linalg/detail/gemm.hpp
```diff
@@ -13,231 +13,135 @@
...
-          const math_t* A,
-          const int lda,
-          const math_t* B,
-          const int ldb,
-          const math_t* beta,
-          math_t* C,
-          const int ldc,
-          cudaStream_t stream)
+template <typename T, bool DevicePointerMode = false>
+void legacy_gemm(raft::resources const& res,
```
cjnolet, line 25: Same here- cublaslt handle + stream or raft resources but not both.

cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp
achirkin: I totally agree we need to do this, but I'm a little bit hesitant to do that in this PR. 
  1. Removing the resources argument is definitely not an option. In cublaslt_wrappers `matmul`, we do need the resources handle, because we get the cache store from it as well, which is better not exposed in the arguments to make sure all users use the same cache.
  2. Removing the stream is doable but tricky. The problem is that both resources handle and the stream are present in all but one (which is not used anywhere) overload of the public gemm in `raft/linalg/gemm.cuh`. If we are to remove the stream in the detail namespace, we need to create a new resources handle in the public functions on each invocation to set it. I'm afraid this can add a bit of unwanted overhead, especially in the solvers like lstsq or rsvd, where gemms are invoked in a tight loop.

Would you like me to proceed with option 2?

achirkin: Another option (3). I've noticed our public gemm still has the `.cuh` extension. I suggest we can do the following:
  - Keep `linalg/detail/cublaslt_wrappers.hpp` as-is in the PR for now.
  - Add couple new overloads to `linalg/detail/gemm.hpp` that do not have the stream argument.
  - Deprecate the whole header `linalg/gemm.cuh`.
  - Add a new public header `linalg/gemm.hpp` with all overloads not having the stream argument.
  - In the future follow-ups switch to using the new gemm header and only update the `linalg/detail/cublaslt_wrappers.hpp` after we remove the deprecated  `linalg/gemm.cuh`.

cpp/include/raft/core/resource/user_resource.hpp
```diff
@@ -0,0 +1,83 @@
...
+    auto pos = map_.find(key);
+    if (pos != map_.end()) { return reinterpret_cast<ResourceT*>(pos->second.get()); }
+    auto store_ptr = new ResourceT{};
+    map_[key] =
+      std::shared_ptr<void>(store_ptr, [](void* ptr) { delete reinterpret_cast<ResourceT*>(ptr); });
+    return store_ptr;
+  }
+
+ private:
+  std::unordered_map<std::type_index, std::shared_ptr<void>> map_{};
```
achirkin, line 46: While I was thinking about overheads of copying the resources handle in the gemm/matmul, it occurred to me that we hadn't discussed what should happen to user resources when the resources handle is copied. I see three sensible options:
  1. [current implementation] The whole map is shared between the old and the new handle. Low overhead, but setting a user resource in one handle affects the other handle as well.
  2. Do a shallow copy of the map when copying the resources handle. Incurs more unwanted overheads, but the behavior of individual user resources is the same as of the  compile-time-registered resources (setting an individual resource affects only the current handle).
  3. Drop the user resource map on copy. Incurs a low overhead during copy. Resets the state of the user resources, not sure for better or worse (e.g. the kernel cache resources  are independent).

@cjnolet, do you have a preference here?

cjnolet: @achirkin maybe we could provide a flag for this? Perhaps even have a predefined element in the map of user resources containing this flag? This way they user could set/change the flag using raft::resource::set_user_resource_copy_mode()` or something? We can default it to the least overhead option.

achirkin: @cjnolet do you mean a runtime flag? I've been thinking about that. At first glance, it seems nice to have this flexibility. However, I think, this has a potential to do more harm than good.
From what I've seen so far, most of the "users" of the user_resource are within raft itself (e.g. caching in the algorithms). When we add these use cases, we keep in mind particular behavior of the resource (e.g. cached kernel parameters are shared across the copied handles or kept private). I can imagine this behavior affecting the performance of the algorithms. If an external user is able to change this behavior at runtime (because they want a particular behavior in their piece of code) it will have an adverse effect on overall performance of the other parts of raft codebase.
Therefore, I think it's reasonable to decide on this behavior once and for all and design the uses cases with the chosen behavior in mind.

cpp/include/raft/core/resource/user_resource.hpp
cjnolet: I don't think this has been added to the docs yet, has it? (along with the one above?)

achirkin: Seems to be present in docs/source/cpp_api/core_resources.rst changeset?

cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp
cjnolet: I'm okay with this third option, and have also been suggesting to folks that we rename the `cuh` extensions to `hpp` as they've been popping up in PR diffs (spgemm in @rhdong's PR as a recent example). 

Though, I'd even go as far as to say that we should deprecate `linalg/gemm.cuh` and only add the mdspan-based functions to `linalg.hpp`, since those are the "official" public API vocabulary going forward.

docs/source/cpp_api/core_resources.rst
```diff
@@ -202,3 +214,22 @@ namespace *raft::resource*
      :project: RAFT
      :members:
      :content-only:
+
+User-defined resources
```
cjnolet, line 218: We should call this something else. Just adding a note so it doesn't get lost with the other rename.

cpp/include/raft/core/resource/user_resource.hpp
cjnolet: > From what I've seen so far, most of the "users" of the user_resource are within raft itself (e.g. caching in the algorithms).

Your comment here makes me struggle with the name a little bit, because I was absolutely thinking we would want to invite users to use a "user resource". One of the reasons we didn't use an unordered map for the other resources is because lookup has shown to be slow, especially when the map is small. 

Rather than a "user resource", this seems to be more of a "runtime shared resource cache". I'd prefer to name it appropriately based on its use and I think that name is generalized enough to tell a potential user (internal or external) what it's for. Docs can help describe it even better for anyone who might have further confusion on the name.

cjnolet (changes requested):
Thanks for bearing with me here @achirkin. It looks like we've made a little progress on the discussions for this PR.

cpp/include/raft/core/resource/user_resource.hpp
cjnolet: Got it, the docs are pulling are pulling it in directly. Thanks.

cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp
achirkin: Sounds good. I added new `linalg/gemm.hpp` for now with the mdspan api only and deprecated the rest.

cpp/include/raft/linalg/detail/gemm.hpp
achirkin: https://github.com/rapidsai/raft/pull/1736#discussion_r1452113436

cpp/include/raft/core/resource/user_resource.hpp
achirkin: I think it's ok to invite users to use this resource, but you're right the name implies it's made specifically for users, which is wrong. How about "custom_resource"?

To me, the only difference of the custom_resource to the other resources is that raft can't allocate a slot for it at compile time. I've tried to reflect it in the updated name and docs. How does it look to you now?

Also regarding the resource copy policy: I suggest we stick to the status quo now (option 1) and revise it when we decide something on `raft::resources` and thread-safety issue.

cpp/include/raft/core/resource/custom_resource.hpp
```diff
@@ -0,0 +1,88 @@
...
+    auto pos = map_.find(key);
+    if (pos != map_.end()) { return reinterpret_cast<ResourceT*>(pos->second.get()); }
+    auto store_ptr = new ResourceT{};
+    map_[key] =
+      std::shared_ptr<void>(store_ptr, [](void* ptr) { delete reinterpret_cast<ResourceT*>(ptr); });
+    return store_ptr;
+  }
+
+ private:
+  std::unordered_map<std::type_index, std::shared_ptr<void>> map_{};
```
achirkin, line 46: @wphicks, as per our discussion today, I was going to replace this with a vector of tuples, but now I'm not sure. We've discussed a little earlier with @cjnolet that the `get_resource` is not likely to be a bottleneck in any of our use cases. Wouldn't it be a little premature optimization to implement this now? Iterating through a vector to find an element and all that - could be a little less readable vs map_.find

tfeher (approved):
Had a quick look, and overall the PR looks good to me!

I am curious whether there is any observable effect of this PR on the following two benchmarks:
- [GramMatrix](https://github.com/rapidsai/raft/blob/branch-24.02/cpp/bench/prims/distance/kernels.cu#L117) linear kernels
- [KMeansBalanced](https://github.com/rapidsai/raft/blob/1e4961e2354afba116e3479c5ec9041937b9922e/cpp/bench/prims/cluster/kmeans_balanced.cu#L97). Currently this test does not run GEMM, but if we [change the distance type](https://github.com/rapidsai/raft/blob/1e4961e2354afba116e3479c5ec9041937b9922e/cpp/bench/prims/cluster/kmeans_balanced.cu#L79) to `InnerProduct` then it would use [linalg::gemm calls](https://github.com/rapidsai/raft/blob/1e4961e2354afba116e3479c5ec9041937b9922e/cpp/include/raft/cluster/detail/kmeans_balanced.cuh#L148), and that gives us a way to benchmark the new GEMM interface.

achirkin:
Thanks, @tfeher  for suggestion. I ran the GramMatrix and KMeansBalanced/InnerProduct tests today; there's no significant difference, even if I set somewhat larger benchmark time the iteration time variance is simply larger than the difference. I guess, one need rather specific, small matrix sizes to not hide the kernel run latency and see the difference.

cpp/include/raft/core/resource/custom_resource.hpp
achirkin: Update: changed this to the vector as suggested. After some testing, it turns out it does make a bit of a difference!

achirkin:
Update: I've changed the resource and the cache implementation to not use unordered_map and rerun the relevant tests with small param adjustments. Now I see some measurable difference (this PR vs current state of 24.02):
benchmark | adjustments | iteration time difference
-- | -- | -- 
`KNN/*/*/ivf_pq_knn/0/0/1` | n_queries = 1 | -5.3±4.0%
`GramMatrix` | | +0.2±0.5%
`KMeansBalanced` |  distance: `InnerProduct` | -2.9±1.8%

cjnolet (approved):
LGTM

tfeher:
Thanks @achirkin for the updated numbers! LGTM!

PR #1743, opened 2023-08-16
Serialize only knn graph with CAGRA bench

tfeher:
This PR changes how the CAGRA ANN benchmarks saves the index. We save the graph only, because the dataset is usually available in a separate file.

### DIFF
--- a/cpp/bench/ann/src/raft/raft_cagra_wrapper.h
+++ b/cpp/bench/ann/src/raft/raft_cagra_wrapper.h
@@ -19,10 +19,13 @@
 #include <fstream>
 #include <iostream>
 #include <memory>
+#include <raft/core/detail/mdspan_numpy_serializer.hpp>
 #include <raft/core/device_mdspan.hpp>
 #include <raft/core/device_resources.hpp>
 #include <raft/core/logger.hpp>
 #include <raft/core/operators.hpp>
+#include <raft/core/resource/cuda_stream.hpp>
+#include <raft/core/serialize.hpp>
 #include <raft/distance/detail/distance.cuh>
 #include <raft/distance/distance_types.hpp>
 #include <raft/linalg/unary_op.cuh>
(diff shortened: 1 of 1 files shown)


cpp/bench/ann/src/raft/raft_cagra_wrapper.h
```diff
@@ -129,15 +135,81 @@ void RaftCagra<T, IdxT>::set_search_param(const AnnSearchParam& param)
 template <typename T, typename IdxT>
 void RaftCagra<T, IdxT>::save(const std::string& file) const
 {
-  raft::neighbors::cagra::serialize(handle_, file, *index_);
+  // 1 orig serialization: save both dataset and knn graph into the file
```
cjnolet, line 138: @benfred and I were literally just talking about this. We have also agreed in the past that we should be making serializing the dataset optional and I believe we also converged on providing a `update_dataset()` method right on the index to alleviate the awkwardness of having to accept a dataset at deserialization time.

tfeher: Alternatively, we can just construct the index from the two arrays like here
https://github.com/rapidsai/raft/pull/1743/files#diff-5f08a0aeb75c8884f5142d218e97ec859de8cc0cffd98d8c08f0e18a45655da0R211-R212

Saving the knn-graph mdspan is trivial with the current helpers, but we could add the following code block as a helper to make it easy to load an mdspan:
https://github.com/rapidsai/raft/pull/1743/files#diff-5f08a0aeb75c8884f5142d218e97ec859de8cc0cffd98d8c08f0e18a45655da0R175-R181

benfred: I have a PR up here https://github.com/rapidsai/raft/pull/1781 which doesn't write out the dataset on serialize (by using the `include_dataset` param from #1755) and `update_dataset` to set the dataset after deserialize.

tfeher:
Alternative solution provided in #1781. Closing this.

PR #1747, opened 2023-08-17
[Experiment] Compile libraft.so with separable compilation

ahendriksen:
With separable compilation, device code is linked together at the end of compilation. This could be advantageous for RAFT, as we have many duplicate kernel templates.

Previously, these duplicate kernel templates would be deduplicated on the host side during linking, but not on device side, causing sometimes up to 80 duplicate kernels to be left in the binary.

With this change, both host and device side weak symbols (e.g., kernel templates) should be deduplicated. Hopefully this has an impact on binary size.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -444,6 +444,7 @@ if(RAFT_COMPILE_LIBRARY)
                CUDA_STANDARD_REQUIRED ON
                POSITION_INDEPENDENT_CODE ON
                INTERFACE_POSITION_INDEPENDENT_CODE ON
+               CUDA_SEPARABLE_COMPILATION ON
   )
 
   target_link_libraries(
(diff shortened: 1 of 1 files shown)


ahendriksen:
While investigating #1722, I found out that kernels with template arguments are not deduplicated at link time on the device side. With separable compilation this issue could be fixed. Let's see what the impact is.. 

Still to investigate (when this experiment turns out to be worthwile):
- For some reason, the fatbin linker script does not work together nicely with CUDA_SEPARABLE_COMPILATION. 
- CMake versions pre-26.1 did not propagate -fvisibility correctly to the device side linker. Not sure if this has an impact on RAFT.


@tfeher

ahendriksen:
For some reason, the binary size increases, rather than decreases. 
Comparing the [PR build artifacts](https://downloads.rapids.ai/ci/raft/pull-request/1747/03da8fe/) with the [nightly artifacts](https://downloads.rapids.ai/nightly/raft/2023-08-18/7aded12/), there is an increase in size for most Conda packages from ~700MB to ~900MB.

We might want to investigate this in the future. Shall I close this branch for now @tfeher ?

tfeher:
Thanks Allard for testing this. We can close it for now.

Issue #1729, opened 2023-08-09
[FEA] Improve CAGRA serialization


lowener:
CAGRA should expose its data type at the beginning of the serialization file for easier deserialization on Python side, just like IVF Flat.
#1717 is introducing a temporary fix on the Python side due to the dtype not being at a fixed offset in the serialization file.

PR #1755, opened 2023-08-18
Update CAGRA serialization

benfred:
This changes the serialization format of saved CAGRA indices by:

* The dtype will now be written in the first 4 bytes of the serialized file, to match the IVF methods and to make it easier to deduce the dtype from python (#1729)
* Writing out the dataset with the index is now optional. Since many use cases will already have the dataset written out separately, this gives us the option to save disk space by not writing out an extra copy of the input dataset. If the include_dataset=false option is given, you will have to call `index.update_dataset` to set the dataset yourself after loading

### DIFF
--- a/cpp/include/raft/neighbors/cagra_types.hpp
+++ b/cpp/include/raft/neighbors/cagra_types.hpp
@@ -258,13 +258,6 @@ struct index : ann::index {
         dataset.data_handle(), dataset.extent(0), dataset.extent(1), dataset.extent(1));
     }
   }
-  void update_dataset(raft::resources const& res,
-                      raft::device_matrix_view<T, int64_t, row_major> dataset)
-  {
-    update_dataset(res,
-                   make_device_matrix_view<const T, int64_t>(
-                     dataset.data_handle(), dataset.extent(0), dataset.extent(1)));
-  }
 
   /**
    * Replace the dataset with a new dataset.
@@ -276,14 +269,6 @@ struct index : ann::index {
   {
     RAFT_LOG_DEBUG("Copying CAGRA dataset from host to device");
     copy_padded(res, dataset);
-  }
-
-  void update_dataset(raft::resources const& res,
-                      raft::host_matrix_view<T, int64_t, row_major> dataset)
-  {
-    update_dataset(res,
-                   make_host_matrix_view<const T, int64_t>(
-                     dataset.data_handle(), dataset.extent(0), dataset.extent(1)));
   }
 
   /**
--- a/cpp/include/raft/neighbors/cagra_types.hpp
+++ b/cpp/include/raft/neighbors/cagra_types.hpp
@@ -258,6 +258,13 @@ struct index : ann::index {
         dataset.data_handle(), dataset.extent(0), dataset.extent(1), dataset.extent(1));
     }
   }
+  void update_dataset(raft::resources const& res,
+                      raft::device_matrix_view<T, int64_t, row_major> dataset)
+  {
+    update_dataset(res,
+                   make_device_matrix_view<const T, int64_t>(
+                     dataset.data_handle(), dataset.extent(0), dataset.extent(1)));
+  }
 
   /**
    * Replace the dataset with a new dataset.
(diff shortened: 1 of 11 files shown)


cpp/include/raft/neighbors/cagra_types.hpp
```diff
@@ -258,6 +258,13 @@ struct index : ann::index {
         dataset.data_handle(), dataset.extent(0), dataset.extent(1), dataset.extent(1));
     }
   }
+  void update_dataset(raft::resources const& res,
```
cjnolet, line 261: We probably want to keep these const mdspans. If this is because of python, can we use make_const_mdspan() in that layer?

python/pylibraft/pylibraft/neighbors/cagra/cagra.pyx
```diff
@@ -162,6 +162,31 @@ cdef class IndexFloat(Index):
...
+        cdef device_resources* handle_ = \
+            <device_resources*><size_t>handle.getHandle()
+
+        dataset_ai = wrap_array(dataset)
+        dataset_dt = dataset_ai.dtype
+        _check_input_array(dataset_ai, [np.dtype("float32")])
+
+        if dataset_ai.from_cai:
+            self.index[0].update_dataset(deref(handle_),
+                                         get_dmv_float(dataset_ai,
```
cjnolet, line 183: Here is where we could use make_const_mdspan. It would simplify things so that we don't need to make non-const functions everywhere (which kind of circumvents the const functions).

python/pylibraft/pylibraft/neighbors/cagra/cagra.pyx
```diff
@@ -195,6 +220,31 @@ cdef class IndexInt8(Index):
...
+        cdef device_resources* handle_ = \
+            <device_resources*><size_t>handle.getHandle()
+
+        dataset_ai = wrap_array(dataset)
+        dataset_dt = dataset_ai.dtype
+        _check_input_array(dataset_ai, [np.dtype("byte")])
+
+        if dataset_ai.from_cai:
+            self.index[0].update_dataset(deref(handle_),
+                                         get_dmv_int8(dataset_ai,
```
cjnolet, line 241: make_const_mdspan here too.

python/pylibraft/pylibraft/neighbors/cagra/cagra.pyx
```diff
@@ -235,6 +285,31 @@ cdef class IndexUint8(Index):
...
+        """
+        cdef device_resources* handle_ = \
+            <device_resources*><size_t>handle.getHandle()
+
+        dataset_ai = wrap_array(dataset)
+        dataset_dt = dataset_ai.dtype
+        _check_input_array(dataset_ai, [np.dtype("ubyte")])
+
+        if dataset_ai.from_cai:
+            self.index[0].update_dataset(deref(handle_),
```
cjnolet, line 305: make_const_mdspan

cjnolet (changes requested):
Changes look great overall. A few minor things.

cpp/include/raft/neighbors/cagra_types.hpp
benfred: I agree that automatically discarding `const` would be bad - but this is doing the opposite and is automatically adding it (like this is converting a non-const mdspan to a const msdpan), which I feel like is something that should be allowed with our API's. 

The issue I have is that Cython kinda sucks with respecting `const` identifiers, which is why all our Cython api's use non-const mdspans right now. Like if I try to add a `get_const_hmv_float` (to parallel the non-const `get_hmv_float` we have now) - I get an error message from Cython, where it doesn't recognize `const float` as a type inside template parameters:

```
      Error compiling Cython file:
      ------------------------------------------------------------
      ...
          if cai.dtype != np.float32:
              raise TypeError("dtype %s not supported" % cai.dtype)
          if check_shape and len(cai.shape) != 2:
              raise ValueError("Expected a 2D array, got %d D" % len(cai.shape))
          shape = (cai.shape[0], cai.shape[1] if len(cai.shape) == 2 else 1)
          return make_host_matrix_view[const float, int64_t, row_major](
                                             ^
      ------------------------------------------------------------
      
      /home/ben/code/raft/python/pylibraft/pylibraft/common/mdspan.pyx:232:39: Expected ']', found 'float'
```

I can get around this by adding a Cython typedef (like `ctypedef const float const_float`) - but that introduces the need for other hacks later on (like cython will treat `const_float` and `const float` as separate types - meaning that when we define the `update_dataset` for Cython in c_cagra.pxd I can't just go `const T` as the type, and have to introduce a new template param =(.  I've done this in the last commit - let me know what you think

cjnolet (approved):
LGTM!

PR #1758, opened 2023-08-21
Add ivf-flat notebook

tfeher:
Example jupyter notebook to demonstrate usage of the IVF-Flat API


achirkin (approved):
I think it looks good. Though as a nitpick: we're using three different ways to print the output across the notebook (varargs, f-strings, .format); I'd suggest we stick to one :) e.g. consistently use the f-strings - to make it look a bit more polished.

notebooks/ivf_flat_example.ipynb
```diff
@@ -0,0 +1,803 @@
+{
```
cjnolet, line 1: <p>Can you add a little more detail here about what the notebook accomplishes? Doesn't need to be extensive, but at least somehting like "It builds and search an index using a dataset from the <a href="https://github.com/erikbern/ann-benchmarks/#data-sets" target="_blank">ann-benchmarks million-scale datasets</a>, saves/loads the index to disk, and explores some of the important parameters for fine-tuning the performance and accuracy of the index."</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1758/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

cjnolet, line 1: <pre class="ql-syntax" spellcheck="false">Line #1.    def calc_recall(ann_idx, true_nn_idx):
</pre><p>We should consider pulling these out into a python file (maybe <code>common.py</code>?) in the <code>notebooks/</code> directory so we can share the code with other notebooks. It'll also clean up the notebook a little bit since it's not a particularly relevant detail for IVF-flat demonstration. It would also be worth quickly updating the ivf-pq notebook to use at least the <code>calc_recall</code> function from this notebook.  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1758/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

cjnolet, line 1: <pre class="ql-syntax" spellcheck="false">Line #1.    #DATASET_URL = "http://ann-benchmarks.com/glove-100-angular.hdf5"
</pre><p>Since we're duplicating this also in the ivf-pq notebook, can we also create a <code>load_dataset()</code> in the <code>common.py</code>?Please also note that our benchmarks python package also contains a python script for both downloading and loading the million-scale datasets from ann-benchmarks and the code might be useful to place in a more general location (such as <code>pylibraft</code>,perhaps?).   </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1758/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

cjnolet, line 1: <pre class="ql-syntax" spellcheck="false">Line #7.            add_data_on_build=True
</pre><p>For users who might not be coming to this notebook from the blog, it would be helpful to provide a link to the pylibraft docs in a text description above the code and maybe mention that some of these are described in more detail later in the notebook. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1758/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

cjnolet, line 1: <p>Can we put a quick link to the corresponding portion of the pylibraft docs here? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/rapidsai/raft/pull/1758/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

tfeher: <p>added</p>

tfeher: <p>moved to utils.py</p>

tfeher: <p>Added <code>load_dataset()</code> to the utils. Yes, I have looked into the benchmarks python package as well. The actual download code that we can use from there  is only a <a href="https://github.com/rapidsai/raft/blob/6bbcf1fef7b60de2f85f055fb1acba7d27515b67/python/raft-ann-bench/src/raft-ann-bench/get_dataset/__main__.py#L28" target="_blank">few lines</a>. I prefer to keep a separate copy in the notebooks folder.</p>

tfeher: <p>Good point, added.</p>

tfeher: <p>added</p>

tfeher:
Thanks @cjnolet for the review, I have addressed the issues.

cjnolet (approved):
LGTM!

PR #1759, opened 2023-08-21
Using expanded distance computations in `pylibraft`

cjnolet:
We are noticing a perf hit of nearly 2x with the unexpanded distance computations vs the expanded sitances.


benfred (approved):

python/pylibraft/pylibraft/test/test_distance.py
```diff
@@ -81,4 +81,4 @@ def test_distance(n_rows, n_cols, inplace, metric, order, dtype):
 
     actual[actual <= 1e-5] = 0.0
 
-    assert np.allclose(expected, actual, rtol=1e-4)
+    assert np.allclose(expected, actual, rtol=1e-3)
```
benfred, line 84: might need to also specify atol here:
```suggestion
    assert np.allclose(expected, actual, atol=1e-3, rtol=1e-3)
```

benfred (approved):

PR #1761, opened 2023-08-22
Adding config files for remaining (relevant) ann-benchmarks million-scale datasets

cjnolet:
Also exposing new `dataset_memtype` option for the RAFT index types that allows us to specify (in the config files) for each index whether we want the dataset to be fed in from "host", "device" or "mmap" (host memory-mapped).

### DIFF
--- a/docs/source/raft_ann_benchmarks.md
+++ b/docs/source/raft_ann_benchmarks.md
@@ -50,6 +50,16 @@ python scripts/ann-benchmarks/data_export.py --output out.csv --dataset glove-10
 python scripts/ann-benchmarks/plot.py --result-csv out.csv
 ```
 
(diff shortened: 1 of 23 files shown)


docs/source/raft_ann_benchmarks.md
```diff
@@ -120,7 +130,7 @@ raft_ivf_pq:
   executable: RAFT_IVF_PQ_ANN_BENCH
   disabled: false
 ```
-`executable` : specifies the binary that will build/search the index. It is assumed to be
+`executable` : specifies the location of the binary that will build/search the index. It is assumed to be
```
divyegala, line 133: It's not the location, it's just the name of the binary/executable that's specified here. By default it's only search for in `${CONDA_PREFIX}/bin/ann` or `cpp/build`

divyegala (approved):
Approved, one comment regarding docs clarity

divyegala (commented):
There's a section in the documentation that describes how to form the configuration for an index. Could you add the new memory type key there?

cjnolet:
Discussed offline- we're going to improve the docs for the benchmarks further by specifying each of the possible indexing, build, and search arguments in a table and further explaining each of them.

PR #1765, opened 2023-08-23
[FEA] Add sample filtering for CAGRA

### LINKED ISSUE #1598: CAGRA Pre-filtering

lowener:
Closes #1598.
This PR introduces `cagra::search_with_filtering` to perform sample filtering in the same way as IVF methods.


cpp/test/neighbors/ann_cagra.cuh
```diff
@@ -15,6 +15,8 @@
  */
 #pragma once
 
+#undef RAFT_EXPLICIT_INSTANTIATE_ONLY  // Search with filter instantiation
```
lowener, line 18: This leads to 3x build time for the CI (30 mins -> 90 mins)

lowener: The CI build time is actually still around the 30 minutes with the instantiation of the search kernels with a test pre-filter.

enp1s0:
Thank you, @lowener, for implementing the filter. However, I found a problem in this implementation. Specifically, nodes for which the filter function returns false do not serve as parent nodes during graph traversal. This leads to recall and throughput degradation. Since the required modification is somewhat complicated, can I take over implementing this feature using your implementation as a reference?

In the CAGRA search, we have a node buffer consisting of internal top-M and candidate list, as shown in Fig. 6 in the [CAGRA paper](https://arxiv.org/abs/2308.15136). Within each search iteration, we select a parent node from the internal top-M list and compute the distance between the parent node's neighbors and the query. The indices and distances of the neighbor nodes are stored in the candidate list (as depicted in steps 2 and 3 of the figure). In the subsequent iteration, we update the internal top-M list and repeat the aforementioned operations (steps 2 and 3) until all nodes in the top-M buffer experience parents or the iteration count reaches the defined `max_iterations` threshold.

Within the context of the filtering implementation, nodes for which the filter function returns false must not be chosen as final results. However, if necessary, these nodes should still be parent nodes during search iterations. Unfortunately, the current PR implementation fails to accomplish the latter requirement. Specifically, the following part of the code is the cause.
```cpp
      if (!sample_filter(query_id, child_id) ||
          (hashmap::insert(visited_hashmap_ptr, hash_bitlen, child_id) == 0)) {
        child_id = invalid_index;
      }
```
https://github.com/lowener/raft/blob/93deac2489882e95eaccaf7e24585bc55904ad7d/cpp/include/raft/neighbors/detail/cagra/compute_distance.hpp#L178-L181

`child_id = invalid_index` leads to excluding the node from the parent candidates in the subsequent iteration.

I have discussed this matter with @anaruse, but we have yet to determine the optimal solution.
One possible approach is to remove the nodes for which the filter function returns false after serving as parent nodes. Specifically,   following the execution of the `compute_distance_to_child_nodes` function, we set `(id, distance) = (invalid, utils::get_max_value<DISTANCE_T>())` for the nodes in the internal top-M buffer if the node index is in the parent node list and `filter(query_id, index) == false`. And in the final step of the search, we filter the nodes in the internal top-M list again before picking up the final top-k results.  The other part of the computation is the same as the original CAGRA. We can attain filtered results by implementing these modifications, provided that the internal top-M length or the `max_iterations` value is sufficiently large. However, this approach does not guarantee a complete set of k results if they are not large enough. To guarantee k results, a new converge criterion is required, but the most effective criterion remains uncertain, and we need more discussion and prototype implementations. And also, we have to impose additional restrictions, such as the forgettable hash table must not be used with filtering, which can lead to throughput degradation.

@anaruse If you have any additional comments, please let us know.

cjnolet:
> Since the required modification is somewhat complicated, can I take over implementing this feature using your implementation as a reference?

Thank you @enp1s0, for volunteering to take over this implementation. One major detail here is that we need to have something implemented for this for the 23.10 release, which is a little over a month away. It doesn't have to be the optimal solution initially, and it can be improved in 23.12, but by the very least we are on the hook to have a pre-filtering and deletion feature implemented for 23.10. Do you think this is possible given the state of the current implementation?

PR #1767, opened 2023-08-23
[WIP] Introducing host API for PCG

vinaydes:
[WIP]


MatthiasKohl (approved):
Thank you very much @vinaydes , LGTM

MatthiasKohl (approved):
reviewed previously, recent changes LGTM

achirkin:
\ok to test

achirkin:
It just occurred to me that the PR breaking, because you move the header providing functions in public namespace `raft::util` (`raft/util/fast_int_div.cuh ->raft/util/integer_utils.cuh`). I think we have two options here:
 
   - a) Add the note to the description what's broke, check whether our downstream projects are affected (at least cuml, cugraph).
   - b) Don't touch `fast_int_div.cuh`, just add your function alone to the new file `integer_utils.cuh` to keep it non-breaking. I see this function cannot go into `.hpp` due to CUDA asm, but maybe it makes sense to add the include of the `integer_utils.hpp` into `integer_utils.cuh`?..

vinaydes:
I think I am inclined to keep this a non-breaking PR. I'll separate out the two functions.

vinaydes:
@MatthiasKohl For now, going with option b) will enable us to call the wide multiply from .cu files for using in host code.

achirkin:
@MatthiasKohl , you're probably right. It didn't occur to me that `__CUDA_ARCH__` would still be available in `*.hpp` included by a `*.cu` file :)

achirkin:
I wonder though. If we start to guard the whole kernels with pre-processor macros, we can end up with replacing all headers in raft with `.hpp` versions. Do we want that? Where's the boundary between what we want to have in `.cuh` and `.hpp`?..

vinaydes:
>I guess, then it's natural to move it to the existing raft/util/integer_utils.hpp.

I was going to say we create a `integer_utils.cuh` but then I took a quick glance at the other .hpp files in the project. Several of them have cuda code. I guess I'll put the function in the `integer_utils.hpp`.

vinaydes:
@achirkin @MatthiasKohl I have tried to accommodate the changes. It should be non-breaking now 🤞.
Also Artem do you mind doing /ok to test one more time?

cpp/include/raft/util/integer_utils.hpp
```diff
@@ -24,6 +24,7 @@
  */
 
 #include <limits>
+#include <raft/core/resources.hpp>
```
achirkin, line 27: What do you need the `core/resources` for? Maybe include `raft/core/detail/macros.hpp` directly?

vinaydes: Fair enough, I will change it. I was not sure about including a file from `detail` directory.

vinaydes: Done.

vinaydes:
@achirkin Can we mark this as non-breaking again?

vinaydes:
Can this be merged now?

achirkin (approved):

cjnolet (approved):

PR #1773, opened 2023-08-25
Add GPU and CPU packages for ANN benchmarks

dantegd:
Builds on top of #1769 

- [x] Removes `libraft-ann-bench` C++ based package
- [x] Creates `raft-ann-bench` packages that includes C++ tests as well as Python scripts
    - [x] `raft-ann-bench` package includes all tests for CPU and GPU
    - [x] `raft-ann-bench-cpu` package that does not depend on CUDA or RAFT GPU code
- [ ] Update docs
- [ ] Test artifacts and scripts in CI  
- [ ] Minor code cleaning

Some changes include:

- Use `RAPIDS_DATASET_ROOT_DIR` env variable to indicate location of datasets (optional) consistent with other repos: https://docs.rapids.ai/maintainers/datasets/
- CPU and GPU packages are built in the existing GPU build GHA. Only the CUDA 12 jobs build the CPU packages. 
- Small change for invocation of scripts, for example: `python bench/ann/run.py --dataset deep-image-96-inner` is now `python -m raft-ann-bench.run --dataset deep-image-96-inner`, but still scripts meant to be invoked from the command line. 

Future improvements: 

- Remove use of popen python scripts from python scripts. 
- Improve printing and logging
- Allow functions of package to be called from python scripts.


ajschmidt8:
This PR has been marked untrusted due to the changes outlined in https://github.com/rapidsai/raft/pull/1774.

It looks like the PR has unsigned commits.

You can rebase your branch to sign the unsigned commits to avoid having to use an `/ok to test` comment for every subsequent push.

The blog post below outlines how you can do this:

- https://duan.ca/2016/12/24/rebase-multiple-git-commits-with-signing/

cpp/bench/ann/src/common/benchmark.hpp
```diff
@@ -506,7 +506,9 @@ inline auto run_main(int argc, char** argv) -> int
     return -1;
   }
 
+#ifndef CPU_ONLY
   if (!CUDART_FOUND) { log_warn("cudart library is not found, GPU-based indices won't work."); }
+#endif
```
achirkin, line 511: I suggest updating the `CUDART_FOUND` definition at the top of the file instead of nesting the pragmas here.

cpp/bench/ann/src/common/ann_types.hpp
```diff
@@ -24,6 +22,9 @@
 
 #ifndef CPU_ONLY
 #include <cuda_runtime_api.h>  // cudaStream_t
+#else
+typedef size_t cudaStream_t;
+typedef size_t cudaEvent_t;
 #endif
```
achirkin, line 28: I've done very similar thing in the `common/cuda_stub.hpp` but only for the case "cpu only" vs "dlopen". Maybe we could try to extend that logic to  "cpu only" vs "dlopen" vs "cudart linked" there and include it here?

Also, in the cuda headers `cudaStream_t` and `cudaEvent_t` are defined as pointers to cuXxx handles. I've typedefed them with `void*`, which I believe is more compatible than `size_t`.  Sure they coincide for x86, but theoretically these could be different types.

cpp/bench/ann/src/common/util.hpp
```diff
@@ -90,9 +90,15 @@ struct buf {
...
  private:
+#ifndef CPU_ONLY
   cudaStream_t stream_{nullptr};
   cudaEvent_t start_{nullptr};
   cudaEvent_t stop_{nullptr};
+#else
+  cudaStream_t stream_{0};
+  cudaEvent_t start_{0};
+  cudaEvent_t stop_{0};
+#endif
```
achirkin, line 101: As I noted above, I think it's better to typedef these two as `void*`, so there's no need for extra pragmas here. Alternatively, you can just change the curly braces to the parentheses to avoid the type conversion error popping up.

cpp/bench/ann/CMakeLists.txt
```diff
@@ -95,10 +128,10 @@ function(ConfigureAnnBench)
     ${BENCH_NAME}
     PRIVATE raft::raft
             nlohmann_json::nlohmann_json
-            $<$<BOOL:${RAFT_ANN_BENCH_USE_MULTIGPU}>:NCCL::NCCL>
+            $<$<BOOL:${GPU_BUILD}>:$<$<BOOL:${RAFT_ANN_BENCH_USE_MULTIGPU}>:NCCL::NCCL>>
```
achirkin, line 131: Question: does anything in ann-bench depend on NCCL at the moment? Maybe we can just remove this line altogether?

cpp/bench/ann/src/hnswlib/hnswlib_benchmark.cpp
```diff
@@ -116,7 +116,9 @@ REGISTER_ALGO_INSTANCE(std::int8_t);
 REGISTER_ALGO_INSTANCE(std::uint8_t);
 
 #ifdef ANN_BENCH_BUILD_MAIN
+#ifndef CPU_ONLY
 #define CPU_ONLY
+#endif
```
achirkin, line 121: What do you think about moving `CPU_ONLY` to the `ConfigureAnnBench`  function in CMakeLists.txt as an option?

cpp/bench/ann/CMakeLists.txt
```diff
@@ -27,14 +27,27 @@ option(RAFT_ANN_BENCH_SINGLE_EXE
...
+  set(RAFT_ANN_BENCH_USE_GGNN OFF)
+else()
+  # Disable faiss benchmarks on CUDA 12 since faiss is not yet CUDA 12-enabled.
+  # https://github.com/rapidsai/raft/issues/1627
+  if(${CMAKE_CUDA_COMPILER_VERSION} VERSION_GREATER_EQUAL 12.0.0)
+    set(RAFT_ANN_BENCH_USE_FAISS_BFKNN OFF)
+    set(RAFT_ANN_BENCH_USE_FAISS_IVF_FLAT OFF)
+    set(RAFT_ANN_BENCH_USE_FAISS_IVF_PQ OFF)
+  endif()
 endif()
```
achirkin, line 51: Extending on my suggestion to add `CPU_ONLY`-like option to `ConfigureAnnBench` function:
besides defining the same-named preprocessor definition, we could use this option to disable the target within the function `ConfigureAnnBench`. This way we can avoid verbose repetition of disabling/enabling the targets here.

achirkin (changes requested):
Can't wait to get the cpu-only packages!
A few comments on the content of `cpp/bench/ann` folder.

cpp/bench/ann/src/common/ann_types.hpp
achirkin: Come to think about it, the only thing you need to do in cuda_stub.hpp is hide `#include <dlfcn.h>` and `dlclose(handle)` behind the `ANN_BENCH_LINK_CUDART` and it's done! Well.. also replace the `#define CPU_ONLY` there with #ifdef CPU_ONLY` and pass CPU_ONLY to the `ANN_BENCH` target.

`ANN_BENCH_LINK_CUDART` is defined only for the `ANN_BENCH` target and gives the CUDA version to link against.

achirkin: update: I've made this small PR https://github.com/rapidsai/raft/pull/1792 with the suggested changes. In theory, you shouldn't need to change any C++ code on top of that.

cpp/bench/ann/CMakeLists.txt
```diff
@@ -53,12 +66,25 @@ if(RAFT_ANN_BENCH_USE_RAFT_IVF_PQ
   set(RAFT_ANN_BENCH_USE_RAFT ON)
 endif()
 
+option(RAFT_ANN_BENCH_USE_MULTIGPU "Use multi-gpus (where possible) in benchmarks" OFF)
```
cjnolet, line 69: We don't support this yet so this can be removed.

cpp/bench/ann/CMakeLists.txt
```diff
@@ -53,12 +66,25 @@ if(RAFT_ANN_BENCH_USE_RAFT_IVF_PQ
...
+
+message(VERBOSE "RAFT: Build ann-bench with FAISS_BFKNN: ${RAFT_ANN_BENCH_USE_FAISS_BFKNN}")
+message(VERBOSE "RAFT: Build ann-bench with FAISS_IVF_FLAT: ${RAFT_ANN_BENCH_USE_FAISS_IVF_FLAT}")
+message(VERBOSE "RAFT: Build ann-bench with FAISS_IVF_PQ: ${RAFT_ANN_BENCH_USE_FAISS_IVF_PQ}")
+message(VERBOSE "RAFT: Build ann-bench with RAFT_IVF_FLAT: ${RAFT_ANN_BENCH_USE_RAFT_IVF_FLAT}")
+message(VERBOSE "RAFT: Build ann-bench with RAFT_IVF_PQ: ${RAFT_ANN_BENCH_USE_RAFT_IVF_PQ}")
+message(VERBOSE "RAFT: Build ann-bench with RAFT_CAGRA: ${RAFT_ANN_BENCH_USE_RAFT_CAGRA}")
+message(VERBOSE "RAFT: Build ann-bench with HNSWLIB: ${RAFT_ANN_BENCH_USE_HNSWLIB}")
+message(VERBOSE "RAFT: Build ann-bench with GGNN: ${RAFT_ANN_BENCH_USE_GGNN}")
+message(VERBOSE "RAFT: Build ann-bench with MULTIGPU: ${RAFT_ANN_BENCH_USE_MULTIGPU}")
```
cjnolet, line 79: This can be removed

python/raft-ann-bench/src/raft-ann-bench/run/algos-cpu.yaml
```diff
@@ -0,0 +1,30 @@
+faiss_gpu_ivf_flat:
```
cjnolet, line 1: I wonder if instead of two files maybe we could add an option here to specify whether the executable runs on the GPU? I'm just thinking it could be easier to maintain.

cjnolet (changes requested):
Just a couple fairly minor things so far. Much easier to review now that we've merged a bunch of the upstream C++ changes!

cjnolet (approved):

build.sh
```diff
@@ -152,7 +154,7 @@ function limitTests {
             # Remove the full LIMIT_TEST_TARGETS argument from list of args so that it passes validArgs function
             ARGS=${ARGS//--limit-tests=$LIMIT_TEST_TARGETS/}
             TEST_TARGETS=${LIMIT_TEST_TARGETS}
-	    echo "Limiting tests to $TEST_TARGETS"
+        echo "Limiting tests to $TEST_TARGETS"
```
divyegala, line 157: I think the original indentation was correct

ci/build_python.sh
```diff
@@ -26,4 +26,21 @@ rapids-mamba-retry mambabuild \
...
+# Build ann-bench for each cuda and python version
+rapids-mamba-retry mambabuild \
+--no-test \
+--channel "${CPP_CHANNEL}" \
+--channel "${RAPIDS_CONDA_BLD_OUTPUT_DIR}" \
+conda/recipes/raft-ann-bench
+
+# Build ann-bench-cpu only in CUDA 12 jobs since it only depends on python
+# version
+if [[ ${CUDA_VERSION} == "11.8.0" ]]; then
```
divyegala, line 38: The comment does not match the conditional. Also, I think the variable name that our CI containers use is `RAPIDS_CUDA_VERSION`. Can you verify if `CUDA_VERSION` is set?

cpp/CMakeLists.txt
```diff
@@ -116,15 +133,28 @@ if(DETECT_CONDA_ENV)
...
+  # * determine GPU architectures
+  # * enable the CMake CUDA language
+  # * set other CUDA compilation flags
+  rapids_find_package(
+    CUDAToolkit REQUIRED
+    BUILD_EXPORT_SET raft-exports
+    INSTALL_EXPORT_SET raft-exports
+  )
+else()
+  add_definitions(-DBUILD_CPU_ONLY)
```
divyegala, line 156: What does this line do? Does it add a `BUILD_CPU_ONLY` macro to every target?

dantegd: yes it does which I've used for quite some time, though would be good to change it to use https://cmake.org/cmake/help/latest/command/add_compile_definitions.html#command:add_compile_definitions which has superseded it

ci/build_python.sh
dantegd: yes, CUDA_VERSION is set, you can see it in the logs https://github.com/rapidsai/raft/actions/runs/6051700515/job/16424916331#step:7:54 which have both with and without `RAPIDS_` prefix

bdice: Better yet, copy this snippet (but use CUDA 12 to match the conditional as @divyegala noted above):

```suggestion
# Build ann-bench-cpu only in CUDA 12 jobs since it only depends on python
# version
RAPIDS_CUDA_MAJOR="${RAPIDS_CUDA_VERSION%%.*}"
if [[ ${RAPIDS_CUDA_MAJOR} == "12" ]]; then
```

https://github.com/rapidsai/cugraph/blob/2b4118aee4af912d74ce1ebe7adc39cf596899ef/ci/build_python.sh#L49-L51

cpp/CMakeLists.txt
divyegala: This is the way I know how to do it, if it has superseded the other method then let's use it!

ci/build_python.sh
dantegd: we require to build in the CUDA 11 jobs for now, but otherwise will use it!

divyegala (approved):

cpp/bench/ann/src/common/util.hpp
```diff
@@ -91,9 +91,15 @@ struct buf {
...
  private:
+#ifndef BUILD_CPU_ONLY
   cudaStream_t stream_{nullptr};
   cudaEvent_t start_{nullptr};
   cudaEvent_t stop_{nullptr};
+#else
+  cudaStream_t stream_{0};
+  cudaEvent_t start_{0};
+  cudaEvent_t stop_{0};
+#endif
```
achirkin, line 102:
```suggestion
  cudaStream_t stream_{nullptr};
  cudaEvent_t start_{nullptr};
  cudaEvent_t stop_{nullptr};
```
Not necessary anymore (merge artifact).

ajschmidt8 (approved):

achirkin (approved):
LGTM, long awaited feature!

Issue #1745, opened 2023-08-16
[DOC] Tutorial for end-to-end vector search in C++


cjnolet:
Need a tutorial for the docs that can demonstrate end-to-end examples of using our indexes to perform vector search. Thinking maybe we could borrow some of the data loading tools from our `ann-bench` suite to demonstrate doing this on real data.

PR #1776, opened 2023-08-26
End-to-end vector search tutorial in docs

cjnolet:
Closes https://github.com/rapidsai/raft/issues/1745


cjnolet:
@wphicks it would be great if you have cycles to add some verbiage and examples around `raft::device_resource_manager` to this tutorial as well. Please also feel free to push directly to my branch. The goal of this tutorial is to cover the basics end to end so someone who's not a CUDA developer can understand the core principals enough to know how to invoke our APIs. Further, they should be able to integrate our vector search APIs end-to-end.

wphicks:
Just pushed some text on `device_resources_manager`. Feel free to edit directly as desired.

benfred (approved):
I really like this tutorial - I think its very clear and well written, and will be a great introduction on how to use raft for ANN.

Some minor comments:

* There is a great introduction on using RAFT and device_resources in the `cuda development` section - but we don't tie it into the example code using the BF and CAGRA example code blocks later. I think we should add a `res.sync_stream()` to the BF/CAGRA example code blocks , and maybe also a snippet that copies the results back to host memory
* The section  on using a `stream_pool` doesn't need to be here (though this is super useful, and we should keep for other parts of the documentation somewhere). Consumers of our vector search indices don't typically need a stream_pool, so this information might be best put elsewhere

PR #1781, opened 2023-08-27
Don't serialize dataset with CAGRA bench

benfred:
As an alternative to #1743, this uses the `include_dataset=False` param in cagra::serialize to avoid writing the dataset to disk with the index. This lets us avoid writing a second copy of the dataset, since it is available in a separate file already


cjnolet (approved):
LGTM. I think this is a clean and maintainable solution that continues integrating the bench-Ann code with the rest of RAFT. Definitely prefer this approach.

PR #1784, opened 2023-08-29
[BUG] Fix search parameter check in CAGRA

enp1s0:
An error occurs when using CAGRA multi-CTA implementation with topk>32. This PR fixes the bug.

### DIFF
--- a/cpp/include/raft/neighbors/detail/cagra/search_multi_cta.cuh
+++ b/cpp/include/raft/neighbors/detail/cagra/search_multi_cta.cuh
@@ -175,6 +175,16 @@ struct search : public search_plan_impl<DATA_T, INDEX_T, DISTANCE_T> {
     topk_workspace.resize(topk_workspace_size, resource::get_cuda_stream(res));
   }
 
+  virtual void check(const uint32_t topk)
+  {
+    RAFT_EXPECTS(num_cta_per_query * 32 >= topk,
+                 "`num_cta_per_query` (%u) * 32 must be equal to or greater than "
+                 "`topk` (%u) when 'search_mode' is \"multi-cta\". "
+                 "(`num_cta_per_query`=max(`search_width`, `itopk_size`/32))",
+                 num_cta_per_query,
+                 topk);
+  }
+
   ~search() {}
 
   void operator()(raft::resources const& res,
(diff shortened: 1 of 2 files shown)


divyegala (approved):

cpp/include/raft/neighbors/detail/cagra/search_plan.cuh
```diff
@@ -250,17 +250,10 @@ struct search_plan_impl : public search_plan_impl_base {
     }
   }
 
-  void check(uint32_t topk)
+  virtual void check(const uint32_t topk)
   {
+    // For multi-CTA and multi kernel
```
lowener, line 255: Isn't it for single-CTA and multi-kernel since it is being overriden in multi-CTA?

cpp/include/raft/neighbors/detail/cagra/search_multi_cta.cuh
```diff
@@ -175,6 +175,16 @@ struct search : public search_plan_impl<DATA_T, INDEX_T, DISTANCE_T> {
     topk_workspace.resize(topk_workspace_size, resource::get_cuda_stream(res));
   }
 
+  virtual void check(const uint32_t topk)
```
lowener, line 178: The check for `topk <= itopk_size` is not needed anymore for multi-CTA?

lowener (changes requested):

cpp/include/raft/neighbors/detail/cagra/search_multi_cta.cuh
lowener:
```suggestion
  void check(const uint32_t topk) override
```
`virtual` is not needed here, but `override` can be used to ensure that the `check` function is virtual.

enp1s0:
Thanks @lowener for reviewing the code. I have modified the comment and `override`.

cpp/include/raft/neighbors/detail/cagra/search_multi_cta.cuh
enp1s0: > The check for topk <= itopk_size is not needed anymore for multi-CTA?

Not needed. In the case of multi-CTA, `itopk_size` is overwritten to 32 and `(actual) itopk_size / 32` CTAs are launched to search approximate neighbors of the same query.
Therefore, https://github.com/enp1s0/raft/blob/496689eb042adf80ab3bbbba8cc546ef573d5f11/cpp/include/raft/neighbors/detail/cagra/search_multi_cta.cuh#L180-L185 is enough.

lowener (approved):

PR #1790, opened 2023-08-30
Cagra memory optimizations

benfred:
When trying to build a CAGRA index with 500M embeddings, we were running out of memory - even when using managed memory. 

This PR contains some changes to reduce the memory usage:
 * For certain large matrices, don't make 2nd copies on the device or host if the memory is already accessible via UVM /ATS/HMM. For instance, we were taking a copy of the intermediate graph from host to device memory - and in certain cases (500M dataset, intermediate_graph_degree=128) the intermediate graph was 256GB alone.
 *  Don’t create a separate ‘pruned_graph’ host matrix in the `optimize` call, and just use host memory passed in by caller
 * Free the intermediate graph before creating the index


cpp/include/raft/neighbors/detail/cagra/utils.hpp
```diff
@@ -150,4 +152,93 @@ struct gen_index_msb_1_mask {
...
+ * In certain situations (UVM/HMM/ATS) host memory might be directly accessible on the
+ * device, and no extra allocations need to be performed. This class checks
+ * if the host_matrix_view is already accessible on the device, and only creates device
+ * memory and copies over if necessary. In memory limited situations this is preferable
+ * to having both a host and device copy
+ */
+template <typename T, typename IdxT>
+class device_matrix_view_from_host {
+ public:
+  device_matrix_view_from_host(raft::resources const& res, host_matrix_view<T, IdxT> host_view)
```
cjnolet, line 167: cc @wphicks this pattern seems a lot like the `mdbuffer` to me. The goal here is to make a `device_mdspan` when the pointer can be accessed from device or copy memory to device when it can't.

cpp/include/raft/neighbors/detail/cagra/utils.hpp
```diff
@@ -150,4 +152,93 @@ struct gen_index_msb_1_mask {
...
+ *
+ * In certain situations (UVM/HMM/ATS) host memory might be directly accessible on the
+ * device, and no extra allocations need to be performed. This class checks
+ * if the host_matrix_view is already accessible on the device, and only creates device
+ * memory and copies over if necessary. In memory limited situations this is preferable
+ * to having both a host and device copy
+ */
+template <typename T, typename IdxT>
+class device_matrix_view_from_host {
+ public:
```
cjnolet, line 166: I'm definitely okay keeping this as an internal utility for now. Could you add a todo to the docs here (and for the host->device conversion function) to use `mdbuffer` for this once it's available?

benfred: added a TODO here https://github.com/rapidsai/raft/pull/1790/commits/6552c66c820431285c5ee01705dbfaaf7b4435ed

cjnolet (approved):
LGTM. Thanks @benfred!

PR #1791, opened 2023-08-31
Add RAFT devcontainers

trxcllnt:
## Description
This PR adds some [devcontainers](https://containers.dev/) to help simplify building the RAFT C++ and Python libraries.

It also adds an optional job to the `pr.yaml` to [build the RAFT libs in each devcontainer](https://github.com/trxcllnt/raft/blob/fea/devcontainers/.github/workflows/pr.yaml#L96-L101), so the build caches are populated for devs by CI.

A devcontainer can be launched by clicking the "Reopen in Container" button that VSCode shows when opening the repo (or by using the "Rebuild and Reopen in Container" command from the command palette):
image

Clicking this button will cause VSCode to prompt the user to select one of these devcontainer variants:
image

On startup, the devcontainer creates or updates the conda/pip environment using `raft/dependencies.yaml`. The envs/package caches are cached on the host via volume mounts, which are described in more detail in [`.devcontainer/README.md`](https://github.com/trxcllnt/raft/blob/fea/devcontainers/.devcontainer/README.md).

The container includes convenience functions to clean, configure, and build the various RAFT components:

```shell
$ clean-raft-cpp # only cleans the C++ build dir
$ clean-pylibraft-python # only cleans the Python build dir
$ clean-raft # cleans both C++ and Python build dirs

$ configure-raft-cpp # only configures raft C++ lib

$ build-raft-cpp # only builds raft C++ lib
$ build-pylibraft-python # only builds raft Python lib
$ build-raft # builds both C++ and Python libs
```

* The C++ build script is a small wrapper around `cmake -S ~/raft/cpp -B ~/raft/cpp/build` and `cmake --build ~/raft/cpp/build`
* The Python build script is a small wrapper around `pip install --editable ~/raft/cpp`

Unlike `build.sh`, these convenience scripts *don't* install the libraries after building them. Instead, they automatically inject the correct arguments to build the C++ libraries from source and use their build dirs as package roots:

```shell
$ cmake -S ~/raft/cpp -B ~/raft/cpp/build
$ CMAKE_ARGS="-Draft_ROOT=~/raft/cpp/build" \ # <-- this argument is automatic
  pip install -e ~/raft/cpp
```

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/rapidsai/raft/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.


.devcontainer/README.md
```diff
@@ -0,0 +1,35 @@
...
+* `${repo}/../.conda:/home/coder/.conda`
+* `${repo}/../.config:/home/coder/.config`
+
+This ensures caches, configurations, dependencies, and your commits are persisted on the host across container runs.
+
+## Launch a Dev Container
+
+To launch a devcontainer from VSCode, open the RAFT repo and select the "Reopen in Container" button in the bottom right:<br/><img src="https://user-images.githubusercontent.com/178183/221771999-97ab29d5-e718-4e5f-b32f-2cdd51bba25c.png"/>
+
+Alternatively, open the VSCode command palette (typically `cmd/ctrl + shift + P`) and run the "Rebuild and Reopen in Container" command.
```
divyegala, line 35: Could you also add commands here for clean/build/configure that you have mentioned in the description of this PR? Also what does `configure` do?

dependencies.yaml
```diff
@@ -146,6 +176,20 @@ dependencies:
           - matrix: # All CUDA 11 versions
             packages:
               - &cuda_python11 cuda-python>=11.7.1,<12.0a0
+      - output_types: [requirements, pyproject]
```
divyegala, line 179: Are we able to write matrices to `pyproject` now or are devcontainers building wheels (or rather, building a python package) by writing matrices to `requirements.txt`?

trxcllnt: Once https://github.com/rapidsai/dependency-file-generator/pull/48 is merged yes, generating pyproject.toml via matrices will work. But the devcontainers currently build a combined `requirements.txt` for every project in the container, and that already works with matrices.

cjnolet:
@trxcllnt Are you ready for this PR to reviewed? Is the target still 23.10?

trxcllnt:
@cjnolet yes, this PR is ready for review/targeting 23.10. I'm fixing the style and pip-based build errors today.

cpp/.clangd
```diff
@@ -0,0 +1,65 @@
+# https://clangd.llvm.org/config
```
cjnolet, line 1: Does this imply that you've actually gotten RAFT compiling w/ clang?

cjnolet (approved):
I'll admit I'm no docker expert but the changes here LGTM in general.

cpp/.clangd
trxcllnt: No, this file is for the clang language server daemon to provide C++ and CUDA intellisense.

.devcontainer/README.md
trxcllnt: Configure does the `cmake` configure step, but doesn't subsequently run `cmake --build`. This can be handy if you're just working on CMake changes and don't want to try to compile after CMake is done configuring.

PR #1800, opened 2023-09-01
Various fixes to reproducible benchmarks

cjnolet:


cpp/bench/ann/src/hnswlib/hnswlib_wrapper.h
```diff
@@ -164,13 +166,13 @@ class HnswLib : public ANN<T> {
...
     int ef_construction;
-    int num_threads{1};
+    int num_threads = omp_get_num_procs();
   };
 
   using typename ANN<T>::AnnSearchParam;
   struct SearchParam : public AnnSearchParam {
     int ef;
-    int num_threads{1};
+    int num_threads = omp_get_num_procs();
```
achirkin, line 175: Does this work well with `num_threads > n_queries`? We've had a similar logic in raft refinement and the performance on small batches was horrible due to overheads of managing many threads compared to the amount of work (n_queries = 1).

cjnolet: Good question. This one is challenging because we don't (and shouldn't) know the number of queries when we set this argument.  We could set it to something small(ish) like 8 or 16, but that would just lower the saturation point for larger batch queries. In general, I know the queries for online systems are going to be long tailed, with 1 being in the main mass and >= 100 being in the tail. 

The problem is that when we run larger batch sizes, we aren't giving hnsw an fair try at all. My thinking was to take the middle ground- setting this to the number of cores. I guess I should measure the impact directly. What kind of perf difference are you seeing for, say, batch size of 10 when the thread pool contains the number of available cores?

achirkin: Maybe this explicit pool does better job than the openmp machinery that we rely upon in the refine operation. But there, I've got something like ~100x boost for a single-query batch `n_queries = 1` (72 cores). I've done some other refactoring at the same time though.

cpp/bench/ann/src/common/util.hpp
```diff
@@ -91,9 +91,15 @@ struct buf {
...
  private:
+#ifndef BUILD_CPU_ONLY
   cudaStream_t stream_{nullptr};
   cudaEvent_t start_{nullptr};
   cudaEvent_t stop_{nullptr};
+#else
+  cudaStream_t stream_{0};
+  cudaEvent_t start_{0};
+  cudaEvent_t stop_{0};
+#endif
```
achirkin, line 102:
```suggestion
  cudaStream_t stream_{nullptr};
  cudaEvent_t start_{nullptr};
  cudaEvent_t stop_{nullptr};
```

divyegala (approved):

PR #1803, opened 2023-09-04
[FEA] Add bitset for ANN pre-filtering and deletion

lowener:
Related to #1600


cpp/include/raft/util/bitset.cuh
```diff
@@ -0,0 +1,202 @@
...
+ * @brief View of a RAFT Bitset.
+ *
+ * This lightweight structure stores a pointer to a bitset in device memory with it's length.
+ * It provides a test() device function to check if a given index is set in the bitset.
+ *
+ * @tparam IdxT Indexing type used. Default is uint32_t.
+ */
+template <typename IdxT = uint32_t>
+struct bitset_view {
+  using BitsetT            = uint32_t;
```
wphicks, line 40: Any reason not to make this a templated type with the indicated default?

cpp/include/raft/util/bitset.cuh
```diff
@@ -0,0 +1,202 @@
...
+  _RAFT_HOST_DEVICE bitset_view(BitsetT* bitset_ptr, IdxT bitset_len)
+    : bitset_ptr_{bitset_ptr}, bitset_len_{bitset_len}
+  {
+  }
+  /**
+   * @brief Create a bitset view from a device vector view of the bitset.
+   *
+   * @param bitset_span Device vector view of the bitset
+   */
+  _RAFT_HOST_DEVICE bitset_view(raft::device_vector_view<BitsetT, IdxT> bitset_span)
```
wphicks, line 52: It seems like most of the logic is host-device compatible. Any reason why we restrict ourselves to an on-device bitset? I ask in particular because I can imagine several common workflows where we construct the bitset on host and then apply it on device by copying the underlying memory. As a matter of fact, this is exactly what we do in FIL.

Separately, it seems like a span, rather than an mdspan, is sufficient for what we're using it for. If we're going to use an mdspan anyway, is there a specific reason to restrict ourselves to 1D? If not, is there some other reason not to just accept a span?

cpp/include/raft/util/bitset.cuh
```diff
@@ -0,0 +1,202 @@
...
+ *
+ * This lightweight structure stores a pointer to a bitset in device memory with it's length.
+ * It provides a test() device function to check if a given index is set in the bitset.
+ *
+ * @tparam IdxT Indexing type used. Default is uint32_t.
+ */
+template <typename IdxT = uint32_t>
+struct bitset_view {
+  using BitsetT            = uint32_t;
+  IdxT bitset_element_size = sizeof(BitsetT) * 8;
```
wphicks, line 41:
```suggestion
  IdxT static constexpr const bitset_element_size = sizeof(BitsetT) * 8;
```
Don't know if the compiler is already converting the division/modulo operations to an efficient bitshift down [here](https://github.com/rapidsai/raft/pull/1803/files#diff-a1da31ebf261e38b85da164b4d187ad8886e11d9c52337c94dd3360487611f97R64), but having it explicitly declared `constexpr` should make that more likely.

I'd also recommend based on how this is being used that the type just be `auto` (std::size_t), but there may be a compelling reason to force it to an IdxT that I'm missing.

cpp/include/raft/util/bitset.cuh
```diff
@@ -0,0 +1,202 @@
...
+  }
+  /**
+   * @brief Device function to test if a given index is set in the bitset.
+   *
+   * @param sample_index Single index to test
+   * @return bool True if index has not been unset in the bitset
+   */
+  inline _RAFT_DEVICE auto test(const IdxT sample_index) const -> bool
+  {
+    const IdxT bit_element = bitset_ptr_[sample_index / bitset_element_size];
```
wphicks, line 64: `bitset_element_size` should be a power of 2, so we can use a bitshift to do the arithmetic here. In all likelihood, the compiler is already doing that (or will with the above constexpr change), but can you just double-check that that is actually happening in the emitted ptxas?

cpp/include/raft/util/bitset.cuh
```diff
@@ -0,0 +1,202 @@
...
+    const IdxT bit_index   = sample_index % bitset_element_size;
+    const bool is_bit_set  = (bit_element & (1ULL << bit_index)) != 0;
+    return is_bit_set;
+  }
+  /**
+   * @brief Get the device pointer to the bitset.
+   */
+  inline _RAFT_HOST_DEVICE auto get_bitset_ptr() -> BitsetT* { return bitset_ptr_; }
```
wphicks, line 72: It might be nice to keep the name of this method more consistent with e.g. `std::vector` and similar (`.data()`) or `mdspan` and similar (`.data_handle()`).

cpp/include/raft/util/bitset.cuh
```diff
@@ -0,0 +1,202 @@
...
+ * device-usable lightweight view of the bitset.
+ * Each index is represented by a single bit in the bitset. The total number of bytes used is
+ * ceil(bitset_len / 4).
+ * The underlying type of the bitset array is uint32_t.
+ * @tparam IdxT Indexing type used. Default is uint32_t.
+ */
+template <typename IdxT = uint32_t>
+struct bitset {
+  using BitsetT            = uint32_t;
+  IdxT bitset_element_size = sizeof(BitsetT) * 8;
```
wphicks, line 97:
```suggestion
  IdxT static constexpr const bitset_element_size = sizeof(BitsetT) * 8;
```

cpp/include/raft/util/bitset.cuh
```diff
@@ -0,0 +1,202 @@
...
+
+  /**
+   * @brief Construct a new bitset object with a list of indices to unset.
+   *
+   * @param res RAFT resources
+   * @param mask_index List of indices to unset in the bitset
```
wphicks, line 103: Any reason why this is a list of indices to unset rather than to set? As a consumer, that would be the opposite of what I would expect.

cpp/include/raft/util/bitset.cuh
```diff
@@ -0,0 +1,202 @@
...
+   * @param bitset_len Length of the bitset
+   */
+  bitset(const raft::resources& res,
+         raft::device_vector_view<const IdxT, IdxT> mask_index,
+         IdxT bitset_len)
+    : bitset_{raft::make_device_vector<BitsetT, IdxT>(
+        res, raft::ceildiv(bitset_len, bitset_element_size))}
+  {
+    RAFT_EXPECTS(mask_index.extent(0) <= bitset_len, "Mask index cannot be larger than bitset len");
```
wphicks, line 112: Is this check necessary? If the provided indices have duplicates, I'm not sure why we should necessarily disallow that.

cpp/include/raft/util/bitset.cuh
```diff
@@ -0,0 +1,202 @@
...
+    cudaMemsetAsync(bitset_.data_handle(),
```
wphicks, line 113: Would it be more efficient to construct the data on host and then just transfer it all at once, rather than setting and unsetting the data like this? It would come at the cost of host memory footprint, but I suspect that it would be worth it for large bitsets.

cpp/include/raft/util/bitset.cuh
```diff
@@ -0,0 +1,202 @@
...
+{
+  auto* bitset_ptr = bitset_view_.get_bitset_ptr();
+  thrust::for_each_n(resource::get_thrust_policy(res),
+                     mask_index.data_handle(),
+                     mask_index.extent(0),
+                     [bitset_ptr] __device__(const IdxT sample_index) {
+                       const IdxT bit_element = sample_index / 32;
+                       const IdxT bit_index   = sample_index % 32;
+                       const uint32_t bitmask = ~(1 << bit_index);
+                       atomicAnd(bitset_ptr + bit_element, bitmask);
```
wphicks, line 177: Can we avoid the atomic by constructing each `bit_element` that we need to touch and then doing a bitwise AND across a whole element (or even several adjacent elements)?

cpp/test/util/bitset.cu
```diff
@@ -0,0 +1,154 @@
...
+};
+
+auto operator<<(std::ostream& os, const test_spec& ss) -> std::ostream&
+{
+  os << "bitset{bitset_len: " << ss.bitset_len << ", mask_len: " << ss.mask_len << "}";
+  return os;
+}
+
+template <typename T>
+void add_cpu_bitset(std::vector<uint32_t>& bitset, const std::vector<T>& mask_idx)
```
wphicks, line 42: Would love to see all this host-side logic encapsulated for use elsewhere by making the bitset buildable from host/device memory.

wphicks (commented):
Love to see this functionality go in! This is something that we currently use in FIL, and I'd love to replace it with a RAFT version.

My general pieces of feedback would be:

- I'd like to make sure that we have a really good understanding of performance and how close to SoL this is. At minimum, adding benchmarks seems like it would be very worthwhile to allow for rapid perf iteration even after the initial PR.
- To the extent possible (and acknowledging the slight differences in purpose), matching the `std::bitset` API would make for a really nice developer experience. Little things like adding a `.flip` method would save downstream users a lot of time.
- Generalizing this as much as possible to different memory locations would also help with a lot of use cases that I can envision.

If we are to use this in FIL, we would also need to be able to compile with or without `nvcc`, but it is fine if we want to leave that off the table for now.

cpp/include/raft/util/bitset.cuh
achirkin: Also a nitpick: if we're to keep this alias, could you please consult with clang-tidy on the naming? I think it should be the snake case:
https://github.com/rapidsai/raft/blob/f691fc9651dc83a2e6b3ce7a8de524f4c2a57c44/cpp/.clang-tidy#L205-L206

lowener: @wphicks Are there use cases for having a different bitset type?

wphicks: For small bitsets, that allows us to avoid wasting memory. In FIL, for instance, we use chars for some of our bitsets and larger data types for others in order to try to back the bitset with a single element when possible but not to use extra memory if we can't. That allows us to perform very efficient bit twiddling operations for the single-element case but only copy exactly as much memory as necessary if we are stuck using multiple elements.

cpp/include/raft/util/bitset.cuh
lowener: It depend on the default value. I made the assumption that the bits are enabled by default that's why I chose `unset`.
I modified the code to be more flexible: The default value of the bitset can be specified at the construction and the bits can be set or unset depending on the value that the user gives.

cpp/include/raft/util/bitset.cuh
lowener: I initially implemented a custom kernel to create the bitset element in shared memory then doing one write to global memory but it was slower than the current approach.
I added benchmarks so that we can iterate later on potential performance improvements.

cpp/include/raft/util/bitset.cuh
lowener: I checked the ptxas and bitshifts are used for arithmetic (I checked after the constexpr changes)

wphicks:
Just to follow up on my previous review, I think all of the code details are in a good place, and we can follow up on any perf questions efficiently using the new benchmarks. My only remaining concern before merging an initial version of this is that it would be nice to make sure that the API does not change significantly in later follow-ups. To that end, it would be really great if we could match the `std::bitset` [API](https://en.cppreference.com/w/cpp/utility/bitset) as much as possible, with any deviations justified by the slight difference in purpose or the need for device-side support. Having e.g. `flip`, `set`, `reset`, etc. methods available on the object itself would make for a very easy transition for folks who are already familiar with `std::bitset`.

lowener:
The reasoning behind this was to have a stateless API as indicated in the [dev guide](https://github.com/rapidsai/raft/blob/branch-23.10/docs/source/developer_guide.md#stateless-c-apis).
Should I do an exception to match `std::bitset` API?

cjnolet:
> The reasoning behind this was to have a stateless API as indicated in the [dev guide](https://github.com/rapidsai/raft/blob/branch-23.10/docs/source/developer_guide.md#stateless-c-apis).
Should I do an exception to match std::bitset API?

@lowener sorry to do this last minute, but I think stateful is okay for these types of reusable objects. We want statelessness across the APIs for compute functions but `raft::bitset` is becoming a pretty core API (like mdspan or mdarray, for example) and I think maintaining state in the bitset is okay. I also think we should move the bitset into `raft/core`. That was my fault- I thought initially it would be a couple utility functions but now it's evolved into a first-class construct, which is great. 

I agree with @wphicks that we should match existing standards as much as possible. The stl are generally very well thought out by large teams of c++ veterans and by following their interfaces, at the very least, we give our users compatbility and familiarity.

docs/source/cpp_api/core.rst
```diff
@@ -20,4 +20,5 @@ expose in public APIs.
    core_nvtx.rst
    core_interruptible.rst
    core_operators.rst
-   core_math.rst
\ No newline at end of file
+   core_math.rst
+   core_bitset.rst
```
cjnolet, line 24: As we continue to build out these filtering primitives, I think we need to consider consolidating them under a single page in the documentation, kind of like what we have done for the math primitives. 

I don't think we need to do that right now, though. But something to think about- maybe something like "fast lookup primitives" or something.

cjnolet (approved):
I think this looks good for a first pass and given the circumstances, I'm going to approve it for now. We can continue to revisit and iterate on it as it gets integrated more into the ANN filtering features.

cpp/include/raft/core/bitset.cuh
```diff
@@ -0,0 +1,308 @@
...
+    const index_t bit_index    = sample_index % bitset_element_size;
+    const bool is_bit_set      = (bit_element & (bitset_t{1} << bit_index)) != 0;
+    return is_bit_set;
+  }
+
+  /**
+   * @brief Get the device pointer to the bitset.
+   */
+  inline _RAFT_HOST_DEVICE auto data_handle() -> bitset_t* { return bitset_ptr_; }
+  inline _RAFT_HOST_DEVICE auto data_handle() const -> const bitset_t* { return bitset_ptr_; }
```
divyegala, line 77:
```suggestion
  inline _RAFT_HOST_DEVICE auto data() -> bitset_t* { return bitset_ptr_; }
  inline _RAFT_HOST_DEVICE auto data() const -> const bitset_t* { return bitset_ptr_; }
```

cpp/include/raft/core/bitset.cuh
```diff
@@ -0,0 +1,308 @@
...
+  }
+
+  inline auto to_mdspan() -> raft::device_vector_view<bitset_t, index_t>
+  {
+    return raft::make_device_vector_view<bitset_t, index_t>(bitset_ptr_, n_elements());
+  }
+  inline auto to_mdspan() const -> raft::device_vector_view<const bitset_t, index_t>
+  {
+    return raft::make_device_vector_view<const bitset_t, index_t>(bitset_ptr_, n_elements());
+  }
```
divyegala, line 98: Can we also add two operators? `inline auto operator raft::device_vector_view...`, this will make the conversion automatic

cpp/bench/prims/core/bitset.cu
```diff
@@ -0,0 +1,74 @@
...
+
+using Uint8_32  = bitset_bench<uint8_t, uint32_t>;
+using Uint16_64 = bitset_bench<uint16_t, uint32_t>;
+using Uint32_32 = bitset_bench<uint32_t, uint32_t>;
+using Uint32_64 = bitset_bench<uint32_t, uint64_t>;
+
+RAFT_BENCH_REGISTER(Uint8_32, "", bitset_input_vecs);
+RAFT_BENCH_REGISTER(Uint16_64, "", bitset_input_vecs);
+RAFT_BENCH_REGISTER(Uint32_32, "", bitset_input_vecs);
+RAFT_BENCH_REGISTER(Uint32_64, "", bitset_input_vecs);
```
divyegala, line 72: Can you please post the benchmark results as a comment in your PR?

cpp/include/raft/core/bitset.cuh
```diff
@@ -0,0 +1,308 @@
...
+ * This structure encapsulates a bitset in device memory. It provides a view() method to get a
+ * device-usable lightweight view of the bitset.
+ * Each index is represented by a single bit in the bitset. The total number of bytes used is
+ * ceil(bitset_len / 8).
+ * @tparam bitset_t Underlying type of the bitset array. Default is uint32_t.
+ * @tparam index_t Indexing type used. Default is uint32_t.
+ */
+template <typename bitset_t = uint32_t, typename index_t = uint32_t>
+struct bitset {
+  index_t static constexpr const bitset_element_size = sizeof(bitset_t) * 8;
```
divyegala, line 117:
```suggestion
  index_t static constexpr bitset_element_size = sizeof(bitset_t) * 8;
```

cpp/include/raft/core/bitset.cuh
```diff
@@ -0,0 +1,308 @@
...
+  void reset(const raft::resources& res)
+  {
+    cudaMemsetAsync(bitset_.data(),
+                    default_value_ ? 0xff : 0x00,
+                    n_elements() * sizeof(bitset_t),
+                    resource::get_cuda_stream(res));
+  }
+
+ private:
+  raft::device_uvector<bitset_t> bitset_;
```
divyegala, line 302: What type is this? I don't recognize it

cpp/include/raft/core/bitset.cuh
```diff
@@ -0,0 +1,308 @@
...
+  [[nodiscard]] inline auto view() const -> raft::core::bitset_view<const bitset_t, index_t>
+  {
+    return bitset_view<const bitset_t, index_t>(to_mdspan(), bitset_len_);
+  }
+
+  /**
+   * @brief Get the device pointer to the bitset.
+   */
+  inline auto data_handle() -> bitset_t* { return bitset_.data(); }
+  inline auto data_handle() const -> const bitset_t* { return bitset_.data(); }
```
divyegala, line 185:
```suggestion
  inline auto data() -> bitset_t* { return bitset_.data(); }
  inline auto data() const -> const bitset_t* { return bitset_.data(); }
```

cpp/include/raft/core/bitset.cuh
```diff
@@ -0,0 +1,308 @@
...
+  bitset(const raft::resources& res,
+         raft::device_vector_view<const index_t, index_t> mask_index,
+         index_t bitset_len,
+         bool default_value = true)
+    : bitset_{std::size_t(raft::ceildiv(bitset_len, bitset_element_size)),
+              raft::resource::get_cuda_stream(res)},
+      bitset_len_{bitset_len},
+      default_value_{default_value}
+  {
+    cudaMemsetAsync(bitset_.data(),
```
divyegala, line 136: Please don't use cuda rt API directly

cpp/include/raft/core/bitset.cuh
```diff
@@ -0,0 +1,308 @@
...
+   *
+   * @param res RAFT resources
+   * @param mask_index List of indices to unset in the bitset
+   * @param bitset_len Length of the bitset
+   * @param default_value Default value to set the bits to. Default is true.
+   */
+  bitset(const raft::resources& res,
+         raft::device_vector_view<const index_t, index_t> mask_index,
+         index_t bitset_len,
+         bool default_value = true)
```
divyegala, line 130: Can we use an enum here instead of boolean? Something like `default_value::ALL_ONES` and `default_value::ALL_ZEROS`

cpp/include/raft/core/bitset.cuh
```diff
@@ -0,0 +1,308 @@
...
+   * @brief Get the number of bits of the bitset representation.
+   */
+  inline auto size() const -> index_t { return bitset_len_; }
+
+  /**
+   * @brief Get the number of elements used by the bitset representation.
+   */
+  inline auto n_elements() const -> index_t
+  {
+    return raft::ceildiv(bitset_len_, bitset_element_size);
```
divyegala, line 196:
```suggestion
    return bitset_.size();
```

cpp/include/raft/core/bitset.cuh
```diff
@@ -0,0 +1,308 @@
...
+
+  /**
+   * @brief Construct a new bitset object with a list of indices to unset.
+   *
+   * @param res RAFT resources
+   * @param mask_index List of indices to unset in the bitset
+   * @param bitset_len Length of the bitset
+   * @param default_value Default value to set the bits to. Default is true.
+   */
+  bitset(const raft::resources& res,
```
divyegala, line 127: Why not store this as a member so that none of the other functions need to accept `raft::resources` on their API?

cpp/include/raft/core/bitset.cuh
```diff
@@ -0,0 +1,308 @@
...
+   * the default value. */
+  void resize(const raft::resources& res, index_t new_bitset_len)
+  {
+    auto old_size = raft::ceildiv(bitset_len_, bitset_element_size);
+    auto new_size = raft::ceildiv(new_bitset_len, bitset_element_size);
+    bitset_.resize(new_size);
+    bitset_len_ = new_bitset_len;
+    if (old_size < new_size) {
+      // If the new size is larger, set the new bits to the default value
+      cudaMemsetAsync(bitset_.data() + old_size,
```
divyegala, line 219: Use RAFT API

divyegala (commented):
Are device functions for `test/set/flip` going to be follow-up work?

cpp/include/raft/core/bitset.cuh
lowener: It's a wrapper around `rmm::device_uvector`

PR #1805, opened 2023-09-06
Fix `raft-dask` naming in wheel builds

divyegala:
Before https://github.com/rapidsai/raft/pull/1804/files, we were using `raft-dask` and `raft_dask` interchangeably because they were hardcoded. Since the package name is now capture as a variable, it's a bug to use `raft_dask` as the pyproject name does not get updated correctly and we end up publishing `raft-dask` wheels without `cu{version}` suffixes

### DIFF
--- a/ci/build_wheel.sh
+++ b/ci/build_wheel.sh
@@ -34,7 +34,7 @@ if ! rapids-is-release-build; then
     alpha_spec=',>=0.0.0a0'
 fi
 
-if [[ ${package_name} == "raft_dask" ]]; then
+if [[ ${package_name} == "raft-dask" ]]; then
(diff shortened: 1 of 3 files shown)


vyasr (changes requested):
This change mostly aligns with the fix we discussed, but I think you may have changed one too many places and you're going to run into some minor issues because of the way that some names get normalized. In particular, I think that the wheel names (like the actual filename) will still use `raft_dask` instead of `raft-dask` because that's the pip convention (convert hyphens to underscores). I could be wrong, but CI should reveal this for me. If you can test locally that might help you get through it faster so that you don't have to wait for pylibraft.

ajschmidt8 (approved):

vyasr (approved):

PR #1808, opened 2023-09-07
ann-bench: miscellaneous improvements

achirkin:
1. IVF-PQ: slightly improve stream ordering
2. IVF-PQ: build param 'codebook_kind' - as per `ivf_pq_types`.
3. FAISS IVF models: build param 'ratio' with the same meaning as in IVF-PQ - the clustering algorithm uses `1/ratio` of the given dataset for training.

### DIFF
--- a/cpp/bench/ann/src/raft/raft_benchmark.cu
+++ b/cpp/bench/ann/src/raft/raft_benchmark.cu
@@ -60,7 +60,6 @@ void parse_build_param(const nlohmann::json& conf,
   if (conf.contains("niter")) { param.kmeans_n_iters = conf.at("niter"); }
   if (conf.contains("ratio")) {
     param.kmeans_trainset_fraction = 1.0 / (double)conf.at("ratio");
-    std::cout << "kmeans_trainset_fraction " << param.kmeans_trainset_fraction;
   }
 }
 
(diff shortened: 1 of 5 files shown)


cpp/bench/ann/src/raft/raft_benchmark.cu
```diff
@@ -82,6 +79,17 @@ void parse_build_param(const nlohmann::json& conf,
   if (conf.contains("ratio")) { param.kmeans_trainset_fraction = 1.0 / (double)conf.at("ratio"); }
   if (conf.contains("pq_bits")) { param.pq_bits = conf.at("pq_bits"); }
   if (conf.contains("pq_dim")) { param.pq_dim = conf.at("pq_dim"); }
+  if (conf.contains("codebook_kind")) {
+    std::string kind = conf.at("codebook_kind");
```
cjnolet, line 83: Need to also document all changes to build and search params in the docs/source/ann_benchmarks_param_tuning.md. Otherwise nobody except us know they exist or how to set them.

achirkin: Thanks for reminder, updated! Ideally though, I think, the parameters should match 1-1 the C++ or python API, so that we could refer the user to the corresponding docs.

cjnolet: I think it's a great idea to keep the parameter names in line w/ the actual arguments in the APIs but the goal for that parameter tuning guide is to make it painfully easy to use the benchmarks specifically, across all libraries and not just RAFT, so that we're not having the users jump all over the place to run their own benchmarks.

cjnolet (approved):

PR #1811, opened 2023-09-08
[FEA] Add pre-filtering to CAGRA

enp1s0:
This PR adds the pre-filtering feature to the CAGRA search implementations.

Rel: taken over from https://github.com/rapidsai/raft/pull/1765

## Algorithm
The pre-filtering algorithm removes a node that should not be in the final result after it has behaved as a parent node. This way, the nodes that should not be in the final result are also used in the graph traversal, avoiding potential performance degradation.

## Changes
- Add filtering operation on a parent node after internal top-M buffer candidate calculation.
- Add filtering operation to result buffer before storing them in the device memory.


enp1s0:
- [ ] Add query id offset
- [ ] Fix test

lowener (commented):
This looks very good! Thanks for this follow up PR

enp1s0:
Thank you, @lowener, for reviewing the code. I have modified the things you pointed out.

cpp/include/raft/neighbors/detail/cagra/search_multi_kernel.cuh
```diff
@@ -436,6 +468,50 @@ void remove_parent_bit(const std::uint32_t num_queries,
...
+                  DISTANCE_T* const result_distances_ptr,
+                  const std::size_t lds,
+                  const std::uint32_t result_buffer_size,
+                  const std::uint32_t num_queries,
+                  const INDEX_T query_id_offset,
+                  SAMPLE_FILTER_T sample_filter,
+                  cudaStream_t cuda_stream)
+{
+  const std::size_t block_size = 256;
+  const std::size_t grid_size  = (num_queries * result_buffer_size + block_size - 1) / block_size;
```
tfeher, line 504:
```suggestion
  const std::size_t grid_size  = ceildiv(num_queries * result_buffer_size, block_size ) ;
```

tfeher (commented):
Thanks @enp1s0 for this PR, I had a quick look at it. The changes in the template instantiations make the PR look large, but the relevant changes are more concise.

Overall the PR looks good, but it should be explained how the filtering works. You gave a detailed description in [this comment](https://github.com/rapidsai/raft/pull/1765#issuecomment-1700519191), and it would be good to add part of it as code comments, or at least include it in the PR description.

lowener (changes requested):
Found those 2 changes while combining pre-filtering with bitset. I think that this behavior is not present in multi-kernel

enp1s0:
Thank you, @tfeher and @lowener, for the review. I have fixed the code and explained how the filtering works in the PR description.

tfeher (approved):
Thanks @enp1s0 for the update LGTM.

lowener (approved):

cpp/include/raft/neighbors/sample_filter_types.hpp
```diff
@@ -37,6 +37,18 @@ struct none_ivf_sample_filter {
   }
 };
 
+/* A filter that filters nothing. This is the default behavior. */
```
cjnolet, line 40: Can we please use one sample filtering API across all the ANN indexes? For example, a user should be able to instantiate a filter and use it in IVF-flat, ivf-pq, brute-force, or cagra. Is there anything stopping us from being able to do this?

cjnolet (changes requested):
Thanks again for getting this feature ready for 23.10, @enp1s0! Others have already done great reviews, but I see just one API thing that I'd like to address before we end up with too many different filtering implementations.

cpp/include/raft/neighbors/sample_filter_types.hpp
enp1s0: @lowener What do you think about the @cjnolet 's suggestion? This code is taken from your [PR](https://github.com/rapidsai/raft/pull/1765). I think we can use `none_ivf_sample_filter` as the default filter instead of `none_cagra_sample_filter`, but the name is not good because CAGRA is not ivf. I prefer to rename `none_ivf_sample_filter` to `none_sample_filter` but it changes the API so I'm not sure it is the best way for users.

cjnolet: We have been communicating to users that sample filtering has been in an experimental state because we haven't had any actual optimized filtering primitives like bitsets, bitmaps, or bloom filters for them to use with the filters. It will be okay to make small adjustments to the APIs at this point, especially if they are consolidating allowing for more reuse.

cjnolet:
@enp1s0 @lowener I'm okay doing the rename as a follow-on to speed up reviews for the PRs that rely on this feature.

cjnolet (approved):

Issue #1813, opened 2023-09-10
[FEA] Add FAISS cpu indexes to `raft-ann-bench`


cjnolet:
Adding the cpu index types for FAISS is fairly straightforward but what becomes challenging is building the faiss targets for both gpu and cpu-only environments since the cmake build targets automatically pass through the linking to the faiss gpu lib.

Still unclear of a good way to handle this but tagging @robertmaynard for thoughts.

PR #1814, opened 2023-09-11
Adding FAISS cpu to `raft-ann-bench`

cjnolet:
Closes https://github.com/rapidsai/raft/issues/1813

This PR adds FAISS CPU indexes to the raft-Ann-bench benchmarks, and adjusts the build accordingly. In addition, this PR also updated the FAISS version to include the RAFT-enabled version, which required removing the FAISS conda packages and building FAISS from source.


cjnolet:
The big challenge with this change is that `faiss` already uses the `faiss::faiss` target and implicitly passes through the dependency on the `libfaiss-gpu` target. It's not clear to me the best approach to pass through the target only sometimes, such as when building `raft-ann-bench` package but not when building the `raft-ann-bench-cpu` package. 

I suppose one way to accomplish this would be to accept an option in `get_faiss.cmake` to set `FAISS_ENABLE_GPU` to `OFF`. Tagging @robertmaynard for any thoughts.

achirkin:
Ha-ha, I had some drafts on this too, but you beat me to it :) I struggled with the FAISS CPU/GPU builds as well and came up with this solution https://github.com/rapidsai/raft/pull/1820 . It's not perfect and needs some refinement, but it works. Any thoughts on this?

cjnolet:
Fixed the handling of the AVX library. @tfeher, upon closer inspection, I realized you were correct that FAISS is passing the `faiss_avx2` linkage through independently from the `faiss::faiss` linkage. I fixed it and we're now properly installing the `libfaiss-avx2` conda package only when the architecture is not aarch64.

cpp/bench/ann/CMakeLists.txt
```diff
@@ -15,9 +15,18 @@
...
 option(RAFT_ANN_BENCH_USE_FAISS_IVF_PQ "Include faiss' ivf pq algorithm in benchmark" ON)
+option(RAFT_ANN_BENCH_USE_FAISS_CPU_FLAT
+       "Include faiss' cpu brute-force knn algorithm in benchmark" ON
+)
+option(RAFT_ANN_BENCH_USE_FAISS_CPU_FLAT "Include faiss' cpu brute-force algorithm in benchmark" ON)
+
+option(RAFT_ANN_BENCH_USE_FAISS_CPU_IVF_FLAT "Include faiss' cpu ivf flat algorithm in benchmark"
+       ON
+)
+option(RAFT_ANN_BENCH_USE_FAISS_CPU_IVF_PQ "Include faiss' cpu ivf pq algorithm in benchmark" ON)
```
achirkin, line 29: Since this is a rather big, breaking change by now, would it make sense to standardize the FAISS algo names?:
```
CMake OPTION                             executable name                algo name
RAFT_ANN_BENCH_USE_FAISS_GPU_FLAT        FAISS_GPU_FLAT_ANN_BENCH       faiss_gpu_flat
RAFT_ANN_BENCH_USE_FAISS_GPU_IVF_FLAT    FAISS_GPU_IVF_FLAT_ANN_BENCH   faiss_gpu_ivf_flat
RAFT_ANN_BENCH_USE_FAISS_GPU_IVF_PQ      FAISS_GPU_IVF_PQ_ANN_BENCH     faiss_gpu_ivf_pq
RAFT_ANN_BENCH_USE_FAISS_CPU_FLAT        FAISS_CPU_FLAT_ANN_BENCH       faiss_cpu_flat
RAFT_ANN_BENCH_USE_FAISS_CPU_IVF_FLAT    FAISS_CPU_IVF_FLAT_ANN_BENCH   faiss_cpu_ivf_flat
RAFT_ANN_BENCH_USE_FAISS_CPU_IVF_PQ      FAISS_CPU_IVF_PQ_ANN_BENCH     faiss_cpu_ivf_pq
```
i.e. consistently add the CPU/GPU prefixes to both versions. This would make it easier to do the algorithm name checks (e.g. executable name -> to lower case).
(At this point, it seems the cmake options specify only "CPU", but algo names in configs specify only "gpu")

tfeher: +1 for explicit CPU/GPU prefix. BTW, what is the motivation to have so many separate executables?

cpp/bench/ann/src/faiss/faiss_cpu_wrapper.h
```diff
@@ -0,0 +1,310 @@
...
+  std::unique_ptr<faiss::IndexRefineFlat> index_refine_;
+  faiss::MetricType metric_type_;
+  int nlist_;
+  double training_sample_fraction_;
+};
+
+template <typename T>
+void Faiss<T>::build(const T* dataset, size_t nrow, cudaStream_t stream)
+{
+  OmpSingleThreadScope omp_single_thread;
```
tfeher, line 145: Please remove this. We do need multi-threaded build.

cpp/bench/ann/CMakeLists.txt
cjnolet: Yes, I plan to fix the naming so that it's more consistent. 

> what is the motivation to have so many separate executables?

The reason we did this is because we had some conflicting compile flags across binaries- specifically between things like HNSW and FAISS. I think it's better to allow each library to have their own individual binaries. I think we should also consolidate the binaries within each library but we're butting up against the release here and I'd like to do that consolidation in 23.12 since it's tech debt and not impacting the operation of the benchmarks.

cpp/bench/ann/CMakeLists.txt
```diff
@@ -213,20 +242,48 @@ if(RAFT_ANN_BENCH_USE_RAFT_CAGRA)
   )
 endif()
 
+set(RAFT_FAISS_TARGETS faiss::faiss)
+if(CXX_AVX_FOUND)
+  list(APPEND RAFT_FAISS_TARGETS faiss::faiss_avx2)
+endif()
```
tfeher, line 248: This way it will pick up non AVX symbols from `faiss::faiss`, resulting in slower binary. The following worked for me:
```suggestion
if(CXX_AVX_FOUND)
  set(RAFT_FAISS_TARGETS faiss::faiss_avx2)
else()
  set(RAFT_FAISS_TARGETS faiss::faiss)
endif()
```

robertmaynard: No need to check a variable, just query if the avx2 target exists.
```cmake
set(RAFT_FAISS_TARGETS faiss::faiss)
if(TARGET faiss::faiss_avx2)
  set(RAFT_FAISS_TARGETS faiss::faiss_avx2)
endif()
```

cpp/cmake/thirdparty/get_faiss.cmake
```diff
@@ -30,36 +30,57 @@ function(find_and_configure_faiss)
             set(CPM_DOWNLOAD_faiss ON)
         endif()
 
+        include(cmake/modules/FindAVX.cmake)
+
+        # Link against AVX CPU lib if it exists
+        set(RAFT_FAISS_GLOBAL_TARGETS faiss::faiss)
```
robertmaynard, line 36: You don't need these conditionals around GLOBAL_TARGETS just provide them all and rapids-cmake will first check that they exist before doing any operations on them

cjnolet:
@robertmaynard looks like this worked! Thanks so much. Honestly, I had no clue that the FAISS side of the cmake didn't include the same search paths as the raft side of the cmake. I just assumed that part was transitively. So happy to see this building in CI finally.

conda/recipes/raft-ann-bench-cpu/meta.yaml
```diff
@@ -47,6 +47,7 @@ requirements:
   host:
     - glog {{ glog_version }}
     - matplotlib
+#    - faiss-cpu {{ faiss_version }}
```
divyegala, line 50: Reminder to remove comments in this file

conda/recipes/raft-ann-bench/meta.yaml
```diff
@@ -71,10 +71,10 @@ requirements:
...
     - nlohmann_json {{ nlohmann_json_version }}
     # Temporarily ignore faiss benchmarks on CUDA 12 because packages do not exist yet
-    {% if cuda_major == "11" %}
-    - faiss-proc=*=cuda
-    - libfaiss {{ faiss_version }}
-    {% endif %}
+#    {% if cuda_major == "11" %}
+#    - faiss-proc=*=cuda
+#    - libfaiss {{ faiss_version }}
+#    {% endif %}
```
divyegala, line 77: Reminder to remove comments in this file

cpp/bench/ann/CMakeLists.txt
```diff
@@ -15,9 +15,18 @@
 # ##################################################################################################
 # * benchmark options ------------------------------------------------------------------------------
 
-option(RAFT_ANN_BENCH_USE_FAISS_BFKNN "Include faiss' brute-force knn algorithm in benchmark" ON)
-option(RAFT_ANN_BENCH_USE_FAISS_IVF_FLAT "Include faiss' ivf flat algorithm in benchmark" ON)
-option(RAFT_ANN_BENCH_USE_FAISS_IVF_PQ "Include faiss' ivf pq algorithm in benchmark" ON)
+option(RAFT_ANN_BENCH_USE_FAISS_GPU_FLAT "Include faiss' brute-force knn algorithm in benchmark" ON)
+option(RAFT_ANN_BENCH_USE_FAISS_GPU_IVF_FLAT "Include faiss' ivf flat algorithm in benchmark" ON)
+option(RAFT_ANN_BENCH_USE_FAISS_GPU_IVF_PQ "Include faiss' ivf pq algorithm in benchmark" ON)
```
divyegala, line 20: I think some of our benchmarks also have `faiss_gpu_ivf_sq` algorithm, can you add a target for that as well? Right now it works because we build all faiss algorithms into all faiss named executables https://github.com/rapidsai/raft/blob/2f08bdd75286688161ca1c8ff3b608985be117bf/cpp/bench/ann/src/faiss/faiss_benchmark.cu#L123

cpp/bench/ann/src/faiss/faiss_cpu_benchmark.cpp
```diff
@@ -0,0 +1,162 @@
...
+  if constexpr (std::is_same_v<T, float>) {
+    raft::bench::ann::Metric metric = parse_metric(distance);
+    if (algo == "faiss_gpu_ivf_flat") {
+      ann = make_algo<T, raft::bench::ann::FaissCpuIVFFlat>(metric, dim, conf, dev_list);
+    } else if (algo == "faiss_gpu_ivf_pq") {
+      ann = make_algo<T, raft::bench::ann::FaissCpuIVFPQ>(metric, dim, conf);
+    } else if (algo == "faiss_gpu_ivf_sq") {
+      ann = make_algo<T, raft::bench::ann::FaissCpuIVFSQ>(metric, dim, conf);
+    } else if (algo == "faiss_gpu_flat") {
+      ann = std::make_unique<raft::bench::ann::FaissCpuFlat<T>>(metric, dim);
```
divyegala, line 126: The algorithm names should not have "gpu" in them. Please replace with "cpu" or nothing

cpp/bench/ann/src/faiss/faiss_cpu_benchmark.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+template <typename T>
+std::unique_ptr<typename raft::bench::ann::ANN<T>::AnnSearchParam> create_search_param(
+  const std::string& algo, const nlohmann::json& conf)
+{
+  if (algo == "faiss_gpu_ivf_flat" || algo == "faiss_gpu_ivf_pq" || algo == "faiss_gpu_ivf_sq") {
+    auto param = std::make_unique<typename raft::bench::ann::FaissCpu<T>::SearchParam>();
+    parse_search_param<T>(conf, *param);
+    return param;
+  } else if (algo == "faiss_gpu_flat") {
```
divyegala, line 145: Same as above

cpp/bench/ann/CMakeLists.txt
divyegala: Also, below we have a CPU IVF SQ benchmark too, please add that target

divyegala (changes requested):
Why was it needed to get `fmt` and `spdlog` for CPU only builds? I know currently RMM brings those, but are we providing them for FAISS?

cpp/bench/ann/src/faiss/faiss_cpu_benchmark.cpp
cjnolet: I did this a couple hours ago. Bad search and replace.

cpp/bench/ann/CMakeLists.txt
cjnolet: I'd prefer not to expose ivf sq at the moment. It's been in the code since the first iteration of the benchmarks and so was copied to the cpu version as well. It's thus far untested and not something we need to benchmark at the moment anyways.

divyegala: Okay, that sounds good to me!

cjnolet: That being said, I'm going to create a GitHub issue to track the ivf sq and reference it in the code so we don't lose sight of it

conda/environments/bench_ann_cuda-118_arch-x86_64.yaml
```diff
@@ -30,11 +29,11 @@ dependencies:
 - libcusolver=11.4.1.48
 - libcusparse-dev=11.7.5.86
 - libcusparse=11.7.5.86
-- libfaiss>=1.7.1
 - nccl>=2.9.9
 - ninja
 - nlohmann_json>=3.11.2
 - nvcc_linux-64=11.8
+- rmm=23.10.*
```
divyegala, line 36: This version is now wrong

cpp/bench/ann/src/faiss/faiss_gpu_wrapper.h
```diff
@@ -107,7 +107,7 @@ class FaissGpu : public ANN<T> {
 
   void build(const T* dataset, size_t nrow, cudaStream_t stream = 0) final;
 
-  void set_search_param(const AnnSearchParam& param) override;
+  virtual void set_search_param(const FaissGpu<T>::AnnSearchParam& param) {}
```
divyegala, line 110: I'm not too sure about this, should this be?
```suggestion
  virtual void set_search_param(const FaissGpu<T>::AnnSearchParam& param) = 0;
```

dependencies.yaml
```diff
@@ -225,9 +225,8 @@ dependencies:
           - nlohmann_json>=3.11.2
           - glog>=0.6.0
           - h5py>=3.8.0
-          - libfaiss>=1.7.1
           - benchmark>=1.8.2
-          - faiss-proc=*=cuda
+          - rmm=23.10.*
```
divyegala, line 229:
```suggestion
          - rmm=23.12.*
```

divyegala (approved):

PR #1817, opened 2023-09-11
Add index class for brute_force knn

benfred:
This adds an index class to match the ANN methods. This allows us to precompute norms for the dataset in `brute_force::build` and then use them in `brute_force::search` - meaning we don't have to compute norms for the entire dataset on every query.


cpp/include/raft/neighbors/brute_force-ext.cuh
```diff
@@ -38,6 +39,21 @@ inline void knn_merge_parts(
   size_t n_samples,
   std::optional<raft::device_vector_view<idx_t, idx_t>> translations = std::nullopt) RAFT_EXPLICIT;
 
+template <typename T,
+          typename Accessor =
+            host_device_accessor<std::experimental::default_accessor<T>, memory_type::host>>
```
cjnolet, line 44: Can we just let the tile get inferred? I don't think we want to default this.

cjnolet (changes requested):
Looks great! Just a few minor things.

cjnolet (approved):
LGTM!

PR #1818, opened 2023-09-12
Provide a raft::copy overload for mdspan-to-mdspan copies

### LINKED ISSUE #1779: [FEA] Add version of `raft::copy()` that accepts `mdspan` and `device_resources` directly.
The mdspan APIs can make interacting with RAFT really clean, but we still have too many remnants of the old pointer-based APIs sitting around that make it a little more nebulus for new users. We should strive to provide `mdspan` variants of these pointer-based APIs wherever possible.

wphicks:
# Purpose
This PR provides a utility for copying between generic mdspans. This includes between host and device, between mdspans of different layouts, and between mdspans of different (convertible) data types

## API
`raft::copy(raft_resources, dest_mdspan, src_mdspan);`

# Limitations

- Currently does not support copies between mdspans on two different GPUs
- Currently not performant for generic host-to-host copies (would be much easier to optimize with submdspan for padded layouts)
- Submdspan with padded layouts would also make it easier to improve perf of some device-to-device copies, though perf should already be quite good for most device-to-device copies.

# Design

- Includes optional `RAFT_DISABLE_CUDA` build definition in order to use this utility in CUDA-free builds (important for use in the FIL backend for Triton)
- Includes a new `raft::stream_view` object which is a thin wrapper around `rmm::stream_view`. Its purpose is solely to provide a symbol that will be defined in CUDA-free builds and which will throw exceptions or log error messages if someone tries to use a CUDA stream in a CUDA-free build. This avoids a whole bunch of ifdefs that would otherwise infect the whole codebase.
- Uses (roughly in order of preference): `cudaMemcpyAsync, std::copy, cublas, custom device kernel, custom host-to-host transfer logic` for the underlying copy
- Provides two different headers: `raft/core/copy.hpp` and `raft/core/copy.cuh`. This is to accommodate the custom kernel necessary for handling completely generic device-to-device copies. See below for more details.

## Details on the header split
For many instantiations, even those which involve the device, we do not require nvcc compilation. If, however, we determine at compilation time that we must use a custom kernel for the copy, then we must invoke nvcc. We do not wish to indicate that a public header file is a C++ header when it is a CUDA header or vice versa, so we split the definitions into separate `hpp` and `cuh` files, with all template instantiations requiring the custom kernel enable-if'd out of the hpp file.

Thus, the cuh header can be used for _any_ mdspan-to-mdspan copy, but the hpp file will not compile for those specific instantiations that require a custom kernel. The recommended workflow is that if a `cpp` file requires an mdspan-to-mdspan copy, first try the `hpp` header. If that fails, the `cpp` file must be converted to a `cu` file, and the `cuh` header should be used. For source files that are already being compiled with nvcc (i.e. `.cu` files), the `cuh` header might as well be used and will not result in any additional compile time penalty.

# Remaining tasks to leave WIP status

- [x] Add benchmarks for copies
- [x] Ensure that new function is correctly added to docs

# Follow-up items

- Optimize host-to-host transfers using a cache-oblivious approach with SIMD-accelerated transposes for contiguous memory
- Test cache-oblivious device-to-device transfers and compare performance
- Provide transparent support for copies between devices.

## Relationship to mdbuffer
This utility encapsulates a substantial chunk of the core logic required for the mdbuffer implementation. It is being split into its own PR both because it is useful on its own and because the mdbuffer work has been delayed by higher priority tasks.

Close #1779


cpp/CMakeLists.txt
```diff
@@ -22,7 +22,8 @@ include(rapids-find)
 
 option(BUILD_CPU_ONLY "Build CPU only components. Applies to RAFT ANN benchmarks currently" OFF)
 
-# workaround for rapids_cuda_init_architectures not working for arch detection with enable_language(CUDA)
+# workaround for rapids_cuda_init_architectures not working for arch detection with
```
wphicks, line 25: These changes seem to be enforced by cmake-format. Not sure why they were not present before.

cpp/include/raft/core/detail/copy.hpp
```diff
@@ -0,0 +1,510 @@
...
+
+template <bool B, typename DstType = void, typename SrcType = void, typename T = void>
+struct mdspan_copyable {};
+
+/*
+ * A helper struct used to determine whether one mdspan type can be copied to
+ * another and if so how
+ */
+template <typename DstType, typename SrcType, typename T>
+struct mdspan_copyable<true, DstType, SrcType, T> {
```
mhoemmen, line 50: Here's how I would implement optimized mdspan copy.

## Dispatch logic

1. If the destination has a nonunique layout, then forbid copying (except for specializations for known nonunique layouts).

2. If the two mdspans have the same layout, then

    a. if the two mdspans have contiguous layout, then dispatch to same-contiguous-copy;

    b. otherwise, dispatch to same-noncontiguous-copy.

3. If the two mdspans have different layout, then

    a. if the layouts and accessors work with `cudaMemcpy*Async`, then dispatch to different-layout-known-copy;

    b. otherwise, dispatch to generic-flattened-copy.

## same-contiguous-copy

1. If it's valid to use `memcpy` (or `cudaMemcpy`, etc.) for the two accessors, then use it.

2. Otherwise, dispatch to same-noncontiguous-copy.

## same-noncontiguous-copy

1. If it's valid to use a parallel algorithm (or device kernel) with the data, then use generic flat copy from `[0, src.mapping().required_span_size())` (transformed through `src.accessor().access(src.data_handle(), _)`) to `[0, dst.mapping().required_span_size())` (transformed analogously through `dst.accessor().access(dst.data_handle(), _)`).

2. Otherwise, use `std::copy` with the same ranges.

## different-layout-known-copy

Use `cudaMemcpy*Async` (where `*` is `2D` or `3D`).

## generic-flattened-copy

Create a `layout_left::mapping` with the same extents as `src` and `dst`.  Invert it using divmods.  The composition of this inverse with the layout mapping of `src` gives a flat (1-D) range for `src`.  Ditto for `dst`.  Apply the same technique as in same-noncontiguous-copy.

cpp/include/raft/core/copy.cuh
```diff
@@ -0,0 +1,69 @@
...
+ * @tparam SrcType An mdspan type for the source container
+ * @param res raft::resources used to provide a stream for copies involving the
+ * device.
+ * @param dst The destination mdspan.
+ * @param src The source mdspan.
+ */
+template <typename DstType, typename SrcType>
+detail::mdspan_copyable_with_kernel_t<DstType, SrcType> copy(resources const& res,
+                                                             DstType&& dst,
+                                                             SrcType const& src)
```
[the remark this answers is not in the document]
wphicks: They're already constrained to be mdspans, but yeah, passing by value is fine here.

cpp/include/raft/core/copy.cuh
```diff
@@ -0,0 +1,69 @@
...
+{
+  detail::copy(res, dst, src);
```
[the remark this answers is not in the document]
wphicks: Whoops! Quite right, thank you.

cpp/include/raft/core/detail/copy.hpp
```diff
@@ -0,0 +1,510 @@
...
+  for (auto i = typename MdspanType::extents_type::rank_type{}; i < md.rank(); ++i) {
+    increment += index_strides[i] * old_indices[i];
+  }
+
+#pragma unroll
+  for (auto i = typename MdspanType::extents_type::rank_type{}; i < md.rank(); ++i) {
+    // Iterate through dimensions in order from slowest to fastest varying
+    auto const real_index = [](auto ind) {
+      if constexpr (std::is_same_v<typename MdspanType::layout_type, layout_f_contiguous>) {
+        return MdspanType::rank() - ind - 1;
```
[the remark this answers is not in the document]
wphicks: The function works just fine for all layouts; the comment just doesn't apply to anything other than `layout_left` and `layout_right`. I'll update accordingly.

cpp/include/raft/core/detail/copy.hpp
```diff
@@ -0,0 +1,510 @@
...
+#else
+    // Not possible to reach this due to enable_ifs. Included for safety.
+    throw(raft::non_cuda_build_error("Copying to from or on device in non-CUDA build"));
+#endif
+  } else if constexpr (config::can_use_cublas) {
+#ifndef RAFT_DISABLE_CUDA
+    auto constexpr const alpha = typename std::remove_reference_t<DstType>::value_type{1};
+    auto constexpr const beta  = typename std::remove_reference_t<DstType>::value_type{0};
+    if constexpr (std::is_same_v<typename config::dst_layout_type, layout_c_contiguous>) {
+      CUBLAS_TRY(linalg::detail::cublasgeam(resource::get_cublas_handle(res),
```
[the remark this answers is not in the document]
wphicks: We need cublas to actually perform the transpose. Note that this code path is only taken if a transpose is required and cublas can do it.

cpp/include/raft/core/detail/copy.hpp
```diff
@@ -0,0 +1,510 @@
...
+  auto static constexpr const use_intermediate_src =
+    std::conjunction_v<std::bool_constant<requires_intermediate>,
+                       std::bool_constant<!use_intermediate_dst>>;
+  auto static constexpr const can_use_device =
+    std::conjunction_v<std::bool_constant<CUDA_ENABLED>,
+                       std::disjunction<std::bool_constant<both_device_accessible>,
+                                        std::bool_constant<requires_intermediate>,
+                                        std::bool_constant<can_use_raft_copy>>>;
+
+  auto static constexpr const can_use_cublas =
```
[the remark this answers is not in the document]
wphicks: Neither of those can perform a transpose, can they?

cpp/include/raft/core/detail/copy.hpp
wphicks: > If the destination has a nonunique layout, then forbid copying
Yep, I think that makes sense. That's a fine check to add to the `mdspan_copyable` helper.

> If it's valid to use memcpy (or cudaMemcpy, etc.) for the two accessors, then use it.
I believe this is exactly what we do already.

> If it's valid to use a parallel algorithm (or device kernel) with the data, then use generic flat copy from [0, src.mapping().required_span_size()) (transformed through src.accessor().access(src.data_handle(), _)) to [0, dst.mapping().required_span_size()) (transformed analogously through dst.accessor().access(dst.data_handle(), _)).

Yep, I thought about this approach, and I don't mind it, but it increases complexity a bit and will _sometimes_ decrease performance. Let me try it out with the benchmarks and just see what domains it actually provides benefit for and how much.

> Use cudaMemcpy*Async (where * is 2D or 3D).
I'm probably missing something, but I don't see how to do transposes with this.

> Create a layout_left::mapping with the same extents as src and dst. Invert it using divmods. The composition of this inverse with the layout mapping of src gives a flat (1-D) range for src. Ditto for dst. Apply the same technique as in same-noncontiguous-copy.

This is not guaranteed to iterate on elements in memory-order, and the current implementation is designed to limit the use of divmods for device code for performance reasons.

cpp/include/raft/core/detail/copy.hpp
```diff
@@ -0,0 +1,510 @@
...
+template <typename DstType, typename SrcType>
+__global__ mdspan_copyable_with_kernel_t<DstType, SrcType> mdspan_copy_kernel(DstType dst,
+                                                                              SrcType src)
+{
+  using config = mdspan_copyable<true, DstType, SrcType>;
+
+  // An intermediate storage location for the data to be copied.
+  __shared__ typename config::dst_value_type tile[mdspan_copy_tile_dim][mdspan_copy_tile_dim + 1];
+
+  // Compute the cumulative product of extents in order from fastest to
```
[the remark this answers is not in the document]
wphicks: Yep! That general approach is how we would iterate in something akin to numpy's ravel. As a matter of fact, I believe a C++17 version of that is pretty close to how RAFT's current ravel works. The goal of this is to (where possible) read from the source in memory-layout order and write to the destination in memory-layout order. This us allows us to perform a coalesced read from global memory and coalesced write to global memory.

We won't always be able to get coalesced reads/writes with the present approach, but we will in the most common cases used by RAFT. Sometimes it will not be possible to get coalesced reads/writes at all, and then this approach will at least be correct even if it is slow.

cpp/include/raft/core/copy.cuh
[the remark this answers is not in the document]
mhoemmen: It's perhaps more important that the two views be handled consistently.  If you want to use forwarding references, that's fine, but then I think both should be forwarding references.  Otherwise, both should be values.

cpp/include/raft/core/detail/copy.hpp
mhoemmen: > This is not guaranteed to iterate on elements in memory-order, and the current implementation is designed to limit the use of divmods for device code for performance reasons.

Right -- this is not the ideal case.  It's a reasonable fall-back for something like `linalg::copy` in P1673.  RAFT controls the resource type, so it has the option to maintain storage for packing and unpacking.

CUTLASS has code that automatically decomposes copies into minimal sequences of hardware-specific copy instructions, if it knows enough about the layouts at compile time.  One could imagine adapting that to work better with run-time information, but it would be a lot of work.

wphicks: > Right -- this is not the ideal case. It's a reasonable fall-back

Yep, makes sense. That's why the current kernel does not try to guarantee maximal coalescing; it's just a best effort setup that will ensure coalescing on both sides (read/write) in the most common cases and on at least one side on the other.

cpp/include/raft/core/detail/copy.hpp
```diff
@@ -0,0 +1,510 @@
...
+  return apply(std::forward<LambdaT>(lambda), std::forward<ContainerT>(args),
+make_index_sequence<IdxT, size>{});
+} */
+
+/*
+ * Given an mdspan and an array of indices, return a reference to the
+ * indicated element.
+ */
+template <typename MdspanType, typename IdxType, IdxType... Idx>
+__device__ auto& get_mdspan_elem(MdspanType& md,
```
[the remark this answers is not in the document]
wphicks: 1. Good catch! Fixed accordingly.
2. Yep, agreed. Note to self: Check that that access gets inlined appropriately
3. Unfortunately, we don't have access to `std::span` in device code.

cpp/include/raft/core/detail/copy.hpp
```diff
@@ -458,6 +471,7 @@ mdspan_copyable_t<DstType, SrcType> copy(resources const& res, DstType&& dst, Sr
 #endif
   } else if constexpr (config::custom_kernel_allowed) {
 #ifdef __CUDACC__
+    config::check_for_unique_dst(dst);
```
wphicks, line 474: @mhoemmen How do we feel about this for the uniqueness check? It should only be necessary in this branch, since it's the only parallelized copy that is not already linked exclusively to layout_left/layout_right mdspans.

cpp/include/raft/core/detail/copy.hpp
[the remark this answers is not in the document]
mhoemmen: I don't think so, but they work for all kinds of data types other than `float` and `double`.

cpp/include/raft/core/detail/copy.hpp
mhoemmen: @wphicks Thanks for asking! : - )  It might be correct given the current state of the code, but I could see future changes breaking this.  For example, some developer might later want to add support for strided layouts (i.e., layouts for which `is_strided()` is `true`) other than `layout_stride`.  `layout_stride` has preconditions that no stride can be zero, but strided layouts in general do not have this precondition.  This means that a strided layout could be nonunique (because it has a zero stride, but a nonzero extent). 

In my view, it would be safer to move this precondition check to the top of the `copy` function.  `is_unique()` is meant to be cheap (that's why it's conservative), and the check is compiled out by the `if constexpr (! m.is_always_unique())` call.  It also makes the code more self-documenting to check all the preconditions at the top of the function.

wphicks: My thought in narrowing the scope of the check was to avoid artificially limiting the cases in which the copy can fail. I agree that the performance impact won't be significant.

Right now the code is structured around determining what method can be used to perform the copy. If we determine that a particular non-unique layout can be safely copied to with e.g. std::copy, I'd rather not prevent that up front. A fine example of this would be copying underlying memory of between identical non-unique layouts. That's safe even if the destination is non-unique.

If a user copies an mdspan from a layout which contains unique values for destination locations which are non-unique, I'd say that's user error and fine to leave as undefined behavior (though we should probably explicitly call that out in documentation).

Our remaining responsibility, then is to ensure that well-defined copies to non-unique destinations are correct. For parallel copies, that is not a simple task, so I think it makes sense to just fail for now. The only two parallel copy methods we have are cuBlas and the custom kernel. cuBlas can inherently only be invoked for unique layouts, so we are left with failing on the custom kernel. If a future developer adds additional parallel techniques, it's up to them to determine whether or not they can handle well-defined copies to non-unique layouts.

cpp/include/raft/core/detail/copy.hpp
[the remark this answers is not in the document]
wphicks: We already use cudaMemcpyAsync for any situation where the layout and data type do not need to change, so I think we have this covered. Other data types are handled by the custom kernel.

cpp/include/raft/core/detail/copy.hpp
wphicks: Just added a warning about ill-defined copies to non-unique layouts to the docs.

mhoemmen: I'll just warn that most developers may not remember that non-unique layouts can exist : - ) .

cpp/include/raft/core/detail/copy.hpp
```diff
@@ -0,0 +1,524 @@
...
+template <bool B, typename DstType = void, typename SrcType = void, typename T = void>
+struct mdspan_copyable {};
+
+/*
+ * A helper struct used to determine whether one mdspan type can be copied to
+ * another and if so how
+ */
+template <typename DstType, typename SrcType, typename T>
+struct mdspan_copyable<true, DstType, SrcType, T> {
+  using dst_type = std::remove_reference_t<DstType>;
```
divyegala, line 51: Can we add `static_assert` here to ensure the types are indeed mdspans, and if not we pass a clear error message to the user? Otherwise there will be a bunch of error messages that will be hard to parse

cpp/include/raft/core/detail/copy.hpp
```diff
@@ -0,0 +1,524 @@
...
+  using index_type =
+    std::conditional_t<(std::numeric_limits<typename dst_extents_type::index_type>::max() >
+                        std::numeric_limits<typename src_extents_type::index_type>::max()),
+                       typename dst_extents_type::index_type,
+                       typename src_extents_type::index_type>;
+
+  // Dtype properties
+  using dst_value_type                   = typename dst_type::value_type;
+  using src_value_type                   = typename src_type::value_type;
+  using dst_element_type                 = typename dst_type::element_type;
```
divyegala, line 66: Maybe another `static_assert` here that `dst_type::element_type` is non-const?

cpp/include/raft/core/detail/copy.hpp
```diff
@@ -0,0 +1,524 @@
...
+                       std::bool_constant<compatible_dtype>,
+                       std::bool_constant<both_contiguous>,
+                       std::bool_constant<same_underlying_layout>>;
+  auto static constexpr const can_use_raft_copy =
+    std::conjunction_v<std::bool_constant<CUDA_ENABLED>,
+                       std::bool_constant<same_dtype>,
+                       std::bool_constant<both_contiguous>,
+                       std::bool_constant<same_underlying_layout>>;
+
+  auto static constexpr const requires_intermediate =
```
divyegala, line 155: Why is an intermediate needed? I am assuming this is the managed memory case

cpp/include/raft/core/detail/copy.hpp
```diff
@@ -0,0 +1,524 @@
...
+  auto static constexpr const custom_kernel_required =
+    std::conjunction_v<std::bool_constant<!can_use_host>,
+                       std::bool_constant<!(can_use_raft_copy || can_use_cublas)>>;
+
+  // Viable overload?
+  auto static constexpr const value =
+    std::conjunction_v<std::bool_constant<is_mdspan_v<src_type>>,
+                       std::bool_constant<is_mdspan_v<dst_type>>,
+                       std::bool_constant<can_use_host || can_use_device>>;
+  using type = std::enable_if_t<value, T>;
```
divyegala, line 192: Is there a scenario where this is not `void`?

divyegala (commented):
First round - reviewing `mdspan_copyable`

cpp/include/raft/core/detail/copy.hpp
wphicks: This is for the case where we have e.g. dst on device and src on hst, _and also_ we cannot just perform a trivial cudaMemcpyAsync. In that case, we need an intermediate device buffer to store the results of the transpose or dtype conversion before we copy back to host or a to store a straight copy from the host before we perform the transpose or dtype conversion.

cpp/include/raft/core/detail/copy.hpp
wphicks: There will be with mdbuffer. In general, this is a convenience just to parallel the behavior of `enable_if_t`, but just like the extra type argument is rarely used with `enable_if_t`, I expect it to be rarely used here as well.

cpp/include/raft/core/detail/copy.hpp
wphicks: Ah, good point! We should really use `is_assignable` rather than `is_convertible` for the compatibility check. That should take care of the const issue.

cpp/include/raft/core/detail/copy.hpp
wphicks: In general, I'm not too concerned about this. `raft::detail::mdspan_copyable` is an implementation detail of `raft::copy` and should never be consumed outside of the context of `raft::copy`. That will be SFINAE'd out for any instantiation that does not take mdspan arguments. If we want to add a `static_assert` for debugging `mdspan_copyable` itself, I'm not opposed, but I don't think it adds anything to the actual safety of the implementation. The compilation error for consumers will still be a SFINAE failure.

cpp/include/raft/core/detail/copy.hpp
wphicks: Added comment explaining this to code.

cpp/include/raft/core/detail/copy.hpp
divyegala: Consider the scenario where I called `raft::copy(p)` where `p` is of type `int*`. Even though `mdspan_copyable` is an implementation detail, the user will get a bunch of errors from all the code internally that tries to access mdspan internals such as `mdspan::layout_type`, etc. `static_assert` will help us keep the error tree clean.

cpp/include/raft/core/stream_view.hpp
```diff
@@ -0,0 +1,108 @@
...
+#ifndef RAFT_DISABLE_CUDA
+    return rmm::cuda_stream_per_thread;
+#else
+    auto static constexpr const default_fail_stream = underlying_view_type{};
+    return default_fail_stream;
+#endif
+  }
+};
+
+auto static const stream_view_per_thread = stream_view{};
```
divyegala, line 106: I'm not quite sure, but should this be `thread_local` instead of `static`?

cpp/test/core/mdspan_copy.cpp
```diff
@@ -0,0 +1,275 @@
...
+#include <raft/core/copy.hpp>
+#ifndef RAFT_DISABLE_CUDA
+#include <raft/core/device_mdarray.hpp>
+#include <raft/core/device_resources.hpp>
+#endif
+#include <raft/core/host_mdarray.hpp>
+#include <raft/core/resources.hpp>
+
+namespace raft {
+TEST(MDSpanCopy, Mdspan1DHostHost)
```
divyegala, line 29: Can you add another test here where you explicitly construct `detail::mdspan_copyable` and for all possible input type combinations, you ensure the right internal booleans that denote which policy to use for copying are being set?

cpp/include/raft/core/stream_view.hpp
divyegala, line 1: I'm not sure if this is the right directory for this. I think the contents of this file can go in `raft/resource/stream_view.hpp`. While we are at that, this is basically a replacement for `raft/resource/cuda_stream.hpp` and I would be open to replacing that entirely in this PR or a follow-up.

cpp/include/raft/core/detail/copy.hpp
mhoemmen: Here's a sketch of an overload and error-checking pattern.

```c++
// Outer interface
template<class InView, class OutView>
void copy(InView in, OutView out) {
  // Check types of in and out individually, and transform each into a "canonical" type.
  // For example, if InView is mdspan<T, ...>, transform to `mdspan<add_const_t<T>, ...>`.
  auto in_canon = detail::canonicalize(in);
  auto out_canon = detail::canonicalize(out);

  // Check that decltype(in_canon) can be copied to decltype(out_canon).
  // Each type has already been checked and canonicalized.
  // For example, we know that out_canon is an mdspan, span,
  // or raw pointer to nonconst.  detail::copyable would check,
  // e.g., that two mdspans have the same ranks and that any
  // extents known at compile time match.
  //
  // We never call detail::copyable for noncanonical types,
  // so it doesn't need to check the input types individually,
  // and it only needs overloads for canonical types.
  static_assert(detail::copyable(in_canon, out_canon));

  // Any run-time precondition checks, e.g., for run-time extents.
  assert(detail::runtime_copyable(in_canon, out_canon));

  // At this point, we know that in_canon can be copied to out_canon.
  // We will never call detail::copy for any types other than
  // individually canonical and mutually copyable types.
  // Thus, we don't need to check errors at compile time.
  // We've also checked run-time preconditions above,
  // so we don't need to check them inside detail::copy.
  // detail::copy can just Do The Copying.
  detail::copy(in_canon, out_canon);
}
```

divyegala: I really like @mhoemmen's approach, it would simplify `detail::mdspan_copyable` by splitting it into multiple type-checkers.

wphicks: Unfortunately, that solves a different (albeit related) problem than `mdspan_copyable` is intended to solve. To give a bit more insight, let me run you through exactly why we need `mdspan_copyable` and how we got to the current design.

First off, we need a way to SFINAE-separate the instantiations that require nvcc from those that down. That's what allows us to correctly construct the public cuh and hpp files, which is important both for enabling usage in non-CUDA builds and for speeding up compilation in CUDA-enabled builds.

Already, that requires all of the following checks:

- Are we in a CUDA-enabled build?
- Are both mdspans device-accessible? If not, is one _only_ device-accessible and the other _only_ host-accessible?
- Are both views on contiguous memory?
- Do both have the same layout? If not, are they actually equivalent layouts spelled differently (e.g. 1D layout_left vs 1D layout_right)?
- Do they both have the same `value_type`?
- If they are not the same layout, are we copying a 2D layout_left to a 2D layout_right?
- If so, are both views on float or both views on doubles?

We need all that information just to correctly set up that top-level SFINAE, because that is what is required to answer "Does this copy involve the device but cannot be performed with cudaMemcpyAsync or cublas?"

Once we've SFINAE-separated the instantiations that require nvcc from those that don't, we still need to figure out exactly which copy implementation we're actually going to use. For deciding if we're going to use cudaMemcpyAsync or cublas, we need to answer all the above questions again, so we have a few possible implementations:

**Possibility 1:** SFINAE-separate each of the `copy` implementations based on the method used.

This path substantially increases the complexity of the code, but it also _massively_ increases compile time. Moving from a single outermost SFINAE to a whole tree of SFINAEs requires the compiler to evaluate a whole bunch more possible instantiations, each with their own (expensive) SFINAE condition. When I tried this, it took so long to compile even with trivial function bodies that I immediately started looking for another path.

**Possibility 2:** Perform all necessary checks for the outermost SFINAE, and then re-evaluate the necessary information in each constexpr if condition.
This makes the code significantly less readable and also increases compile time somewhat. I had assumed the compiler would cleverly handle the redundant checks, but given the actual compile times I was seeing, that does not seem to be the case.

**Possibility 3:** Separate `mdspan_copyable` into several separate objects and `std::conjunction/std::disjunction` them together.
This had a lot of appeal for me, and I tried several different ways to get it to work as desired. It's the equivalent of breaking a complex function down into many smaller well-named functions. Unfortunately, for the version that gets semi-decent compile time, we actually hit the same problem that @divyegala noticed right at the top of this thread over and over again in separate objects, each of which have to be SFINAE-protected. This significantly increases complexity, requires some redundant checks, and ultimately gives mediocre compile time.

**Final version:** What I ultimately landed on was a single object with clear, well-named type aliases and static constexpr values which could both be used for the necessary SFINAEs and re-used in the constexpr if conditions. From the compiler's perspective, the nice thing about this is that it's easy to re-use that template instantiation everywhere it's needed in the translation unit, and we perform all necessary checks for a specific instantiation exactly once. From the developer's perspective, though there's a lot going on in this object, we can follow the logic through the naming of the constexpr values. E.g. We see that the original `raft::copy` (equivalent to cudaMemcpyAsync) is used when:

1. CUDA is enabled for the build overall
2. The mdspans contain the same dtype
3. Both mdspans are contiguous
4. Both mdspans have the same underlying layout

Furthermore, we just need one top-level SFINAE check to protect everything correctly and avoid weird, intractable compiler errors. As @divyegala pointed out, I did miss something in that check, but that has been corrected in the version I just pushed. Now any `raft::copy` instantiation that we can't support gives a nice, familiar, and succinct SFINAE failure from the compiler.

If there's some other way to structure this that is more readable without messing with compile time, I'd be open to a PR to my branch, but I've already tried a bunch of different approaches, so I have no confidence I could come up with something better myself. I'd be even happier to see this as a follow-up PR to clean up implementation details if we're generally happy with the runtime, compile time, and public API. We'll need a follow-up to reach SoL copies for all cases anyway, so this can continue to evolve as we use it.

cpp/include/raft/core/stream_view.hpp
wphicks: I had exactly the same question when I was writing it! I had to go into the [rmm source](https://github.com/rapidsai/rmm/blob/branch-23.10/include/rmm/cuda_stream_view.hpp#L119) to confirm, but this is actually correct.

cpp/test/core/mdspan_copy.cpp
wphicks: I'd prefer not to include implementation details in unit tests. Ideally, the tests should ensure correctness for every scenario that can be handled by the public API, and if we ultimately are taking the wrong path (but still giving correct results), then that's a benchmarking question. That is to say, so long as we have the correct results, all we care about is whether the copy is as fast as it should be and not using up system resources unnecessarily.

cpp/include/raft/core/detail/copy.hpp
divyegala: Thanks for the answer @wphicks that details everything you tried and why you arrived at this design. Your reasons for your current approach seem more than satisfactory to me now.

Could you show the error stacktrace that the compiler throws when we try to call this function with a non-mdspan object or with a const destination?

Pending what that would look like and response to a couple of other open questions, I think we are very close to finalizing this PR.

cpp/include/raft/core/stream_view.hpp
wphicks: I'm assuming you mean `raft/core/resource/stream_view.hpp`? My thought in putting it here was that this defines the actual object, not the resource. That's particularly useful to have as a type in a self-contained header for use in other CUDA-free builds (looking at you, FIL backend...).

That being said, I'd be 100% on-board with having `cuda_stream_resource` return an object of this type if everyone's okay with it. I was trying not to make any changes that would have such a far-reaching impact on the code base in this PR, but I really like this notion. It would be a big step toward setting up CUDA-free builds of other stuff that we want to make available in CUDA-free environments.

My vote would be to leave that as a follow-up, since I suspect other people will have opinions on exactly how we do that, but I'm open to other opinions. The other reason to leave it for a follow up is that we'd probably want to add `noexcept` flags correctly depending on the build type, but that shouldn't take too much thought/effort.

cpp/include/raft/core/detail/copy.hpp
wphicks: Thanks very much! Really appreciate your help in thinking through this so carefully. Here's a gist where we attempt a copy between two ints: https://gist.github.com/wphicks/668681c85bf59c1d565bcaa0f20ed74b, and here is a gist where we attempt to copy into a mdspan of consts: https://gist.github.com/wphicks/2c791336955a33f9f74dbb50629e2fd4

Also useful to note is that my static analyzer correctly recognizes that these are not viable overloads (unsurprising but still very satisfying!) and offers a nice warning before I even attempt to compile.

Does all that seem okay to you? @divyegala, feel free to hit the "Resolve conversation" button if you're satisfied; I don't want to cut off discussion prematurely.

divyegala: The error gists look great! Thank you for these, this is looking really good.

cpp/include/raft/core/stream_view.hpp
divyegala: I agree, let's leave it as a follow-up item. Can you draft an issue for follow-up tasks? For this PR, I'd still like to see it in `raft/core/resource/stream_view.hpp` and we can take it from there

cpp/test/core/mdspan_copy.cpp
divyegala: I think I disagree here that arriving at the correct result using the wrong path is a correct test. I would very much like to see a unit test that checks the internals of `mdspan_copyable` as it's not only a way to ensure type correctness, but also helps any person who modifies this code figure out that they did not change an internal that leads them to take the wrong path by mistake and spend time on benchmarks figuring out why perf suddenly degraded.

A test with clear compile-time asserts will make changes easier to iterate on for any future dev working on this.

cpp/include/raft/core/stream_view.hpp
wphicks: Sorry, just want to make sure we're on the same page in terms of location. My understanding is that this object should definitely _not_ go in `raft/core/resource/stream_view.hpp` because it's not itself a resource but an object returned by a resource. It's also not strictly coupled to its use as a resource but usable as an independent object. In e.g. the FIL backend, that's exactly how it will (and must) be used, so it would be preferably not to transitively include all of the resource machinery.

Is the argument that it _should_ actually be strictly coupled to its use with the resource? Or is there some other consideration here that I'm missing?

divyegala: I was suggesting originally that it be strictly coupled with its resource machinery since those are essentially pretty lightweight wrappers with no other dependencies. But in general, I could be swayed either way here. In terms of RAFT types, what's an object without the mirroring resource?

But then, if you are going to use the object by itself in the FIL backend then that changes the meaning. Let's keep it in `raft/core/stream_view.hpp` for now.

divyegala (approved):

wphicks:
CI failing for same reason as other 23.12 PRs. I believe this should be unblocked after #1868 goes through and gets merged back into this PR.

cpp/include/raft/core/detail/copy.hpp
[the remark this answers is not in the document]
mhoemmen: @wphicks wrote:
> Unfortunately, we don't have access to std::span in device code.

What about `cuda::std::span`?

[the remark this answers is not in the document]
divyegala: @mhoemmen unrelated to this PR and out of curiosity, how would you go about adding a proxy reference to `mdspan`?

PR #1821, opened 2023-09-13
matrix::select_k: extra tests and benchmarks

achirkin:
Add a few extra test and benchmark cases; in particular:
  1. Allow specifying non-trivial input indices
  2. Allow filling the input data with infinities to see how algorithms perform in edge cases

These tests are borrowed from the controversial workaround https://github.com/rapidsai/raft/pull/1742

### DIFF
--- a/cpp/test/matrix/select_k.cuh
+++ b/cpp/test/matrix/select_k.cuh
@@ -49,14 +49,16 @@ auto gen_simple_ids(uint32_t batch_size, uint32_t len) -> std::vector<IdxT>
 template <typename KeyT, typename IdxT>
 struct io_simple {
  public:
-  bool not_supported = false;
+  bool not_supported               = false;
+  std::optional<select::Algo> algo = std::nullopt;
 
   io_simple(const select::params& spec,
             const std::vector<KeyT>& in_dists,
+            const std::optional<std::vector<IdxT>>& in_ids,
             const std::vector<KeyT>& out_dists,
             const std::vector<IdxT>& out_ids)
     : in_dists_(in_dists),
-      in_ids_(gen_simple_ids<IdxT>(spec.batch_size, spec.len)),
+      in_ids_(in_ids.value_or(gen_simple_ids<IdxT>(spec.batch_size, spec.len))),
       out_dists_(out_dists),
       out_ids_(out_ids)
   {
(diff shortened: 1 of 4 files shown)


cpp/test/matrix/select_k.cuh
```diff
@@ -223,32 +227,62 @@ struct SelectK  // NOLINT
...
       return false;
     };
     ASSERT_TRUE(hostVecMatch(ref.get_out_ids(), res.get_out_ids(), compare_ids));
   }
+
+  auto forgive_algo(const std::optional<select::Algo>& algo, IdxT ix) const -> bool
+  {
+    if (!algo.has_value()) { return false; }
+    switch (algo.value()) {
+      // not sure which algo this is.
```
tfeher, line 262: What does this mean: "not clear which one is selected because it might depend on input parameters"? or something else?

cpp/test/matrix/select_k.cuh
```diff
@@ -223,32 +227,62 @@ struct SelectK  // NOLINT
...
+    // due to non-deterministic nature of some implementations.
+    auto compare_ids = [this](const IdxT& i, const IdxT& j) {
       if (i == j) return true;
+      auto& in_ids   = ref.get_in_ids();
+      auto& in_dists = ref.get_in_dists();
       auto ix_i = static_cast<int64_t>(std::find(in_ids.begin(), in_ids.end(), i) - in_ids.begin());
       auto ix_j = static_cast<int64_t>(std::find(in_ids.begin(), in_ids.end(), j) - in_ids.begin());
-      if (static_cast<size_t>(ix_i) >= in_ids.size() || static_cast<size_t>(ix_j) >= in_ids.size())
-        return false;
+      auto forgive_i = forgive_algo(ref.algo, i);
```
tfeher, line 238: Could you add a comment on the forgiveness logic here?

tfeher (changes requested):
Thanks Artem for improving the tests. Looks good, just two small questions.

cpp/test/matrix/select_k.cuh
achirkin: I explain it in more detail in an issue: https://github.com/rapidsai/raft/issues/1822

cpp/test/matrix/select_k.cuh
achirkin: Yes, we don't have access to the algo selection logic here, thus we don't know what behavior to expect.

cpp/test/matrix/select_k.cuh
tfeher: Thanks the issue helps, would be great to link in the comment below.

tfeher (approved):
Thanks @achirkin for answering my questions. LGTM.

cpp/test/matrix/select_k.cuh
```diff
@@ -223,32 +227,62 @@ struct SelectK  // NOLINT
...
+      auto forgive_j = forgive_algo(res.algo, j);
+      // Some algorithms return invalid indices in special cases.
+      // This can be considered as TODO for us to fix.
```
achirkin, line 241:
```suggestion
      // TODO: https://github.com/rapidsai/raft/issues/1822
```

cpp/test/matrix/select_k.cuh
achirkin: Thanks, good point, I added the link.

PR #1825, opened 2023-09-13
Attempt removing dependency on conda-forge channel.

bdice:
Experimental PR. Do not merge.


bdice:
This is blocked for now on a few things:
- some missing packages (`cuda-version` is currently only on `nvidia/label/tst` and is missing 12.0)
- the `nvidia` channel compiler package doesn't play nicely with CMake and needs additional flags (we identified this earlier while working on CUDA 12 support, and can probably copy some of those fixes in for testing purposes)

PR #1828, opened 2023-09-18
Add IVF-Flat C++ example

tfeher:
This PR adds a C++ example program that demonstrate the usage of IVF-Flat vector search.


cpp/template/src/ivf_flat_example.cu
```diff
@@ -0,0 +1,173 @@
...
+  // Create output arrays.
+  int64_t topk      = 10;
+  int64_t n_queries = queries.extent(0);
+  auto neighbors    = raft::make_device_matrix<int64_t, int64_t>(dev_resources, n_queries, topk);
+  auto distances    = raft::make_device_matrix<float, int64_t>(dev_resources, n_queries, topk);
+
+  // Search K nearest neighbors for each queries.
+  ivf_flat::search(
+    dev_resources, search_params, index, queries, neighbors.view(), distances.view());
+}
```
achirkin, line 140: Would it make sense to return the `make_tuple(neighbors, distances)` here or just print to `std::cout` a few neighbors?
This would also be educational to the viewer in that we'd have to call `dev_resources.sync_stream()` before accessing the results from the host.

tfeher: Good point about highlighting `sync_stream`! Added together with a print function that list neighbors of first queries.

achirkin (approved):
LGTM!

ajschmidt8:
> /ok to test

For what it's worth, the first `/ok to test` comment above should have worked.

I checked the application logs and the GitHub API request was sent successfully and received a 200 response, but for whatever reason the https://github.com/rapidsai/raft/tree/pull-request/1828 branch was not actually updated.

Looks like this was a fluke on GitHub's end.

cpp/template/CMakeLists.txt
```diff
@@ -36,3 +36,8 @@ include(cmake/thirdparty/get_raft.cmake)
 # -------------- compile tasks ----------------- #
 add_executable(TEST_RAFT src/test_vector_search.cu)
 target_link_libraries(TEST_RAFT PRIVATE raft::raft raft::compiled)
+target_compile_definitions(TEST_RAFT PRIVATE "RAFT_EXPLICIT_INSTANTIATE_ONLY")
+
+add_executable(IVF_FLAT_EXAMPLE src/ivf_flat_example.cu)
+target_link_libraries(IVF_FLAT_EXAMPLE PRIVATE raft::raft raft::compiled)
+target_compile_definitions(IVF_FLAT_EXAMPLE PRIVATE "RAFT_EXPLICIT_INSTANTIATE_ONLY")
```
cjnolet, line 43: I think we should pass an option back to the main build target to set this value on `raft::raft` target, which should get passed through transitively. It's an extra step that is, quite frankly, not self explanatory and confusing to see. Users should just be able to include their `get_raft.cmake` and add the needed raft targets.

cpp/template/src/ivf_flat_example.cu
```diff
@@ -0,0 +1,208 @@
...
+#include <raft/random/make_blobs.cuh>
+#include <raft/random/sample_without_replacement.cuh>
+#include <raft/util/cudart_utils.hpp>
+
+#include <thrust/copy.h>
+#include <thrust/device_ptr.h>
+#include <thrust/iterator/counting_iterator.h>
+
+// Copy the results to host and print a few samples
+void print_results(raft::device_resources const& dev_resources,
```
cjnolet, line 33: This is super useful and I think we could use it in the CAGRA example as well, could we pull this into a common header?

cpp/template/src/ivf_flat_example.cu
```diff
@@ -0,0 +1,208 @@
...
+  raft::random::sample_without_replacement(
+    dev_resources, rng, data_indices, std::nullopt, train_indices.view(), std::nullopt);
+
+  raft::matrix::copy_rows(
+    dev_resources, dataset, trainset.view(), raft::make_const_mdspan(train_indices.view()));
+
+  return trainset;
+}
+
+void ivf_flat_build_extend_search(raft::device_resources const& dev_resources,
```
cjnolet, line 120: Super simple but powerful change- could we copy this into the CAGRA example as well and use this pattern for a quick CAGRA demo?

cpp/template/src/ivf_flat_example.cu
```diff
@@ -0,0 +1,208 @@
...
+
+  // The call to ivf_flat::search is asynchronous. Before accessing the data, sync using:
+  // raft::resource::sync_stream(dev_resources);
+
+  print_results(dev_resources, neighbors.view(), distances.view());
+}
+
+int main()
+{
+  raft::device_resources dev_resources;
```
cjnolet, line 179: Same here, can we copy this into the CAGRA example?

cpp/template/CMakeLists.txt
```diff
@@ -36,3 +36,8 @@ include(cmake/thirdparty/get_raft.cmake)
 # -------------- compile tasks ----------------- #
 add_executable(TEST_RAFT src/test_vector_search.cu)
```
cjnolet, line 37: Can we change this to `CAGRA_EXAMPLE`? It's no longer really a RAFT test and after consolidating with the ivf-flat example, it'll be consistent.

cpp/template/src/ivf_flat_example.cu
```diff
@@ -0,0 +1,208 @@
...
+    dev_resources, search_params, index, queries, neighbors.view(), distances.view());
+
+  // The call to ivf_flat::search is asynchronous. Before accessing the data, sync by calling
+  // raft::resource::sync_stream(dev_resources);
+
+  print_results(dev_resources, neighbors.view(), distances.view());
+}
+
+/** Subsample the dataset to create a training set*/
+raft::device_matrix<float, int64_t> subsample(
```
cjnolet, line 96: This could probably be added to the common header as well, since it could be used for a future ivf-pq example (and others).

cpp/template/src/ivf_flat_example.cu
```diff
@@ -0,0 +1,208 @@
...
+void ivf_flat_build_extend_search(raft::device_resources const& dev_resources,
+                                  raft::device_matrix_view<const float, int64_t> dataset,
+                                  raft::device_matrix_view<const float, int64_t> queries)
+{
+  using namespace raft::neighbors;
+
+  // Define dataset indices.
+  auto data_indices = raft::make_device_vector<int64_t, int64_t>(dev_resources, dataset.extent(0));
+  thrust::counting_iterator<int64_t> first(0);
+  thrust::device_ptr<int64_t> ptr(data_indices.data_handle());
```
cjnolet, line 129: Does this get used anywhere?

cpp/template/src/ivf_flat_example.cu
```diff
@@ -0,0 +1,208 @@
...
+
+  print_results(dev_resources, neighbors.view(), distances.view());
+}
+
+int main()
+{
+  raft::device_resources dev_resources;
+
+  // Set pool allocator with 2 GiB limit for temporary arrays.
+  raft::resource::set_workspace_to_pool_resource(dev_resources, 2 * 1024 * 1024 * 1024ull);
```
cjnolet, line 182: Can we please just set the default rmm allocator to a pool? I'd like to avoid carving memory resources up like this because it's very nebulus to users and they don't know how much is enough.

cjnolet (changes requested):
Really happy to have these examples for users to reference. Mostly minor things, but want to make sure we're keeping the examples consistent (converting CAGRA to your example format shouldn't take much effort, AFAIK).

cpp/template/src/ivf_flat_example.cu
tfeher: Yes, added to `common.cuh`

cpp/template/src/ivf_flat_example.cu
tfeher: added.

cpp/template/src/ivf_flat_example.cu
tfeher: Changed the cagra example to follow the same pattern.

cpp/template/src/ivf_flat_example.cu
tfeher: In the next line we will the data_indices using this ptr.

cpp/template/src/ivf_flat_example.cu
tfeher: I have changed the CAGRA example to follow the same logic as IVF_FLAT example. Let me know if I missed something.

cpp/template/src/ivf_flat_example.cu
tfeher: I understand your concern, and I have changed it. But advanced users should still know about how to set workspace resource, therefore I have kept the workspace allocator setting in a comment. Setting the workspace resource to a separate pool (with limited size) has a few advantages:
- less chance fragmentation
- the size limit allows us to adjust batch size to available workspace.

tfeher (commented):
Thanks @cjnolet for the good suggestions, I have updated the PR accordingly.

cpp/template/CMakeLists.txt
cjnolet: I don't mean passing it through implicitly by default, I mean that we should provide a cmake option that users can set so that the flag is set the raft::raft target by the raft cmake. It follows the standardized way that we enable such options.

tfeher: I would treat this separately from the current PR (especially that I do not know what is the right way to achieve this in cmake). I have created issue #1840. For the current PR, I have removed the flag (and comments) from CMakelist.txt.

cjnolet (approved):

cjnolet:
@tfeher given how we want to get this in quickly for the blog, I think we can do the cmake change to expose the explicit instantiation as a follow-on in 23.12.

PR #1832, opened 2023-09-19
Fixes for OOM during CAGRA benchmarks

benfred:
Running the CAGRA benchmarks and there could be OOM errors on GPU memory with large datasets. This is caused by holding multiple copies of the dataset in GPU memory. Fix by:

* Free existing memory for the dataset/graph before allocating new memory during update_dataset/update_grph
* On deserialize, if the serialized index doesn't contain the dataset - don't allocate GPU memory for it
* Don't call update_dataset repeatedly in the benchmarking code with the same dataset

### DIFF
--- a/cpp/bench/ann/src/common/benchmark.hpp
+++ b/cpp/bench/ann/src/common/benchmark.hpp
@@ -189,6 +189,7 @@ void bench_search(::benchmark::State& state,
   ANN<T>* algo;
   std::unique_ptr<typename ANN<T>::AnnSearchParam> search_param;
   try {
+    search_param = ann::create_search_param<T>(index.algo, sp_json);
     if (!current_algo || (algo = dynamic_cast<ANN<T>*>(current_algo.get())) == nullptr) {
       auto ualgo = ann::create_algo<T>(
         index.algo, dataset->distance(), dataset->dim(), index.build_param, index.dev_list);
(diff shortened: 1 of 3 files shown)


cpp/bench/ann/src/common/benchmark.hpp
```diff
@@ -189,14 +189,27 @@ void bench_search(::benchmark::State& state,
...
   try {
+    search_param = ann::create_search_param<T>(index.algo, sp_json);
     if (!current_algo || (algo = dynamic_cast<ANN<T>*>(current_algo.get())) == nullptr) {
       auto ualgo = ann::create_algo<T>(
         index.algo, dataset->distance(), dataset->dim(), index.build_param, index.dev_list);
       algo = ualgo.get();
       algo->load(index_file);
       current_algo = std::move(ualgo);
+
+      if (search_param->needs_dataset()) {
```
achirkin, line 200: It could happen that for the same algorithm in the config file, some run configurations need dataset, and some don't. A common use-case: IVF-PQ, where the first search configs don't use refinement, but the following search configs use it. `algo->set_search_dataset` being placed inside the condition that does not depend on the search params would never be called and IVF-PQ would get a `nullptr` instead of the dataset.

Note, the Dataset type caches the `base_set()` pointer anyway, so it won't allocate or move data twice. If there is some logic inside CAGRA benchmark wrapper that does any heavy/memory work in the `set_search_dataset` method, I'd suggest fixing that on the that side rather than modifying the benchmarking logic.

achirkin (changes requested):

cjnolet (approved):

benfred:
Created an issue to track the dataset copies here https://github.com/rapidsai/raft/issues/1848 -

cpp/bench/ann/src/common/benchmark.hpp
benfred: reverted this code, and created an issue here to track https://github.com/rapidsai/raft/issues/1848.

PR #1844, opened 2023-09-22
raft-ann-bench package fixes for plotting and conf files

dantegd:
Number of improvements and fixes for different million scale datasets that are supported by default by the python benchmarking package.

### DIFF
--- a/python/raft-ann-bench/src/raft-ann-bench/run/conf/nytimes-256-inner.json
+++ b/python/raft-ann-bench/src/raft-ann-bench/run/conf/nytimes-256-inner.json
@@ -1,9 +1,9 @@
-c{
+{
   "dataset": {
-    "name": "mnist-784-euclidean",
-    "base_file": "mnist-784-euclidean/base.fbin",
-    "query_file": "mnist-784-euclidean/query.fbin",
-    "groundtruth_neighbors_file": "mnist-784-euclidean/groundtruth.neighbors.ibin",
+    "name": "nytimes-256-inner",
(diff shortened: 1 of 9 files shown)


python/raft-ann-bench/src/raft-ann-bench/run/conf/nytimes-256-inner.json
```diff
@@ -0,0 +1,1352 @@
...
+        {"ef":40, "numThreads":1},
+        {"ef":60, "numThreads":1},
+        {"ef":80, "numThreads":1},
+        {"ef":120, "numThreads":1},
+        {"ef":200, "numThreads":1},
+        {"ef":400, "numThreads":1},
+        {"ef":600, "numThreads":1},
+        {"ef":800, "numThreads":1}
+      ],
+      "search_result_file" : "result/nytimes-256-inner/hnswlib/M12"
```
divyegala, line 32: This key is not needed anymore, it was necessary in the old benchmarks

dantegd: we still have it in a lot of files, perhaps I can do a follow up to remove it everywhere?

divyegala (approved):
Approved, pending successful local runs

python/raft-ann-bench/src/raft-ann-bench/run/conf/nytimes-256-inner.json
divyegala: Please!

PR #1861, opened 2023-09-28
[FEA] Helpers for identifying contiguous layouts.

trivialfis:
- Use strides to determine whether we can use a contiguous layout.

Useful for accepting inputs with stride information like numpy array or legate stores. The code is mostly copied and pasted from mdspan layout_left/right  constructors. These constructors can optionally accept a stride layout if the underlying memory is contiguous, otherwise throw an exception. We change the exception into a boolean return value.


cpp/include/raft/core/mdspan.hpp
```diff
@@ -312,6 +312,41 @@ RAFT_INLINE_FUNCTION auto unravel_index(Idx idx,
...
+
+/**
+ * @brief Whether the strides imply a c-contiguous layout.
+ */
+template <typename Extents, typename Strides>
+[[nodiscard]] auto is_c_contiguous(Extents const& extents, Strides const& strides) -> bool
+{
+  std::uint64_t stride = 1;
+  for (auto r = extents.rank(); r > 0; r--) {
+    if (stride != strides[r - 1]) { return false; }
```
divyegala, line 328: In your test case, in the first iteration isn't `stride = 1` and `strides[r - 1] = 8` where `r = 3`? I don't understand how this doesn't exit in the first iteration itself

cpp/test/core/mdspan_utils.cu
```diff
@@ -247,4 +247,20 @@ void test_const_mdspan()
 
 TEST(MDSpan, ConstMDSpan) { test_const_mdspan(); }
 
+void test_contiguous_predicates()
+{
+  extents<std::int64_t, dynamic_extent, dynamic_extent, dynamic_extent> exts{10, 10, 10};
+  {
+    std::array<std::size_t, 3> strides{128, 32, 8};
```
divyegala, line 254: Shouldn't this be `{800, 80, 8}`? I must be really misunderstanding something here

trivialfis:
@divyegala Apologies for the error, I got a bit confused now that I pick up the PR, how did I get the tests to pass (along with CI) ...

- I have added an additional test to ensure the same unit is used between mdarray and the predicates.
- Some changes in type consistency.

divyegala (approved):

PR #1869, opened 2023-10-03
[BUG] Fix a bug in NN descent

enp1s0:
I got errors when compiling a program using raft NN-descent. This PR fixes the bug.

## Error
e.g.
```
/home/.../include/raft/neighbors/detail/nn_descent.cuh(1158): error: invalid narrowing conversion from "unsigned long" to "int"
      h_rev_graph_old_{static_cast<size_t>(nrow_ * NUM_SAMPLES)},
```

- nvcc

```
Built on Tue_Aug_15_22:02:13_PDT_2023
Cuda compilation tools, release 12.2, V12.2.140
Build cuda_12.2.r12.2/compiler.33191640_0
```

- gcc
```
gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
```

- thrust : 2.2 (This is the cause of this error [[detail](https://github.com/rapidsai/raft/pull/1869#issuecomment-1746637944)])

# Change
Use `Class(...)` instead of `Class{...}`.

# Cause
The NN-descent code calls constructors of `thrust::host_vector` as shown below:
```cpp
graph_host_buffer_{static_cast<size_t>(nrow_ * DEGREE_ON_DEVICE)},
```
However, this constructor is regarded as a list initialization.
This is the same as the following code outputting 1 instead of 2.
```cpp
#include <iostream>
#include <vector>

int main() {
  std::vector<float> list{2};

  std::cout << list.size() << std::endl;
}
```

[detail](https://en.cppreference.com/w/cpp/language/list_initialization)


enp1s0:
Thank you, @divyegala and @RayWang96, for porting NN-descent to raft. Is this modification correct?

RayWang96:
I believe you are right. And I'm curious why it can pass the unit test with this bug.

cjnolet (approved):

divyegala:
@enp1s0 thanks for this PR. Could you post your NVCC and GCC versions in the description of the PR for hindsight? It seems like RAFT's CI NVCC+GCC are able to compile fine, so we should make note of what causes this compilation failure

cpp/include/raft/neighbors/detail/nn_descent.cuh
```diff
@@ -979,10 +979,10 @@ GnndGraph<Index_t>::GnndGraph(const size_t nrow,
     node_degree(node_degree),
     num_samples(num_samples),
     bloom_filter(nrow, internal_node_degree / segment_size, 3),
-    h_dists{raft::make_host_matrix<DistData_t, size_t, raft::row_major>(nrow, node_degree)},
-    h_graph_new{nrow * num_samples},
-    h_list_sizes_new{nrow},
-    h_graph_old{nrow * num_samples},
+    h_dists(raft::make_host_matrix<DistData_t, size_t, raft::row_major>(nrow, node_degree)),
```
divyegala, line 982: This can keep brace initializer

cpp/include/raft/neighbors/detail/nn_descent.cuh
```diff
@@ -1143,21 +1143,21 @@ GNND<Data_t, Index_t>::GNND(raft::resources const& res, const BuildConfig& build
...
-    h_rev_graph_old_{static_cast<size_t>(nrow_ * NUM_SAMPLES)},
-    d_list_sizes_new_{raft::make_device_vector<int2, Index_t>(res, nrow_)},
-    d_list_sizes_old_{raft::make_device_vector<int2, Index_t>(res, nrow_)}
+    d_data_(raft::make_device_matrix<__half, Index_t, raft::row_major>(
+      res, nrow_, build_config.dataset_dim)),
+    l2_norms_(raft::make_device_vector<DistData_t, Index_t>(res, nrow_)),
+    graph_buffer_(
+      raft::make_device_matrix<ID_t, Index_t, raft::row_major>(res, nrow_, DEGREE_ON_DEVICE)),
+    dists_buffer_(
+      raft::make_device_matrix<DistData_t, Index_t, raft::row_major>(res, nrow_, DEGREE_ON_DEVICE)),
```
divyegala, line 1152: Same as above

cpp/include/raft/neighbors/detail/nn_descent.cuh
```diff
@@ -1143,21 +1143,21 @@ GNND<Data_t, Index_t>::GNND(raft::resources const& res, const BuildConfig& build
...
+    graph_host_buffer_(nrow_ * DEGREE_ON_DEVICE),
+    dists_host_buffer_(nrow_ * DEGREE_ON_DEVICE),
+    d_locks_(raft::make_device_vector<int, Index_t>(res, nrow_)),
+    h_rev_graph_new_(nrow_ * NUM_SAMPLES),
+    h_graph_old_(nrow_ * NUM_SAMPLES),
+    h_rev_graph_old_(nrow_ * NUM_SAMPLES),
+    d_list_sizes_new_(raft::make_device_vector<int2, Index_t>(res, nrow_)),
+    d_list_sizes_old_(raft::make_device_vector<int2, Index_t>(res, nrow_))
```
divyegala, line 1160: Same as above

divyegala (commented):
`raft::device_*` variants of `mdarray` do not have an initializer-list based constructor, so I prefer to keep brace initialization where possible

enp1s0:
Thank you, @divyegala, @RayWang96, and @cjnolet, for reviewing the PR. I have modified the code based on the comments from @divyegala.

The compilation of the unit test passes because it uses `int64_t` as the index type, while I use `int32_t` in my program. The error is `invalid narrowing conversion from "unsigned long" to "int"` and `"int"` here is the index type, `int32_t`. In the case of the unit test, no narrowing conversion occurs because the index type is `int64_t`. But I'm not sure why the unit test works well while the memory is not allocated correctly. Let me look into this a little more. And I think this PR can be merged without waiting for the investigation.

cjnolet:
Thanks @enp1s0. Release 23.10 code freeze has passed, though, and the branch is being prepped for release. Can you base your branch on the 23.12? I'm in the process of fixing some merge conflicts to our forward merger, but we can get your changes merged shortly after.

divyegala:
@enp1s0 I think it's quite likely that this issue is a result of your nvcc version. NN-Descent hard-codes internal index types to int32 so whether you use int64 or int32 index types for CAGRA it should not matter.

Are you saying that the original code compiles with int64 index type for CAGRA with your nvcc version?

enp1s0:
@divyegala Thank you for your comment, and yes, you are right. I can compile with nvcc 12.0.1, the same version as the one used by CI, and 12.1.2, while the error occurs with 12.2.

enp1s0:
Thank you, @divyegala, for your comment. I have found that the difference in the thrust versions is causing the compilation error. Raft uses thrust 1.17.2 while my program uses thrust 2.2, which is included in the CUDA 12.2 toolkit. And the list initializer was introduced in thrust 2.1 [[release note](https://github.com/NVIDIA/thrust/releases/tag/2.1.0)].

So this is not a bug. Is it better to close this PR or merge it for the future thrust version update in raft? (cc: @cjnolet)

divyegala:
@enp1s0 thanks for the investigations! We should merge this PR. Can you also add the thrust version to the description of this PR?

divyegala (approved):

PR #1874, opened 2023-10-05
Always linking gbench statically

cjnolet:
This just makes more sense. There's no inherent reason folks should have to have gbench installed just to run benchmarks.

### DIFF
--- a/cpp/bench/ann/CMakeLists.txt
+++ b/cpp/bench/ann/CMakeLists.txt
@@ -86,6 +86,14 @@ endif()
 # ##################################################################################################
 # * Configure tests function-------------------------------------------------------------
 
+# Build and link static version of the GBench to keep ANN_BENCH self-contained.
+get_target_property(TMP_PROP benchmark::benchmark SOURCES)
+add_library(benchmark_static STATIC ${TMP_PROP})
+get_target_property(TMP_PROP benchmark::benchmark INCLUDE_DIRECTORIES)
+target_include_directories(benchmark_static PUBLIC ${TMP_PROP})
+get_target_property(TMP_PROP benchmark::benchmark LINK_LIBRARIES)
+target_link_libraries(benchmark_static PUBLIC ${TMP_PROP})
+
 function(ConfigureAnnBench)
 
   set(oneValueArgs NAME)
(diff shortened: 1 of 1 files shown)


benfred (approved):

vyasr (changes requested):
Does this make sense to do more generally? @robertmaynard are there known issues with statically linking gbench? If not, should we move this logic into `rapids_cpm_gbench` and make this target available for all of RAPIDS to use in this way?

bdice:
> Does this make sense to do more generally? @robertmaynard are there known issues with statically linking gbench? If not, should we move this logic into `rapids_cpm_gbench` and make this target available for all of RAPIDS to use in this way?

I'd also like to know if we should do the same for nvbench, if applicable.

cjnolet:
@bdice @vyasr @robertmaynard given that we are almost always linking gbench (and gtest) directly into binaries and not into libraries, it would be ideal to make the binaries as self-contained as possible, with the exception of libraft. 

As for the reasoning behind this change- there's currently a gbench version conflict happening in some environments where a different version is installed on the system and we get an library not found error as a result. Gbench is lightweight enough that we shouldn't even have to require it as a runtime dependency. Not many systems are already going to have it installed for runtime anyways. 

Whether we do this at rapids-cmake or do this directly for raft doesn't matter to me, but we need to alleviate this burden because too many users (including our own developers) have encountered this problem.

cpp/bench/ann/CMakeLists.txt
```diff
@@ -86,6 +86,14 @@ endif()
 # ##################################################################################################
 # * Configure tests function-------------------------------------------------------------
 
+# Build and link static version of the GBench to keep ANN_BENCH self-contained.
```
robertmaynard, line 89: This won't work when google benchmark is pre-built and coming from the environment. The sources in that case will be empty and you will fail to have any gtest/gbench in your final executable.

robertmaynard (changes requested):
The proper approach will be to request rapids-cmake to provide you a static version of gtest/gbench. Since this approach doesn't work with an installed version

robertmaynard:
> Does this make sense to do more generally? @robertmaynard are there known issues with statically linking gbench? If not, should we move this logic into `rapids_cpm_gbench` and make this target available for all of RAPIDS to use in this way?

The thing we lose is the ability to use the conda version of gbench/gtest since they only come as shared. So as long as we are fine with that we can update rapids-cmake to provide a static version when requested.

robertmaynard:
> > Does this make sense to do more generally? @robertmaynard are there known issues with statically linking gbench? If not, should we move this logic into `rapids_cpm_gbench` and make this target available for all of RAPIDS to use in this way?
> 
> I'd also like to know if we should do the same for nvbench, if applicable.

NVBench explicitly builds shared we will need to look into this

PR #1882, opened 2023-10-10
Fix ann-bench conda recipe to use faiss only in x86 for 11.8 in recipe

dantegd:
Even though conda packages are being produced for AARCH64 with CUDA 11.8, their dependency is broken due to there not being packages available in the rapidsai-nightly channel. Doesn't affect CUDA 12.0 packages or release packages since the latter ones use the rapidsai channel.


dantegd:
Closing, this will be fixed by #1814

PR #1894, opened 2023-10-12
Add check-json to pre-commit linters, and fix invalid ann-bench JSON config

benfred:
The ann-bench conf files had a couple of datasets with invalid JSON config files, causing them to fail to run.

This adds a check-json pre-commit hook to detect invalid json, which would have failed on these files with:

```
python/raft-ann-bench/src/raft-ann-bench/run/conf/mnist-784-euclidean.json: Failed to json decode (Expecting value: line 1 column 1 (char 0))
python/raft-ann-bench/src/raft-ann-bench/run/conf/glove-100-inner.json: Failed to json decode (Expecting ',' delimiter: line 686 column 7 (char 15218))
```

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -28,7 +28,7 @@ concurrency:
 jobs:
   cpp-build:
     secrets: inherit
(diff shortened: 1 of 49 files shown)


cjnolet (approved):
LGTM!

Issue #1819, opened 2023-09-13
[FEA] Introduce `universal_mdspan/mdarray` vocabulary to RAFT


divyegala:
Working on https://github.com/rapidsai/raft/pull/1748, I realized the need to introduce types that allow for memory access from host and device both, as more algorithms need to support allocating either `managed` or `pinned` memory.

How I came up with an RAII solution to allocating pinned host memory:
```
using pinned_memory_resource = thrust::universal_host_pinned_memory_resource;
template <typename T>
using pinned_memory_allocator = thrust::mr::stateless_resource_allocator<T, pinned_memory_resource>;

thrust::host_vector<Index_t, pinned_memory_allocator<Index_t>> vec;
```

**RMM**
1. Issue with using an RMM based allocator (`rmm/mr/host/pinned_memory_resource.hpp`):
-`thrust` allocates elements by keeping track of a template type `T` for the element to be allocated whereas RMM allocates bytes directly
- RMM has [thrust_allocator_adaptor.h](https://github.com/rapidsai/rmm/blob/branch-23.10/include/rmm/mr/device/thrust_allocator_adaptor.hpp) which accounts for the above conversion, but unfortunately derives from `thrust::device_malloc_allocator`
- Furthermore, `rmm::device_uvector` only allows an allocator derived from a `device_memory_resource`

2. Fix
- Modify `thrust_allocator_adaptor.h` to allow `rmm::thrust_allocator` for inheriting from any arbitrary `thrust::allocator` and using it with `thrust::host_vector`

3. Consequences
- Still relying on `thrust::host_vector` when RAFT should be able to natively support these solutions

**RAFT**
1. Why a `universal_*` vocabulary makes the most sense
- The standard for https://en.cppreference.com/w/cpp/container/vector defines the type for allocator to be similar to thrust's type based allocator of the form `std::allocator<ElementType` 
- In `raft::host_mdarray` we ourselves use `std::allocator<ElementType>` for allocations with `std::vector` under the hood

2. Fix
- Adding a `pinned_memory_allocator` and `managed_memory_allocator` to RAFT does not make much sense in terms of using simple factories to create RAFT types
- Keeping in spirit with the current design, adding a `pinned_container_policy` and `managed_container_policy` in `raft/core/universal_container_policy.hpp` and creating equivalent `universal_mdspan/mdarray` types gives us the most flexibility and removes confusion for users and devs of having to work with allocators and memory resources

3. Consequences
[...]

PR #1896, opened 2023-10-12
ANN bench options to specify CAGRA graph and dataset locations

cjnolet:
CAGRA index can be constructed using existing device_mdarrays, in which case just reference to these arrays are stored. 
This way allocations are managed outside the index, and we can customize how these allocations are made.

This PR
- modifies the CAGRA ANN bench wrapper to manage the allocations locally,
- add options for the json file to specify whether the graph / dataset is allocated in device / host_pinned / host_huge_page memory.


tfeher:
Putting here Corey's original comments in the PR description:

>These are changes that were spearheaded by @tfeher for profiling and we think they've demonstrated enough performance benefit that we should incorporate them into RAFT. The big question is, how do we do this? What should the ideal design look like? This is the first time enabling a feature like this where only certain allocations are using pinned memory while other allocations aren't so we want to focus on a scalable and maintainable solution. 

After the latest changes, the PR does not change CAGRA API. It enables us to benchmark with data in different memory locations, and also demonstrates that it is possible to do so with the existing API.

cpp/bench/ann/src/raft/raft_cagra_wrapper.h
```diff
@@ -118,18 +146,77 @@ void RaftCagra<T, IdxT>::build(const T* dataset, size_t nrow, cudaStream_t)
...
+    // Move graph to correct memory space
+    graph_mem_ = search_param.graph_mem;
+    std::cout << "Moving graph to new memory space " << allocator_to_string(graph_mem_)
+              << std::endl;
+    // We create a new graph and copy to it from existing graph
+    auto mr        = get_mr(graph_mem_);
+    auto new_graph = make_device_mdarray<IdxT, int64_t>(
+      handle_, mr, make_extents<int64_t>(index_->graph().extent(0), index_->graph_degree()));
+
+    std::cout << "new_grap " << new_graph.extent(0) << "x" << new_graph.extent(1) << std::endl;
```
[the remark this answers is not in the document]
tfeher: Debug comment, removed.

cpp/bench/ann/src/raft/raft_cagra_wrapper.h
```diff
@@ -118,18 +146,77 @@ void RaftCagra<T, IdxT>::build(const T* dataset, size_t nrow, cudaStream_t)
...
-  index_->update_dataset(handle_,
-                         raft::make_host_matrix_view<const T, int64_t>(dataset, nrow, this->dim_));
+  // It can happen that we are re-using a previous algo object which already has
+  // the dataset set. Check if we need update.
+  if (index_->size() != nrow || need_dataset_update_) {
+    // First free up existing memory
+    dataset_ = make_device_matrix<T, int64_t>(handle_, 0, 0);
+    index_->update_dataset(handle_, make_const_mdspan(dataset_.view()));
+
+    // Allocate space using the correcct memory resource
```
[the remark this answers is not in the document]
tfeher: Fixed.

cpp/bench/ann/src/common/cuda_huge_page_resource.hpp
```diff
@@ -0,0 +1,136 @@
...
+#include <sys/mman.h>
+
+#include <cstddef>
+
+namespace rmm::mr {
+/**
+ * @brief `device_memory_resource` derived class that uses cudaMallocHost/Free for
+ * allocation/deallocation.
+ */
+class cuda_huge_page_resource final : public device_memory_resource {
```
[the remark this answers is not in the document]
tfeher: Updated the documentation.
> Is there a reason why it is not derived from rmm::mr::host_memory_resource?

The goal is to use this allocator in a [device_container_policy](https://github.com/rapidsai/raft/blob/branch-23.12/cpp/include/raft/core/device_container_policy.hpp#L177), and that requires a `device_memory_resource`.
What we have here is not ideal, more like a quick solution to avoid experimenting / benchmarking with such settings. A related issue https://github.com/rapidsai/raft/issues/1819.

tfeher (commented):
Thanks @lowener for the review, I have addressed the issues.

lowener (approved):

cpp/bench/ann/src/raft/raft_cagra_wrapper.h
```diff
@@ -112,18 +143,70 @@ void RaftCagra<T, IdxT>::build(const T* dataset, size_t nrow, cudaStream_t)
...
+
+    // Allocate space using the correct memory resource.
+    auto mr                 = get_mr(dataset_mem_);
+    auto input_dataset_view = make_device_matrix_view<const T, int64_t>(dataset, nrow, this->dim_);
+    raft::neighbors::cagra::detail::copy_with_padding(handle_, dataset_, input_dataset_view, mr);
+
+    index_->update_dataset(handle_, make_const_mdspan(dataset_.view()));
+
+    // Ideally, instead of dataset_.view(), we should pass a strided matrix view to update.
+    // auto dataset_view = make_device_strided_matrix_view<const T, int64_t>(
```
cjnolet, line 205: Can you make an issue for this and reference it here so we can keep track of it?

cpp/bench/ann/src/common/cuda_huge_page_resource.hpp
```diff
@@ -0,0 +1,134 @@
+/*
+ * Copyright (c) 2019-2023, NVIDIA CORPORATION.
```
cjnolet, line 2: I should have noticed this the first time around- dates for new files should be 2023.

cpp/bench/ann/src/common/cuda_pinned_resource.hpp
```diff
@@ -0,0 +1,130 @@
+/*
+ * Copyright (c) 2019-2023, NVIDIA CORPORATION.
```
cjnolet, line 2: Here too- 2023

cjnolet (commented):
2 minor things. I can't approve my own PR but please feel free to approve it. Tagging @divyegala for approval as well.

cpp/bench/ann/src/raft/raft_benchmark.cu
```diff
@@ -215,6 +215,21 @@ void parse_build_param(const nlohmann::json& conf,
   }
 }
 
+AllocatorType parse_allocator(std::string mem_type)
+{
+  if (mem_type == "device") {
```
cjnolet, line 220: Can you also update the Ann bench param tuning guide to add the men types?

divyegala (approved):

cpp/bench/ann/src/common/cuda_huge_page_resource.hpp
tfeher: fixed

cpp/bench/ann/src/raft/raft_benchmark.cu
tfeher: updated

tfeher (commented):
Thanks @cjnolet for the review, I have addressed the issues.

cpp/bench/ann/src/common/cuda_huge_page_resource.hpp
```diff
@@ -0,0 +1,132 @@
...
+   *
+   * @throws `rmm::bad_alloc` if the requested allocation could not be fulfilled
+   *
+   * @param bytes The size, in bytes, of the allocation
+   * @return void* Pointer to the newly allocated memory
+   */
+  void* do_allocate(std::size_t bytes, rmm::cuda_stream_view) override
+  {
+    void* _addr{nullptr};
+    _addr = mmap(NULL, bytes, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
```
tfeher, line 75: There is a request provide an alternative methods for allocation (without `mmap`) so that we can use Transparent Huge Pages. I will submit a follow up PR with these changes. I am thinking about the following design:

```c++
enum class AllocatorType { HostPinned, HostMmap, HostMalloc }

class host_device_allocator final : public rmm::mr::device_memory_resource {
  host_device_allocator(AllocatorType type): type_(type) {}
  //...
  void* do_alocate(std::size_t bytes, rmm::cuda_stream_view) override {
    switch(type_) {
      case HostPinned:
        cudaMallocHost(&ptr, bytes));
        return ptr;
      case HostMmap:
        ptr = mmap(...);
        //...
        return ptr;
      case HostMalloc:
        //...
    } 
  }
 private:
  AllocatorType type_;
//...
}
```
In this case we would have a single custom `mr` class that we would configure according to the benchmark input parameter. Let me know what do you think.

cjnolet: Hmm, I'll have to think about this. I don't think I generally mind the approach of having a `composite_memory_resource` or `delegating_memory_resource` but I think we should continue using RMM's abstraction layers everywhere possible instead of, for example, calling `cudaMallocHost` directly. My concern is a pretty easy fix, though, by just storing the corresponding memory_resource instances inside the `delegating_memory_resource` and delegating as necessary. That would maintain the compatibility w/ RMM (this could, for example, still be set as the `current_device_resource` if we so wanted).

cjnolet: Also cc'ing @harrism for his thoughts.

cpp/bench/ann/src/common/cuda_pinned_resource.hpp
```diff
@@ -0,0 +1,130 @@
...
+ * This is almost the same as rmm::mr::host::pinned_memory_resource, but it has
+ * device_memory_resource as base class. Pinned memory can be accessed from device,
+ * and using this allocator we can create device_mdarray backed by pinned allocator.
+ *
+ * TODO(tfeher): it would be preferred to just rely on the existing allocator from rmm
+ * (pinned_memory_resource), but that is incompatible with the container_policy class
+ * for device matrix, because the latter expects a device_memory_resource. We shall
+ * revise this once we progress with Issue https://github.com/rapidsai/raft/issues/1819
+ */
+class cuda_pinned_resource final : public rmm::mr::device_memory_resource {
```
cjnolet, line 39: It would be nice to see these new memory resources worked back into RMM eventually.

cpp/bench/ann/src/raft/raft_benchmark.cu
```diff
@@ -236,6 +251,8 @@ void parse_search_param(const nlohmann::json& conf,
       THROW("Invalid value for algo: %s", tmp.c_str());
     }
   }
+  if (conf.contains("graph_mem")) { param.graph_mem = parse_allocator(conf.at("graph_mem")); }
```
cjnolet, line 254: Can we spell out `*_memory_type` to match the other properties that we parse for other algorithms? Will get confusing to set these if we use different spellings.

cpp/bench/ann/src/common/cuda_pinned_resource.hpp
tfeher: RMM has a pinned memory resource. But it is a subclass of `host_memory_resource`. The problem is on our side: our API for [managing allocators for mdspan ](https://github.com/rapidsai/raft/blob/061c0cf7a50bff8600ca9ecef86edee81a354233/cpp/include/raft/core/device_container_policy.hpp#L177) requires a device_memory_resource. We would need to enable host allocator there, and then we could use `rmm::mr::host_memory_resource`.

cpp/bench/ann/src/common/cuda_huge_page_resource.hpp
tfeher: The class above is a subclass of `rmm::mr::device_memory_resource`. We are using it to allocate `mdarray`s that are accessible from both host and device. It is a workaround to unblock some tests.

I feel that RAFT's distinction between `host_mdarry` and `device_mdarray`  stands in the way. What our algorithm needs is an `mdarray` which we can configured with any allocator (host, device). In a few places our algorithm does not care whether the input is a `host_mdarray` or a `device_mdarray`, just passes the pointer to `cudaMemcpy`, to fetch a chunk of the array we are working on. 

How we allocate `mdarray` is described by a `ContainerPolicy` class. So probably the only thing we need is to implement a contaner policy that could be initialized by both `device_memory_resource` and `host_memory_resource`. Afterwards we can simply utilize existing RMM memory resources.

The only thing we are missing from RMM side is a memory resource that uses `mmap` and `madvise`.

raydouglass:

cpp/bench/ann/src/common/cuda_pinned_resource.hpp
harrism: We have a plan too add a device-accessible pinned memory resource. Ideally using the new cuda::memory_resource refactoring.

cpp/bench/ann/src/common/cuda_huge_page_resource.hpp
harrism: Look, it's fine to provide your own base-level `memory_resource` classes that derive from RMM resource base classes. The problem today is that there is no common base for host and device MRs. We are starting to refactor on top of `cuda::memory_resource` which abstracts memory `kind` as a template parameter, and adds `resource_ref` and `async_resource_ref` which have template parameters to specify memory kind. This means you can have a function that accepts any memory resource that is device accessible, host accessible, or both, or whatever kind you can define. And we can start to reuse all the MR and adaptor types in RMM for different kinds of memory.

The problem with @tfeher 's suggestion at the top is that it appears to mix host- and device-accessible memory kinds under `device_memory_resource`, which is supposed to allocate only device-accessible memory.

As to @cjnolet 's comment, I agree that you should have separate resources for different kinds of allocation, and then have a higher level MR that takes pointers/refs to MRs for each kind you want to dispatch to, rather than hard-coding different allocation calls within a single MR.

tfeher (approved):

PR #1898, opened 2023-10-12
Make all cuda kernels have hidden visibility

robertmaynard:
Effect on binary size of libraft.a
23.12: 133361630
pr: 129748904

Effect on binary size of libraft.so
23.12: 83603224
pr: 83873088


robertmaynard:
Replaces https://github.com/rapidsai/raft/pull/1723 

I am working on a follow PR that will marks all kernel call sites as hidden visibility which will allow us to close #1722

cjnolet (approved):

PR #1904, opened 2023-10-16
Fixing Googletests and re-enabling in CI

cjnolet:
Recently, the CI script to run the C++ gtests was changed to call `ctest` but it looks like this command has been failing in CI for quite some time. It's a little scary as most of RAFT's C++ APIs are not exposed through Python and so we cannot rely on those Python tests to guarantee quality of the C++ APIs. 

I'm reverting the change back to invoking the C++ binaries directly so that we can get the tests running again. Once we figure out why `ctest` isn't running the tests then we can make the change to use that again.


wphicks (approved):

robertmaynard (approved):

cjnolet:
It turns out at least the expanded Euclidean issue is related to the precision of clamping from round-off errors between the subtraction of the norms and the dot product (there can be an error of 1e-5, for example, which would end up becoming 0.003 once the sqrt is taken. Thus tiny errors get magnified from the sqrt. 

Prior to this change, we were just taking any small value < 1e-4 and clamping it to 0. The problem is that we were limiting the effective range of inputs to the algorithms. This change adds an explicit check that the norms are equal as well. Be cautioned that in order to determine the vectors are equal, we need to test that the dot product = normX = normY. The challenge is that this roundoff error makes `normX == normY` but the dot product might only be approximately equal to them. Adding this extra check for the equality of the norms should at least narrow down the potential for the clamping to affect other numbers, however, it can still effect things like true orthognality, which could still have equal norms.

raydouglass (approved):

cpp/include/raft/distance/detail/distance_ops/l2_exp.cuh
```diff
@@ -86,10 +104,16 @@ struct l2_exp_distance_op {
...
+        DataT accVal = acc[i][j];
+        DataT val    = regxn[i] + regyn[j] - (DataT)2.0 * accVal;
+
+        /**
+         * Self-neighboring points should have (aNorm == bNorm) == accVal and the dot product
+         * (accVal) can sometimes have round-off errors, which will cause (aNorm == bNorm) ~ accVal
+         * instead.
+         */
+        acc[i][j] =
+          val * (val > 0) * !((val * val < get_clamp_precision<DataT>()) * (regxn[i] == regyn[j]));
```
benfred, line 116: should we update bfknn to have the same clamping logic here https://github.com/rapidsai/raft/blob/50a9081254829177af34b70dc79178aeb16be37b/cpp/include/raft/neighbors/detail/knn_brute_force.cuh#L199-L203 ?

cjnolet: I think that's a good idea. I almost wonder if we should consolidate this w/ the current `l2_exp` distance op so that we only have 1 code path to maintain there.

python/pylibraft/pylibraft/test/test_distance.py
```diff
@@ -63,7 +63,7 @@ def test_distance(n_rows, n_cols, inplace, metric, order, dtype):
     else:
         expected = cdist(input1, input1, metric)
 
-    expected[expected <= 1e-5] = 0.0
+    # expected[expected <= 1e-5] = 0.0
```
benfred, line 66: Can we just delete rather than comment out?

python/pylibraft/pylibraft/test/test_distance.py
```diff
@@ -79,6 +79,8 @@ def test_distance(n_rows, n_cols, inplace, metric, order, dtype):
 
     actual = output_device.copy_to_host()
 
-    actual[actual <= 1e-5] = 0.0
+    # actual[actual <= 1e-5] = 0.0
+    # if metric == "euclidean":
+    #     np.fill_diagonal(actual, 0.0)
```
benfred, line 84:
```suggestion
```

benfred (approved):
lgtm!

python/pylibraft/pylibraft/test/test_distance.py
cjnolet: Ah yes. Those comments can be removed. If this run passes, let me get this merged just for the sake of getting CI (and the numerical issue) fixed and then I'll follow-up with a PR to remove the commented out code.

PR #1910, opened 2023-10-18
Remove dynamic entry-points from raft-ann-bench

benfred:
Building raft-ann-bench can throw a

```
  distutils.errors.DistutilsOptionError: No configuration found for dynamic 'entry-points'.
  Some dynamic fields need to be specified via `tool.setuptools.dynamic`
  others must be specified via the equivalent attribute in `setup.py`.
```

error sometimes depending on distutils version.

This is becase we were specifying dynamic=entry-points, but not including the relevant `tool.setuptools.dynamic` section in pyproject.toml. Remove this to allow building


cjnolet (approved):
LGTM!

PR #1911, opened 2023-10-18
Fix filepath in `raft-ann-bench/split_groundtruth` module

divyegala:


cjnolet (approved):

dantegd (approved):
latest change for using normpath works great!

PR #1912, opened 2023-10-18
CAGRA ANN bench: parse build options for IVF-PQ build algo

tfeher:
This PR enables fine tuning the the parameters for CAGRA index building using the IVF-PQ build algo.

This is only affects ANN bechmarks (an example json file is added). The public API of CAGRA is not changed because:
- Advanced users can achieve the same effect by calling `build_knn_graph` and `optimize` individually.
- Long term we consider NN descent to be the preferred build algo, which does not need IVF-PQ parameters.


cpp/include/raft/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -264,4 +264,82 @@ void build_knn_graph(raft::resources const& res,
...
+template <typename T,
+          typename IdxT = uint32_t,
+          typename Accessor =
+            host_device_accessor<std::experimental::default_accessor<T>, memory_type::host>>
+index<T, IdxT> build(raft::resources const& res,
+                     const index_params& params,
+                     mdspan<const T, matrix_extent<int64_t>, row_major, Accessor> dataset,
+                     std::optional<float> refine_rate                   = std::nullopt,
+                     std::optional<ivf_pq::index_params> build_params   = std::nullopt,
+                     std::optional<ivf_pq::search_params> search_params = std::nullopt)
```
tfeher, line 301: Allowing these parameters is the difference between the public and the detail:: version of build.

divyegala: This is a bit conflicting because you could also build `NN_DESCENT` with this method. Could we allow the base class `ann::index_params` as the function argument here, and then cast it to either IVF_PQ or NN_DESCENT index_params based on `cagra::index_params::graph_build_algo`?

tfeher: That would require us to change `ann::index_params` into a polymorphic class. If we change that, then [this assertion would fail](https://github.com/rapidsai/raft/blob/branch-23.12/cpp/include/raft/neighbors/cagra_types.hpp#L120) (and also similar assertions for ivf_flat and ivf_pq).

I do not know what was the motivation behind keeping the index params as aggregate type. I would prefer to return to such question in a separate issue, and keep this PR a non-breaking change.

I have added an `optional<nn_descent_params>`, so that we can fine set these through the config file.

cpp/include/raft/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -264,4 +264,86 @@ void build_knn_graph(raft::resources const& res,
...
+          typename IdxT = uint32_t,
+          typename Accessor =
+            host_device_accessor<std::experimental::default_accessor<T>, memory_type::host>>
+index<T, IdxT> build(
+  raft::resources const& res,
+  const index_params& params,
+  mdspan<const T, matrix_extent<int64_t>, row_major, Accessor> dataset,
+  std::optional<experimental::nn_descent::index_params> nn_descent_params = std::nullopt,
+  std::optional<float> refine_rate                                        = std::nullopt,
+  std::optional<ivf_pq::index_params> pq_build_params                     = std::nullopt,
```
cjnolet, line 302: After we figure out the way best way to capture these parameters, we also need to update the parameter tuning guide so that it's immediately obvious to users how to specify these options.

python/raft-ann-bench/src/raft-ann-bench/run/conf/wiki_all.json
```diff
@@ -0,0 +1,248 @@
...
+        "graph_degree": 32,
+        "intermediate_graph_degree": 48 ,
+        "ivf_pq_build_params": {
+          "pq_dim": 32,
+          "pq_bits": 8,
+          "nlist": 16384,
+          "niter": 10,
+          "ratio": 10
+        },
+      "ivf_pq_search_params": { "nprobe": 10, "internalDistanceDtype": "half", "smemLutDtype": "half", "refine_ratio": 4 }},
```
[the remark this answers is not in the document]
tfeher: Currently both `ivf_pq_build_params` and `ivf_pq_search_params` are part of CAGRA's `build_param`. 

I do not see how flattening these would change anything in our ability to do grid search. 

I think keeping the current structure is easier to read: parameters are grouped according to their function, and names are shorter.

Btw, you probably do not want to run a full grid search over `graph_degree` and full combination of `ivf_pq_build_params` and `ivf_pq_search_params`. The right approach is to run an `ivf_pq` grid search to find reasonable build / search config, and then just copy the best candidates over here. Keeping the same structure makes it simpler do so so.

Having said that, if you are convinced that the flattened names are better, then we can change them.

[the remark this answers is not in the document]
tfeher: One more thing: I will delete this file, and add the new cagra configs to `wiki_all_1M` / `wiki_all_10M` and `wiki_all_88M.json` files.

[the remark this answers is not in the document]
cjnolet: We need to expose these properties through the yaml config files and we can't do that today because we cant specify nested objects as arguments to the build/search params. We can, however, specify simple objects in flattened keys (and even just lock them to a single value instead of specifying a range). The nice thing about this is that a range could be specified easily if so desired. 

We're pretty much past specifying this stuff manually in json files at this point. At a minimum, we need to make sure we can specify these things in the yaml files.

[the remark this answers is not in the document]
tfeher: Even developing this PR is includes manually tweaking json files. But I agree that we need to specify these params at a higher level as well.

I agree that we should keep things flat in the yaml config. So the question is: what is the simplest way to generate the json config from yaml? The python wrapper for the benchmark is not aware of special needs of individual algorithms, so it is indeed the C++ layer which can map the parameters into the correct structure the easiest way. I have modfied the PR accordingly.

python/raft-ann-bench/src/raft-ann-bench/run/conf/wiki_all_1M.json
```diff
@@ -169,7 +170,23 @@
...
+        "ivf_pq_build_niter": 10,
+        "ivf_pq_build_ratio": 10,
+        "ivf_pq_search_nprobe": 30,
+        "ivf_pq_search_internalDistanceDtype": "half",
+        "ivf_pq_search_smemLutDtype": "half",
+        "ivf_pq_search_refine_ratio": 8,
+        "nn_descent_max_iterations": 10,
+        "nn_descent_graph_degree": 48,
+        "nn_descent_intermediate_graph_degree": 72,
+        "nn_descent_termination_threshold": 0.001
```
tfeher, line 188: These are just examples showing how to specify build params.
- IVF-PQ params are not far from optimal but would need a little more fine tuning. 
- I did not experiment with NN descent parameters.
- Note than `nn_descent_graph_degree` is actually redundant, it has to agree with the `intermediate_graph_deree` (in line 174).

cpp/include/raft/neighbors/detail/cagra/cagra_build.cuh
tfeher: Added description in the tuning guide.

python/raft-ann-bench/src/raft-ann-bench/run/conf/wiki_all_1M.json
cjnolet: I believe @divyegala had the nn descent graph degree being set to the cagra intermediate graph degree in the c++ layer already. If not, we probably want to do this to remove user confusion here.

tfeher: Yes `nn_descent_graph_degree` is correctly set on the C++ side. If user tries to rewrite it, then we emit a warning and restore the correct value. I have removed `nn_descent_graph_degree` from the json config. The cagra parameter tuning guide omits this param.

I would still keep the ability of the json config reader to process this parameter, in case we wan to build simple knn-graphs in the future using nn descent.

cjnolet: Yep, that approach sounds great

cjnolet (approved):
LGTM!

PR #1918, opened 2023-10-20
Add `wiki_all` dataset config and documentation. 

cjnolet:
…et more clarify on how the dataset was generated.

### DIFF
--- a/docs/source/raft_ann_benchmarks.md
+++ b/docs/source/raft_ann_benchmarks.md
@@ -18,6 +18,8 @@ This project provides a benchmark program for various ANN search implementations
   - [Running with Docker containers](#running-with-docker-containers)
 - [Creating and customizing dataset configurations](#creating-and-customizing-dataset-configurations)
 - [Adding a new ANN algorithm](#adding-a-new-ann-algorithm)
+- [Parameter tuning guide](https://docs.rapids.ai/api/raft/nightly/ann_benchmarks_param_tuning/)
+- [Wiki-all RAG/LLM Dataset](https://docs.rapids.ai/api/raft/nightly/wiki_all_dataset/)
 
 ## Installing the benchmarks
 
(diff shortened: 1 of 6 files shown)


docs/source/raft_ann_benchmarks.md
```diff
@@ -242,15 +244,18 @@ Configuration files already exist for the following list of the million-scale da
...
 
 ### End to end: large-scale benchmarks (>10M vectors)
+
 `raft-ann-bench.get_dataset` cannot be used to download the [billion-scale datasets](ann_benchmarks_dataset.md#billion-scale)
 due to their size. You should instead use our billion-scale datasets guide to download and prepare them.
 All other python commands mentioned below work as intended once the
 billion-scale dataset has been downloaded.
 To download billion-scale datasets, visit [big-ann-benchmarks](http://big-ann-benchmarks.com/neurips21.html)
 
+We also provide a new dataset called `wiki-all` containing 88 million 768-dimensional vectors. This dataset is meant for benchmarking a realistic RAG/LLM embedding size at scale. It also contains 1M and 10M vector subsets for smaller-scale experiments. See our [Wiki-all Dataset Guide](https://docs.rapids.ai/api/raft/nightly/wiki_all_dataset/) for more information and to download the dataset.
```
dantegd, line 257: Have we defines the RAG acronym somewhere before? Otherwise it'd be nice to have a `Retrieval-Augmented Generation (RAG)` mentioned somewhere for some users that might not be familiar at first sight with it.

docs/source/wiki_all_dataset.md
```diff
@@ -0,0 +1,46 @@
...
+### Full dataset
+
+A version of the dataset is made available in the binary format that can be used directly by the [raft-ann-bench](https://docs.rapids.ai/api/raft/nightly/raft_ann_benchmarks/) tool. The full 88M dataset is ~251GB and the download link below contains tarballs that have been split into multiple parts.
+
+The following will download all 10 the parts and untar them to a `wiki_all_88M` directory:
+```bash
+curl -s https://data.rapids.ai/raft/datasets/wiki_all/wiki_all.tar.{00..9} | tar -xf - -C /datasets/wiki_all_88M/
+```
+
+The above has the unfortunate drawback that if the command should fail for any reason, it cannot be restarted. The files can also be downloaded individually and then untarred to the directory. Each file is ~27GB and there are 10 of them.
```
dantegd, line 25:
```suggestion
The above has the unfortunate drawback that if the command should fail for any reason, all the parts need to be redownloaded. The files can also be downloaded individually and then untarred to the directory. Each file is ~27GB and there are 10 of them.
```

dantegd (approved):
Looks great! Left 2 small suggestions

PR #1927, opened 2023-10-25
Relax ucx pinning

vyasr:
The latest builds of libarrow are built against ucx 1.15, so the upper bound constraint in raft is causing failures downstream when conda tries to solve environments containing both raft-dask and other packages that transitively bring in ucx via libarrow (namely cudf). This PR also bumps the minimum requirement to 1.14.1 to match ucx-py.

### DIFF
--- a/conda/recipes/raft-dask/conda_build_config.yaml
+++ b/conda/recipes/raft-dask/conda_build_config.yaml
@@ -14,7 +14,7 @@ sysroot_version:
   - "2.17"
 
 ucx_version:
-  - ">=1.13.0,<1.15.0"
+  - ">=1.13.0,<1.16.0"
 
 ucx_py_version:
   - "0.35.*"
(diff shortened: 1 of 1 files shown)


vyasr:
For an example of such an error, see https://github.com/rapidsai/cuml/actions/runs/6631385329/job/18021706812

conda/recipes/raft-dask/conda_build_config.yaml
```diff
@@ -14,7 +14,7 @@ sysroot_version:
   - "2.17"
 
 ucx_version:
-  - ">=1.13.0,<1.15.0"
+  - ">=1.13.0,<1.16.0"
```
pentschev, line 17:
```suggestion
  - ">=1.14.1,<1.16.0"
```

pentschev: Just to make it [consistent with UCX-Py](https://github.com/rapidsai/ucx-py/blob/c34aad0e163a5f656a58d57dcf03530a948ec700/conda/recipes/ucx-py/meta.yaml#L38).

pentschev (changes requested):
Looks mostly good to me, I've left a request just to make it consistent with UCX-Py.

pentschev (approved):
Thanks @vyasr !

PR #1929, opened 2023-10-25
Add YAML config files to run parameter sweeps for ANN benchmarks

divyegala:
Also generate CUDA 12.0 development environments for `bench-ann`

### DIFF
--- a/python/raft-ann-bench/src/raft-ann-bench/run/__main__.py
+++ b/python/raft-ann-bench/src/raft-ann-bench/run/__main__.py
@@ -243,7 +243,7 @@ def main():
         "--search-mode",
         help="run search in 'latency' (measure individual batches) or "
         "'throughput' (pipeline batches and measure end-to-end) mode",
-        default="throughput",
+        default="latency",
     )
 
     args = parser.parse_args()
@@ -415,7 +415,12 @@ def main():
                             func = importable[-1]
                             validator = import_module(module)
                             search_validator = getattr(validator, func)
-                            if search_validator(search_dict, k, batch_size):
+                            if search_validator(
+                                search_dict,
+                                index["build_param"],
+                                k,
+                                batch_size,
+                            ):
                                 index["search_params"].append(search_dict)
                 executables_to_run[executable]["index"].append(index)
 
--- a/python/raft-ann-bench/src/raft-ann-bench/run/conf/datasets.yaml
+++ b/python/raft-ann-bench/src/raft-ann-bench/run/conf/datasets.yaml
@@ -29,6 +29,7 @@
   distance: euclidean
 
 - name: fashion-mnist-784-euclidean
+  dims: 784
   base_file: fashion-mnist-784-euclidean/base.fbin
   query_file: fashion-mnist-784-euclidean/query.fbin
   groundtruth_neighbors_file: fashion-mnist-784-euclidean/groundtruth.neighbors.ibin
--- a/python/raft-ann-bench/src/raft-ann-bench/validators/__init__.py
+++ b/python/raft-ann-bench/src/raft-ann-bench/validators/__init__.py
@@ -19,16 +19,22 @@ DTYPE_SIZES = {"float": 4, "half": 2, "fp8": 1}
 def raft_ivf_pq_build_validator(params, dims):
     if "pq_dim" in params:
         return params["pq_dim"] <= dims
+    return True
 
 
-def raft_ivf_pq_search_validator(params, k, batch_size):
+def raft_ivf_pq_search_validator(params, build_params, k, batch_size):
+    ret = True
     if "internalDistanceDtype" in params and "smemLutDtype" in params:
-        return (
+        ret = (
             DTYPE_SIZES[params["smemLutDtype"]]
-            >= DTYPE_SIZES[params["internalDistanceDtype"]]
+            < DTYPE_SIZES[params["internalDistanceDtype"]]
         )
 
+    if "nlist" in build_params and "nprobe" in params:
+        ret = build_params["nlist"] <= params["nprobe"]
+    return ret
 
-def raft_cagra_search_validator(params, k, batch_size):
+
+def raft_cagra_search_validator(params, build_params, k, batch_size):
     if "itopk" in params:
         return params["itopk"] >= k
--- a/python/raft-ann-bench/src/raft-ann-bench/run/conf/algos/raft_ivf_pq.yaml
+++ b/python/raft-ann-bench/src/raft-ann-bench/run/conf/algos/raft_ivf_pq.yaml
@@ -5,12 +5,13 @@ validators:
 groups:
   base:
     build:
-      nlist: [1024, 2048, 4096, 8192, 16000, 32000]
-      pq_dim: [384, 256, 128, 64, 32]
+      nlist: [1024, 2048, 4096, 8192]
+      pq_dim: [64, 32]
       pq_bits: [8, 6, 5, 4]
       ratio: [1, 10, 25]
       niter: [25]
     search:
-      nprobe: [1, 5, 10, 50, 100, 200, 500, 1000, 2000]
+      nprobe: [1, 5, 10, 50, 100, 200, 500]
       internalDistanceDtype: ["float", "half"]
       smemLutDtype: ["float", "fp8", "half"]
+      refine_ratio: [1, 2, 4]


python/raft-ann-bench/src/raft-ann-bench/plot/__main__.py
```diff
@@ -379,7 +411,7 @@ def main():
         formatter_class=argparse.ArgumentDefaultsHelpFormatter
     )
     parser.add_argument(
-        "--dataset", help="dataset to download", default="glove-100-inner"
+        "--dataset", help="dataset to plots", default="glove-100-inner"
     )
```
cjnolet, line 415: should this be singular?

python/raft-ann-bench/src/raft-ann-bench/plot/__main__.py
```diff
@@ -397,6 +429,15 @@ def main():
                               algorithms",
         default=None,
     )
+    parser.add_argument(
+        "--groups",
+        help="plot only comma separated groups of parameters",
```
cjnolet, line 434: Can you also update the algorithms help string to add that unless specified with 'groups' or 'algo-groups', the parameters in the 'base' group of the yml config are executed by default.

python/raft-ann-bench/src/raft-ann-bench/plot/__main__.py
```diff
@@ -397,6 +429,15 @@ def main():
...
+        default="base"
+    )
+    parser.add_argument(
+        "--algo-groups",
+        help="add comma separated algorithm+groups to the plot",
+    )
```
cjnolet, line 440: Probably work providing an example of this (using groups that are already specified.) Something like  "raft_ivf_pq.large,hnswlib.large"

python/raft-ann-bench/src/raft-ann-bench/run/__main__.py
```diff
@@ -198,10 +213,13 @@ def main():
         default=None,
     )
     parser.add_argument(
-        "--indices",
-        help="run only comma separated list of named indices. \
-                              parameter `algorithms` is ignored",
-        default=None,
+        "--groups",
+        help="run only comma separated groups of parameters",
```
cjnolet, line 217: Can you match the help in plot above (also with my comments)

python/raft-ann-bench/src/raft-ann-bench/run/__main__.py
```diff
@@ -198,10 +213,13 @@ def main():
...
+        default="base"
+    )
+    parser.add_argument(
+        "--algo-groups",
```
cjnolet, line 221: Same for this and 'algorithms'

python/raft-ann-bench/src/raft-ann-bench/run/__main__.py
```diff
@@ -173,13 +184,17 @@ def main():
...
         help="number of query vectors to use in each query trial",
     )
+    parser.add_argument(
+        "--dataset-configuration",
+        help="path to configuration file for datasets",
+    )
     parser.add_argument(
         "--configuration",
-        help="path to configuration file for a dataset",
+        help="path to configuration file or directory for algorithms",
```
cjnolet, line 193: Can you just briefly point out that any groups found in the specified files/directories wil automatically override groups of the same name for corresponding algorithms? (Including 'base')

python/raft-ann-bench/src/raft-ann-bench/run/conf/datasets.yaml
```diff
@@ -0,0 +1,85 @@
+- name: bigann-100M
```
cjnolet, line 1: My PR brought in new json configs for wiki-all-88M, wiki-all-10M, and wiki-all-1M. Can you add those here?

python/raft-ann-bench/src/raft-ann-bench/validators/__init__.py
```diff
@@ -0,0 +1,29 @@
...
+
+def raft_ivf_pq_search_validator(params, k, batch_size):
+    if "internalDistanceDtype" in params and "smemLutDtype" in params:
+        return (
+            DTYPE_SIZES[params["smemLutDtype"]]
+            >= DTYPE_SIZES[params["internalDistanceDtype"]]
+        )
+
+
+def raft_cagra_search_validator(params, k, batch_size):
```
cjnolet, line 27: Did you verify these actually filter corresponding parameters?

cjnolet (changes requested):
I haven't gone through the readme yet but so far everything else looks good. Mostly minor things in the help guides. 

I'll review the docs when ready. Please also remember we need a copy/paste example of what a custom yaml file should look like. It doesn't have to be real, and doesn't have to include the validators since those should always be loaded first in the included configs.  At least something that shows multiple groups an the association between a couple parameter names and values (for both build and search)

docs/source/raft_ann_benchmarks.md
```diff
@@ -183,18 +198,21 @@ CSV file in `<dataset-path/<dataset>/result/search/<-k{k}-batch_size{batch_size}
...
   --dataset-path DATASET_PATH
-                        path to dataset folder (default: ${RAPIDS_DATASET_ROOT_DIR})
+                        path to dataset folder (default: /raid/dgala/raft/datasets/)
   --output-filepath OUTPUT_FILEPATH
-                        directory for PNG to be saved (default: os.getcwd())
+                        directory for PNG to be saved (default: /raid/dgala/raft)
   --algorithms ALGORITHMS
                         plot only comma separated list of named algorithms (default: None)
+  --groups GROUPS       plot only comma separated groups of parameters (default: base)
+  --algo-groups ALGO_GROUPS
```
cjnolet, line 214: Reminder to update these (for all scripts) when the "help" is updated

docs/source/raft_ann_benchmarks.md
```diff
@@ -206,6 +224,11 @@ options:
                         Scale to use when drawing the Y-axis (default: linear)
   --raw                 Show raw results (not just Pareto frontier) in faded colours (default: False)
 ```
+`algorithms`: plots all algorithms that it can find results for the specified `dataset`
```
cjnolet, line 227: Mention base group (copy from ru n script instructions above)

docs/source/raft_ann_benchmarks.md
```diff
@@ -357,40 +380,44 @@ Additionally, the containers can be run in detached mode without any issue.
...
-   "name": "algo_name.unique_index_name",
-   "algo": "algo_name",
-   "file": "sift-128-euclidean/algo_name/param1_val1-param2_val2",
-   "build_param": { "param1": "val1", "param2": "val2" },
-   "search_params": [{ "search_param1": "search_val1" }]
-}
+<a id='yaml-algo-config'></a>Configuration files for ANN algorithms supported by `raft-ann-bench` are provided in `${RAFT_HOME}/python/raft-ann-bench/src/raft-ann-bench/run/conf`. `raft_cagra` algorithm configuration looks like:
+```yaml
+name: raft_cagra
+validators:
```
cjnolet, line 398: Let's remove validators from here.  If we need to, we can discuss them in a different section for more advanced topics. These are algo specific and like you pointed out aren't easily changed. They shojkd also remain specified by default because we should be loading the base installed configurations first.

docs/source/raft_ann_benchmarks.md
```diff
@@ -357,40 +380,44 @@ Additionally, the containers can be run in detached mode without any issue.
...
+      graph_degree: [32, 64]
+    search:
+      itopk: [32, 64, 128]
 ```
+It is mandatory that every algorithm have a `base` group.
 
-The table below contains the possible settings for the `algo` field. Each unique algorithm will have its own set of `build_param` and `search_params` settings. The [ANN Algorithm Parameter Tuning Guide](ann_benchmarks_param_tuning.md) contains detailed instructions on choosing build and search parameters for each supported algorithm.
+There's several things going on here:
+1. `groups` - define a run group which has a particular set of parameters. Each group has `build` parameters wherein a cross-product of all parameters will be used to build a unique index. Each group also has `search` parameters, and a cross-product of all search parameters will be used to search all unique indices that we create.
+2. `validators` - This has two entries `build` and `search`. It's a string denoting a Python importable function of a module that will be used to verify whether there are any bogus parameters in any `groups` that do not make sense. If such parameters exist, they will be silently ignored.
```
cjnolet, line 418: Remove validators. Users don't need to care about this

docs/source/raft_ann_benchmarks.md
```diff
@@ -357,40 +380,44 @@ Additionally, the containers can be run in detached mode without any issue.
...
+    search:
+      itopk: [32, 64, 128]
+
+  large:
+    build:
+      graph_degree: [32, 64]
+    search:
+      itopk: [32, 64, 128]
 ```
+It is mandatory that every algorithm have a `base` group.
```
cjnolet, line 414: Users don't need to care about this because it's an internal detail. Just say that they can override the "default" parameters by creating a group called "base" in their config file.

docs/source/raft_ann_benchmarks.md
```diff
@@ -357,40 +380,44 @@ Additionally, the containers can be run in detached mode without any issue.
...
 
-The table below contains the possible settings for the `algo` field. Each unique algorithm will have its own set of `build_param` and `search_params` settings. The [ANN Algorithm Parameter Tuning Guide](ann_benchmarks_param_tuning.md) contains detailed instructions on choosing build and search parameters for each supported algorithm.
+There's several things going on here:
```
cjnolet, line 416: Instead of saying this, please say something like "The config file above has two fields:" (the list below should contain "name" and "groups"

docs/source/raft_ann_benchmarks.md
```diff
@@ -357,40 +380,44 @@ Additionally, the containers can be run in detached mode without any issue.
...
+1. `groups` - define a run group which has a particular set of parameters. Each group has `build` parameters wherein a cross-product of all parameters will be used to build a unique index. Each group also has `search` parameters, and a cross-product of all search parameters will be used to search all unique indices that we create.
```
cjnolet, line 417: "cross-product of all hyper-parameter fields for build and search". 

Can you please link out to the parameter tuning guide and mention that some parameters in the tuning  guide are required, but otherwise don't need to be specified.

cjnolet (changes requested):

docs/source/raft_ann_benchmarks.md
divyegala: The tuning guide is linked in the sentence below

cjnolet (approved):
LGTM!

raydouglass (approved):

PR #1935, opened 2023-10-27
Verify conda-cpp-post-build-checks

robertmaynard:
Adds a CI check to verify the the libraft.so doesn't have any public weak kernel symbols.


.github/workflows/build.yaml
```diff
@@ -34,6 +34,17 @@ jobs:
       branch: ${{ inputs.branch }}
       date: ${{ inputs.date }}
       sha: ${{ inputs.sha }}
+  cpp-checks:
+    needs: [cpp-build]
+    secrets: inherit
+    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-post-build-checks.yaml@conda-cpp-post-build-checks
+    with:
+      build_type: nightly
```
ajschmidt8, line 42:
```suggestion
      build_type: ${{ inputs.build_type || 'branch' }}
```

This `build.yaml` workflow gets triggered upon multiple events:

- pushes to `branch-*` branches (e.g. PR merges)
- nightly builds (called via the top-level workflow [here](https://github.com/rapidsai/workflows/blob/main/.github/workflows/nightly-pipeline.yaml))

Because of this, we can't hardcode `nightly` otherwise the job won't work correctly for PR merges.

To fix this, we can apply this suggestion, which will make the job work correctly for both events.

Alternatively, we can make this job conditional to only run nightly.

However, if this job uses CPU nodes (which it appears to), then it's probably easier to just run this on both events since we're not constrained on CPU machines.

robertmaynard: That makes sense! Yeah this check can run on CPU only nodes so having it run on both events is fine.

.github/workflows/build.yaml
```diff
@@ -34,6 +34,17 @@ jobs:
       branch: ${{ inputs.branch }}
       date: ${{ inputs.date }}
       sha: ${{ inputs.sha }}
+  cpp-checks:
```
ajschmidt8, line 37: Now that I think about it, this job is probably better suited for the `test.yaml` workflow since it's not technically a build job.

In our nightly workflow, the `test.yaml` workflow is triggered after a successful `build.yaml` workflow completion.

So the consequence of including this in `build.yaml` is that it will prevent `test.yaml` from being triggered in our nightly pipelines if `cpp-checks` is the only failing job, which is probably not what we want.

robertmaynard: done

.github/workflows/test.yaml
```diff
@@ -14,6 +14,16 @@ on:
         type: string
 
 jobs:
+  conda-cpp-checks:
+    secrets: inherit
+    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-post-build-checks.yaml@branch-23.12
+    with:
+      build_type: ${{ inputs.build_type || 'branch' }}
```
ajschmidt8, line 21:
```suggestion
      build_type: nightly
```

`test.yaml` is only used for nightlies, so we can hard code this value here.

ajschmidt8 (approved):

robertmaynard:
@cjnolet This introduces a new CI job that checks the libraft.so binary for public weak CUDA kernels. When one is found it will produce an error report so that the PR can be corrected before merging.

This ensures that the work done in https://github.com/rapidsai/raft/pull/1898 doesn't regress when developers add new kernels.

PR #1920, opened 2023-10-20
Adding `throughput` and `latency` modes to `raft-ann-bench`

cjnolet:
Separating the way the benhcmarks are measured into `throughput` and `latency` modes. 

- `latency` mode accumulates the times for each batch to be processed and then estimates QPS and provides the average time spent doing processing on the GPU. For batch size of 1, this becomes a fairly estimate of average latency per query. For larger batches, it becomes a fairly accurate estimate of time spent per batch. 

- `throughput` mode pipelines the individual batches using a thread pool (and stream pool for the GPU algos). For both smaller and larger batches, this gives a good estimate of the amount of data we can push through the hardware in a period of time. 

A good comprehensive comparison will include both of these numbers.

PR #1940, opened 2023-10-31
Fix ANN bench latency

tfeher:
This adds explicit latency column to the benchmark.


tfeher:
The initial version of this PR adds a helper ANN benchmark class: FixLatencyWorkload. This can be configured to spend 
10ms on CPU or GPU time sleeping, and it is useful to illustrate the shortcommings of the current benchmark counters.
```
./RAFT_IVF_PQ_ANN_BENCH --search --data_prefix=/data --benchmark_filter=fix_latency --benchmark_min_time=100x --benchmark_counters_tabular=true --mode=latency fix_latency.json
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Benchmark                                               Time             CPU   Iterations        GPU    Latency     Recall end_to_end items_per_second          k  n_queries total_queries    use_gpu
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
fix_latency/0/process_time/real_time/threads:1       10.2 ms        0.032 ms          100  0.0101664  0.0101838          0    1.01838        982.414/s         10         10          1000          0
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Benchmark                                               Time             CPU   Iterations        GPU    Latency     Recall end_to_end items_per_second          k  n_queries sync_stream total_queries    use_gpu
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
fix_latency/1/process_time/real_time/threads:1      0.024 ms        0.024 ms          100   13.5014u  0.0104866          0    1.04866       408.261k/s         10         10           0          1000          1
fix_latency/2/process_time/real_time/threads:1       10.5 ms         10.5 ms          100  0.0104682  0.0104769          0    1.04769        954.496/s         10         10           1          1000          1
```
We can see the following issues:
- CPU: measures [user time and system time.](https://github.com/google/benchmark/blob/b93f5a592972b9017539cf15a5d299149c1cc2f4/src/timers.cc#L88-L92) This does not include idle time. Idle time can also occur when we [wait for GPU sync](https://github.com/rapidsai/raft/blob/9ad76faeec798a45c359fb358c414f1d3a19eb3f/cpp/include/raft/core/interruptible.hpp#L299), therefore this is not a correct measure for latency. 

- GPU: If the GPU lib does not sync stream, then GPU time will not work.

tfeher:
In throughput mode `Time` is not a correct measure of latency. `Time` is the wall clock time diff divided by the number of iterations. Since we run iterations in parallel, time becomes effectively latency of single iteration divided by number of threads. 

I believe we want to define vector search `Latency` as wall-clock time for finishing a batch.  This PR adds such a column to the benchmark output. `Time` gives us almost what we need, but it is divided by the number of threads. Unfortunately the timers are private, so we cannot access them from the benchmark `state`. We can use the `end_to_end` duration variable, and divide by the iteration count to get the average latency (notice the `kAvgThreads`).
```
/RAFT_IVF_PQ_ANN_BENCH --search --data_prefix=/data --benchmark_filter=fix_latency --benchmark_min_time=100x  --benchmark_counters_tabular=true --mode=throughput /workspace1/config/fix_latency.json

------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Benchmark                                                Time             CPU   Iterations        GPU    Latency     Recall end_to_end items_per_second          k  n_queries total_queries    use_gpu
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
fix_latency/0/process_time/real_time/threads:1        10.2 ms        0.032 ms          100  0.0101596  0.0101772          0    1.01772        983.058/s         10         10          1000          0
fix_latency/0/process_time/real_time/threads:2        5.10 ms        0.046 ms          200  0.0101801  0.0102127          0    1.02172       1.95936k/s         10         10            2k          0
fix_latency/0/process_time/real_time/threads:4        2.55 ms        0.057 ms          400  0.0101708   0.010227          0    1.02396       3.91614k/s         10         10            4k          0
fix_latency/0/process_time/real_time/threads:8        1.28 ms        0.064 ms          800  0.0101425  0.0102338          0    1.02196       7.84236k/s         10         10            8k          0
fix_latency/0/process_time/real_time/threads:16      0.633 ms        0.039 ms         1600  0.0101036  0.0101894          0    1.01989       15.8025k/s         10         10           16k          0
fix_latency/0/process_time/real_time/threads:32      0.316 ms        0.033 ms         3200  0.0100886  0.0102426          0    1.01987       31.6601k/s         10         10           32k          0
fix_latency/0/process_time/real_time/threads:48      0.211 ms        0.033 ms         4800   0.010088  0.0102979          0    1.02507       47.4775k/s         10         10           48k          0
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Benchmark                                                Time             CPU   Iterations        GPU    Latency     Recall end_to_end items_per_second          k  n_queries sync_stream total_queries    use_gpu
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
fix_latency/1/process_time/real_time/threads:1       0.025 ms        0.025 ms          100   13.8858u  0.0104901          0    1.04901       404.242k/s         10         10           0          1000          1
fix_latency/1/process_time/real_time/threads:2       0.013 ms        0.023 ms          200   8.14784u  0.0104827          0    1.04866       797.795k/s         10         10           0            2k          1
fix_latency/1/process_time/real_time/threads:4       0.014 ms        0.041 ms          400   10.9731u  0.0104859          0    1.04857       709.049k/s         10         10           0            4k          1
fix_latency/1/process_time/real_time/threads:8       0.368 ms        0.657 ms          800   105.973u  0.0296742          0    2.96675       27.1756k/s         10         10           0            8k          1
fix_latency/1/process_time/real_time/threads:16      0.945 ms         12.0 ms         1600   1.43132m  0.0289549          0    2.89458       10.5867k/s         10         10           0           16k          1
fix_latency/1/process_time/real_time/threads:32      0.516 ms         16.1 ms         3200   1023.67u  0.0180644          0    1.80312       19.3707k/s         10         10           0           32k          1
fix_latency/1/process_time/real_time/threads:48      0.354 ms         16.4 ms         4800   790.397u   0.017988          0    1.79329        28.237k/s         10         10           0           48k          1
fix_latency/2/process_time/real_time/threads:1        10.5 ms         10.5 ms          100  0.0104754  0.0104842          0    1.04842        953.833/s         10         10           1          1000          1
fix_latency/2/process_time/real_time/threads:2        5.24 ms         10.5 ms          200  0.0104645  0.0104835          0    1.04876       1.90857k/s         10         10           1            2k          1
fix_latency/2/process_time/real_time/threads:4        2.62 ms         10.5 ms          400  0.0104443  0.0104819          0    1.04864       3.81809k/s         10         10           1            4k          1
fix_latency/2/process_time/real_time/threads:8        1.31 ms         10.4 ms          800   0.010427  0.0104937          0    1.04861       7.63035k/s         10         10           1            8k          1
fix_latency/2/process_time/real_time/threads:16      0.655 ms         10.4 ms         1600  0.0103763  0.0105217          0    1.05113       15.2569k/s         10         10           1           16k          1
fix_latency/2/process_time/real_time/threads:32      0.328 ms         10.3 ms         3200  0.0103018  0.0105683          0     1.0542       30.4933k/s         10         10           1           32k          1
fix_latency/2/process_time/real_time/threads:48      0.219 ms         10.3 ms         4800  0.0103413   0.010683          0    1.06139        45.568k/s         10         10           1           48k          1

File: cpp/bench/ann/src/common/benchmark.hpp [hunk 1]
```diff
@@ -268,24 +267,22 @@ void bench_search(::benchmark::State& state,
         state.SkipWithError(std::string(e.what()));
       }
 
-      auto end = std::chrono::high_resolution_clock::now();
-
-      auto elapsed_seconds = std::chrono::duration_cast<std::chrono::duration<double>>(end - start);
       // advance to the next batch
       batch_offset = (batch_offset + n_queries) % query_set_size;
       queries_processed += n_queries;
-      state.SetIterationTime(elapsed_seconds.count());
```
tfeher, line 277: I have removed this because it is only useful, if we call `UseManualTime()` while adding the benchmark.  We could keep this and switch to ManualTime. This would more or less agree with the current `Time` column, and would have the same issues.

cpp/bench/ann/src/common/benchmark.hpp
```diff
@@ -268,24 +267,22 @@ void bench_search(::benchmark::State& state,
...
-      total_time += elapsed_seconds.count();
     }
   }
-  auto end = std::chrono::high_resolution_clock::now();
-  if (state.thread_index() == 0) {
-    auto duration = std::chrono::duration_cast<std::chrono::duration<double>>(end - start).count();
-    state.counters.insert({{"end_to_end", duration}});
-  }
+  cudaDeviceSynchronize();
```
tfeher, line 275: Continuing the discussion from here: https://github.com/rapidsai/raft/pull/1920#discussion_r1376622215
Not sure if there is a smaller hammer to guarantee proper `GPU` time.

cpp/bench/ann/src/common/benchmark.hpp
```diff
@@ -268,24 +267,22 @@ void bench_search(::benchmark::State& state,
...
+  auto end      = std::chrono::high_resolution_clock::now();
+  auto duration = std::chrono::duration_cast<std::chrono::duration<double>>(end - start).count();
+  if (state.thread_index() == 0) { state.counters.insert({{"end_to_end", duration}}); }
+  state.counters.insert(
+    {"Latency", {duration / double(state.iterations()), benchmark::Counter::kAvgThreads}});
```
tfeher, line 280: Continuing discussion https://github.com/rapidsai/raft/pull/1920#discussion_r1376318275

cpp/bench/ann/src/raft/fix_latency_workload.h
```diff
@@ -0,0 +1,121 @@
...
+{
+  if (use_gpu) {
+    kernel_sleep<<<1, 1, 0, stream>>>(sleep_ms);
+    if (sync_stream) { cudaStreamSynchronize(stream); }
+  } else {
+    std::this_thread::sleep_for(std::chrono::milliseconds(sleep_ms));
+  }
+}
+
+class FixLatencyWorkload : public ANN<float> {
```
tfeher, line 52: This is just a tool to test the timers. Shall we remove before merging the PR?

tfeher:
- [ ] add command line arg to control the number of threads. This would be useful when running larger set of benchmarks / profiling.

cpp/bench/ann/src/common/benchmark.hpp
cjnolet: I'm confused by this- why would we synchronize the whole device when the whole goal w/ the multithreading is to allow all the threads to do their work independently?

cpp/bench/ann/src/raft/fix_latency_workload.h
cjnolet: Yeah, I don't think we need to keep this.

cpp/bench/ann/src/common/benchmark.hpp
cjnolet: I think this is fine. Average latency is what we want and it's probably best to have our own computed time here, rather than relying on the time computed from Gbench. I think maybe we should add to the docs that users ignore Time and CPU, and just rely on Latency, `end_to_end`. and `items_per_second`.

cpp/bench/ann/src/common/benchmark.hpp
achirkin: NB: this `cudaDeviceSynchronize();` happens once per invocation of the benchmark case, not each iteration (which is the inner loop on the `state`). This is to ensure the implementation has finished all work in all streams it may have created (e.g. destructors release gpu memory.).

tfeher: Yes, `cudaDeviceSynchronize()` happens outside the benchmark loop. It to ensure that `Latency` is correct even if the benchmarked algorithm does not behave nicely (forget to sync). It does add an overhead, but it is amortized away if we have many benchmark iterations.

tfeher: Many thanks @achirkin for the [detailed explanation](https://github.com/rapidsai/raft/pull/1920#discussion_r1377232547) for the `cuda_timer` and the sync mechanism. It does make sense now.

cpp/bench/ann/src/common/benchmark.hpp
tfeher: I have improved the docs along these lines.

cpp/bench/ann/src/common/benchmark.hpp
cjnolet: Sorry, I meant to respond to this earlier. Yeah, I was just looking at the diff in my browser and it almost appeared as if the sync was happening on each iteration. Totally fine with doing an initial device sync at the beginning

cjnolet:
> TODO: add command line arg to control the number of threads. This would be useful when running larger set of benchmarks / profiling.

It would be nice to even be able to set the min/max threads (and set them equal if one desires to run only a single threaded experiment)

cpp/bench/ann/src/raft/raft_benchmark.cu
```diff
@@ -47,10 +47,19 @@ extern template class raft::bench::ann::RaftCagra<uint8_t, uint32_t>;
...
+#include "fix_latency_workload.h"
 #include <nlohmann/json.hpp>
 
 namespace raft::bench::ann {
 
+inline void parse_search_param(const nlohmann::json& conf,
+                               typename raft::bench::ann::FixLatencyWorkload::SearchParam& param)
+{
+  if (conf.contains("use_gpu")) param.use_gpu = conf.at("use_gpu");
+  if (conf.contains("sync_stream")) param.sync_stream = conf.at("sync_stream");
```
cjnolet, line 59: Are these also renmants of the above `fix_latency_workload`? We aren't planning on adding new params for the user to specify, are we?

cpp/bench/ann/src/raft/fix_latency_workload.h
tfeher: removed

cjnolet (approved):
LGTM!

tfeher:
I am still investigating on bug. Please do not merge until its fixed.

tfeher:
Fixed the problem. Pending CI it is ready to merge. Example output
```
./RAFT_IVF_FLAT_ANN_BENCH --search --data_prefix=/data --benchmark_filter=raft_ivf_flat --benchmark_out_format=csv --benchmark_out=res.csv  --override_kv=n_queries:10 --benchmark_counters_tabular=true --mode=throughput --threads=1:24 --benchmark_min_warmup_time=1 /workspace1/config/deep-10M.json
...
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Benchmark                                                          Time             CPU   Iterations        GPU    Latency     Recall end_to_end items_per_second          k  n_queries     nprobe total_queries
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
raft_ivf_flat.nlist5K/0/process_time/real_time/threads:1       0.111 ms        0.111 ms         6335   101.801u   110.565u    0.90378   0.700427       90.4461k/s         10         10         50        63.35k
raft_ivf_flat.nlist5K/0/process_time/real_time/threads:2       0.075 ms        0.146 ms         9250   133.635u   151.153u    0.90378   0.699437       132.519k/s         10         10         50         92.5k
raft_ivf_flat.nlist5K/0/process_time/real_time/threads:4       0.052 ms        0.187 ms        13560   177.321u   206.465u    0.90378   0.699672       194.025k/s         10         10         50        135.6k
raft_ivf_flat.nlist5K/0/process_time/real_time/threads:8       0.040 ms        0.250 ms        17696   276.013u   324.323u    0.90378   0.717858       249.676k/s         10         10         50       176.96k
raft_ivf_flat.nlist5K/0/process_time/real_time/threads:16      0.037 ms        0.331 ms        18528    508.35u   596.448u    0.90378   0.689932       271.407k/s         10         10         50       185.28k
raft_ivf_flat.nlist5K/0/process_time/real_time/threads:24      0.037 ms        0.400 ms        18144   757.152u   892.142u    0.90377   0.671828       272.899k/s         10         10         50       181.44k
```

It is recommended to use  `--benchmark_min_warmup_time=1`, otherwise the first benchmark is not warmed up and the realtive perf with different number of threads is misleading.

python/raft-ann-bench/src/raft-ann-bench/run/__main__.py
```diff
@@ -243,6 +247,16 @@ def main():
         default="latency",
     )
 
+    parser.add_argument(
+        "-t",
+        "--search-threads",
+        help="specify the number threads to use for throughput benchmark."
+        " Power of 2 values between 'min' and 'max' will be used. If only "
```
tfeher, line 255:
```suggestion
        " Single value or a pair of min and max separated by ':'. Example --threads=1:4"
        " Power of 2 values between 'min' and 'max' will be used. If only "
```

cpp/bench/ann/src/common/benchmark.hpp
```diff
@@ -278,24 +291,22 @@ void bench_search(::benchmark::State& state,
...
-      state.SetIterationTime(elapsed_seconds.count());
-      total_time += elapsed_seconds.count();
     }
   }
-  auto end = std::chrono::high_resolution_clock::now();
-  if (state.thread_index() == 0) {
-    auto duration = std::chrono::duration_cast<std::chrono::duration<double>>(end - start).count();
-    state.counters.insert({{"end_to_end", duration}});
-  }
+  cudaDeviceSynchronize();
```
tfeher, line 299: CPU only builds fail at this point
```
benchmark.hpp:299:3: error: there are no arguments to 'cudaDeviceSynchronize' that depend on a template parameter, so a declaration of 'cudaDeviceSynchronize' must be available [-fpermissive]
  299 |   cudaDeviceSynchronize()
```

I thought we [cuda_stub.hpp](https://github.com/rapidsai/raft/blob/branch-23.12/cpp/bench/ann/src/common/cuda_stub.hpp) to help with this, but it does not work. If we are in a hurry we could disable the sync here, but it would be better to fix the `cuda_stubs`.

PR #1956, opened 2023-11-02
CAGRA build + HNSW search

divyegala:


cpp/bench/ann/src/raft/raft_benchmark.cu
```diff
@@ -182,6 +188,37 @@ void parse_search_param(const nlohmann::json& conf,
 }
 #endif
 
+#ifdef RAFT_ANN_BENCH_USE_RAFT_CAGRA_HNSWLIB
+template <typename T, typename IdxT>
+void parse_build_param(const nlohmann::json& conf,
+                       typename raft::bench::ann::RaftCagraHnswlib<T, IdxT>::BuildParam& param)
+{
+  if (conf.contains("graph_degree")) {
```
cjnolet, line 196: Since these are all CAGRA-specific, can we just use the existing CAGRA version? This is going to increase maintenance burden.

cpp/bench/ann/src/raft/raft_benchmark.cu
```diff
@@ -182,6 +188,37 @@ void parse_search_param(const nlohmann::json& conf,
...
+  }
+  if (conf.contains("nn_descent_niter")) { param.nn_descent_niter = conf.at("nn_descent_niter"); }
+}
+
+template <typename T, typename IdxT>
+void parse_search_param(const nlohmann::json& conf,
+                        typename raft::bench::ann::RaftCagraHnswlib<T, IdxT>::SearchParam& param)
+{
+  param.ef = conf.at("ef");
+  if (conf.contains("numThreads")) { param.num_threads = conf.at("numThreads"); }
```
cjnolet, line 218: Can we try to reuse the search params in hnswlib's source file for this?

cpp/bench/ann/src/raft/raft_cagra_hnswlib.cu
```diff
@@ -0,0 +1,22 @@
...
+//  *
+//  * Unless required by applicable law or agreed to in writing, software
+//  * distributed under the License is distributed on an "AS IS" BASIS,
+//  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  * See the License for the specific language governing permissions and
+//  * limitations under the License.
+//  */
+// #include "raft_cagra_hnswlib_wrapper.h"
+
+// namespace raft::bench::ann {
```
cjnolet, line 18: Blank file?

cpp/bench/ann/src/raft/raft_cagra_hnswlib_wrapper.h
```diff
@@ -0,0 +1,216 @@
...
+
+  ~RaftCagraHnswlib() noexcept {}
+
+  void build(const T* dataset, size_t nrow, cudaStream_t stream) final;
+
+  void set_search_param(const AnnSearchParam& param) override;
+
+  // TODO: if the number of results is less than k, the remaining elements of 'neighbors'
+  // will be filled with (size_t)-1
+  void search(const T* queries,
```
cjnolet, line 93: We should be able to reuse all this from hnswlib's wrapper, right? We should consolidate this stuff because having duplicated code all over the place is going to make this hard to maintain.

cpp/bench/ann/src/raft/raft_cagra_hnswlib_wrapper.h
```diff
@@ -0,0 +1,216 @@
...
+    index_.emplace(raft::neighbors::cagra::build(handle_, index_params_, dataset_view));
+    return;
+  }
+}
+
+template <typename T, typename IdxT>
+void RaftCagraHnswlib<T, IdxT>::set_search_param(const AnnSearchParam& param_)
+{
+  auto param        = dynamic_cast<const SearchParam&>(param_);
+  appr_alg_->ef_    = param.ef;
```
cjnolet, line 148: Just have this call `Hnswlib<T,IdxT>::set_search_param(param_);`  (or pull the logic into a shared helper functino).

cpp/cmake/modules/ConfigureCUDA.cmake
```diff
@@ -20,12 +20,12 @@ endif()
 # Be very strict when compiling with GCC as host compiler (and thus more lenient when compiling with
 # clang)
 if(CMAKE_COMPILER_IS_GNUCXX)
-  list(APPEND RAFT_CXX_FLAGS -Wall -Werror -Wno-unknown-pragmas -Wno-error=deprecated-declarations)
-  list(APPEND RAFT_CUDA_FLAGS -Xcompiler=-Wall,-Werror,-Wno-error=deprecated-declarations)
+  # list(APPEND RAFT_CXX_FLAGS -Wall -Werror -Wno-unknown-pragmas -Wno-error=deprecated-declarations)
```
cjnolet, line 23: I assume these changes will get reverted before this is merged?

cpp/cmake/thirdparty/get_hnswlib.cmake
```diff
@@ -26,6 +26,11 @@ function(find_and_configure_hnswlib)
                 COMMAND git clone --branch=v0.6.2 https://github.com/nmslib/hnswlib.git hnswlib-src
                 WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/_deps )
 
+        message("SOURCE ${CMAKE_CURRENT_SOURCE_DIR}")
+        execute_process (
+                COMMAND git apply ${CMAKE_CURRENT_SOURCE_DIR}/cmake/patches/hnswlib.patch
```
cjnolet, line 31: So how does this work w/ the original hnswlib benchmarks? We don't want to modify hnswlib for those, right? Should this patch only get applied for cagra+hnswlib? Maybe we need to check the code out into a different working directory for each and only apply the patch to one?

cpp/include/raft/neighbors/cagra_serialize.cuh
```diff
@@ -93,6 +93,23 @@ void serialize(raft::resources const& handle,
...
+               const index<T, IdxT>& index)
+{
+  detail::serialize_to_hnswlib<T, IdxT>(handle, os, index);
+}
+
+template <typename T, typename IdxT>
+void serialize_to_hnswlib(raft::resources const& handle,
+               const std::string& filename,
+               const index<T, IdxT>& index,
+               bool include_dataset = true)
```
cjnolet, line 108: `include_dataset` doesn't seem to be used. We should probalby just remove it.

cjnolet (changes requested):

cpp/cmake/thirdparty/get_hnswlib.cmake
divyegala: We introduce a parameter in hnswlib called `base_layer_only`. When `false`, the original hnswlib is unchanged and that's what hnswlib benchmarks use

cjnolet: That's fine so long as it's not modifying the basic behavior.

cjnolet (approved):

PR #1971, opened 2023-11-07
Add build and search params to `raft-ann-bench.data_export` CSVs

divyegala:
Additionally, build params are also appended to search CSVs.


python/raft-ann-bench/src/raft-ann-bench/data_export/__main__.py
```diff
@@ -62,6 +92,27 @@ def convert_json_to_csv_search(dataset, dataset_path):
                 "qps": df["items_per_second"],
             }
         )
+        for name in df:
+            if name not in skip_search_cols:
+                write[name] = df[name]
+
+        if os.path.exists(build_file):
```
cjnolet, line 99: Can we do an else here and print a warning when the corresponding build file could not be found?

cjnolet (approved):

PR #1981, opened 2023-11-09
ANN bench: use different offset for each thread

tfeher:
In ANN benchmark, when throughput mode is enabled, it is important that each thread works with a separate set of queries, otherwise cache effects can distort the results. This PR sets a different starting offset for each thread.


cjnolet (approved):
LGTM!

PR #1983, opened 2023-11-09
Add UCXX support

pentschev:
Add support for [UCXX](https://github.com/rapidsai/ucxx). It is our intention to soon switch from UCX-Py to UCXX and archive the former.

This PR adds support for UCXX on the C++ backend but retains the original UCX implementation for now (based on the UCP layer), moving to UCXX will simplify RAFT code a bit given the UCXX implementation requires fewer lines of boilerplate code.

On the Python front raft-dask tests are added for both UCX-Py (which there weren't any) and UCXX. The UCX-Py tests continue to use the UCX (UCP layer) implementation, whereas the UCXX tests use the UCXX C++ implementation.

When the switch is complete we can remove all previous UCX/UCX-Py code from the RAFT codebase. If for some reason using the UCX (UCP layer) is preferred on the C++ backend instead of the UCXX C++ implementation this is possible, but UCX-Py code will be archived and dropped in favor of the UCXX Python backend.


cjnolet:
@pentschev are you ready for this to be merged by 24.02?

pentschev:
> @pentschev are you ready for this to be merged by 24.02?

I'm hoping for that but we still need UCXX wheels which is still ongoing work. I can't say whether they'll be ready yet. On the UCXX side this PR is working as expected locally so if we're able to get wheels built I think we won't have much trouble getting this in.

vyasr:
It looks like ucx's config is unfortunately not idempotent. We'll need to implement a workaround for that. Otherwise everything is looking solid though.

pentschev:
> It looks like ucx's config is unfortunately not idempotent. We'll need to implement a workaround for that. Otherwise everything is looking solid though, pending tests of raft-dask wheels to see how well ucxx works at runtime.

@vyasr Can you elaborate in layman's terms what that means in this context? Is this something we can fix in UCXX or not possible given the need/use of specific properties in UCXX?

Thanks so much for the help in pushing this forward!

vyasr:
I'm not yet sure how to fix this in UCXX, will think on it this morning. There's probably some way it can be done. The longer-term fix really ought to be in UCX itself.

The failure in conda Python builds is fixable in UCXX's CMake.

pentschev:
Since we got blocked on releasing UCXX wheel for 24.02 this cannot make it either, retargeting to 24.04.

vyasr:
OK, I was able to fix the issues by modifying UCXX. We'll eventually need to undo those fixes once UCX is patched (see https://github.com/rapidsai/ucxx/issues/173), but until then what we have here should work.

pentschev:
> OK, I was able to fix the issues by modifying UCXX. We'll eventually need to undo those fixes once UCX is patched (see [rapidsai/ucxx#173](https://github.com/rapidsai/ucxx/issues/173)), but until then what we have here should work.

Thanks a lot Vyas! Those changes probably won't make their way in until UCX 1.17.x, maybe 1.16.x but I'll have to check. So I'll keep an eye on that and revert when we move to newer UCX.

AyodeAwe (approved):

quasiben:
@cjnolet is it too late to review this for 24.04 ?  We'll retarget for 24.06 but we'd still like your review before merging in

bdice (commented):
I only reviewed the CMake and versioning/dependency updates. They seem fine to me, but there are pending conflicts to resolve in `ci/release/update-version.sh` and `python/raft-dask/CMakeLists.txt`. This should probably be retargeted to 24.06, since the 24.04 release is happening this week, so I'll hold off on approving until the versions are changed.

pentschev:
> I only reviewed the CMake and versioning/dependency updates. They seem fine to me, but there are pending conflicts to resolve in `ci/release/update-version.sh` and `python/raft-dask/CMakeLists.txt`. This should probably be retargeted to 24.06, since the 24.04 release is happening this week, so I'll hold off on approving until the versions are changed.

Thanks @bdice , branch was retargeted, conflicts are resolved and versions updated. 🙂

pentschev:
The CI failures are not problems with the PR but rather recent changes in `rapids-dask-dependency`/`dask-cuda`. We're discussing how to better address it (short-term probably means adding `dask[dataframe]` as a `dask-cuda` dependency) and we should have a solution to make tests pass by tomorrow.

cjnolet:
I think these changes look fine but the real telling piece will be if cuML and cugraph both pass CI with the change.

pentschev:
> I think these changes look fine but the real telling piece will be if cuML and cugraph both pass CI with the change.

Agreed, once this PR is merged then https://github.com/rapidsai/cuml/pull/5697 can also be merged so we can have a better overview of how things are looking.

pentschev:
The error is now gone and all tests are passing again, so should be good for a review. @divyegala mentioned he would do that, no absolutely immediate pressure but would be good if we could have this merged some time this week or next the latest.

divyegala (approved):

ci/test_wheel_raft_dask.sh
```diff
@@ -6,12 +6,15 @@ set -euo pipefail
 mkdir -p ./dist
 RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"
 RAPIDS_PY_WHEEL_NAME="raft_dask_${RAPIDS_PY_CUDA_SUFFIX}" rapids-download-wheels-from-s3 ./dist
+ucxx_wheelhouse=$(RAPIDS_PY_WHEEL_NAME="ucxx_${RAPIDS_PY_CUDA_SUFFIX}" rapids-download-wheels-from-s3 ./local-raft-dask-dep)
```
vyasr, line 9:
```suggestion
ucxx_wheelhouse=$(RAPIDS_PY_WHEEL_NAME="ucxx_${RAPIDS_PY_CUDA_SUFFIX}" rapids-get-pr-wheel-artifact ucxx 226 python)
```

ci/test_wheel_raft_dask.sh
```diff
@@ -6,12 +6,15 @@ set -euo pipefail
 mkdir -p ./dist
 RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"
 RAPIDS_PY_WHEEL_NAME="raft_dask_${RAPIDS_PY_CUDA_SUFFIX}" rapids-download-wheels-from-s3 ./dist
+ucxx_wheelhouse=$(RAPIDS_PY_WHEEL_NAME="ucxx_${RAPIDS_PY_CUDA_SUFFIX}" rapids-get-pr-wheel-artifact ucxx 226 python)
```
vyasr, line 9:
```suggestion
ucxx_wheelhouse=$(RAPIDS_PY_WHEEL_NAME="ucxx_${RAPIDS_PY_CUDA_SUFFIX}" rapids-get-pr-wheel-artifact ucxx 226 python)
distributed_ucxx_wheelhouse=$(RAPIDS_PY_WHEEL_NAME="distributed_ucxx_${RAPIDS_PY_CUDA_SUFFIX}" rapids-get-pr-wheel-artifact ucxx 226 python)
```

ci/test_wheel_raft_dask.sh
```diff
@@ -6,12 +6,15 @@ set -euo pipefail
...
+
 
 # Download the pylibraft built in the previous step
 RAPIDS_PY_WHEEL_NAME="pylibraft_${RAPIDS_PY_CUDA_SUFFIX}" rapids-download-wheels-from-s3 ./local-pylibraft-dep
 python -m pip install --no-deps ./local-pylibraft-dep/pylibraft*.whl
 
-# echo to expand wildcard before adding `[extra]` requires for pip
-python -m pip install $(echo ./dist/raft_dask*.whl)[test]
+python -m pip install "raft_dask-${RAPIDS_PY_CUDA_SUFFIX}[test]>=0.0.0a0" --find-links dist/ --find-links "${ucxx_wheelhouse}"
```
vyasr, line 16:
```suggestion
python -m pip install "raft_dask-${RAPIDS_PY_CUDA_SUFFIX}[test]>=0.0.0a0" --find-links dist/ --find-links "${ucxx_wheelhouse}" --find-links "${distributed_ucxx_wheelhouse}"
```

conda/recipes/raft-dask/meta.yaml
vyasr, line 1: The `ucx` version specification in here isn't quite right. The `conda_build_config.yaml` file has the looser version pinning, so we'll always end up pulling 1.16 to build against as well as run. I think we want to always build against the oldest right? So we need to specify `==1.15` for the build.

ci/test_wheel_raft_dask.sh
```diff
@@ -11,7 +11,13 @@ RAPIDS_PY_WHEEL_NAME="raft_dask_${RAPIDS_PY_CUDA_SUFFIX}" rapids-download-wheels
...
 
-python -m pytest ./python/raft-dask/raft_dask/test
+# rapids-logger "pytest raft-dask"
+# python -m pytest ./python/raft-dask/raft_dask/test
+
+# rapids-logger "pytest raft-dask (ucx-py only)"
+# python -m pytest ./python/raft-dask/raft_dask/test --run_ucx
+
+rapids-logger "pytest raft-dask (ucxx only)"
+python -m pytest ./python/raft-dask/raft_dask/test --run_ucxx
```
vyasr, line 23: These flag names are a bit confusing. Do they really mean only run the ucx-py/ucxx tests? What happens if you don't pass either flag? Do _all_ tests run?

vyasr (changes requested):
One change on the ucx dependency in conda, otherwise I'm generally fine with getting this PR merged.

ci/test_wheel_raft_dask.sh
pentschev: Passing no flags run all Dask tests with `protocol="tcp"`, including those that are not specifically meant to test comms. With `--run_ucx`/`run_ucxx` _only_ tests specifically mean to test comms run with `protocol="ucx"`(UCX-Py)/`protocol="ucxx"`(UCXX)`. The reason for running them separately is that we need to create full Dask clusters with the respective comms configuration, setting clusters up take both time and, more importantly, memory, thus we cannot run all in one pass.

conda/recipes/raft-dask/meta.yaml
pentschev: As discussed offline, I don't think it truly makes sense to explicitly include `ucx` in the conda package as `raft-dask` will not be used without `ucx-py` or `ucxx` which will both bring `ucx` transitively, and thus reduce one more place where we have to explicitly synchronize UCX versions in the RAPIDS stack.

vyasr (approved):

Issue #1975, opened 2023-11-08
[ENH] Re-apply the matrix::select_k heuristics


achirkin:
Re-run and apply [select_k/generate_heuristic](https://github.com/rapidsai/raft/blob/branch-23.12/cpp/scripts/heuristics/select_k/generate_heuristic.ipynb).

After #1878 is merged, the radix-based implementation should be faster than before for some inputs.
If the radix-based implementation is faster than the legacy FAISS implementation in all cases, this would be a good moment to get rid of the old code. Otherwise, we should investigate further the reasons for being slower in the remaining cases.

benfred:
Done w/ #1985

PR #1985, opened 2023-11-10
Update select-k heuristic

benfred:
With https://github.com/rapidsai/raft/pull/1878 merged, the performance of the radix select algorithms is much improved and we no longer need to incorporate the faiss block select algorithm. With #1878 merged, faiss block select goes from being the 3rd ranked selection algorithm, to the 5th.

This regenerates the heuristic function with the latest benchmark times, and removes the faiss block select in favour of kWarpImmediate and kRadix11bitsExtraPass.


tfeher (approved):
Thanks @benfred for the update, the PR looks good to me!

achirkin:
@benfred Cool, just noticed this change! Does this close https://github.com/rapidsai/raft/issues/1975 ?  Also, is there anything still blocking us from removing the lefacy faiss neighbors::detail::select_k ?

benfred:
@achirkin I think we can close out #1975 with this change - and I've also removed the legacy neibhors::detail::select_k code in #2027

PR #1989, opened 2023-11-13
Adjusting end-to-end start time so it doesn't include stream creation time

cjnolet:
We can safely assume all cuda streams have been created before we run benchmark experiments. With the sensitivity to latencies for smaller batch sizes, this has an impact on overall throughput.


benfred (approved):
Lgtm!

PR #1993, opened 2023-11-14
SNMG ANN

viclafargue:
The goal of this PR is to implement a distributed (single-node-multiple-GPUs) implementation of ANN indexes. It will allow to `build`, `extend` and `search` an index on multiple GPUs.

Before building the index, the user has to choose between two modes :
- **Sharding mode** : The index dataset is split, each GPU trains its own index with its respective share of the dataset. This is intended to both increase the search throughput and the maximal size of the index.
- **Index duplication mode** : The index is built once on a GPU and then copied over to others. Alternatively, the index dataset is sent to each GPU to be built there. This intended to increase the search throughput.


cpp/include/raft/neighbors/ann_mg.cuh
```diff
@@ -0,0 +1,24 @@
...
+ */
+#pragma once
+
+#ifndef RAFT_EXPLICIT_INSTANTIATE_ONLY
+#include "ann_mg-inl.cuh"
+#endif
+
+#ifdef RAFT_COMPILED
+#include "ann_mg-ext.cuh"
+#endif
```
tfeher, line 24: Thanks @viclafargue for working on this. Just a quick comment on header organization.

I am not convinced that we need to complicate the interface with separate `-inl` and `-ext` headers. According to our [developer guide](https://github.com/rapidsai/raft/blob/branch-23.12/docs/source/developer_guide.md#header-organization-of-expensive-function-templates):
>Any header file that defines an expensive function template (say expensive.cuh) should be split in three parts: expensive.cuh, expensive-inl.cuh, and expensive-ext.cuh

I do not think that we are defining expensive to compile functions here.  I believe we will simply wrap the existing ANN methons and dispatch them onto multiple devices. I expect the code here to be relatively quick to compile, and therfore we could stay with a simple header structure.

viclafargue:
The PR is ready for a first review. In its current state, it implements the build, extend and search ANN methods (IVF-Flat and IVF-PQ only for now) in index duplication and sharding mode. For now, the index duplication mode only works by copying the index dataset over and building the index on each GPU separately. I am now looking to improve the API in such a way that it would allow to build the index on a GPU and copy it over. Serialization on disk would work, but does not seem ideal. Then, transferring the indexes attributes through NCCL seem like not very safe. What would you recommend?

cpp/include/raft/neighbors/detail/ann_mg.cuh
```diff
@@ -0,0 +1,645 @@
...
+  }
+
+  void build(const ann::index_params* index_params,
+             raft::host_matrix_view<const T, IdxT, row_major> index_dataset)
+  {
+    if (mode_ == INDEX_DUPLICATION) {
+      for (int rank = 0; rank < num_ranks_; rank++) {
+        RAFT_CUDA_TRY(cudaSetDevice(dev_ids_[rank]));
+        auto& ann_if = ann_interfaces_.emplace_back();
+        ann_if.build(dev_resources_[rank], index_params, index_dataset);
```
lijinf2, line 299: Is it every GPU will copy the host dataset into device so the total number of copies will be num_ranks_?

lijinf2: Another related question, will GPU 1 not start and wait until GPU 0 build finishes? If that's the case, the total runtime of the for loop seems to be single GPU build time * num_ranks_.

cpp/include/raft/neighbors/detail/ann_mg.cuh
```diff
@@ -0,0 +1,645 @@
...
+      RAFT_CUDA_TRY(cudaSetDevice(dev_ids_[rank]));
+      dev_resources_.emplace_back();
+    }
+  }
+
+  void init_nccl_clique() {
+    RAFT_NCCL_TRY(ncclCommInitAll(nccl_comms_.data(), num_ranks_, dev_ids_.data()));
+    for (int rank = 0; rank < num_ranks_; rank++) {
+      RAFT_CUDA_TRY(cudaSetDevice(dev_ids_[rank]));
+      raft::comms::build_comms_nccl_only(&dev_resources_[rank], nccl_comms_[rank], num_ranks_, rank);
```
lijinf2, line 281: The NCCL initialization seems to be "one process multiple GPUs". 

Is it possible to adapt it to "one process or thread one GPU"? May have to use something like std::thread. But the benefit is to enable the APIs of the PR to be reusable to Dask/Spark. Both Dask and Spark currently comply with one process one GPU when initializing NCCL.

lijinf2:
Thank you Victor. I have learned a lot from the code! I like the idea of combining three algorithms into one unified interface. A few questions to make myself more familiar with the PR and design choice. Will be wonderful if Spark Rapids ML can leverage the APIs in this PR.

cpp/include/raft/neighbors/detail/ann_mg.cuh
viclafargue: Yes exactly, in the index duplication mode the dataset is copied in full to each GPU for training. An alternative method is to train a model locally, serialize it and distribute it with either one of the `distribute_flat`, `distribute_pq` or `distribute_cagra` functions.

> Another related question, will GPU 1 not start and wait until GPU 0 build finishes? If that's the case, the total runtime of the for loop seems to be single GPU build time * num_ranks_.

The `build`, `extend` and `search` functions take in a `handle` parameter containing the CUDA stream on which the kernels should be launched. This is supposed to be asynchronous and allow fast switching of GPUs. However, this has not yet been tested. An actual benchmark would be necessary to confirm that things scale as expected.

cpp/include/raft/neighbors/detail/ann_mg.cuh
viclafargue: The single process solution was better suited to implement the much requested feature in RAFT for now. But, I agree that in the end we should definitely look into making it possible to make things run on Dask/Spark. This would probably involve the use of multiple processes/threads and a much broader use of NCCL.
cc @cjnolet

python/raft-ann-bench/src/raft-ann-bench/run/conf/sift-128-euclidean.json
```diff
@@ -472,6 +472,19 @@
         {"nprobe": 2000}
       ]
     },
+    {
+      "name": "raft_ann_mg.nlist16384",
+      "algo": "raft_ann_mg",
```
cjnolet, line 477: I think `raft_ivf_flat_mg` and `raft_ivf_pq_mg` might make more sense here.

cjnolet (changes requested):
Hey Victor. This is a pretty sizeable PR so my suggestions will come in a few different passes through the changes. I took an initial look. Overall I think it's headed in the right direction. A lot of my suggestions so far are mechanical things. I'll take a closer look at the impl next.

cpp/include/raft/neighbors/detail/ann_mg.cuh
tfeher: We have a problem here. Building a GPU index is not only a GPU operation. It can have significant CPU work (e.g. CAGRA graph optimization, NN descent data pre/post proc, host side sub-sampling of IVF-methods).

Furthermore there are cases where our algorithms block CPU thread while waiting for GPU kernels to finish ( e.g. wait for return values that determine memory allocation size).

We cannot launch `build`  on a single CPU thread and expect that it will run parallel, because the GPU ops are asynchronous. Most are, but the the few that I cite above will essentially serialize the whole process.

At least we would need different worker threads for each GPU stream. But I would recommend one process per GPU.

We should also keep in mind that `build` is multi-threaded. It spawns OpenMP threads to help shuffle data in host memory (singe thread is not enough to saturate mem bandwidth). We should document that this can be controlled with the `OMP_NUM_THREADS` variable.

cpp/include/raft/neighbors/detail/ann_mg.cuh
```diff
@@ -0,0 +1,712 @@
...
+      IdxT n_rows           = index_dataset.extent(0);
+      IdxT n_cols           = index_dataset.extent(1);
+      IdxT n_rows_per_shard = (n_rows + num_ranks_ - 1) / num_ranks_;
+      IdxT offset           = 0;
+      for (int rank = 0; rank < num_ranks_; rank++) {
+        RAFT_CUDA_TRY(cudaSetDevice(dev_ids_[rank]));
+        n_rows_per_shard       = std::min(n_rows_per_shard, n_rows - offset);
+        const T* partition_ptr = index_dataset.data_handle() + (offset * n_cols);
+        auto partition         = raft::make_host_matrix_view<const T, IdxT, row_major>(
+          partition_ptr, n_rows_per_shard, n_cols);
```
tfeher, line 354: Is it really the job of `ann_mg_index` to shard the data? An alternative is to provide this sharding code as a helper function, that the user has to explicitly call to provide already sharded input to `ann_mg_index::build`

On single node, it make sense to add sharding inside `build`. The hole dataset can be loaded (`mmap`-ed) by the host, and we can shard the data here.

On multiple node the data has to be loaded by each node. Due to its large size, each node would load only their portion of the data. If we allow `ann_mp_index::build` to take that sharded data, then we can already support multi node.

(Yes, I know that multi note is out of scope of the current PR, but if we plan to add that in the future, then we should probably agree already now on a build API that would easily enables that).

cpp/include/raft/neighbors/detail/ann_mg.cuh
```diff
@@ -0,0 +1,712 @@
...
+        auto h_trans = std::vector<IdxT>(num_ranks_);
+        IdxT translation_offset = 0;
+        for (int rank = 0; rank < num_ranks_; rank++) {
+          h_trans[rank] = translation_offset;
+          translation_offset += ann_interfaces_[rank].size();
+        }
+        auto d_trans = raft::make_device_vector<IdxT, IdxT>(root_handle_, num_ranks_);
+        raft::copy(d_trans.data_handle(), h_trans.data(), num_ranks_, resource::get_cuda_stream(root_handle_));
+        auto translations = std::make_optional<raft::device_vector_view<IdxT, IdxT>>(d_trans.view());
+        raft::neighbors::brute_force::knn_merge_parts<float, IdxT>(root_handle_,
```
tfeher, line 498: Out of scope for the current PR, but we might consider as a follow up: IVF-PQ or CAGRA-Q only return approximate distances. While merging parts based on the approximate distances, we might be throwing out good neighbors due to innacurate distance values. If we plan to do refinement, then we can treat the `in_neighbors` as candidates for refinement, and run refinement directly instead of calling `knn_merge_parts`.

tfeher (changes requested):
Thanks Victor for the PR! The code is well structured and clean, but I want to point out a few issues that we need to discuss (see below). I think these can be conceptually easily fixed by adhering to our One Process per GPU principle.

cpp/include/raft/neighbors/detail/ann_mg.cuh
```diff
@@ -0,0 +1,712 @@
...
+ public:
+  ann_mg_index(const std::vector<int>& dev_list, dist_mode mode = SHARDING)
+    : mode_(mode),
+      root_rank_(0),
+      num_ranks_(dev_list.size()),
+      dev_ids_(dev_list),
+      nccl_comms_(dev_list.size())
+  {
+    init_device_resources();
+    init_nccl_clique();
```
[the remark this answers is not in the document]
tfeher: Tagging @cjnolet for input.

cpp/test/neighbors/ann_mg.cuh
```diff
@@ -0,0 +1,386 @@
...
+        h_index_dataset.data(), ps.num_db_vecs, ps.dim);
+      auto query_dataset = raft::make_host_matrix_view<const DataT, uint32_t, row_major>(
+        h_query_dataset.data(), ps.num_queries, ps.dim);
+      auto neighbors = raft::make_host_matrix_view<uint32_t, uint32_t, row_major>(
+        indices_ann_32bits.data(), ps.num_queries, ps.k);
+      auto distances = raft::make_host_matrix_view<float, uint32_t, row_major>(
+        distances_ann.data(), ps.num_queries, ps.k);
+
+      /*
+      TODO : fix CAGRA serialization issue
```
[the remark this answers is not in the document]
viclafargue: Actually solved

cpp/include/raft/neighbors/detail/ann_mg.cuh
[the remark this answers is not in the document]
viclafargue: Created the `nccl_clique` helper struct to solve this.

jameslamb (approved):
Giving this a `packaging-codeowners` approval... building/packaging changes look fine to me.

Issue #1948, opened 2023-11-01
[BUG] Cagra index search segfaults for vectors with 1024+ dimensions


Ngalstyan4:
**Describe the bug**
Cagra index creation via the CPP api (`cagra::build`) seems to succeed for vectors of dimension 1025 but search (`cagra::search`) results in a Segmentation Fault.

The issues exists both in the cuda API and in the runtime API of cagra.

**Steps/Code to reproduce bug**

Change [this line](https://github.com/rapidsai/raft/blob/branch-23.12/cpp/template/src/cagra_example.cu#L80) in cagra example  to `int64_t n_dim   = 1025;`, compile and run. This should cause a segmentation fault.

To test on the runtime API, use `raft::runtime::neighbors::cagra::{build, search}` in place of `raft::neighbors::cagra::{build, search}`.

**Expected behavior**
Expected the index creation and search to succeed without any issues, as it does for smaller dimensional vectors.

**Environment details (please complete the following information):**
 - Environment location: GCP, Debian 11, Cuda compilation tools, release 12.2, V12.2.140, Tesla V100-SXM2-16GB
 - Method of RAFT install: install from source

**Additional context**
This is important because OpenAI embeddings are 1536 dimensional. There are many other models that produce 1024+ dimensional embeddings.

The fix might be as simple as adding an element to [this array](https://github.com/rapidsai/raft/blob/945355d707d60c2e507899f262e87eb00d8624d8/cpp/src/neighbors/detail/cagra/search_single_cta_00_generate.py#L84) but I have not tried it.

cjnolet:
@Ngalstyan4 CAGRA CUDA kernels are templated on the dataset dimension and we need to define a new template argument for larger dimensions.

Some of these RAG embeddings go up to extremely large numbers of dimensions. Any specific range you'd are targeting or just as high as we can go?

Ngalstyan4:
1536 dimensional openai embeddings are probably most widely used these days, so having support for that out of the box would be the primary ask.

It would also be great to throw an informative error when the templates are used with unsupported vector sizes instead of segfaulting, as it happens now.

cjnolet:
> It would also be great to throw an informative error when the templates are used with unsupported vector sizes instead of segfaulting, as it happens now.

[...]

PR #1994, opened 2023-11-15
Add support for 1024+ dim vectors in CAGRA search

enp1s0:
This PR updates the CAGRA search implementation to support 1024+ dim vectors.
For 1024+ dim vectors, the distance between a vector and the query vector is calculated by splitting the vector into multiple 1024 dim vectors and accumulating the distances of each sub-vector.

Rel: #1948


enp1s0:
- [ ] single-CAT
- [ ] multi-CAT
- [ ] multi-kernel

enp1s0:
No significant performance drop is observed
s1024+

cjnolet (approved):
LGTM! This is an important feature and we have folks from Lucene and Lantern waiting on this. These changes look good, but we can follow up if anyone else sees any issues.

tfeher (commented):
Thanks @enp1s0 for the PR, it looks good to me.

PR #1996, opened 2023-11-15
Add filtering to vector search tutorial

lowener:


docs/source/vector_search_tutorial.md
```diff
@@ -340,4 +340,32 @@ The below example specifies the total number of bytes that RAFT can use for temp
 
 std::shared_ptr<rmm::mr::managed_memory_resource> managed_resource;
 raft::device_resource res(managed_resource, std::make_optional<std::size_t>(3 * 1024^3));
-```
\ No newline at end of file
+```
+
+### Filtering
+
+As of RAFT 23.10, support for pre-filtering of neighbors has been added to ANN index. This feature is useful for applications that want to remove vectors already added to the index, or to enable access control in searches. The filtering is done by applying a predicate function on the GPU. The following example demonstrates how to use the filtering API:
```
cjnolet, line 347: Can you add some more information to this description?  Think about it like walking the user through the use-case like they are reading a book on the topic. THings such as:
1. Moreso than just removing vectors already added to the index, what types of use-cases does this enable? (e.g. hybrid search, multi-valued keys, deletion, security label filtering, etc... 
2. What types of functions are available and why? 
3. What's a bitset and why would one use that primitive?

A section like this is where it's really important that we spell out the capabilities and potential here.

cjnolet (approved):

PR #1997, opened 2023-11-15
Remove static checks for serialization size

cjnolet:
This PR removes static checks for serialization size. Upstream changes like https://github.com/rapidsai/rmm/pull/1370 have altered these sizes and break RAFT CI. An alternative approach to verifying serialization will be developed.

### DIFF
--- a/cpp/include/raft/neighbors/detail/cagra/cagra_serialize.cuh
+++ b/cpp/include/raft/neighbors/detail/cagra/cagra_serialize.cuh
@@ -31,19 +31,7 @@
 
 namespace raft::neighbors::cagra::detail {
 
-constexpr int serialization_version = 4;
-
-// NB: we wrap this check in a struct, so that the updated RealSize is easy to see in the error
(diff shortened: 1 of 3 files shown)


bdice:
There are more places to fix:
```
  /home/coder/raft/cpp/include/raft/neighbors/detail/cagra/cagra_serialize.cuh(40): error: static assertion failed with "The size of the index struct has changed since the last update; paste in the new size and consider updating the serialization logic"
            detected during instantiation of class "raft::neighbors::cagra::detail::check_index_layout<RealSize, ExpectedSize> [with RealSize=216UL, ExpectedSize=200UL]" 
```

Also I believe the serialization version is expected to increment every time this changes. See https://github.com/rapidsai/raft/pull/1964 for an example of the changes we'll need.

divyegala (approved):

harrism (approved):
Thanks for removing these checks. Will make life easier upstream.

Issue #1842, opened 2023-09-22
[FEA] Use cache-oblivious copies for arbitrary copies in raft::copy


wphicks:
As a follow up to PR #1818, the logic for arbitrary host-to-host copies as well as the custom device kernel would benefit from cache-oblivious transfers. This would be almost trivial with #501, which (as I understand it) would extend submdspan support to arbitrary layouts. On the host side, the smallest submdspan accessed in the cache-oblivious transfer could (generally speaking) be SIMD accelerated, as well.

PR #1999, opened 2023-11-15
Implement maybe-owning multi-dimensional container (mdbuffer)

### LINKED ISSUE #1602: [FEA] mdbuffer implementation in RAFT
Linking to #1408. `mdbuffer` differs from `buffer` in that it supports multiple dimenions up front. This is needed to ease integration of RAFT into various libraries downstream.

wphicks:
### What is mdbuffer?

This PR introduces a maybe-owning multi-dimensional abstraction called `mdbuffer` to help simplify code that _may_ require an `mdarray` but only if the data are not already in a desired form or location.

As a concrete example, consider a function `foo_device` which operates on memory accessible from the device. If we wish to pass it data originating on the host, a separate code path must be created in which a `device_mdarray` is created and the data are explicitly copied from host to device. This leads to a proliferation of branches as `foo_device` interacts with other functions with similar requirements.

As an initial simplification, `mdbuffer` allows us to write a single template that accepts an `mdspan` pointing to memory on either host _or_ device and routes it through the same code:
```c++
template <typename mdspan_type>
void foo_device(raft::resources const& res, mdspan_type data) {
  auto buf = raft::mdbuffer{res, raft::mdbuffer{data}, raft::memory_type::device};
  // Data in buf is now guaranteed to be accessible from device.
  // If it was already accessible from device, no copy was performed. If it
  // was not, a copy was performed.

  some_kernel<<<...>>>(buf.view<raft::memory_type::device>());

  // It is sometimes useful to know whether or not a copy was performed to
  // e.g. determine whether the transformed data should be copied back to its original
  // location. This can be checked via the `is_owning()` method.
  if (buf.is_owning()) {
    raft::copy(res, data, buf.view<raft::memory_type::device>());
  }
}

foo_device(res, some_host_mdspan);  // Still works; memory is allocated and copy is performed
foo_device(res, some_device_mdspan);  // Still works and no allocation or copy is required
foo_device(res, some_managed_mdspan);  // Still works and no allocation or copy is required
```

While this is a useful simplification, it still leads to a proliferation of template instantiations. If this is undesirable, `mdbuffer` permits a further consolidation through implicit conversion of an mdspan to an mdbuffer:

```c++
void foo_device(raft::resources const& res, raft::mdbuffer<float, raft::matrix_extent<int>>&& data)
{ auto buf = raft::mdbuffer{res, data, raft::memory_type::device};
  some_kernel<<<...>>>(buf.view<raft::memory_type::device>());
  if (buf.is_owning()) {
    raft::copy(res, data, buf.view<raft::memory_type::device>());
  }
}

// All of the following work exactly as before but no longer require separate template instantiations
foo_device(res, some_host_mdspan);
foo_device(res, some_device_mdspan);
foo_device(res, some_managed_mdspan);
```

`mdbuffer` also offers a simple way to perform runtime dispatching based on the memory type passed to it using standard C++ patterns. While mdbuffer's `.view()` method takes an optional template parameter indicating the mdspan type to retrieve as a view, that parameter can be omitted to retrieve a `std::variant` of all mdspan types which may provide a view on the `mdbuffer`'s data (depending on its memory type). We can then use `std::visit` to perform runtime dispatching based on where the data are stored:

```c++
void foo(raft::resources const& res, raft::mdbuffer<float, raft::matrix_extent<int>>&& data) {
  std::visit([](auto view) {
    if constexpr (typename decltype(view)::accessor_type::is_device_accessible) {
      // Do something with these data on device
    } else {
      // Do something with these data on host
    }
  }, data.view());
}
```

In addition to moving data among various memory types (host, device, managed, and pinned currently), `mdbuffer` can be used to coerce data to a desired in-memory layout or to a compatible data type (e.g. floats to doubles). As with changes in the memory type, a copy will be performed if and only if it is necessary.

```c++
template <typename mdspan_type>
void foo_device(raft::resources const& res, mdspan_type data) {
  auto buf = raft::mdbuffer<float, raft::matrix_extent<int>, raft::row_major>{res,
raft::mdbuffer{data}, raft::memory_type::device};
  // Data in buf is now guaranteed to be accessible from device, and
  // represented by floats in row-major order.

  some_kernel<<<...>>>(buf.view<raft::memory_type::device>());

  // The same check can be used to determine whether or not a copy was
  // required, regardless of the cause. I.e. if the data were already on
  // device but in column-major order, the is_owning() method would still
  // return true because new storage needed to be allocated.
  if (buf.is_owning()) {
    raft::copy(res, data, buf.view<raft::memory_type::device>());
  }
}
```

### What mdbuffer is **not**
`mdbuffer` is **not** a replacement for either `mdspan` or `mdarray`. `mdspan` remains the standard object for passing data views throughout the RAFT codebase, and `mdarray` remains the standard object for allocating new multi-dimensional data. This is reflected in the fact that `mdbuffer` can _only_ be constructed from an existing `mdspan` or `mdarray` or another `mdbuffer`. `mdbuffer` is intended to be used solely to simplify code where data _may_ need to be copied to a different location.

### Follow-ups

-  I have omitted the mdbuffer-based replacement for and generalization of `temporary_device_buffer` since this PR is already enormous. I have this partially written however, and I'll post a link to its current state to help motivate the changes here.
- For all necessary copies, `mdbuffer` uses `raft::copy`. For _some_ transformations that require a change in data type or layout, `raft::copy` is not fully optimized. See #1842 for more information. Optimizing this will be an important change to ensure that `mdbuffer` can be used with absolutely minimal overhead in all cases. These non-optimized cases represent a small fraction of the real-world use cases we can expect for `mdbuffer`, however, so there should be little concern about beginning to use it as is.
- `std::visit`'s performance for a small number of variants is sometimes non-optimal. As a followup, it would be good to benchmark `mdbuffer`'s current performance and compare to internal use of a `visit` implementation that uses a `switch` on the available memory types.

Resolve #1602


cpp/include/raft/neighbors/detail/ivf_pq_serialize.cuh
```diff
@@ -42,9 +42,10 @@ constexpr int kSerializationVersion = 3;
 // message.
 template <size_t RealSize, size_t ExpectedSize>
 struct check_index_layout {
-  static_assert(RealSize == ExpectedSize,
+  /* static_assert(RealSize == ExpectedSize,
```
wphicks, line 45: REVERT BEFORE MERGING. Workaround for current serialization compilation issues.

cpp/include/raft/neighbors/detail/ivf_flat_serialize.cuh
```diff
@@ -40,9 +40,10 @@ constexpr int serialization_version = 4;
 // message.
 template <size_t RealSize, size_t ExpectedSize>
 struct check_index_layout {
-  static_assert(RealSize == ExpectedSize,
+  /*static_assert(RealSize == ExpectedSize,
```
wphicks, line 43: REVERT BEFORE MERGING. Workaround for current serialization compilation issues.

cpp/include/raft/neighbors/detail/cagra/cagra_serialize.cuh
```diff
@@ -37,9 +37,10 @@ constexpr int serialization_version = 3;
 // message.
 template <size_t RealSize, size_t ExpectedSize>
 struct check_index_layout {
-  static_assert(RealSize == ExpectedSize,
+  /*static_assert(RealSize == ExpectedSize,
```
wphicks, line 40: REVERT BEFORE MERGING. Workaround for current serialization compilation issues.

wphicks:
Link to partial and incomplete `copyback_buffer` just as a motivating example of how `mdbuffer` can be used to create higher-order objects: https://github.com/wphicks/raft/pull/1

wphicks:
@divyegala asked for a motivating example of when we might want to use runtime-dispatching based on the current memory type of an mdbuffer as opposed to just accepting the mdspan template parameter. To provide a concrete real-world example, in the FIL Backend for Triton, the core Triton server allocates the I/O memory and passes it to the backend as a pointer with extents and an enum indicating whether the memory is on host or device.

Because FIL offers optimized CPU and GPU implementations, and CPU execution for simple models with small batches _can_ be faster than the D2H round-trip, we do one of three things with the I/O pointers:

1. If the data are already on device, we immediately launch GPU execution.
2. If the data are on host, and the batch size is below a model-specific threshold, we immediately launch CPU execution.
3. If the data are on host, and the batch size is above that threshold, we copy the data to device, launch GPU execution, and then copy the data back to host.

Let's look at how we would do (a simplified version of) that with and without runtime dispatching. First, using just the `view` template parameter:
```
enum struct triton_mem_type {
  device,
  host
};

template<mdspan_type>
fil_infer();

template<>
fil_infer(host_matrix) {
  // CPU implementation
}

template<>
fil_infer(device_matrix) {
  // GPU implementation
}

void process_data_from_triton(float* data, int rows, int cols, triton_mem_type mtype) {
  if(mtype == triton_mem_type::host) {
    auto input_raw = rafft::make_host_mdspan(...);
    if (rows < THRESHOLD) {
      fil_infer(input_raw);
    } else {
      auto input_device = mdbuffer(input_raw, raft::memory_type::device);
      fil_infer(input_device.view<raft::memory_type::device>());
    }
  } else {
    auto input = raft::make_device_mdspan(...);
    fil_infer(input);
  }
}
```
There's absolutely nothing wrong with this code, but it is also possible to restructure it a bit with runtime dispatching:

```
void process_data_from_triton(float* data, int rows, int cols, triton_mem_type mtype) {
  auto input = mdbuffer<float, raft::matrix_extent<int>>{};
  if(mtype == triton_mem_type::host) {
    input = mdbuffer{res, mdbuffer{raft::make_host_mdspan(...)}, rows > THRESHOLD ? memory_type::device : memory_type::host};
  } else {
    input = mdbuffer{raft::make_device_mdspan(...)};    
  }
  std::visit([](auto&& view) { fil_infer(view); }, input);
}
```

Neither one is clearly superior in all situations. In the specific case of the FIL backend, however, the processing of Triton input into an mdspan happens in a completely different library, so the runtime dispatching offers the chance to do this:

```
// Other Library
auto get_triton_data(float* data, int rows, int cols, triton_mem_type mtype) {
  auto data = mdbuffer<float, raft::matrix_extent<int>>{};
  if(mtype == triton_mem_type::host) {
    data = mdbuffer{raft::make_host_mdspan(...)};
  } else {
    data = mdbuffer{raft::make_device_mdspan(...)};    
  }
  return data;
}

// FIL Backend
void process_batch(float* data, int rows, int cols, triton_mem_type mtype) {
  auto final_mem_type = (mtype == triton_mem_type::host && rows <= THRESHOLD) ? memory_type::host : memory_type::device;
  auto input = mdbuffer{get_triton_data(...), final_mem_type};
  std::visit([](auto&& view) { fil_infer(view); }, input);
}
```

In general, within code internal to RAFT, we probably _do_ want to generally stick with compile-time dispatching based on a templated mdspan type, since that is consistent with existing code. For external code, though, especially when there is a separation of concerns between coercing data into the desired form and actually calling inference on it, this form of runtime dispatching can be useful. Each strategy can be used when and where it best simplifies the code.

cpp/include/raft/core/memory_type.hpp
```diff
@@ -49,4 +79,29 @@ auto constexpr memory_type_from_access()
...
+  auto result = memory_type::host;
+#ifndef RAFT_DISABLE_CUDA
+  auto* void_ptr = static_cast<void*>(nullptr);
+  if constexpr (std::is_const_v<T>) {
+    void_ptr = const_cast<void*>(static_cast<void const*>(ptr));
+  } else {
+    void_ptr = static_cast<void*>(ptr);
+  }
+  auto attrs = cudaPointerAttributes{};
+  RAFT_CUDA_TRY(cudaPointerGetAttributes(&attrs, void_ptr));
```
divyegala, line 95: Does this throw a compiler error if the pointer is not cast to `void`?

cpp/include/raft/core/managed_mdspan.hpp
```diff
@@ -0,0 +1,96 @@
...
+template <typename ElementType,
+          typename Extents,
+          typename LayoutPolicy   = layout_c_contiguous,
+          typename AccessorPolicy = std::experimental::default_accessor<ElementType>>
+using managed_mdspan = mdspan<ElementType, Extents, LayoutPolicy, managed_accessor<AccessorPolicy>>;
+
+template <typename T, bool B>
+struct is_managed_mdspan : std::false_type {};
+template <typename T>
+struct is_managed_mdspan<T, true> : std::bool_constant<T::accessor_type::is_managed_accessible> {};
```
divyegala, line 41: Maybe we should remove `is_managed_accessible` and just check for host and device accessibility

cpp/include/raft/core/pinned_mdspan.hpp
```diff
@@ -0,0 +1,164 @@
...
+
+/**
+ * @brief std::experimental::mdspan with pinned tag to avoid accessing incorrect memory location.
+ */
+template <typename ElementType,
+          typename Extents,
+          typename LayoutPolicy   = layout_c_contiguous,
+          typename AccessorPolicy = std::experimental::default_accessor<ElementType>>
+using pinned_mdspan = mdspan<ElementType, Extents, LayoutPolicy, pinned_accessor<AccessorPolicy>>;
+
```
divyegala, line 38: No type traits?

cpp/include/raft/util/variant_utils.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+template <typename... types1, typename... types2>
+struct concatenated_variant<std::variant<types1...>, std::variant<types2...>> {
+  using type = std::variant<types1..., types2...>;
+};
+
+template <typename variant1, typename variant2>
+using concatenated_variant_t = typename concatenated_variant<variant1, variant2>::type;
+
+template <typename visitor_t, typename variant_t, std::size_t index = std::size_t{}>
```
divyegala, line 33: What is the optimization here? Being able to provide the starting point? For example, providing `2` if we only want `managed` or `pinned` views

divyegala (commented):
First pass of reviews for everything except mdbuffer/tests

divyegala:
Let's encapsulate the run-time dispatch via a type-dispatcher inspired by libcudf.

```
struct foo {

  template <typename mdspan, typename = std::enable_if_t<mdspan::is_device_accessible>
  void operator()() { /* do something */}

  template <typename mdspan, typename = std::enable_if_t<not mdspan::is_device_accessible>
  void operator()() { throw ...; /* unreachable */ }
};

void bar() {
  visiter(foo{}, mdbuffer, Args... /* for foo() */, index /* optional for fast visit optimization */);
}
```

cpp/include/raft/util/variant_utils.hpp
wphicks: That's an interesting thought, and I had not even thought about that potential use case (though I imagine it might get confusing without some additional wrapping). Cool! But not what I was going for. ;)

No, this is to deal with a limit on inlining that is accidentally imposed by the computational complexity requirements in the standard for `std::visit`. The short version is that the requirement essentially forces stdlib implementations to create a jump table when they could inline instead.

This implementation converts the dispatch to a bunch of if/else branches which _may_ still permit inlining. That being said, I actually do not currently use this anywhere in the current implementation, so I'm happy to just remove it from the PR.

Later on, I'd like to review each place that we use `std::visit` in the actual mdbuffer implementation and see if they could benefit from this optimization. I suspect for most of them, they just wouldn't benefit from inlining anyway, but it would be good to benchmark and make sure that the core mdbuffer implementation is as close to speed-of-light as possible.

wphicks:
> Let's encapsulate the run-time dispatch via a type-dispatcher inspired by libcudf.

Love this idea! Can you point me to the implementation in libcudf? I'm having trouble visualizing how to structure this generically.

Ideas that I've toyed with so far:

- Have the template accept a variant of lambdas and dispatch according to variant index (unwieldy; may require specifying extra unused lambdas or repeating logic)
- Have the template accept a generic lambda (no clear advantage relative to a visit call)

cpp/include/raft/core/pinned_mdspan.hpp
wphicks: I'm just realizing now that we should probably eliminate the type traits for `managed` as well, or else introduce a memory type constexpr member to the accessor. There's no way to distinguish between managed and pinned memory on the basis of host/device access, which brings us back to the `universal` vocabulary we had been discussing.

Hmmm... I'm very conflicted on this one. On the one hand, the memory type is not really relevant to how the data are accessed, so it feels like it doesn't belong on the accessor type. On the other hand, practically speaking, we may actually really care about the distinction between managed and pinned mdspans in certain cases. Thoughts?

I'm slightly leaning toward just eliminating the managed type traits. That would encourage people to dispatch based on memory accessibility, not the specific memory type. If we later added other memory types, that might be the more sustainable way to go.

divyegala: I think we are facing a conundrum here that I'm unable to come up with a good answer for as of now. While I agree that we need to let users pick from these types when the usability so desires that the difference between managed/pinned matters, what about the cases that don't?
Problems that will work as-is:
1. Algorithms (like NN Descent) which have sections that **only** work with pinned/managed memory. This is easier to solve as these APIs can overload on type-traits of `is_host_accessible && is_device_accessible`
2. `host/device` functions that can work with any kind of memory, so you should be able to pass any variant of `mdspan` (again type-trait of point 1 suffices)

Problems that won't work after PR merge:
1. Current APIs that explicitly use `device_matrix_view`: we should be able to pass `managed/pinned_matrix_view` seamlessly to this but it won't work directly constructing a `managed/pinned_mdarray`, the user will have to use `mdbuffer` and get the right `mdspan` variant

This problem is specifically why I first thought of a `universal_mdspan` but that is an incomplete solution as well. What we need to figure out is if there is a way to implicitly convert `pinned/managed_mdspan` to `device/host_mdspan`. If we can't do it implicitly, then explicitly is easy enough. We can add overloads to `make_device/host_*` that accept any `mdspan` that satisfies the correct type-trait and returns a pure device or host-only view.

The advantage of the explicit function is that we formalize the back-bone of our APIs as device/host/both variants, and give users the tools to interchange our types when possible for APIs that DO NOT provide overloads for all 3. For APIs that already provide device/host overloads, the task then becomes as simple as adding a third overload.

cpp/include/raft/util/variant_utils.hpp
wphicks: Aha! Now I'm understanding a bit better. There are actually a couple different issues that are getting pulled together here. Let's start with this one:

> For example, pinned should be able to provide an mdspan for all APIs that accept host/device/pinned.

There's actually a bug in the current implementation because [this line here](https://github.com/rapidsai/raft/blob/branch-23.12/cpp/include/raft/core/memory_type.hpp#L23) does not label pinned memory as device-accessible (Which blames to me. Oops!). I'll fix that up, but the point still stands for both pinned and managed memory.

Let's imagine you have three functions: `foo_host`, `foo_device`, and `foo_universal`. You want to invoke those functions with four different mdspans: `host_md`, `device_md`, `managed_md`, and `pinned_md`. The following wrappers will work for all 12 combinations:

```
template <typename mdspan_type>
void wrapper_host(mdspan_type input) {
   auto buf = mdbuffer(res, input, memory_type::host); // No copy for host/managed/pinned input; copy for device input
   foo_host(buf.view<memory_type::host>());
}

template <typename mdspan_type>
void wrapper_device(mdspan_type input) {
   auto buf = mdbuffer(res, input, memory_type::device); // No copy for device/managed/pinned input; copy for host input
   foo_device(buf.view<memory_type::device>());
}

template <typename mdspan_type>
void wrapper_universal(mdspan_type input) {
   auto buf = mdbuffer(res, input, memory_type::managed); // No copy for managed/pinned input; copy for host/device input
   foo_universal(buf.view<memory_type::managed>());
}
```
Those wrappers ensure that each of the foo functions can accept _any_ memory type, but let's take a look at how we would do runtime dispatching to whichever wrapper avoids copies. We'll use `std::visit` here first and then look at a custom dispatcher.

```
void wrapper(mdbuffer input) {
  std::visit([](auto view) {
    if constexpr (typename decltype(view)::accessor_type::is_host_device_accessible) {
      foo_universal(input.view<memory_type::managed>);
    } else if constexpr (typename decltype(view)::accessor_type::is_device_accessible) {
      foo_device(input.view<memory_type::device>());
    } else {
      foo_host(input.view<memory_type::host>());
    }
  }, data.view());
}
```
Or more practically speaking, since we rarely specialize these kinds of function calls for host/device accessibility but instead just dispatch to device methods for managed/pinned memory:
```
void wrapper(mdbuffer input) {
  std::visit([](auto view) {
    if constexpr (typename decltype(view)::accessor_type::is_device_accessible) {
      foo_device(input.view<memory_type::device>()); // invoked for device/managed/pinned input
    } else {
      foo_host(input.view<memory_type::host>()); // invoked for host input
    }
  }, data.view());
}
```
So really, we want to set up something that optimally avoids this boilerplate for the common task of dispatching based on where data are accessible from. I started to write out something based on cuDF's type dispatcher here in the comments, but it's complex enough, I'm going to post it separately and see what you think of the whole thing.

cpp/include/raft/core/mdbuffer.cuh
```diff
@@ -0,0 +1,831 @@
...
+   * corresponding to its current memory type.
+   *
+   * This method is useful for writing generic code to handle any memory type
+   * that might be contained in an mdbuffer at a particular point in a
+   * workflow. By performing a `std::visit` on the returned value, the caller
+   * can easily dispatch to the correct code path for the memory type.
+   */
+  [[nodiscard]] auto view() const { return view<memory_type_constant<>>(); }
+};
+
```
tarang-jain, line 828: This is just a suggestion and perhaps others can weigh in. I was thinking if there should be a helper to access the underlying mdarray from the mdbuffer object. However, I cannot think of a use-case where this will be helpful.

On the other hand if we want the mdbuffer to act like a fundamental container (wherein the fact that the underlying container is an mdarray is unknown to the user), the `data_handle` method can be useful to fetch the ptr to the underlying data (rather than the user going through the view using `.view().data_handle()`

cpp/include/raft/core/pinned_container_policy.hpp
```diff
@@ -0,0 +1,142 @@
...
+namespace raft {
+#ifndef RAFT_DISABLE_CUDA
+
+/**
+ * @brief A thin wrapper over thrust::host_vector for implementing the pinned mdarray container
+ * policy.
+ *
+ */
+template <typename T>
+struct pinned_container {
```
tarang-jain, line 37: Can the pinned container policy not be created by passing the pinned memory allocator as a template arg to the host_container_policy? The host_container_policy has a template argument for the allocator type.

cpp/include/raft/core/memory_type.hpp
tarang-jain: Yes I think the cudaPointerGetAttributes requires a void pointer.

tarang-jain (commented):
Looks great to me overall. Are there plans to modify the `temporary_device_buffer` to use the mdbuffer internally.

cpp/include/raft/core/mdbuffer.cuh
wphicks: I'd prefer to avoid exposing the underlying `mdarray` if at all possible. It won't always exist, and if the caller wants to directly manipulate the `mdarray`, I think that suggests that they should not be using an `mdbuffer` in the first place. The general purpose of `mdbuffer` is to get you an `mdspan` of the correct type when you have an `mdspan` of the possibly-correct-but-maybe-incorrect type.

In terms of exposing `data_handle`, I'm less strongly opposed, but I would prefer not to. Basically, the goal is to encourage folks to write `mdspan` based code wherever possible. `mdbuffer` is used to translate between mdspans, but it should not be used to manipulate the underlying data directly.

cpp/include/raft/core/memory_type.hpp
divyegala: A pointer should be able to be implicitly cast to `void*`. My question is if a compiler error occurs without the explicit cast.

cpp/include/raft/core/mdbuffer.cuh
```diff
@@ -0,0 +1,831 @@
...
+    std::variant<typename container_policy_at_index<0>::container_type,
+                 typename container_policy_at_index<1>::container_type,
+                 typename container_policy_at_index<2>::container_type,
+                 typename container_policy_at_index<3>::container_type>;
+
+  template <raft::memory_type MemType>
+  using container_type = alternate_from_mem_type<MemType, container_type_variant>;
+
+  using accessor_policy_variant =
+    std::variant<host_device_accessor<typename container_policy_at_index<0>::accessor_policy,
```
divyegala, line 110: Should we directly be able to use `typename container_policy_at_index<0>::accessor_policy` without the wrap of `host_device_accessor` here?

cpp/include/raft/core/mdbuffer.cuh
```diff
@@ -0,0 +1,831 @@
...
+  using const_view_type_variant = std::variant<const_view_type<static_cast<memory_type>(0)>,
+                                               const_view_type<static_cast<memory_type>(1)>,
+                                               const_view_type<static_cast<memory_type>(2)>,
+                                               const_view_type<static_cast<memory_type>(3)>>;
+
+  using storage_type_variant = concatenated_variant_t<view_type_variant, owning_type_variant>;
+
+  template <memory_type MemType, bool is_owning>
+  using storage_type =
+    std::variant_alternative_t<std::size_t{is_owning} * std::variant_size_v<view_type_variant> +
```
divyegala, line 366: Can you write a small example in commented code here explaining this arithmetic?

cpp/include/raft/core/mdbuffer.cuh
```diff
@@ -0,0 +1,831 @@
...
+  }
+
+  template <std::size_t FromIndex, typename FromT, std::size_t... Is>
+  auto static constexpr get_copyable_combinations(std::index_sequence<Is...>)
+  {
+    return std::array{is_copyable_combination<FromT, FromIndex, Is>()...};
+  }
+
+  template <typename FromT, std::size_t... Is>
+  auto static constexpr get_copyable_combinations(bool, std::index_sequence<Is...>)
```
divyegala, line 394: What is `bool` parameter here for?

cpp/include/raft/core/mdbuffer.cuh
```diff
@@ -0,0 +1,831 @@
...
+   * on the stored data. If the mdbuffer does not contain data of the indicated
+   * memory type, a std::bad_variant_access will be thrown.
+   */
+  template <memory_type mem_type>
+  [[nodiscard]] auto view()
+  {
+    return view<memory_type_constant<mem_type>>();
+  }
+  /**
+   * @brief Return an mdspan containing const elementgs of the indicated memory type representing a
```
divyegala, line 796:
```suggestion
   * @brief Return an mdspan containing const elements of the indicated memory type representing a
```

cpp/include/raft/core/mdbuffer.cuh
```diff
@@ -0,0 +1,831 @@
...
+   * @brief Return a std::variant representing the possible mdspan types that
+   * could be returned as views on the mdbuffer. The variant will contain the mdspan
+   * corresponding to its current memory type.
+   *
+   * This method is useful for writing generic code to handle any memory type
+   * that might be contained in an mdbuffer at a particular point in a
+   * workflow. By performing a `std::visit` on the returned value, the caller
+   * can easily dispatch to the correct code path for the memory type.
+   */
+  [[nodiscard]] auto view() { return view<memory_type_constant<>>(); }
```
divyegala, line 815: What does `memory_type_constant<>` return when template parameter is not specified?

cpp/include/raft/core/memory_type.hpp
wphicks: Nope, it doesn't throw a compiler error, but I generally prefer to avoid implicit casts unless there's a good reason for them. Not religious about that, though.

cpp/include/raft/core/pinned_mdspan.hpp
wphicks: Going to mark this as resolved for now, since I think this discussion is really leading us back to the memory type dispatcher, and we can discuss that in its implementation. Happy to revisit anything here after we've brought the dispatcher into the conversation.

cpp/include/raft/util/variant_utils.hpp
wphicks: Resolving to shift conversation to the dispatcher implementation.

cpp/include/raft/core/mdbuffer.cuh
```diff
@@ -0,0 +1,831 @@
...
+ * container policies into a container policy that can be used by an mdbuffer.
+ */
+template <typename ElementType,
+          typename ContainerPolicyVariant = default_container_policy_variant<ElementType>>
+struct default_buffer_container_policy {
+  using element_type = ElementType;
+  using value_type   = std::remove_cv_t<element_type>;
+
+  using container_policy_variant =
+    std::variant<host_device_accessor<std::variant_alternative_t<0, ContainerPolicyVariant>, static_cast<memory_type>(0)>, host_device_accessor<std::variant_alternative_t<1, ContainerPolicyVariant>, static_cast<memory_type>(1)>, host_device_accessor<std::variant_alternative_t<2, ContainerPolicyVariant>, static_cast<memory_type>(2)>, host_device_accessor<std::variant_alternative_t<3, ContainerPolicyVariant>, static_cast<memory_type>(3)>, >;
```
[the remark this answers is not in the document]
wphicks: Oh, whoops! Just realized that this is actually the source of truth for `container_policy_at_index`, so we can't use it quite yet at this point in the code.

cpp/include/raft/core/mdbuffer.cuh
wphicks: Potentially, but not definitively. If someone passed a custom container policy variant that does not wrap the accessor policy, we want to wrap it here because we need it wrapped for all the subsequent machinery.

cpp/include/raft/core/mdbuffer.cuh
wphicks: `std::nullopt` (of type `std::optional<memory_type>`)

cpp/include/raft/core/pinned_container_policy.hpp
wphicks: I went down that route initially, but because the thrust and rmm allocators behave a bit differently, this gets tricky. This keeps things a bit less hacky for now, and we can use a pinned rmm allocator when it's available.

cpp/include/raft/core/managed_mdarray.hpp
```diff
@@ -0,0 +1,178 @@
...
+ * @param mr rmm memory resource used for allocating the memory for the array
+ * @param exts dimensionality of the array (series of integers)
+ * @return raft::managed_mdarray
+ */
+template <typename ElementType,
+          typename IndexType    = std::uint32_t,
+          typename LayoutPolicy = layout_c_contiguous,
+          size_t... Extents>
+auto make_managed_mdarray(raft::resources const& handle,
+                          rmm::mr::managed_memory_resource* mr,
```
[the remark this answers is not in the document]
wphicks: This overload shouldn't exist at all. I've removed it. If someone needs a more specialized allocator, they should handle that in the `container_policy`, since there is no way to introspect the memory resource and confirm that it's returning managed memory.

cpp/include/raft/core/mdbuffer.cuh
wphicks: Just added a comment in the code because it took me a minute for me to figure that out for myself.

wphicks:
I've just updated this PR based on initial review. @divyegala if you have a moment to check out the final version of the memory type dispatcher, I'd love to have your thoughts on it.

wphicks:
> Looks great to me overall. Are there plans to modify the `temporary_device_buffer` to use the mdbuffer internally?

Yep! The above-referenced `copyback_buffer` will replace `temporary_device_buffer` and be more general.

cpp/include/raft/core/mdbuffer.cuh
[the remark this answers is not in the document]
divyegala: What do you mean by source of truth?

cpp/include/raft/core/mdbuffer.cuh
```diff
@@ -0,0 +1,897 @@
...
+ * to ensure that the data are in a particular memory space (e.g. device,
+ * host, managed, pinned), but those functions may be called with data that
+ * may or may not already be in the desired memory space. For instance, when
+ * called in one workflow, the data may have been previously transferred to
+ * device, rendering a copy unnecessary. In another, the function may be
+ * directly invoked on host data.
+ *
+ * Even when working strictly with host memory, it is often necessary to
+ * ensure that the data are in a particular layout for efficient access (e.g.
+ * column major vs row major) or that the the data are of a particular type
```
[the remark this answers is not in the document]
wphicks: There is an example of how to do this at line 246, but here is an example with a bit more commentary:
```
auto buf = raft::mdbuffer<float, raft::matrix_extent<int>, raft::row_major>{res, raft::mdbuffer{data}, raft::memory_type::device};
```
Here `data` is any mdspan with an `element_type` compatible with `float`. If `data` is column-major, an allocation will be made, and the data will be copied. In this case `buf` will be owning. If `data` is already a row-major `mdspan` of `float` on device, then `buf` will be non-owning, and no copy will be performed.

cpp/include/raft/core/managed_mdarray.hpp
```diff
@@ -0,0 +1,152 @@
...
+auto make_managed_mdarray(raft::resources const& handle, extents<IndexType, Extents...> exts)
+{
+  using mdarray_t = managed_mdarray<ElementType, decltype(exts), LayoutPolicy>;
+
+  typename mdarray_t::mapping_type layout{exts};
+  typename mdarray_t::container_policy_type policy{};
+
+  return mdarray_t{handle, layout, policy};
+}
+
```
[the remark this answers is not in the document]
wphicks: I don't think we had this prior to this PR, and if I previously introduced that overload, it was probably a mistake. There's no way to enforce that an `rmm` device memory resource actually returns managed memory, so it would get fairly dangerous to label an `mdarray` as being `managed` in that case. I think the best options for downstream users would indeed be the alternative you mentioned and then explicitly wrap it in an mdspan with a managed accessor _or_ to create their own container policy that definitely provides managed memory. Without a better type-based enforcement for a memory resource to make managed allocations, I think that's as far as we should go for now.

cpp/include/raft/core/mdbuffer.cuh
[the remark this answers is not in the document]
wphicks: Also, just noticed that the example at 246 was using `vector_extent` which is indeed misleading (since row/column major is the same for single-dimensional arrays). Pushing a one-word change to change it to `matrix_extent`.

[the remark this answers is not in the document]
wphicks: Final thought on this to explain the reasoning for the current design: One of the reasons it's good to avoid a named `transpose` utility function is that in the context of an `mdspan`, it is not clear what a `transpose` will actually do. Will it change the memory layout? Will it swap the indexing (in the sense of a matrix transpose)? Will it do both?

The way mdbuffer is currently written is to avoid questions like that because mdbuffer construction is always about "I want this data that already have, but I want to make sure it has _these_ mdspan properties." That may be memory location, data type, or layout, but what you get out is always some alternate representation of exactly what you put in.

cpp/include/raft/core/mdbuffer.cuh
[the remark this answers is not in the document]
wphicks: Thanks so much for following up on this! Looking over it again made me realize not only that my response was incorrect but that there was confusing variable naming that made it an easy mistake to make. Updated accordingly, and we now do use the shorthand here.

divyegala (approved):

cpp/include/raft/core/managed_mdarray.hpp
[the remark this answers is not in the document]
tarang-jain: Yes, sounds good!

wphicks:
> I noticed it's missing the factory functions to create the mdbuffer object

The idea in omitting them is to encourage people to still use `mdspan` and `mdarray` as the fundamental objects and to use `mdbuffer` as a translation tool. As I mentioned:

> This is reflected in the fact that mdbuffer can only be constructed from an existing mdspan or mdarray or another mdbuffer.

To expound on that a bit, if you're directly building an mdbuffer from e.g. a pointer, you probably want an `mdspan`. If you are trying to create an empty chunk of memory, you probably want an `mdarray`. You can immediately use `mdbuffer` to perform transfers or dispatching from there if necessary, but it's not the fundamental object you're working with.

This already saved me from a bad design decision when I was starting to play around with `copyback_buffer`. I found myself reaching for a `make_*_mdbuffer` function and realized that the object that I really wanted was not an `mdbuffer` but a `std::optional<device_mdarray>`. It was a situation where there was no ambiguity or runtime condition for where the memory was located or whether or not it had already been allocated; I just sometimes needed data on device and sometimes did not.

The one counterargument I could see to that is that it would be very nice to encapsulate the pattern of introspecting a pointer to determine its memory type, creating an `mdspan` around it, and then outputting an `mdbuffer` which expresses that initial uncertainty about the memory type. I'm less opposed to adding a `make_*_mdbuffer` for that use case alone, but it might be better to push it to a follow-on if we decide we really want it. I imagine there might be some discussion to be had around edge cases for distinguishing memory types that would be good to work out fully.

@divyegala I'd love your thoughts on this too since we chatted a bit about how to help keep `mdspan` and `mdarray` as the fundamental objects in the codebase.

cjnolet:
Thanks for the explanation, @wphicks. Yes I do understand your reasoning. 

> The one counterargument I could see to that is that it would be very nice to encapsulate the pattern of introspecting a pointer to determine its memory type, creating an mdspan around it, and then outputting an mdbuffer which expresses that initial uncertainty about the memory type. 

I do agree with this, wholeheartedly. In general, we've found it quite challenging to maintain stable C++ APIs that can still evolve over time and providing the additional layer of indirection has helped us to maintain a consistent API experience, while affording ourselves the flexibility to modify things like class constructors underneath, as well as other implementation-specific features. 

I'm okay with adding this in a follow-on. Do you already have a Github issue with follow-on items written up? If not, can you create one? It's helpful for users and other developers to prioritize and know what things are already in the plans.

wphicks:
Thanks! Just created issues for all the follow-up items, including this one.

cjnolet (approved):
LGTM. Thanks @wphicks. I'm hoping @benfred, @achirkin, and @tfeher will provide their feedback/reviews shortly. This is going to be a super nice feature to have for 24.02.

achirkin:
Thanks @wphicks, I'm really excited to see this in raft!

One question from a user point of view. There's a common problem with const-ness of data in the msdpan apis, since the `mdspan<T>` doesn't implicitly convert into `mdspan<const T>`. In raft, it's common to see a workaround of `raft::make_const_mdspan` before passing the mdspan into a function that requires const.
Does the mdbuffer relate to this? i.e. let's say a function declares it does not change the input data; would the argument look something like this: `raft::mdbuffer<const float, raft::matrix_extent<int>>&&`? How difficult would it be to pass an `raft::mdbuffer<float, raft::matrix_extent<int>>` in there?

wphicks:
> i.e. let's say a function declares it does not change the input data; would the argument look something like this: raft::mdbuffer<const float, raft::matrix_extent<int>>&&? How difficult would it be to pass an raft::mdbuffer<float, raft::matrix_extent<int>> in there?

That is _such_ a good question! I have run into that issue many times with mdspan and while I understand the constraint, `mdbuffer` is supposed to help elide exactly those sorts of annoying conversions.

The short answer is I'm honestly not sure what the current design does for that because I had not thought to even consider the question. I'm going to go experiment a bit and if this use case is not currently supported, I'll see how complex it would be to add.

In a similar vein, I'd like to take a look at if `mdbuffer` can help with situations where two mdspan types are otherwise compatible but vary in their extent type. That will probably need to be a follow-on, but it's the other major "this should be easy!" conversion that I have not really thought about support for.

cpp/include/raft/core/mdbuffer.cuh
```diff
@@ -0,0 +1,948 @@
...
+
+/**
+ * @brief A variant of container policies for each memory type which can be
+ * used to build the default container policy for a buffer.
+ */
+template <typename T>
+using default_container_policy_variant = std::variant<host_vector_policy<T>,
+                                                      device_uvector_policy<T>,
+                                                      managed_uvector_policy<T>,
+                                                      pinned_vector_policy<T>>;
```
achirkin, line 79: I feel a little bit uncomfortable that this variant ordering seems to be tied to the `variant_index_from_memory_type` and the `raft::memory_type` declared in `core/memory_type.hpp`, yet there's no static (compile time) enforcing of that.
Have you considered somehow addressing this? e.g. some kind of `static_assert` just below the declaration that would combine `variant_index_from_memory_type` and `std::get (std::variant)` for every alternative to ensure the connection and catch any changes to one of these if they ever happen?

cpp/include/raft/core/mdbuffer.cuh
```diff
@@ -0,0 +1,948 @@
...
+  /**
+   * @brief Construct one mdbuffer from another mdbuffer lvalue with matching
+   * element type, extents, layout, and container policy.
+   *
+   * Unlike when constructing from an rvalue, the new mdbuffer will take a
+   * non-owning view whenever possible, since it is assumed that the caller
+   * will manage the lifetime of the lvalue input.
+   */
+  mdbuffer(raft::resources const& res,
+           mdbuffer<ElementType, Extents, LayoutPolicy, ContainerPolicy>& other,
```
achirkin, line 684: Should the `other` be here a `const mdbuffer&`?

cpp/include/raft/core/pinned_mdspan.hpp
```diff
@@ -0,0 +1,270 @@
...
+ * @tparam LayoutPolicy policy for strides and layout ordering
+ * @tparam IndexType the index type of the extents
+ * @param[in] ptr to pinned memory to wrap
+ * @param[in] n_rows number of rows in pointer
+ * @param[in] n_cols number of columns in pointer
+ */
+template <typename ElementType,
+          typename IndexType    = std::uint32_t,
+          typename LayoutPolicy = layout_c_contiguous>
+auto make_pinned_matrix_view(ElementType* ptr, IndexType n_rows, IndexType n_cols)
```
achirkin, line 185: Same question as with managed view: would you consider making them `constexpr` for device access?

achirkin (approved):
LGTM and a few minor nitpicks

cpp/include/raft/core/mdbuffer.cuh
wphicks: Agree entirely! Take a look at the change I just pushed and see what you think. I believe it should ensure that we get the correct policy even if the enum is reordered.

cpp/include/raft/core/managed_mdspan.hpp
```diff
@@ -0,0 +1,271 @@
...
+ * @tparam LayoutPolicy must be of type layout_{left/right}_padded
+ * @tparam IndexType the index type of the extents
+ * @param[in] ptr to managed memory to wrap
+ * @param[in] n_rows number of rows in pointer
+ * @param[in] n_cols number of columns in pointer
+ */
+template <typename ElementType,
+          typename IndexType    = std::uint32_t,
+          typename LayoutPolicy = layout_right_padded<ElementType>>
+auto make_managed_aligned_matrix_view(ElementType* ptr, IndexType n_rows, IndexType n_cols)
```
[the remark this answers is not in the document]
wphicks: For sure! Done.

wphicks:
@achirkin Following up on the const translation question, it turns out to be more complicated than I thought on the surface, but fixing it pointed out a couple issues with how I was handling const element types anyway. I pushed a commit that now allows you to do [this](https://github.com/wphicks/raft/blob/fea-mdbuffer/cpp/test/core/mdbuffer.cu#L319-L327) and also fixes handling of const element types.

cjnolet:
@wphicks super excited about the implicit conversion. That's going to be very powerful.

cpp/include/raft/core/mdbuffer.cuh
wphicks: No, this constructor should only apply to non-const lvalues. The `mdbuffer` constructors are all intended to take some data and make it accessible in exactly the same format modulo whatever changes you explicitly asked for. If `other` were passed as a const lvalue reference, we could only obtain const views from it (i.e. mdspans with const element type), so the mdbuffer constructed from it would have const element type rather than the original type passed in.

In this case, the caller almost certainly means to have constructed an mdbuffer of const _data_ rather than a const mdbuffer. If we provided an overload for const lvalue reference, it would also be ambiguous whether the caller was tryng to construct that non-owning mdbuffer of const data or if they actually wanted to perform a copy of the (possibly) underlying mdarray.

By simply not supporting a constructor from a const lvalue reference, we avoid a bunch of potential confusion and encourage users to use mdbuffer as a highly temporary object, relying on mdspans obtained from the mdbuffer for actual data manipulation.

Callers can still work with const lvalue references to mdbuffers if necessary, but they will need to be explicit about how the ambiguity should be resolved. For instance, if they want to allow a copy of the (possibly) underlying mdarray, they can make use of the default copy constructor:
```
void foo(raft::resources const& res, mdbuffer const& buf) {
  // Explicit use of copy constructor, which behaves according to familiar rules for mdspan and mdarray copies
  auto dev_buf = mdbuffer(res, mdbuffer(buf), memory_type::device);
}

void bar(raft::resources const& res, mdbuffer const& buf) {
  auto dev_buf = std::visit(
    [](auto view) {
      // mdbuffer constructed from a view of const data following familiar rules of mdspan-to-mdbuffer construction
      return mdbuffer(res, view, memory_type::device);
    },
    buf.view()
  );
}
```

Regardless of which the user chooses, it forces them to be explicit about the behavior they expect if they really, really, really meant to use a const mdbuffer and not an mdbuffer of const data.

cpp/include/raft/core/mdbuffer.cuh
```diff
@@ -0,0 +1,999 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+#include <algorithm>
+#include <execution>
+#include <optional>
+#include <raft/core/cuda_support.hpp>
```
cjnolet, line 21: Sorry, Will, I happened to be looking through here more closely as I'm working on a PR to add the factory functions for creating the mdbuffer and I realized is that this file should have an `hpp`, not a `cuh` extension. `cuh` signifies that `nvcc` is needed to compile the file, whereas `hpp` extension might require cuda runtime, but doesn't need nvcc to compile. 

If the expectation that's being used here is that `cuh` signifies anything that requires `nvcc` or `cuda runtime` then we need to change that.

cjnolet (changes requested):

cpp/include/raft/core/mdbuffer.cuh
wphicks: Unfortunately, we need nvcc for [this guy right here](https://github.com/rapidsai/raft/pull/1999/files/fca74aa18c0a21ab9a689e9c26b6f582edbd3528#diff-7684a375b65e461c3fe89da16b806c1a9c9169cadf03bb25a0eff104a600fad2R38). That includes a kernel definition for some of the copies that mdbuffer requires.

wphicks: Sorry, that link doesn't seem to be taking me to the right line. Try [this](https://github.com/wphicks/raft/blob/fea-mdbuffer/cpp/include/raft/core/mdbuffer.cuh#L38) instead.

cjnolet: Fair enough- yeah the copy does complicate things there and I admit I didn't notice that the first time around. I think what really stinks, though, that the copy itself forces device functions in the code for what could otherwise often just be plain cuda symbols (for eg. we are most often using standard strided layouts in the mdspans/mdarrays these days so the standard `cudaMemcpy` would get the job done most of the time). 

Anyways, you already have an `hpp` file for `mdbuffer` and I think we can (and should) revisit the potential for making the device-function-based copy an as-needed option at some point to avoid the propagation of the device functions. It's also not immediately obvious to me yet what that would look like.

cjnolet (approved):
Giving my approval again. Let's give this another day to see if @tfeher and @benfred have any feedback and then we can get it merged for the weekend!

cpp/include/raft/core/mdbuffer.cuh
cjnolet: The closer I look, the more I realize we are getting inconsistent with this.. 

Take [math.hpp](https://github.com/wphicks/raft/blob/fea-mdbuffer/cpp/include/raft/core/math.hpp), which marks functions as device when compiling with `nvcc`. 

And also take [operators.hpp](https://github.com/wphicks/raft/blob/fea-mdbuffer/cpp/include/raft/core/operators.cuh) which includes the device functions so that they aren't included in `operators.hpp` (though `operators.hpp` includes `math.hpp`! 

I'm going to create an issue to revisit this for sure- we should definitely strive to be consistent everywhere.

cpp/include/raft/core/mdbuffer.cuh
```diff
@@ -0,0 +1,999 @@
...
+using memory_type_to_default_policy = std::conditional_t<
+  MemType == raft::memory_type::host,
+  host_vector_policy<T>,
+  std::conditional_t<
+    MemType == raft::memory_type::device,
+    device_uvector_policy<T>,
+    std::conditional_t<
+      MemType == raft::memory_type::managed,
+      managed_uvector_policy<T>,
+      std::conditional_t<MemType == raft::memory_type::pinned, pinned_vector_policy<T>, void>>>>;
```
achirkin, line 91: This is totally subjective, but maybe partial instantiation here is a little bit more readable by avoiding nested conditional_t?
```suggestion
template <typename T, raft::memory_type MemType>
struct memory_type_to_default_policy {};
template <typename T>
struct memory_type_to_default_policy<T, raft::memory_type::host> {
    using type = typename raft::host_vector_policy<T>;
};
template <typename T>
struct memory_type_to_default_policy<T, raft::memory_type::device> {
    using type = typename raft::device_uvector_policy<T>;
};
template <typename T>
struct memory_type_to_default_policy<T, raft::memory_type::managed> {
    using type = typename raft::managed_uvector_policy<T>;
};
template <typename T>
struct memory_type_to_default_policy<T, raft::memory_type::pinned> {
    using type = typename raft::pinned_vector_policy<T>;
};

template <typename T, raft::memory_type MemType>
using memory_type_to_default_policy_t = typename memory_type_to_default_policy<T, MemType>::type;
```

cpp/include/raft/core/mdbuffer.cuh
achirkin: This should work, thanks!

cpp/include/raft/core/mdbuffer.cuh
achirkin: Ok, container const-ness is always a little bit confusing, but I think I've got your point :)
Maybe just add a short `/* NOLINT */` to the argument, so that the reader sees this is intentional?

cpp/include/raft/core/mdbuffer.cuh
wphicks: Sure! I don't mind it

cpp/include/raft/core/mdbuffer.cuh
wphicks: Yeah, I agree entirely. I'm wondering if for headers that are "optionally cuh" we might benefit from having a consistent flag that enables or disables the components that require nvcc. That would need to be separate from the current RAFT_DISABLE_CUDA definition, but it would give us an easy way to compose these files without an absurd propagation of options. Anyway, something we can discuss further in the issue, but I'm absolutely behind you on finding a more consistent global solution.

cpp/include/raft/core/mdbuffer.cuh
wphicks: Just did so and updated the docs in a few places. I agree on container const-ness being tricky! If you hadn't asked the question above about converting to const views without going through `make_const_mdspan`, I wouldn't have realized that the technique I expected people to use for _this_ question (const element type) wasn't working. Thanks very much for pushing me on both these!

tfeher (approved):
Thanks @whpicks for this feature! Due to lack of time I just skimmed through a large part of implementation details, so my review is restricted to the high level API.

 The API is well polished and seems convenient to use. I think it will be a really helpful utility!

 You are handling `host`, `device`, `managed`, and `pinned` memory types. I agree that these are the variants that we expect to encounter in practice. 
 
We have APIs in RAFT, specifically some of the ANN build algorithms, that just care about the data shape, and host/device dispatch is happenning only deeper in the call chain. At that point `mdbuffer` is a great help to simplify the code when we finally need to guarantee accessibility from device. And I see that actually all four allocation types could be useful when we create an `mdbuffer`. So having the vocabulary to express this using `raft::memory_type` seems helpful. 
 
Still, I am a little bit concerned by the proliferation of `mdarray` types. The allocation mechanism of `mdarray` can be defined by simply passing the desired `memory_resource` during creation. When we want to use the data, we care about whether it is accessible from host or device memory space, and that is expressed by the `accessor_type`. We shall ask the question: it is really useful to have so many `mdarray` / `mdspan` variants? I am not sure about this, but I do not have any objection for the current proposal. We should still keep in mind that other memory requirements will could come up (e.g. array allocated with huge page attributes) and we probably don't want to define new `mdarray` type for each of these.

 Thanks once more for this work, it is a great addition to RAFT! The PR looks good to me.

cpp/include/raft/core/memory_type.hpp
```diff
@@ -14,13 +14,28 @@
...
+#include <cuda_runtime.h>
+#include <raft/util/cuda_rt_essentials.hpp>
+#include <type_traits>
+#else
+#include <raft/core/logger.hpp>
+#endif
 
 namespace raft {
-enum class memory_type { host, device, managed, pinned };
+enum class memory_type : std::uint8_t {
```
benfred, line 28: nit: I sorta wonder if we should align this enum with the cudaMemoryType enum:
https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1g13de56a8fe75569530ecc3a3106e9b6d

afaict the cudaMemoryType enum has the exact same values (aside from some naming differences with pinned memory, like cudaMemoryTypeUnregistered being unpinned host memory).

benfred (approved):
this looks great! I think this code will really help simplify some of our code, and I'm looking forward to taking advantage of it

cpp/include/raft/core/memory_type.hpp
wphicks: Agreed! Time to see how well I did at ensuring that the order of the enum does not matter ;). I'm kicking off a build now with the enum values aligned.

A general note for RAFT's memory_type enum is that it was designed to be extensible beyond the usual CUDA memory types. For instance, we might choose to add a `cuda_ipc` memory type in the future that allows us to easily flag memory that can be shared across processes. That does not mean that the base values shouldn't currently align with cudaMemoryType; just that we may need to be careful if we do decide to extend these values in the future.

wphicks:
> Still, I am a little bit concerned by the proliferation of mdarray types.

I actually strongly agree with this. As you say, there are use cases where we _really_ only care about whether or not we can access those data at a particular point in runtime and what the shape/layout is. In that case, we pay fairly dearly for the proliferation of compile-time types. The compile-time specification of memory type also runs counter to some of the general ways I see folks thinking about the purpose of `mdspan`.

I think there may be room for both constructs to exist alongside one another, however. There's no question that having the ability to dispatch to radically different compile-time paths based on the runtime condition of memory type is useful, and I'm hoping that's what mdbuffer will support. On the other hand, it would be good to revisit how we're using mdspan in cases where we explicitly wish to elide the difference between memory types for e.g. GPU/CPU compatible code. I had proposed something along those lines quite some time ago, but at the time others objected to it. Maybe we can review some of those ideas again as a follow-on to this PR and see if there is greater clarity on the different use cases now that `mdbuffer` has solidified.

wphicks:
Thank you everyone for all of the valuable feedback on this! I'm going to try to implement @benfred 's suggestion to align the memory_type values with cudaMemoryType, and if there are no objections, I'll trigger a merge after that.

cjnolet:
> The compile-time specification of memory type also runs counter to some of the general ways I see folks thinking about the purpose of mdspan.

This is counter to the original use of mdspan within RAFT also. We intentionally created only two different types (device and host) with a general accessor wrapper that specifies whether or not the memory can be read from device or host. Note that the "both" case can still be specified through that pattern, which also covers managed memory, for example. The idea is to allow functions to define which access level they are interested in without having to introduce new template types for each one.

I also agree that we should be able to simplify mdarray in this way also- since the template itself should only care about what mdspan it needs to produce in order to interact with APIs that require it. Ideally, the mdarray would accept the memory type as a runtime argument so long as it produces an mdspan that is either device or host accessible.

cjnolet:
> I had proposed something along those lines quite some time ago, but at the time others objected to it. Maybe we can review some of those ideas again as a follow-on to this PR and see if there is greater clarity on the different use cases now that mdbuffer has solidified.

The decision to make separate "device" and "host" mdspan typedefs adheres to the design of the other c++ vocabulary types in RAPIDS and in the CCCL (including Thrust). As an organization, we should be striving to follow CCCL's lead here. 

cc @jrhemstad @mhoemmen @harrism

wphicks:
> The decision to make separate "device" and "host" mdspan typedefs adheres to the design of the other c++ vocabulary types

Thanks! That's not really what I was getting at, but I think this is getting beyond the scope of this PR. We can discuss more elsewhere.

cjnolet:
@wphicks, my prior response was also in reply to @tfeher's earlier comment:

> We shall ask the question: it is really useful to have so many mdarray / mdspan variants? I am not sure about this, but I do not have any objection for the current proposal

(I can only speak to the decision for "device"-accessible and "host"-accessible variants)

PR #2002, opened 2023-11-16
Fix pyproject versions

vyasr:
The version doesn't need to be hardcoded into pyproject.toml files anymore, but it looks like update-version.sh wasn't updated to account for that.


cjnolet (approved):
LGTM!

raydouglass (approved):

raydouglass:
Going to admin merge this to get it in before the nightly run later tonight. If this is problematic, we will address it tomorrow.

PR #2009, opened 2023-11-17
Export Pareto frontier in `raft-ann-bench.data_export`

divyegala:
This PR now exports 3 CSVs from search result JSON files of the suffixes:
1. `raw`: All results
2. `throughput`: Pareto frontier of throughput results
3. `latency`: Pareto frontier of latency results

The Pareto frontier is now no more created in `raft-ann-bench.plot`.


cjnolet (approved):
Looks great! I know you are adding the guard for the read_file but pre-approving.

PR #2013, opened 2023-11-17
pylibraft only depends on numpy at runtime, not build time.

bdice:
pylibraft specifies its numpy pinning incorrectly in conda packages. It is only a runtime dependency, not a build dependency. Because of run-exports from numpy, the resulting pinning on pylibraft requires a very new numpy version, and is thus causing cuml to fall back to old raft builds.


dantegd (approved):

harrism (approved):

vyasr (approved):

vyasr:
Run exports are great... until they're not.

raydouglass (approved):

PR #2017, opened 2023-11-20
Backport numpy pinning fix #2013 to 23.10.

bdice:
Backport of https://github.com/rapidsai/raft/pull/2013 to 23.10.

pylibraft specifies its numpy pinning incorrectly in conda packages. It is only a runtime dependency, not a build dependency. Because of run-exports from numpy, the resulting pinning on pylibraft requires a very new numpy version, and is thus causing cuml to fall back to old raft builds.


bdice:
Closing. This change is not needed for a backport because the existing 23.10 packages have a numpy pinning compatible with the rest of RAPIDS 23.10. The 23.12 release will have a wider numpy pinning as desired.

PR #2021, opened 2023-11-22
[WIP] Fix ann-bench multithreading

achirkin:


achirkin:
Notes:
  - At the moment, [a new `raft::resources` handle is created for each thread](https://github.com/rapidsai/raft/blob/885519782cac6d7375b45c5e4494c524b1f9e4dd/cpp/bench/ann/src/raft/raft_ann_bench_utils.h#L83-L86), it seems to be not thread-safe. This can be quickly fixed by modifying the copy-constructor of the configured_raft_resources.
  - The PR needs more testing!

cpp/bench/ann/src/faiss/faiss_gpu_wrapper.h
```diff
@@ -147,18 +146,33 @@ class FaissGpu : public ANN<T> {
...
+  /** [NOTE Multithreading]
+   *
+   * `gpu_resource_` is a shared resource:
+   *   1. It uses a shared_ptr under the hood, so the copies of it refer to the same
+   *      resource implementation instance
+   *   2. GpuIndex is probably keeping a reference to it, as it's passed to the constructor
+   *
+   * To avoid copying the index (database) in each thread, we make both the index and
+   * the gpu_resource shared.
+   * This means faiss GPU streams are possibly shared among the CPU threads;
```
wphicks, line 158: Do we have an issue tracking this?

cpp/bench/ann/src/faiss/faiss_gpu_wrapper.h
```diff
@@ -105,7 +105,6 @@ class FaissGpu : public ANN<T> {
     RAFT_CUDA_TRY(cudaGetDevice(&device_));
     RAFT_CUDA_TRY(cudaEventCreate(&sync_, cudaEventDisableTiming));
     faiss_default_stream_ = gpu_resource_.getDefaultStream(device_);
-    raft::resource::set_cuda_stream(handle_, faiss_default_stream_);
```
wphicks, line 108: Any reason we're not using the FAISS default stream here anymore?

wphicks (approved):
Just one substantive question, but this looks great. Love the design of `configured_raft_resources`.

cjnolet (changes requested):
Making sure we discuss this before merging.

cpp/bench/ann/src/raft/raft_ann_bench_utils.h
```diff
@@ -41,4 +41,72 @@ inline raft::distance::DistanceType parse_metric_type(raft::bench::ann::Metric m
...
+             rmm::mr::get_current_device_resource(), 1024 * 1024 * 1024ull};
+           rmm::mr::set_current_device_resource(mr);
+           return mr;
+         }(),
+         [](rmm::mr::pool_memory_resource<rmm::mr::device_memory_resource>* mr) {
+           if (rmm::mr::get_current_device_resource()->is_equal(*mr)) {
+             rmm::mr::set_current_device_resource(mr->get_upstream());
+           }
+           delete mr;
+         }}}
```
harrism, line 76: suggestion: Initialize the members of the class by name, not as a struct.
suggestion: don't bury the re-setting of the previous device resource inside an unnamed deleter lambda. Just reset it in your class destructor.
question: does the mr_ really need to be shared? 
suggestion: use a unique_ptr.

```suggestion
  configured_raft_resources()
     : mr_{make_unique<rmm::mr::pool_memory_resource<rmm::mr::device_memory_resource>>(
             rmm::mr::get_current_device_resource(), 1024 * 1024 * 1024ull)}
{
}

~configured_raft_resources() 
{
  rmm::mr::set_current_device_resource(mr_->get_upstream());
}

```

Note that this also removes the call to `is_equal()` -- is there a reason in this benchmark that you are worried about the possibility of the current resource changing during the lifetime of this class?

In the future, please use `operator==` on MRs, because `is_equal` is likely to be removed.

harrism (changes requested):
@cjnolet pointed me to this PR because I was asking about uses of MR comparisons. Looking at the call to `is_equal` used in this PR I had a few other suggestions to improve the surrounding code.

cpp/bench/ann/src/raft/raft_ann_bench_utils.h
achirkin: Hi Mark, thanks for having a look at this!

First of all, yes, we must use the shared pointer here. We create a new `configured_raft_resources` *at most* once per benchmark case, and let multiple concurrent threads spawned by gbench copy it. We do this because we need some of the resources not be shared due to thread safety (the raft handle and the event). However, we must to have exactly one rmm pool resource shared by everything, which is dictated by rmm current_device_resource semantics. Hence we need the pool resource be initialized with the original `configured_raft_resources` and destructed after the last copy of `configured_raft_resources` is destructed.

The same applies to setting/unsetting the rmm current_device_resource. Hence I cannot just reset it my class destructor and bury it inside an unnamed deleter lambda. An alternative would be to create a new class with the destructor that does the same and put it in the shared pointer, but why bother?

Also, I'm just curious as I'm not that experienced with the latest C++ best practices. What are the downsides of using curly braces to initialize the members "as a struct" as I did in this code?

harrism: The downsides are that you aren't initializing them by name, so it is less readable, and if the order of the class members changes this initialization will be invalid. 

My suggestions were aimed at improving the readability and maintainability of the code, which currently I find obfuscated and error prone. But really this is not a library I have to read or maintain so it's up to you and the RAFT reviewers.

The one thing that I ask is that you remove the call to `is_equal()` for the resources, as this is probably going away.  If you must compare the resources, use `==`.

harrism: See https://github.com/rapidsai/rmm/issues/1402

cjnolet: @achirkin the big difference between the rmm example you've pointed out and the example here is that the rmm initializers are still naming the members so if the names should change or he rearranged, maintenance becomes much more straightforward.

I do agree with Mark here, and to be quite honest, I find the use of the pattern here (unnamed parameter initialization through lambdas) to be quite hard to read myself. I also prefer that we at least name the parameters, especially for a list that includes multiple different members. It seems like not much additional work in proportion to the significant gain in readability.

harrism (approved):

cpp/bench/ann/src/raft/raft_ann_bench_utils.h
achirkin: Ok, now I really feel dumb that I don't see what you both mean :) I'm just calling here another constructor with a single argument that takes the pool, am I? If you're referring to the aggregate initialization, it's not even possible for this class, because it has two user-declared constructors.

cjnolet (approved):
LGTM!

PR #2022, opened 2023-11-23
`libraft` and `pylibraft` API for CAGRA build and HNSW search

### LINKED ISSUE #1772: [FEA] Explore potential for a CAGRA trained model on GPU to be inferenced on CPU
There's been discussion and some light research around this but I want to capture it in an issue so we dan discuss it further, track its progress, and potentially garner some user feedback along the way. 

CAGRA's a state-of-the-art graph-based algorithm, and while it can be much faster to train on a GPU than HNSW on the CPU, there has been much interest, especially with Grace hardware, in being able to train the CAGRA model on the GPU and transfer it to the CPU for fast inference. One way in which we are interested in exploring is whether or not the underlying CAGRA graph might have similar enough properties to an HNSW graph that we could either use HNSW for inference or, without too much additional pre-processing, be able to convert the CAGRA graph into a structure that could be queried w/ HNSW. 

Adding usual suspects for reference: @nwstephens @tfeher @anaruse @enp1s0

divyegala:
Closes #1772


python/pylibraft/pylibraft/neighbors/cagra_hnswlib.pyx
```diff
@@ -0,0 +1,445 @@
...
+        return self.params.ef
+
+    @property
+    def num_threads(self):
+        return self.params.num_threads
+
+
+@auto_sync_handle
+@auto_convert_output
+def search(SearchParams search_params,
```
lowener, line 322: Add all these new functions to the pylibraft API documentation

cpp/include/raft/neighbors/detail/cagra_hnswlib.hpp
```diff
@@ -0,0 +1,186 @@
...
+#include "../cagra_hnswlib_types.hpp"
+
+#include <raft/core/host_mdspan.hpp>
+#include <raft/core/resources.hpp>
+
+#include <atomic>
+#include <future>
+#include <memory>
+#include <mutex>
+#include <omp.h>
```
lowener, line 28: Unused?

lowener (changes requested):

cpp/cmake/thirdparty/get_hnswlib.cmake
```diff
@@ -27,11 +27,20 @@ function(find_and_configure_hnswlib)
                 WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/_deps )
 
         message("SOURCE ${CMAKE_CURRENT_SOURCE_DIR}")
+        message("WORKING DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}")
         execute_process (
                 COMMAND git apply ${CMAKE_CURRENT_SOURCE_DIR}/cmake/patches/hnswlib.patch
                 WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/_deps/hnswlib-src
         )
     endif ()
+    # rapids_cpm_find(hnswlib ${PKG_VERSION}
```
cjnolet, line 36: Can we remove this until we have e solution?

cpp/cmake/thirdparty/get_hnswlib.cmake
```diff
@@ -27,11 +27,20 @@ function(find_and_configure_hnswlib)
                 WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/_deps )
 
         message("SOURCE ${CMAKE_CURRENT_SOURCE_DIR}")
+        message("WORKING DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}")
```
cjnolet, line 30: Is this just for debugging?

cpp/include/raft/neighbors/cagra_hnswlib_types.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+#include "ann_types.hpp"
+#include <memory>
+#include <raft/distance/distance_types.hpp>
+
+#include <cstdint>
+#include <sys/types.h>
+#include <type_traits>
+
+namespace raft::neighbors::cagra_hnswlib {
```
cjnolet, line 27: Rather than a separate index, namespace, and file for cagra_hnsw, can't we just use the cagra index, provide a helper in raft::neighbors::hnsw to convert it to the HNSW index wrapper, and then provide a raft::neighbors::hnsw::search()? I understand using the cagra_HNSW algo name for the benchmarks but I think HNSW itself is more appropriate for the actual public api to do this.

cpp/include/raft/neighbors/detail/cagra/cagra_serialize.cuh
```diff
@@ -103,6 +103,8 @@ void serialize_to_hnswlib(raft::resources const& res,
                           std::ostream& os,
                           const index<T, IdxT>& index_)
 {
+  // static_assert(std::is_same_v<IdxT, int> or std::is_same_v<IdxT, uint32_t>,
```
cjnolet, line 106: Can we move this to hnsw_serialize?

cpp/include/raft/neighbors/detail/cagra_hnswlib.hpp
```diff
@@ -0,0 +1,82 @@
+/*
+ * Copyright (c) 2023, NVIDIA CORPORATION.
```
cjnolet, line 2: I think raft/neighbors/detail/hnsw.hpp would be more appropriate. By this point we have removed cagra completely from the story.

cpp/include/raft/neighbors/hnswlib_types.hpp
```diff
@@ -0,0 +1,101 @@
...
+#include <memory>
+#include <raft/core/error.hpp>
+#include <raft/distance/distance_types.hpp>
+
+#include <cstdint>
+#include <hnswlib.h>
+#include <sys/types.h>
+#include <type_traits>
+
+namespace raft::neighbors::cagra_hnswlib {
```
cjnolet, line 29: Let's just use `raft::neighbors::hnsw` for the namespace here.

cpp/include/raft/neighbors/hnswlib_types.hpp
```diff
@@ -0,0 +1,101 @@
...
+    appr_alg_->base_layer_only = true;
+  }
+
+  /**
+  @brief Get hnswlib index
+  */
+  auto get_index() const -> void const* override { return appr_alg_.get(); }
+
+ private:
+  std::unique_ptr<hnswlib::HierarchicalNSW<typename hnsw_dist_t<T>::type>> appr_alg_;
```
cjnolet, line 95: This is leaning hnswlib types through the public API. This should just be a pure virtual class and the impl should be in detail (and produced via a factory function).

cpp/include/raft_runtime/neighbors/cagra.hpp
```diff
@@ -27,48 +27,56 @@
...
+                  raft::host_matrix_view<const T, int64_t, row_major> dataset,         \
+                  raft::neighbors::cagra::index<T, IdxT>& idx);                        \
+                                                                                       \
+  void search(raft::resources const& handle,                                           \
+              raft::neighbors::cagra::search_params const& params,                     \
+              const raft::neighbors::cagra::index<T, IdxT>& index,                     \
+              raft::device_matrix_view<const T, int64_t, row_major> queries,           \
+              raft::device_matrix_view<IdxT, int64_t, row_major> neighbors,            \
+              raft::device_matrix_view<float, int64_t, row_major> distances);          \
+  void serialize_file(raft::resources const& handle,                                   \
```
cjnolet, line 57: I think these belong in dedicated hnsw API.

cpp/include/raft_runtime/neighbors/cagra_hnswlib.hpp
```diff
@@ -0,0 +1,43 @@
...
+#pragma once
+
+#include <raft/core/host_mdspan.hpp>
+#include <raft/core/resources.hpp>
+#include <raft/neighbors/cagra_hnswlib_types.hpp>
+
+namespace raft::runtime::neighbors::cagra_hnswlib {
+
+#define RAFT_INST_CAGRA_HNSWLIB_FUNCS(T)                                      \
+  void search(raft::resources const& handle,                                  \
```
cjnolet, line 26: Let's just put this still in an hnsw.hpp file (and namespace)

cpp/src/raft_runtime/neighbors/cagra_hnswlib.cpp
```diff
@@ -0,0 +1,49 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <raft/neighbors/cagra_hnswlib.hpp>
+#include <raft_runtime/neighbors/cagra_hnswlib.hpp>
+
```
cjnolet, line 19: Same here- hnswlib.cpp

cpp/include/raft_runtime/neighbors/cagra_hnswlib.hpp
```diff
@@ -0,0 +1,43 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <raft/core/host_mdspan.hpp>
+#include <raft/core/resources.hpp>
+#include <raft/neighbors/cagra_hnswlib_types.hpp>
```
cjnolet, line 21: This is going to leak hnswlib types through the headers. Not good

cjnolet (changes requested):
I think the general idea is getting there, but there's still a lot of api leakage and we have an opportunity here for a dedicated HNSW API (independent from hnswlib).

cpp/include/raft_runtime/neighbors/cagra_hnswlib.hpp
divyegala: That file does not include any hnswlib headers.

cpp/include/raft/neighbors/hnswlib_types.hpp
divyegala: The pure virtual class is elsewhere. This class should be in detail.

cjnolet: No, the types in these files should not be leaking any other dependencies. This is where the virtual class (eg interface) belongs so that hnswlib is guaranteed never to leak outside of here.

divyegala: Yes, you are absolutely right. I put this class in detail.

divyegala:
@cjnolet I addressed your suggestions and reworked the filenames/namespaces/structure to a pure `hnsw` namespace. There's one thing that I'd like your opinion on and something I would consider a nice-to-have. As you requested, I moved the serialization and de-serialization functions into `hnsw_serialize.cuh`, however, here's the problem:
1. `serialize` accepts a CAGRA index and needs to be CUDA compatible
2. `deserialize` returns an hnsw index and does not need to be CUDA compatible

As a result of putting these two functions in the same header, they need to be compiled in the same `.cu` translation unit because `hnswlib` is not true header-only: it does not have `inline` functions hence all headers that contains an hnswlib included header must be compiled in the same TU.

By splitting `deserialize` and `serialize` into two TUs, we will be able to:
1. Compile `serialize` in `hnswlib_serialize.cu`
2. Compile `deserialize` and `search` in `hnsw.cpp`

Another benefit of this splitting is that users will be able to build CPU-only RAFT and still be able to load hnsw indexes and search them. If this change is acceptable to you, then I will go ahead and make it.

lowener:
One last remark on my side: HNSW supports Filtering functions in `searchKnn`, we should also expose this.

divyegala:
@lowener I filed an [issue](https://github.com/rapidsai/raft/issues/2069) for filtering functions as a follow-on, this PR is pretty big already

cjnolet (changes requested):

cpp/include/raft_runtime/neighbors/cagra.hpp
```diff
@@ -27,48 +27,53 @@
...
+             const raft::neighbors::cagra::index_params& params,                       \
+             raft::device_matrix_view<const T, int64_t, row_major> dataset)            \
+    ->raft::neighbors::cagra::index<T, IdxT>;                                          \
+                                                                                       \
+  auto build(raft::resources const& handle,                                            \
+             const raft::neighbors::cagra::index_params& params,                       \
+             raft::host_matrix_view<const T, int64_t, row_major> dataset)              \
+    ->raft::neighbors::cagra::index<T, IdxT>;                                          \
+                                                                                       \
+  void build_device(raft::resources const& handle,                                     \
```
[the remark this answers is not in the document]
divyegala: Not sure, this change predates this PR

cjnolet (approved):

PR #2027, opened 2023-11-29
Remove selection_faiss

benfred:
Remove the selection_faiss instantiations. Since #1985, we haven't been using the faiss select_k code and these aren't necessary anymore. This should lead to a 70MB improvement in libraft.so binary size.

This also removes the raft::spatial::select_k code in favour of matrix:: select_k - the spatial version was marked deprecated, and didn't switch between the best selection algorithms for the input size.


tfeher (approved):
Thanks @benfred LGTM! The size of libraft.so goes from 486 MB to 425 MB! (cuda11 x86_64 build).

Issue #1984, opened 2023-11-10
[FEA] Wrap `epsilon_neighbors` through pylibraft


cjnolet:
As the title suggests, RAFT's c++ API contains an `epsilon_neighbors` function, which is extremely useful for computing neighborhoods within a radius ball, and their corresponding vertex degrees. 

This is an important primitive and it shouldn't be hard to wrap through pylibraft.

Issue #517, opened 2022-02-19
[FEA] Use RBC to compute Epsilon neighborhoods


cjnolet:

cjnolet:
Hey @mfoerste4,  RBC functions similarly to IVF methods, but with the added benefit that it knows the number of probes that need to be selected in order to find the exact nearest neighbors. The other (more subtle) benefit to the algortihm is that since we know the closest IVF clusters that we need to scan through, we can also issue a warning or raise an appropriate error up front. 

I don't see your first item above as being prohibitive to constructing a CSR- the reason being that the adjacency list / mask is boolean, with a naive implementation taking up 1/8th the space of a single-precision pairwise matrix, but also because we can build the adjacency matrix in batches if memory becomes an issue. Another option could also be to use a bitmap instead of a boolean matrix (this will require atomics, but at least each value only needs to be written once). 

The current epsilon neighbors primitive constructs a boolean mask and vertex degree matrix, which essentially can be used to construct the `indptr` and `column` arrays of the CSR. Only if a user needs the distances would a second pass be needed, and in this case, I don't think we would want to do a second pass w/ the RBC algorithm, but we would want to expose (and use) a masked matrix multiplication primitive like cusparse's SDDMM to compute the final distances. This detail is important- while in dbscan we want the distances, not all user want that, but the good news is that we have primitives that are optimized for computing selective distances (at least for dot-product-based distances) and so we don't need to run a full GEMM to get them. 

If there are users that set their epsilon poorly and end up with a full pairwise distance matrix, we can let them know this up front by testing how many RBC clusters are within `eps` and the total number of points that belong to those closest balls. We could even go a step further to make the heuristic more accurate and use the variance of the balls to determine the confidence ranges. For example, this would allow us to issue a message like "WARNING: The requested eps will cause an adjacency list that is XXX dense. This is going to put all points into the same cluster". The benefit here is that we can implement this message in DBSCAN itself, just based on the adjacency matrix. 

BTW, for further proof that not all users care about the distances of an epsilon neighborhood, https://github.com/rapidsai/raft/issues/1984 is a direct request form a user who only wants the neighbors (e.g. adjacency matrix) returned and not the actual distances.

PR #2028, opened 2023-11-29
Add eps-neighbor search via RBC

mfoerste4:
This PR adds support for an eps-neighborhood search based on random ball cover. The algorithm re-uses the existing rbc index creation.

Changes:
* add C++ API `raft::neighbors::ball_cover::epsUnexpL2NeighborhoodRbc`
* lifted the 2/3-D limitation for eps-neighborhod via RBC (limitation still in place for k-nn queries)
* pylibraft support for dense brute-force eps-neighborhood
* pylibraft support for sparse/dense rbc epsneighborhood

Note: The PR also contains a fix for the vertex degree computation in the brute force algorithm `spatial::knn::detail::epsUnexpL2SqNeighborhood`.

Related to #1984 and #517


cjnolet:
> lifted the 2/3-D limitation for RBC (only valid for k-nn)

This is awesome! So can we support arbitrarily large dimensions now or are we still limited to what can be fused into the knn with k-selection? 

(I admit I haven't skimmed the whole PR yet)

python/pylibraft/pylibraft/neighbors/eps_neighborhood.pyx
```diff
@@ -0,0 +1,341 @@
...
+                raise ValueError("Unsupported method %s" % method)
+    else:
+        raise TypeError("dtype %s not supported" % dataset_cai.dtype)
+
+    return (adj, vd)
+
+
+@auto_sync_handle
+@auto_convert_output
+def eps_neighbors_sparse(RbcIndex rbc_index, queries, eps, handle=None):
```
cjnolet, line 232: I'm a little conflicted about the layout in this namespace. I think it might be better and more consistent with the other `pylibraft` algorithms to separate the APIs by their associated index type. So, for example, `pylibraft.neighbors.rbc.eps_neighbors` and `pylibraft.neighbors.brute_force.eps_neighbors`. This is because, similar to knn, there are mutliple different algorithms for computing it and we separate each one of those different algorithms. `eps-nn` is similar in this way.

cpp/include/raft/neighbors/ball_cover-ext.cuh
```diff
@@ -121,4 +167,9 @@ void knn_query(raft::resources const& handle,
 
 instantiate_raft_neighbors_ball_cover(int64_t, float, uint32_t, uint32_t);
 
+instantiate_raft_neighbors_ball_cover(int32_t, float, int32_t, int32_t);
```
cjnolet, line 170: Do you have an idea of how much all these new instantiations are increasing the binary size? We've been watching it pretty closely since it's been a point of contention for many users.

cpp/include/raft/neighbors/ball_cover-inl.cuh
```diff
@@ -295,6 +294,103 @@ void knn_query(raft::resources const& handle,
...
+ *                    `vd + m` stores the total number of edges in the adjacency
+ *                    matrix. Pass a nullptr if you don't need this info.
+ * @param[in]  x      first matrix [row-major] [on device] [dim = m x k]
+ * @param[in]  m      number of rows in x
+ * @param[in]  n      number of columns in x and index
+ * @param[in]  eps    defines epsilon neighborhood radius
+ */
+template <typename idx_t, typename value_t, typename int_t, typename matrix_idx = std::uint32_t>
+void epsUnexpL2NeighborhoodRbc(raft::resources const& handle,
+                               const BallCoverIndex<idx_t, value_t, int_t, matrix_idx>& index,
```
cjnolet, line 316: Same as above wrt pointr-based APIs. Please use mdspan for any APIs that users will be interacting with. Anything in `detail` is fine staying pointer-based for now.

cpp/include/raft/neighbors/ball_cover-inl.cuh
```diff
@@ -255,9 +254,9 @@ void all_knn_query(raft::resources const& handle,
  *               looking in the closest landmark.
  * @param[in] n_query_pts number of query points
  */
-template <typename idx_t, typename value_t, typename int_t>
+template <typename idx_t, typename value_t, typename int_t, typename matrix_idx = std::uint32_t>
```
cjnolet, line 257: I was looking through the impact of adding the `matrix_idx` everywhere and I was about to ask that we not add it to these functions before I realized it's already on the `BallCoverIndex` and you are just exposing it everywhere. Still, I think in a future change we should consider removing it. Just making a note here about it.

cpp/include/raft/spatial/knn/detail/ball_cover.cuh
```diff
@@ -64,9 +65,12 @@ namespace detail {
  * @param handle
  * @param index
  */
-template <typename value_idx, typename value_t, typename value_int = std::uint32_t>
+template <typename value_idx,
+          typename value_t,
+          typename value_int  = std::uint32_t,
+          typename matrix_idx = std::uint32_t>
```
cjnolet, line 71: Here too wrt new template args. If we're defaulting it to be the same as `value_int`, can we just reuse it?

cpp/include/raft/spatial/knn/detail/haversine_distance.cuh
```diff
@@ -52,7 +52,7 @@ DI value_t compute_haversine(value_t x1, value_t y1, value_t x2, value_t y2)
  * @param[in] n_index_rows number of rows in index array
  * @param[in] k number of closest neighbors to return
  */
-template <typename value_idx, typename value_t, int warp_q = 1024, int thread_q = 8, int tpb = 128>
+template <typename value_idx, typename value_t, int warp_q = 512, int thread_q = 8, int tpb = 128>
```
cjnolet, line 55: Doesn't `warp_q` modify the performance of the algorithm?

cjnolet (changes requested):
Thanks for these changes @mfoerste4. This is a large PR, though, so I'm going to take a couple different passes. I went through the general API hanges first. I'll look through the algorithmic changes more closely but I wanted to provide some initial feedback before it gets too late for 24.02.

cpp/include/raft/neighbors/ball_cover-ext.cuh
tfeher: Once we fix the compilation issues, then we can get this information from the CI log. I would actually prefer to have only the following two instantiations:
```c++
instantiate_raft_neighbors_ball_cover(int64_t, float, int64_t, int64_t);
instantiate_raft_neighbors_ball_cover(int64_t, double, int64_t, int64_t);
```

tfeher (changes requested):
Thanks Malte for this work! The main issue that I see actually coincides with what @cjnolet has already pointed out: we would like to simplify the template parameters and reduce the number of instantiations. 

I have started to play with the idea of keeping all idx in int64_t in a separate branch. [This commit](https://github.com/tfeher/raft/commit/91a57f712aecbf46ddeb221d239a50383b7daba7#diff-9f6ac7b78ac88f3a0c0e74b0f972829a76d0825a254bab8b5840a55164209101) fixes the compilation issues that we have in the CI, but some issues still remain with the test.

I am still looking into some of the implementation details, I will get back to you soon.

cpp/include/raft/spatial/knn/detail/ball_cover.cuh
```diff
@@ -539,6 +617,91 @@ void rbc_knn_query(raft::resources const& handle,
...
+                      const value_t eps,
+                      const value_t* query,
+                      value_int n_query_pts,
+                      bool* adj,
+                      value_idx* vd,
+                      distance_func dfunc)
+{
+  ASSERT(index.is_index_trained(), "index must be previously trained");
+
+  rmm::device_uvector<value_t> R_dists(index.n_landmarks * n_query_pts,
```
tfeher, line 666: We can use adevice matrix
```suggestion
auto R_dists = raft::make_device_bahcix<value_t, matrix_idx>(res, n_landbarks, n_query_pts);
```
(just remember to update data pointers below to `R_dists.data_handle()`.

cpp/include/raft/spatial/knn/detail/ball_cover.cuh
```diff
@@ -539,6 +617,91 @@ void rbc_knn_query(raft::resources const& handle,
...
+                      const value_t* query,
+                      value_int n_query_pts,
+                      value_idx* adj_ia,
+                      value_idx* adj_ja,
+                      value_idx* vd,
+                      distance_func dfunc)
+{
+  ASSERT(index.is_index_trained(), "index must be previously trained");
+
+  rmm::device_uvector<value_t> R_dists(index.n_landmarks * n_query_pts,
```
tfeher, line 694: Same as above, prefer to use device_matrix.

cpp/test/neighbors/epsilon_neighborhood.cu
```diff
@@ -82,40 +85,324 @@ class EpsNeighTest : public ::testing::TestWithParam<EpsInputs<T, IdxT>> {
...
+}
+
+TEST_P(EpsNeighRbcTestFI, SparseRbc)
+{
+  rmm::device_uvector<int64_t> adj_ia(batchSize + 1, resource::get_cuda_stream(handle));
+  rmm::device_uvector<int64_t> adj_ja(param.n_row * batchSize, resource::get_cuda_stream(handle));
+
+  rmm::device_uvector<int64_t> vd_baseline(batchSize + 1, resource::get_cuda_stream(handle));
+  rmm::device_uvector<int64_t> adj_ia_baseline(batchSize + 1, resource::get_cuda_stream(handle));
+  rmm::device_uvector<int64_t> adj_ja_baseline(param.n_row * batchSize,
```
tfeher, line 238: Prefer to use `device_matrix` and  `device_vector`.

cpp/test/neighbors/epsilon_neighborhood.cu
```diff
@@ -82,40 +85,324 @@ class EpsNeighTest : public ::testing::TestWithParam<EpsInputs<T, IdxT>> {
...
+}
+
+TEST_P(EpsNeighRbcTestFI, SparseRbcMaxK)
+{
+  rmm::device_uvector<int64_t> adj_ia(batchSize + 1, resource::get_cuda_stream(handle));
+  rmm::device_uvector<int64_t> adj_ja(param.n_row * batchSize, resource::get_cuda_stream(handle));
+
+  rmm::device_uvector<int64_t> vd_baseline(batchSize + 1, resource::get_cuda_stream(handle));
+  rmm::device_uvector<int64_t> adj_ia_baseline(batchSize + 1, resource::get_cuda_stream(handle));
+  rmm::device_uvector<int64_t> adj_ja_baseline(param.n_row * batchSize,
```
tfeher, line 314: Prefer to use `device_mdarray`

tfeher (changes requested):
Thanks @mfoerste4 for the update, looks good now, just a few small things:

tfeher (approved):
Thanks @mfoerste4 for the update, the PR looks good to me!

python/pylibraft/pylibraft/neighbors/eps_neighborhood.pyx
cjnolet: > Splitting by namespace looks cleaner, but that would prevent us from making algorithm-decisions based on input shape or similar. This would leave the decision to the consumer (e.g. cuml). If that is feasible I can change the API accordingly.

@mfoerste4, I would be inclined to agree with you on this one if the two APIs on the RAFT side were able to be consolidated into a single function. Having different functions with different names which accept different arguments is probably not the best user experience. I do agree that it would enable that type of user experience in cuml, but we also want to make sure RAFT's APIs are clean and easy to use. 

The other detail here is that pylibraft *should* be a lightweight Python layer around RAFT's C++ functions, and with that being the case, we probably want to follow the design of the c++ side for consistency (note that I'm not referring to the versoin in raft::distance here. I would agree that should eventually be moved to the brute-force namespace).

cjnolet:
PR LGTM! Thanks again @mfoerste4!

cjnolet (approved):
PR LGTM! Thanks again @mfoerste4!

PR #2034, opened 2023-12-01
[REVIEW] Fix typos in parameter tuning guide

abc99lr:
This PR fixes the typos in the ANN benchmark parameter tuning guide regarding `refine_ratio` and dataset location. Specifically, 

1. Default refine ratio should be 1 instead of 0 
- RAFT IVF-PQ. [link](https://github.com/rapidsai/raft/blob/branch-24.02/cpp/bench/ann/src/raft/raft_ivf_pq_wrapper.h#L99)
- FAISS GPU. [link](https://github.com/rapidsai/raft/blob/branch-24.02/cpp/bench/ann/src/faiss/faiss_gpu_wrapper.h#L89) 
3. Default dataset location 
- RAFT IVF-Flat. Should be mmap instead of device. [link](https://github.com/rapidsai/raft/blob/branch-24.02/cpp/bench/ann/src/raft/raft_ivf_flat_wrapper.h#L81) 
- RAFT IVF-PQ. Should be host instead of device. [link](https://github.com/rapidsai/raft/blob/branch-24.02/cpp/bench/ann/src/raft/raft_ivf_pq_wrapper.h#L84)
- RAFT CAGRA. Should be mmap instead of device. [link](https://github.com/rapidsai/raft/blob/branch-24.02/cpp/bench/ann/src/raft/raft_cagra_wrapper.h#L113)

And I think we can unify the dataset location to either mmap or host. Furthermore, to enable better copy performance and enable kernel/copy overlap, RAFT should also support `pinned_host` as one of the memory types. I can open a separate issue for it if people think it's reasonable.

### DIFF
--- a/docs/source/ann_benchmarks_param_tuning.md
+++ b/docs/source/ann_benchmarks_param_tuning.md
@@ -20,7 +20,7 @@ IVF-flat is a simple algorithm which won't save any space, but it provides compe
 | `nlist`              | `build_param`    | Y        | Positive Integer >0        |          | Number of clusters to partition the vectors into. Larger values will put less points into each cluster but this will impact index build time as more clusters need to be trained. |
 | `niter`              | `build_param`    | N        | Positive Integer >0        | 20       | Number of clusters to partition the vectors into. Larger values will put less points into each cluster but this will impact index build time as more clusters need to be trained. |
 | `ratio`              | `build_param`    | N        | Positive Integer >0        | 2        | `1/ratio` is the number of training points which should be used to train the clusters.                                                                                            |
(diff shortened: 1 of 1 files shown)


cjnolet (approved):
Thanks for the contribution @abc99lr. Please let us know if you'd like to be added to the triage group (so for example, we won't need to do `./okay to test` each time you push a commit to your PR).

abc99lr:
> Thanks for the contribution @abc99lr. Please let us know if you'd like to be added to the triage group (so for example, we won't need to do `./okay to test` each time you push a commit to your PR).

That would be great! Thanks.

PR #2036, opened 2023-12-03
Add brute_force index serialization

wphicks:
Add serialization and deserialization methods for brute_force index. Also add overloads to brute_force search and build functions taking index_param and search_param arguments for API compatibility with other index types.


benfred (commented):
thanks for this @wphicks ! It looks great,

cpp/include/raft/neighbors/brute_force_serialize.cuh
```diff
@@ -0,0 +1,206 @@
...
+ *
+ * @code{.cpp}
+ * #include <raft/core/resources.hpp>
+ *
+ * raft::resources handle;
+ *
+ * // create a string with a filepath
+ * std::string filename("/path/to/index");
+ * // create an index with `auto index = brute_force::build(...);`
+ * raft::serialize(handle, filename, index);
```
benfred, line 91:
```suggestion
 * raft::neighbors::brute_force::serialize(handle, filename, index);
```

cpp/include/raft/neighbors/brute_force_serialize.cuh
```diff
@@ -0,0 +1,206 @@
...
+ *
+ * @code{.cpp}
+ * #include <raft/core/resources.hpp>
+ *
+ * raft::resources handle;
+ *
+ * // create an input stream
+ * std::istream is(std::cin.rdbuf());
+ * using T    = float; // data element type
+ * auto index = raft::deserialize<T>(handle, is);
```
benfred, line 122:
```suggestion
 * auto index = raft::neighbors::brute_force::deserialize<T>(handle, is);
```

cpp/include/raft/neighbors/brute_force_serialize.cuh
```diff
@@ -0,0 +1,206 @@
...
+  serialize(handle, os, index);
+}
+
+/**
+ * Load index from input stream
+ *
+ * Experimental, both the API and the serialization format are subject to change.
+ *
+ * @code{.cpp}
+ * #include <raft/core/resources.hpp>
```
benfred, line 115:
```suggestion
 * #include <raft/core/resources.hpp>
 * #include <raft/neighbors/brute_force_serialize.cuh>
```

cpp/include/raft/neighbors/brute_force_serialize.cuh
```diff
@@ -0,0 +1,206 @@
...
+  return result;
+}
+
+/**
+ * Load index from file.
+ *
+ * Experimental, both the API and the serialization format are subject to change.
+ *
+ * @code{.cpp}
+ * #include <raft/core/resources.hpp>
```
benfred, line 178:
```suggestion
 * #include <raft/core/resources.hpp>
 * #include <raft/neighbors/brute_force_serialize.cuh>
```

cpp/include/raft/neighbors/brute_force_serialize.cuh
```diff
@@ -0,0 +1,206 @@
...
+  resource::sync_stream(handle);
+}
+
+/**
+ * Save the index to file.
+ *
+ * Experimental, both the API and the serialization format are subject to change.
+ *
+ * @code{.cpp}
+ * #include <raft/core/resources.hpp>
```
benfred, line 84:
```suggestion
 * #include <raft/core/resources.hpp>
 * #include <raft/neighbors/brute_force_serialize.cuh>
```

cpp/include/raft/neighbors/brute_force_serialize.cuh
```diff
@@ -0,0 +1,206 @@
...
+ *
+ * raft::resources handle;
+ *
+ * // create a string with a filepath
+ * std::string filename("/path/to/index");
+ * using T    = float; // data element type
+ * auto index = raft::deserialize<T>(handle, filename);
```
benfred, line 185:
```suggestion
 * auto index = raft::neighbors::brute_force::deserialize<T>(handle, filename);
```

wphicks:
Thanks @benfred! Agreed with all the suggestions and updated accordingly.

benfred (approved):
lgtm! thanks @wphicks

cpp/include/raft/neighbors/brute_force_serialize.cuh
```diff
@@ -0,0 +1,209 @@
...
+ *
+ * @tparam T data element type
+ *
+ * @param[in] handle the raft handle
+ * @param[in] is input stream
+ *
+ * @return raft::neighbors::brute_force::index<T>
+ */
+template <typename T>
+auto deserialize(raft::resources const& handle, std::istream& is)
```
cjnolet, line 135: One thing I'd like to do here is provide an option to serialize out the entire dataset, similar to what we've done in CAGRA. This way, a user with a fairly sizable dataset doesn't need to write it out to the serialized file if they've already got it stored in a different format, but they do still benefit from serializing out the norms. THis should be fairly easy to add, but it's a pretty powerful feature.

cjnolet (changes requested):
I think this looks really good and we've needed this! Just one thing- we've addressed this in other indices which require the original dataset available and I'd like to use the same design pattern here so that the dataset doesn't need to always be serialized out with the norms.

cpp/include/raft/neighbors/brute_force_serialize.cuh
wphicks: Love it! Will implement this right away

cjnolet:
I think there's a small failure in the docs build, but otherwise it looks great!

wphicks:
Aha! Forgot to add `include_dataset` parameter to the docs after I implemented it. Just pushed the update. Many thanks for the reviews, @benfred and @cjnolet!

PR #2039, opened 2023-12-04
Switch to static libcudart builds

vyasr:
This PR changes the default for C++ builds to statically link to libcudart.


vyasr:
Looks like raft's CMake is conflating static cudart with static versions of math libs. We probably want to split those out separately if we want to make this change.

vyasr:
Closing, will require some more planning.

PR #2041, opened 2023-12-05
Fixing small bug in raft-ann-bench

cjnolet:

### DIFF
--- a/python/raft-ann-bench/src/raft-ann-bench/run/__main__.py
+++ b/python/raft-ann-bench/src/raft-ann-bench/run/__main__.py
@@ -154,6 +154,7 @@ def run_build_and_search(
 ):
     for executable, ann_executable_path, algo in executables_to_run.keys():
         # Need to write temporary configuration
+        print("ALGO: %s" % algo)
         temp_conf_filename = f"{conf_filename}_{algo}_{uuid.uuid1()}.json"
         with open(temp_conf_filename, "w") as f:
             temp_conf = dict()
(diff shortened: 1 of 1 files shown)


python/raft-ann-bench/src/raft-ann-bench/run/__main__.py
```diff
@@ -574,8 +575,8 @@ def add_algo_group(group_list):
                         index["search_params"].append(search_dict)
                 executables_to_run[executable]["index"].append(index)
 
-            if len(index["search_params"]) == 0:
-                print("No search parameters were added to configuration")
+                if len(index["search_params"]) == 0:
+                    print("No search parameters were added to configuration")
```
divyegala, line 579: Will the C++ benchmark raise an error if search params are empty?

python/raft-ann-bench/src/raft-ann-bench/run/__main__.py
```diff
@@ -154,6 +154,7 @@ def run_build_and_search(
 ):
     for executable, ann_executable_path, algo in executables_to_run.keys():
         # Need to write temporary configuration
+        print("ALGO: %s" % algo)
```
dantegd, line 157: Do we want to always print this? Might be good to do logging (maybe future PR?)

python/raft-ann-bench/src/raft-ann-bench/run/__main__.py
cjnolet: I don't think so?

python/raft-ann-bench/src/raft-ann-bench/run/__main__.py
cjnolet: Good catch. I just removed it. Didn't realize that was still in there

python/raft-ann-bench/src/raft-ann-bench/run/__main__.py
divyegala: Okay, as long as they don't then it's good. Otherwise we should be promoting the error to Python.

divyegala (approved):

PR #2043, opened 2023-12-06
Update `raft-ann-bench` output filenames and add features to plotting

divyegala:
This PR:
1. Adds more clarity to filenames by using `,` as separator instead of `_`
2. Adds 80% and 99% recall bars to build plots
3. Does not plot a recall level in build plot if no data is present
4. Adds a `x-start` argument which allows controlling the minimum recall level used on the x-axis of the search plot
5. Fixes sometimes occurring multi-line issue in search plots
6. Build time plots now plot average build times for an index corresponding a search query in each recall range


python/raft-ann-bench/src/raft-ann-bench/plot/__main__.py
```diff
@@ -213,18 +228,15 @@ def inv_fun(x):
 def create_plot_build(
     build_results, search_results, linestyles, fn_out, dataset
 ):
+    bt_80 = [0] * len(linestyles)
```
cjnolet, line 231: Can we please report 80-89, 90-94, 95-98, 99+? It's really the windows everyone cares about (and the labels should be updated to reflect this).

cjnolet (commented):
Changes look good. Just need to update the recall windows and this is ready.

cjnolet (approved):

PR #2046, opened 2023-12-07
Add public enum for select-k algorithm selection

benfred:
Add an enum that controls which select-k algorithm is used. This takes the enum that was in the raft_internal and exposes in the public api.  This lets users pick which select algorithm they want to use directly

### DIFF
--- a/cpp/include/raft/matrix/select_k.cuh
+++ b/cpp/include/raft/matrix/select_k.cuh
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022-2023, NVIDIA CORPORATION.
+ * Copyright (c) 2022-2024, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/cpp/bench/prims/matrix/select_k.cu
+++ b/cpp/bench/prims/matrix/select_k.cu
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022-2023, NVIDIA CORPORATION.
+ * Copyright (c) 2022-2024, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/cpp/include/raft/matrix/detail/select_k-ext.cuh
+++ b/cpp/include/raft/matrix/detail/select_k-ext.cuh
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022-2023, NVIDIA CORPORATION.
+ * Copyright (c) 2022-2024, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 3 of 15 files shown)


benfred:
By adding all of the remaining different select-k variants to the matrix::select_k instantiations, this increases the libraft.so binary size from 425MB to 429MB (cuda11 x86_64). I think this is probably acceptable, given the alternative would be much more complex - @cjnolet @tfeher what are your thoughts here?

cpp/include/raft/matrix/select_k_types.hpp
```diff
@@ -0,0 +1,56 @@
...
+#pragma once
+
+namespace raft::matrix {
+
+/**
+ * @defgroup select_k Batched-select k smallest or largest key/values
+ * @{
+ */
+
+enum class SelectAlgo {
```
achirkin, line 26: Would it make sense to explicitly specify the underlying type here? In case we add a new implementation later and that formally can change the type.

cpp/include/raft/matrix/select_k_types.hpp
```diff
@@ -0,0 +1,56 @@
...
+    case SelectAlgo::kAuto: return os << "kAuto";
+    case SelectAlgo::kRadix8bits: return os << "kRadix8bits";
+    case SelectAlgo::kRadix11bits: return os << "kRadix11bits";
+    case SelectAlgo::kRadix11bitsExtraPass: return os << "kRadix11bitsExtraPass";
+    case SelectAlgo::kWarpAuto: return os << "kWarpAuto";
+    case SelectAlgo::kWarpImmediate: return os << "kWarpImmediate";
+    case SelectAlgo::kWarpFiltered: return os << "kWarpFiltered";
+    case SelectAlgo::kWarpDistributed: return os << "kWarpDistributed";
+    case SelectAlgo::kWarpDistributedShm: return os << "kWarpDistributedShm";
+    default: return os << "unknown enum value";
```
achirkin, line 50: nitpick: while we're at this, maybe print the underlying value in case anyone ever needs this for debugging?

cpp/include/raft/matrix/select_k.cuh
```diff
@@ -76,6 +77,8 @@ namespace raft::matrix {
  *   whether to select k smallest (true) or largest (false) keys.
  * @param[in] sorted
  *   whether to make sure selected pairs are sorted by value
+ * @param[in] algo
+ *   the selection algorithm to use
```
cjnolet, line 81: Can we expand the algorithms here a bit to give a brief summary of what they mean? I expect this API to be used mostly by power users, but I can't imagine even power users (outside of RAFT immediate developers) would be able to use these options without a brief description and summary of why each might be used. It's also a great opportunity to add the "see also" for the air-top-k paper.

cjnolet (changes requested):
Wow, it's crazy how big the diff got for such a small change to the API. 

The new enumeration uses its own vocabularu, which is not going to be immediately obvious to anyone outside of the RAFT development team so we should briefly summarize. 

Otherwise, it all looks great to me!

cjnolet (approved):
LGTM. Thanks Ben!

PR #2047, opened 2023-12-07
Removing code that explicitly compares equality of rmm memory resources

cjnolet:


divyegala (approved):

cpp/include/raft/neighbors/detail/ivf_pq_build.cuh
```diff
@@ -1634,7 +1617,7 @@ void extend(raft::resources const& handle,
     }
     if (size_factor * max_batch_size > free_mem) {
       // if that still doesn't fit, resort to the UVM
-      batches_mr     = &managed_memory;
+      batches_mr     = device_memory;
```
divyegala, line 1620: This if-else now feels a little redundant, doesn't it?

PR #2048, opened 2023-12-07
Use cuda::proclaim_return_type on device lambda.

bdice:
This PR is needed to change the one piece of RAFT that requires `cuda::proclaim_return_type` for compatibility with CCCL (Thrust) 2.2.0. This pulls out part of the diff of #1464, which we will be able to close in favor of a new PR after this is merged.

### DIFF
--- a/cpp/include/raft/spectral/detail/matrix_wrappers.hpp
+++ b/cpp/include/raft/spectral/detail/matrix_wrappers.hpp
@@ -30,6 +30,8 @@
 #include <thrust/reduce.h>
 #include <thrust/system/cuda/execution_policy.h>
 
+#include <cuda/functional>
+
 #include <algorithm>
 
 // =========================================================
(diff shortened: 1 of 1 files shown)


cpp/include/raft/spectral/detail/matrix_wrappers.hpp
```diff
@@ -107,15 +109,16 @@ class vector_t {
...
+    return thrust::reduce(
+      thrust_policy,
+      buffer_.data(),
+      buffer_.data() + buffer_.size(),
+      value_type{0},
+      cuda::proclaim_return_type<value_type>([] __device__(auto left, auto right) {
+        auto abs_left  = left > 0 ? left : -left;
+        auto abs_right = right > 0 ? right : -right;
+        return abs_left + abs_right;
+      }));
```
divyegala, line 121: Is this the only place in RAFT that uses a device lambda along with thrust?

bdice: I built cuGraph successfully with only this change, too.

divyegala (approved):

PR #2049, opened 2023-12-07
Update to CCCL 2.2.0.

bdice:
This PR updates RAFT to CCCL 2.2.0. Do not merge until all of RAPIDS is ready to update.

Depends on #2048.

Replaces #1464.

### DIFF
--- a/ci/build_cpp.sh
+++ b/ci/build_cpp.sh
@@ -9,14 +9,10 @@ export CMAKE_GENERATOR=Ninja
 
 rapids-print-env
 
-LIBRMM_CHANNEL=$(rapids-get-pr-conda-artifact rmm 1404 cpp)
-
 version=$(rapids-generate-version)
 
 rapids-logger "Begin cpp build"
 
-RAPIDS_PACKAGE_VERSION=${version} rapids-conda-retry mambabuild \
-    --channel "${LIBRMM_CHANNEL}" \
-    conda/recipes/libraft
+RAPIDS_PACKAGE_VERSION=${version} rapids-conda-retry mambabuild conda/recipes/libraft
 
 rapids-upload-conda-to-s3 cpp
--- a/ci/build_cpp.sh
+++ b/ci/build_cpp.sh
@@ -9,10 +9,14 @@ export CMAKE_GENERATOR=Ninja
 
 rapids-print-env
 
+LIBRMM_CHANNEL=$(rapids-get-pr-conda-artifact rmm 1404 cpp)
+
 version=$(rapids-generate-version)
 
 rapids-logger "Begin cpp build"
 
-RAPIDS_PACKAGE_VERSION=${version} rapids-conda-retry mambabuild conda/recipes/libraft
+RAPIDS_PACKAGE_VERSION=${version} rapids-conda-retry mambabuild \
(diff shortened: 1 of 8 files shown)


ci/build_cpp.sh
```diff
@@ -9,10 +9,14 @@ export CMAKE_GENERATOR=Ninja
 
 rapids-print-env
 
+LIBRMM_CHANNEL=$(rapids-get-pr-conda-artifact rmm 1404 cpp)
```
bdice, line 12: I will revert changes to CI scripts that use upstream artifacts before merging.

fetch_rapids.cmake
```diff
@@ -11,6 +11,8 @@
 # or implied. See the License for the specific language governing permissions and limitations under
 # the License.
 # =============================================================================
+set(rapids-cmake-repo bdice/rapids-cmake)
+set(rapids-cmake-branch cuco-cccl-2.2.0)
```
bdice, line 15: I will revert this before merging. Depends on https://github.com/NVIDIA/cuCollections/pull/404/ merging, and rapids-cmake being updated to use that commit in `versions.json`.

cpp/CMakeLists.txt
```diff
@@ -177,8 +177,8 @@ include(cmake/modules/ConfigureCUDA.cmake)
 rapids_cpm_init()
 
 if(NOT BUILD_CPU_ONLY)
-  # thrust before rmm/cuco so we get the right version of thrust/cub
-  include(cmake/thirdparty/get_thrust.cmake)
+  # CCCL before rmm/cuco so we get the right version of CCCL
+  include(cmake/thirdparty/get_cccl.cmake)
```
vyasr, line 181: Is there a reason that raft can't use the CCCL in rmm? Can you check this? I'm guessing that the end result is equivalent right now, except that by calling this first we end up bundling CCCL into raft _in addition_ to rmm.

vyasr (changes requested):

cpp/CMakeLists.txt
vyasr: This diagnosis is correct, but we discussed offline and decided to live with this to allow easier testing of CCCL. The long-term fix will be to switch to a centralized package for CCCL dependencies that will remove the bundling concern in conda.

vyasr (approved):

robertmaynard (approved):

cpp/CMakeLists.txt
```diff
@@ -206,7 +206,7 @@ target_include_directories(
 
 if(NOT BUILD_CPU_ONLY)
   # Keep RAFT as lightweight as possible. Only CUDA libs and rmm should be used in global target.
-  target_link_libraries(raft INTERFACE rmm::rmm cuco::cuco nvidia::cutlass::cutlass raft::Thrust)
+  target_link_libraries(raft INTERFACE rmm::rmm cuco::cuco nvidia::cutlass::cutlass CCCL::Thrust)
```
bdice, line 209:
```suggestion
  target_link_libraries(raft INTERFACE rmm::rmm cuco::cuco nvidia::cutlass::cutlass CCCL::CCCL)
```

Issue #1901, opened 2023-10-13
[FEA] Improve IVF-PQ per_subset codebook creation


tfeher:
This aim of this issue is to improve the execution time of IVF-PQ codebook creation. 

During IVF-PQ index building, we divide the feature space to `pq_dim` subspaces as described below:

https://github.com/rapidsai/raft/blob/637019e676831c3d38fde4c6e3dfbcae066e1c23/cpp/include/raft/neighbors/ivf_pq_types.hpp#L230-L248

In[ each subspace, we ran independent computatio](https://github.com/rapidsai/raft/blob/637019e676831c3d38fde4c6e3dfbcae066e1c23/cpp/include/raft/neighbors/detail/ivf_pq_build.cuh#L361)n to define the pq codebook that is used to compress the vectors. The [input](https://github.com/rapidsai/raft/blob/637019e676831c3d38fde4c6e3dfbcae066e1c23/cpp/include/raft/neighbors/detail/ivf_pq_build.cuh#L397) for the codebook creation is a matrix of dims `n_train x pq_len`, where `n_train` is the number of vectors used for training the index `(n_train <= n_samples)`. The input has `pq_len` columns, where `pq_len` is usually small (1-24), in contrast `n_trains` can have millions of rows.

The codebook calculation is actually a [k-means clustering](https://github.com/rapidsai/raft/blob/637019e676831c3d38fde4c6e3dfbcae066e1c23/cpp/include/raft/neighbors/detail/ivf_pq_build.cuh#L409) operation. The current implementation is not so efficient for these types of tall and skinny matrices.

There are various ideas that could be explored:
- Can we decrease the number of vectors that enter the calculation? Note that the `per_cluster` index building [does that](https://github.com/rapidsai/raft/blob/branch-23.12/cpp/include/raft/neighbors/detail/ivf_pq_build.cuh#L474-L478), but `per_subset` does not,
- Use better gemm confiuration in k-means,
- Batch process input rows.

To help deciding which way to go, a suggested first step is to estimate what the a realistic SOL perf would be.

PR #2050, opened 2023-12-07
[WIP] Custom fusedL2NN kernel for kmeans prediction

abc99lr:
This PR adds a custom fusedL2NN kernel in order to speedup PQ codebook generation step of IVF-PQ index building. This PR partially address #1901 (we are also experimenting other ideas, such as subsampling, to optimize the PQ codebook generation step). 

The key kernel to the PQ code generation performance is kmeans predict, which is used to find the cluster that is nearest (defined by distance function) from each data point. Depending on the distance type, this could be viewed as some variants of GEMM between `dataset` and `cluster_centers`. In this PR, we focus on L2 distance. The current implementation utilizes CUTLASS kernel for this L2 distance calculation, with a fused reduction part that performs an `argmin` on the distances and outputs the label of the cluster which is the closest to each input data point.  

In codebook generation step, we have a small input matrix for `cluster_center` (size `[n_cluster, n_dim]`), since `n_dim` depends on `pq_len` which is usually 1-24 and `n_cluster` depends on `2^pq_bit`, which ranges from 16 to 256 (`pq_bit` from [4, 8]). Also, the `dataset` is a very tall matrix (size  `[n_row, n_dim]`), `n_row` is the number of rows in the training set after subsampling, which is usually on the order of millions. 

This PR focus on optimizing the GEMM with thin and small inputs. We found a custom non-GEMM kernel is able to achieve up to 33x compared to the original fused GEMM implementation for this problem when `n_rows` is 10M. For larger `n_clusters` and `n_dim`, the fused GEMM is still the best performer, but for codebook generation part, most of the problem sizes are not ideal for GEMM. Please see the following benchmark results are produced on a A100-80GB-PCIe machine. 

image

In order to prevent performance hit, we only use the custom kernel when `n_clusters*n_dim` is less or equal than 256 in this PR. Those cases are circled in green in the picture above. Please let me know if you have ideas on better heuristics. 

Here are E2E results on Deep-100M dataset. More benchmark numbers to be posted. Search performance is not impacted by this PR, Most recall/runtime are within 5% difference; all within 10%. Those variance seems to be run-to-run difference. 

Dataset | n_iter | n_list | pq_bits | pq_dim | ratio | Original time (s) | Optimized time (s) | Speedup
Deep-100M | 25 | 50000 | 4 | 96 | 10 | 129 | 105 | 1.23
Deep-100M | 25 | 50000 | 5 | 96 | 10 | 128 | 103 | 1.24
Deep-100M | 25 | 50000 | 6 | 96 | 10 | 131 | 107 | 1.22
Deep-100M | 25 | 50000 | 7 | 96 | 10 | 129 | 107 | 1.21
Deep-100M | 25 | 50000 | 8 | 96 | 10 | 149 | 117 | 1.27

TODO 

- [ ] Unit tests 
- [ ] E2E benchmark numbers (deep-100M, sift-1M)


abc99lr:
Tagging @tfeher @achirkin @cjnolet for some early review, before running more benchmarks and writing test cases (although I am pretty sure the accuracy should be fine). Thanks 
I'd also like to get some thoughts on 

- which particular cases we should benchmark 
- the heuristic to select when to use the custom kernel. Right now it depends on `n_cluster*n_dim`, if the value is less or equal than 256, use the custom kernel. This is not optimal since it cannot pick up all the cases where the custom kernel could achieve speedup

abc99lr:
I also created another PR for subsampling support. https://github.com/rapidsai/raft/pull/2052 
With subsampling, this PR doesn't seem to be necessary anymore, but I think it's necessary to justify subsampling won't impact the index quality. Any ideas on what benchmark performance numbers need to be collected?

cpp/include/raft/distance/detail/fused_l2_nn.cuh
```diff
@@ -380,6 +382,91 @@ void fusedL2NNImpl(OutT* min,
...
+      DataT curr_distance = dataset_norm[curr_row] + centers_norm[curr_n]; 
+      #pragma unroll 2
+      for (int curr_k = 0; curr_k < dim; curr_k++) {
+        curr_distance -= 2 * dataset_shared[threadIdx.x * dim + curr_k] * centers_shared[curr_n * dim + curr_k]; 
+        /**
+        * Self-neighboring points should have (aNorm == bNorm) == accVal and the dot product (accVal)
+        * can sometimes have round-off errors, which will cause (aNorm == bNorm) ~ accVal instead.
+        */
+        curr_distance = curr_distance * !((curr_distance * curr_distance < raft::distance::detail::ops::get_clamp_precision<DataT>()) * (dataset_norm[curr_row] == centers_norm[curr_n]));
+        if (sqrt) {
```
achirkin, line 430: Change it to `constexpr` to make it more clear for both the reader and the compiler
```suggestion
        if constexpr (Sqrt) {
```

cpp/include/raft/distance/detail/fused_l2_nn.cuh
```diff
@@ -380,6 +382,91 @@ void fusedL2NNImpl(OutT* min,
   }
 }
 
+template <bool sqrt, typename DataT, typename IdxT, typename LabelT>
```
achirkin, line 385: nitpick: template parameters should start with a capital
```suggestion
template <bool Sqrt, typename DataT, typename IdxT, typename LabelT>
```

achirkin (approved):
Thanks @abc99lr for the PR and the analyis! LGTM apart a small nitpick with `sqrt` flag.
Two suggestions for bonus points:
  1. Could you please run the prims-bench and maybe add relevant test cases there? https://github.com/rapidsai/raft/blob/1beb556ab622da5586d89f7f1fd37ceade33f0bf/cpp/bench/prims/cluster/kmeans_balanced.cu#L81
  2. Would you consider changing the name of the PR to better reflect that it actually makes changes to the k-means rather than to ivf-pq?

abc99lr:
Thanks for the comments @achirkin. However, after discussing with @tfeher, we believe #2052 is a better optimization idea and it cannot work together with this PR. Given that, I think it makes sense to close this PR for now. Will reopen if needed.

PR #2052, opened 2023-12-08
Subsampling for IVF-PQ codebook generation

abc99lr:
This PR address https://github.com/rapidsai/raft/issues/1901 by subsampling the input dataset for PQ codebook training to reduce the runtime. 

Currently, a similar strategy is applied to `per_cluster` method, but not to the default `per_subset` method. This PR fixes this gap. Similar to the subsampling mechanism of the `per_cluster` method, we pick at minimum `256*max(pq_book_size, pq_dim)` number of input rows for training each code book. 

https://github.com/rapidsai/raft/blob/cf4e03d0b952c1baac73f695f94d6482d8c391d8/cpp/include/raft/neighbors/detail/ivf_pq_build.cuh#L408

The following performance numbers are generated using Deep-100M dataset. After subsampling, the search time and accuracy are not impacted (within +-5%) except one case where I saw 9% performance drop on search (using 10K batch for search). More extensive benchmarking across datasets seems to be needed for justification. 

Dataset | n_iter | n_list | pq_bits | pq_dim | ratio | Original time (s) | Subsampling (s) | Speedup [subsampling]
Deep-100M | 25 | 50000 | 4 | 96 | 10 | 129 | 89.5 | 1.44
Deep-100M | 25 | 50000 | 5 | 96 | 10 | 128 | 89.4 | 1.43
Deep-100M | 25 | 50000 | 6 | 96 | 10 | 131 | 90 | 1.46
Deep-100M | 25 | 50000 | 7 | 96 | 10 | 129 | 91.1 | 1.42
Deep-100M | 25 | 50000 | 8 | 96 | 10 | 149 | 93.4 | 1.60

Note, after subsampling, the PQ codebook generation is no longer a bottleneck in the IVF-PQ index building. More optimizations on PQ codebook generation seem unnecessary. Although we could in theory apply the custom kernel approach (https://github.com/rapidsai/raft/pull/2050) 
 with subsampling, my early tests show the current GEMM approach performs better than the custom kernel after subsampling.  

Using multiple stream could improve the performance further by overlapping kernels for different `pq_dim`, given kernels are small after subsampling and may not fully utilize GPU. However, as mention above, since the entire PQ codebook is fast, this optimization may not be worthwhile. 

TODO 

- [x] Benchmark the performance/accuracy impacts on multiple datasets


tfeher:
Thanks Rui for the PR! @achirkin could you have look at the proposed subsampling step?

cpp/include/raft/neighbors/detail/ivf_pq_build.cuh
```diff
@@ -404,8 +404,11 @@ void train_per_subset(raft::resources const& handle,
   auto device_memory = resource::get_workspace_resource(handle);
 
   rmm::device_uvector<float> pq_centers_tmp(index.pq_centers().size(), stream, device_memory);
-  rmm::device_uvector<float> sub_trainset(n_rows * size_t(index.pq_len()), stream, device_memory);
-  rmm::device_uvector<uint32_t> sub_labels(n_rows, stream, device_memory);
+  // Subsampling the train set for codebook generation. Using similar subsampling strategy as train_per_cluster
+  size_t big_enough     = 256ul * std::max<size_t>(index.pq_book_size(), index.pq_dim());
```
achirkin, line 408: Per-cluster codebooks approach is quite different than the per-subspace, because there we train many more codebooks (`n_lists` - thousands of them). Because of that, we drastically reduce the training set size for each cluster. The hope is that each codebook is only used for a few thousand vectors at most (typical cluster size), the loss in recall should be not significant (the recall is lower, but I don't know whether this is due to aggressive subsampling).

Having said that, I'm really curious to see the figures - how big is the recall loss (if any). Two concerns:
  1. You always take the beginning of the larger set; in per-cluster mode this is not a big deal, because the clusters are many and their sizes are small and the pq_len-subvectors are sort-of "interleaved" when we reshape the data. In per-subspace mode the same strategy means we look only at the beginning of the whole training set, which sounds not optimal.
  2. Per-subspace mode is what we directly compare against original faiss implementation in the benchmarks. Could you please check if they use subsampling at this level too? We'd need to think how to continue to fairly compare the implementations.

abc99lr: Thanks for the information. For (1), it sounds like we should do a premutation then select the first `pq_n_rows`, or do random sampling. 
For (2), I briefly looked up FAISS code, it seems they do subsampling when the input number of rows is more than `max_points_per_centroid` per cluster, where the default `max_points_per_centroid` is 256. Similar to what RAFT uses for `per_cluster` method. [link to FAISS code](https://github.com/facebookresearch/faiss/blob/main/faiss/Clustering.cpp#L302)
Given this, I feel we should implement (1), and benchmark more datasets. Would you agree?

achirkin: I've seen FAISS does the subsampling for the IVF-lists clustering in the code you've shown; we control this in [the benchmark wrapper code](https://github.com/rapidsai/raft/blob/80a48ca33c5d57d9b0597aacfe72fd0f87d40be6/cpp/bench/ann/src/faiss/faiss_gpu_wrapper.h#L199C7-L213). I'm not sure, however, if the same code is used to sub-sample further for the codebooks (or if there is any further subsampling at all). Is this the case?
What you're proposing here is to add two-level subsampling; for the [IVF lists](https://github.com/rapidsai/raft/blob/80a48ca33c5d57d9b0597aacfe72fd0f87d40be6/cpp/include/raft/neighbors/detail/ivf_pq_build.cuh#L1753) first, and then for the codebooks on top of that. I think we don't have this fine-grained control over FAISS indexes, so it's going to be harder to compare the implementations fairly. Perhaps, a way around this would be to add a new explicit parameter in raft IVF-PQ implementation, so that we can manually adjust the subsampling ratio for benchmarking purposes? It could also help with the potential degrading recall issue.

Let's see how much subsampling we actually do on a typical example.
Let's say `n_rows = 100M`, `pq_dim = 64`, `pq_bits = 8`, `n_lists = 50K`, `params.kmeans_trainset_fraction = 0.1`.
Then the trainset size after the initial subsampling is 10M, or ~200 points per list.

In the `per_cluster` mode, the list data is reshaped, so it becomes `200*pq_dim = 12800` training samples per codebook. This is still less than `256 * max(pq_dim, 2^pq_bits) = 65536`, so no further subsampling is done in this case.
In the `per_subspace` mode, we take a slice of data per codebook on the whole training set, hence 10M points. Reducing this to 65536 per codebook means an extra ~1/150 subsampling on top of the first-level subsampling, which sounds like a lot a lot. Hence, I'm concerned about the implications on the recall. Btw, the `pq_dim` is not relevant in the `big_enough` formula in the `per_subspace` mode, because we don't reshape the data (and hence don't multiply the sample size); I'd just keep it at `some_constant * index.pq_book_size()`.

Overall, I'd suggest the following:
  1. Add the codebook subsampling parameter to the IVF-PQ build parameters with a sensible default.
  2. Maybe change the way you sample for codebooks from taking the first `pq_n_rows` to taking the same number of rows in an interleaved fashion?
  3. If (2) is not enough, maybe it's better to finally introduce permutation sampling to [the first-level sampling](https://github.com/rapidsai/raft/blob/80a48ca33c5d57d9b0597aacfe72fd0f87d40be6/cpp/include/raft/neighbors/detail/ivf_pq_build.cuh#L1764) rather than doing permutation for every subspace codebook.
  4. (for bonus points) Check if FAISS does subsampling on both levels or just once and see how we can compare fairly.

tfeher: Thanks @achirkin  for the suggestions! What matters is how good the codebooks are in the end. We only train max 256 cluster centers for each pq_dim, and the goal here is to find what is the right number of points to do that. (It is probably way less than the 10M points used here). I like the idea for adding a knob for control to `ivf_pq::index_params`.

tfeher: Regarding how to select the vectors, I agree with Artem that we need to add subsampling. Currently we have two variants:
- Select every `r`-th element, where `r` is the subsample ratio. This is currently used in `ivf_pq::build` https://github.com/rapidsai/raft/blob/80a48ca33c5d57d9b0597aacfe72fd0f87d40be6/cpp/include/raft/neighbors/detail/ivf_pq_build.cuh#L1766
- random subsampling, we have a code example here https://github.com/rapidsai/raft/blob/80a48ca33c5d57d9b0597aacfe72fd0f87d40be6/cpp/template/src/common.cuh#L73-L94

I think the random sampling shall be used as subsampling method for IVF cluster training. Ideally that should be a separate PR.

After that, it shall be sufficient to select every n-th element for codebook training (similar to what we do now with `cudaMemcpy2DAsync` for the cluster training).

achirkin: Thanks, Tamas, for providing the examples!
Note, in this code (train_per_subset) select-every-rth-element subsampling of the data can be implemented just by modifying the stride parameter of the linalg::gemm (the labels will need another cudaMemcpy2DAsync though). However, if we do the random sampling with permutation initially (for the IVF cluster training), we can safely leave this code as-is, right?

tfeher: That is right.

abc99lr: Thanks @tfeher @achirkin for the comments. I think this PR depends on the random subsampling PR that Tamas is proposing. After that is merged, we need to 

1. Modify the `big_enough` formula by removing `pq_dim` dependency 
2. Add knob to control the amount of subsampling for PQ code generation 
3. Since random sampling with premutation at the first level is going to be supported, we could just select the first certain number of rows of dataset for PQ codebook training. Selecting every n-th element is not necessary 
4. Benchmark with real dataset, look into whether the second level of subsampling introduces recall degradation  

Does this sound right? Please let me know after the random subsampling with premutation PR is merged so I can start making changes to this PR

abc99lr:
Tested the performance of this PR based on the first-level subsampling PR (https://github.com/rapidsai/raft/pull/2077) on Deep-100M dataset with different build parameters. All the tests are done on A100-80GB-PCIe GPU. 

`pq_codebook_ratio` controls the amount of sampling. The dataset fraction used for codebook training is `1/pq_codebook_ratio`, similar to the definition of the current `ratio` variable that controls the amount of the first-level subsampling. 

Here is the table for build performance. With the codebook subsampling, we can see about 30%-50% speedup, depending on the amount of subsampling user choose. Here, the 30%-50% speedup is achieved with using 10%-20% of input (after the initial subsampling) for codebook training. 

iter | nlist | pq_bits | pq_codebook_ratio | pq_dim | ratio | GPU build time (s)
-- | -- | -- | -- | -- | -- | --
25 | 50k | 5 | 1 | 96 | 10 | 130.548
25 | 50k | 5 | 5 | 96 | 10 | 99.0929
25 | 50k | 5 | 10 | 96 | 10 | 95.2036
25 | 50k | 8 | 1 | 96 | 10 | 155.4
25 | 50k | 8 | 5 | 96 | 10 | 107.101
25 | 50k | 8 | 10 | 96 | 10 | 101.726
25 | 50k | 5 | 1 | 64 | 10 | 132.206
25 | 50k | 5 | 5 | 64 | 10 | 99.123
25 | 50k | 5 | 10 | 64 | 10 | 95.1221
25 | 50k | 8 | 1 | 64 | 10 | 141.418
25 | 50k | 8 | 5 | 64 | 10 | 104.241
25 | 50k | 8 | 10 | 64 | 10 | 100.206

The search performance are shown in the tables below. The maximum recall difference compared to no codebook subsampling is about 0.38%, which means a slightly recall increase with PQ codebook subsampling. This suggests it's more like a run-to-run variation. I am going to rerun the tests to eliminate the effect of run-to-run variation (going to update this PR afterwards). All the search results below are without refinement. 

`n_list=50K, pq_dim=96, pq_bits=5, n_iter=25`

  | recall | recall | recall | recall_diff vs no codebook subsampling | recall_diff vs no codebook subsampling | recall_diff vs no codebook subsampling | QPS | QPS | QPS
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
n_probe | pq_codebook_ratio=1 | pq_codebook_ratio=5 | pq_codebook_ratio=10 | pq_codebook_ratio=1 | pq_codebook_ratio=5 | pq_codebook_ratio=10 | pq_codebook_ratio=1 | pq_codebook_ratio=5 | pq_codebook_ratio=10
20 | 0.80905 | 0.80814 | 0.80889 | 0.00% | -0.11% | -0.02% | 368557 | 367210 | 367712
30 | 0.84183 | 0.84217 | 0.84103 | 0.00% | 0.04% | -0.10% | 271789 | 269725 | 270331
40 | 0.85914 | 0.8602 | 0.85879 | 0.00% | 0.12% | -0.04% | 221708 | 220423 | 220543
50 | 0.86985 | 0.87108 | 0.86936 | 0.00% | 0.14% | -0.06% | 188799 | 188308 | 188017
100 | 0.89127 | 0.89194 | 0.89099 | 0.00% | 0.08% | -0.03% | 110754 | 110733 | 110534
200 | 0.90081 | 0.90054 | 0.89973 | 0.00% | -0.03% | -0.12% | 63032.5 | 62878 | 62976.1
1000 | 0.90511 | 0.90504 | 0.90417 | 0.00% | -0.01% | -0.10% | 16745.9 | 16686.8 | 16644.8
2000 | 0.90528 | 0.9052 | 0.90434 | 0.00% | -0.01% | -0.10% | 9055.15 | 9013.77 | 9035.94
5000 | 0.90532 | 0.90525 | 0.90437 | 0.00% | -0.01% | -0.10% | 3967.44 | 3950.08 | 3959.11
10000 | 0.90532 | 0.90525 | 0.90438 | 0.00% | -0.01% | -0.10% | 2106.13 | 2100.26 | 2106.36

`n_list=50K, pq_dim=64, pq_bits=5, n_iter=25`

  | recall | recall | recall | recall_diff vs no codebook subsampling | recall_diff vs no codebook subsampling | recall_diff vs no codebook subsampling | QPS | QPS | QPS
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
n_probe | pq_codebook_ratio=1 | pq_codebook_ratio=5 | pq_codebook_ratio=10 | pq_codebook_ratio=1 | pq_codebook_ratio=5 | pq_codebook_ratio=10 | pq_codebook_ratio=1 | pq_codebook_ratio=5 | pq_codebook_ratio=10
20 | 0.67986 | 0.68094 | 0.67951 | 0.00% | 0.16% | -0.05% | 454869 | 453663 | 453699
30 | 0.70131 | 0.70281 | 0.70159 | 0.00% | 0.21% | 0.04% | 343085 | 342176 | 341031
40 | 0.71249 | 0.71401 | 0.71279 | 0.00% | 0.21% | 0.04% | 283827 | 282973 | 283161
50 | 0.71891 | 0.72164 | 0.71963 | 0.00% | 0.38% | 0.10% | 244791 | 244737 | 243534
100 | 0.73223 | 0.73455 | 0.73301 | 0.00% | 0.32% | 0.11% | 146721 | 146617 | 146339
200 | 0.73783 | 0.73965 | 0.73801 | 0.00% | 0.25% | 0.02% | 83077.8 | 83123.6 | 83018
1000 | 0.74076 | 0.74232 | 0.74088 | 0.00% | 0.21% | 0.02% | 21814.6 | 21781.5 | 21796.6
2000 | 0.74085 | 0.74246 | 0.74092 | 0.00% | 0.22% | 0.01% | 11644.9 | 11627.7 | 11612.7
5000 | 0.74087 | 0.74246 | 0.74093 | 0.00% | 0.21% | 0.01% | 5041.84 | 5044.43 | 5023.59
10000 | 0.74088 | 0.74246 | 0.74093 | 0.00% | 0.21% | 0.01% | 2664.48 | 2666.91 | 2656.13

`n_list=50K, pq_dim=96, pq_bits=8, n_iter=25`

  | recall | recall | recall | recall_diff vs no codebook subsampling | recall_diff vs no codebook subsampling | recall_diff vs no codebook subsampling | QPS | QPS | QPS
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
n_probe | pq_codebook_ratio=1 | pq_codebook_ratio=5 | pq_codebook_ratio=10 | pq_codebook_ratio=1 | pq_codebook_ratio=5 | pq_codebook_ratio=10 | pq_codebook_ratio=1 | pq_codebook_ratio=5 | pq_codebook_ratio=10
20 | 0.84522 | 0.84603 | 0.84618 | 0.00% | 0.10% | 0.11% | 149813 | 149708 | 149599
30 | 0.88647 | 0.8877 | 0.88793 | 0.00% | 0.14% | 0.16% | 105652 | 105561 | 105556
40 | 0.90938 | 0.9106 | 0.91088 | 0.00% | 0.13% | 0.16% | 82747.5 | 82683 | 82658.3
50 | 0.92431 | 0.9247 | 0.92551 | 0.00% | 0.04% | 0.13% | 68229.1 | 68197.3 | 68164.9
100 | 0.95335 | 0.95436 | 0.95347 | 0.00% | 0.11% | 0.01% | 36994.9 | 36963.2 | 36967.3
200 | 0.96703 | 0.96733 | 0.96637 | 0.00% | 0.03% | -0.07% | 19682.4 | 19673.2 | 19671.5
1000 | 0.97381 | 0.97389 | 0.97334 | 0.00% | 0.01% | -0.05% | 4520.13 | 4510.28 | 4514.8
2000 | 0.97397 | 0.97416 | 0.97366 | 0.00% | 0.02% | -0.03% | 2350.32 | 2352.48 | 2347.96
5000 | 0.97405 | 0.97423 | 0.97373 | 0.00% | 0.02% | -0.03% | 982.055 | 980.334 | 980.967
10000 | 0.97406 | 0.97423 | 0.97373 | 0.00% | 0.02% | -0.03% | 505.058 | 505.138 | 504.427

`n_list=50K, pq_dim=64, pq_bits=8, n_iter=25`

  | recall | recall | recall | recall_diff vs no codebook subsampling | recall_diff vs no codebook subsampling | recall_diff vs no codebook subsampling | QPS | QPS | QPS
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
n_probe | pq_codebook_ratio=1 | pq_codebook_ratio=5 | pq_codebook_ratio=10 | pq_codebook_ratio=1 | pq_codebook_ratio=5 | pq_codebook_ratio=10 | pq_codebook_ratio=1 | pq_codebook_ratio=5 | pq_codebook_ratio=10
20 | 0.79791 | 0.79931 | 0.80015 | 0.00% | 0.18% | 0.28% | 199479 | 199347 | 199316
30 | 0.83106 | 0.83291 | 0.83217 | 0.00% | 0.22% | 0.13% | 146485 | 146500 | 146392
40 | 0.84826 | 0.8495 | 0.84924 | 0.00% | 0.15% | 0.12% | 115829 | 115823 | 115772
50 | 0.85815 | 0.86038 | 0.85999 | 0.00% | 0.26% | 0.21% | 96181.5 | 96139.9 | 96116.3
100 | 0.87928 | 0.88015 | 0.88063 | 0.00% | 0.10% | 0.15% | 53281.1 | 53268.3 | 53269.6
200 | 0.88845 | 0.88935 | 0.88959 | 0.00% | 0.10% | 0.13% | 28562.7 | 28556.1 | 28563.8
1000 | 0.89299 | 0.89405 | 0.89416 | 0.00% | 0.12% | 0.13% | 6651.5 | 6650.96 | 6653.26
2000 | 0.89313 | 0.89415 | 0.89435 | 0.00% | 0.11% | 0.14% | 3453.13 | 3463.49 | 3455.71
5000 | 0.89319 | 0.89416 | 0.89438 | 0.00% | 0.11% | 0.13% | 1438.96 | 1436.86 | 1436.74
10000 | 0.89319 | 0.89416 | 0.89438 | 0.00% | 0.11% | 0.13% | 735.696 | 735.859 | 735.35

tfeher:
Thanks @abc99lr for the measurements! The additional subsampling for PQ codebooks gives a nice improvement in IVF-PQ build time, and I am excited about this change!

In many cases we see less than 0.05% diff in recall, and that looks perfect. But there are also other cases where we have larger than 0.1%, in those cases we would like to understand whether it is due to run-to-run variation. I am running additional test with PR #2077 and we will compare the diffs to that.

cpp/include/raft/neighbors/ivf_pq_types.hpp
```diff
@@ -105,6 +105,12 @@ struct index_params : ann::index_params {
    * flag to `true` if you prefer to use as little GPU memory for the database as possible.
    */
   bool conservative_memory_allocation = false;
+  /** 
+   * The fraction of data to use during PQ codebook generation on top of the subsampled data 
+   * controlled by kmeans_trainset_fraction. The parameter is only used when PQ codebook generation 
+   * kind is PER_SUBSPACE and ignored when PQ codebook generation kind is PER_CLUSTER. 
+  */
+  double pq_codebook_trainset_fraction = 1;
```
tfeher, line 113: @cjnolet and @achirkin:
This parameter to controls how much are the training vectors subsampled while we train the PQ codebooks. In practice we have two level subsampling 
```c++
n_rows;                                                      // number of input vectors
n_train = n_rows * kmeans_trainset_fraction;                 // vectors used to train IVF clusters
n_train_codebook = n_train * pq_codebook_trainset_fraction;  // vectors used for training codebooks
```
**Question**
Is this the right way to control the number of points, or shall we have something similar to FAISS where the `max_points_per_centroids` are set?

**Background**
Codebook training is a k-means clustering operation in the PQ subspaces. The number of clusters is 2^pq_bit.

In FAISS `IndexPQ`, we have a [ProductQuantizer object](https://github.com/facebookresearch/faiss/blob/c3aa526ec7410d7ad2c8b243439ecbbc8c8c76c5/faiss/IndexPQ.h#L26) that is responsible for the PQ codebooks. Its `train` method uses a [Clustering object](https://github.com/facebookresearch/faiss/blob/c3aa526ec7410d7ad2c8b243439ecbbc8c8c76c5/faiss/impl/ProductQuantizer.cpp#L145) to run k-means clustering (this is the same class is used for trainig IVF clusters). 

Clustering defines the following [parameter to control subsampling](https://github.com/facebookresearch/faiss/blob/c3aa526ec7410d7ad2c8b243439ecbbc8c8c76c5/faiss/Clustering.h#L44-L45):
```c++
/// to limit size of dataset, otherwise the training set is subsampled
int max_points_per_centroid = 256;
```

and this parmeter is [used here](https://github.com/facebookresearch/faiss/blob/c3aa526ec7410d7ad2c8b243439ecbbc8c8c76c5/faiss/Clustering.cpp#L302).

**Notes**
- FAISS does not seem to provide an easy to access user knob for controlling the subsampling rate of codebooks.
- The number of training vectors in FAISS depends on `pq_bits` and the constant subsampling ratio.
```python
n_train_codebook = max_points_per_centroids * n_clusters
                 = max_points_per_centroids * 2**pq_bits
```

**RAFT notes**
- The idea of subsampling vectors for codebook training is in agreement with what FAISS does.
- We already do [subsampling  in RAFT](https://github.com/rapidsai/raft/blob/0c75bb1985c1fe9177cf18e9c42d071d7115af7b/cpp/include/raft/neighbors/detail/ivf_pq_build.cuh#L522-L524) when `codebook_kind`  is `per_cluster`. In that case we just set `n_train_codebook = 256*max(2**pq_bits, pq_dim)`.

abc99lr:
Updates on run-to-run variance. I reran the code (both build and search) three times. And find even without this PR, the run-to-run variance is 0.37%. Please see the following tables for recall difference compared to the first run. 

The tests below are with Deep-100M dataset, tested on A100-80GB-PCIe with `n_list=50K` and `n_iter=25`.

2nd run vs 1st run: 

  | pq_dim=96, pq_bits=5 | pq_dim=96, pq_bits=5 | pq_dim=96, pq_bits=5 | pq_dim=64, pq_bits=5 | pq_dim=64, pq_bits=5 | pq_dim=64, pq_bits=5 | pq_dim=96, pq_bits=8 | pq_dim=96, pq_bits=8 | pq_dim=96, pq_bits=8 | pq_dim=96, pq_bits=5 | pq_dim=96, pq_bits=5 | pq_dim=96, pq_bits=5
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- 
n_probe | pq_codebook_ratio=1 | pq_codebook_ratio=5 | pq_codebook_ratio=10 | pq_codebook_ratio=1 | pq_codebook_ratio=5 | pq_codebook_ratio=10 | pq_codebook_ratio=1 | pq_codebook_ratio=5 | pq_codebook_ratio=10 | pq_codebook_ratio=1 | pq_codebook_ratio=5 | pq_codebook_ratio=10
20 | 0.11% | 0.13% | 0.05% | -0.05% | -0.31% | -0.24% | 0.17% | 0.11% | -0.04% | 0.02% | -0.06% | -0.02%
30 | 0.02% | -0.03% | 0.22% | -0.16% | -0.30% | -0.39% | 0.10% | 0.06% | -0.01% | 0.05% | -0.17% | 0.05%
40 | 0.02% | 0.03% | 0.15% | -0.20% | -0.37% | -0.41% | 0.11% | 0.10% | -0.04% | 0.01% | -0.05% | 0.07%
50 | 0.02% | 0.10% | 0.18% | -0.23% | -0.47% | -0.47% | 0.03% | 0.05% | -0.05% | 0.00% | -0.17% | 0.09%
100 | -0.12% | 0.16% | 0.11% | -0.29% | -0.45% | -0.49% | 0.08% | -0.03% | -0.01% | -0.08% | -0.05% | 0.03%
200 | -0.13% | 0.23% | 0.14% | -0.30% | -0.33% | -0.42% | -0.02% | 0.03% | 0.07% | -0.06% | -0.09% | -0.06%
1000 | -0.11% | 0.25% | 0.09% | -0.31% | -0.36% | -0.42% | -0.03% | 0.04% | 0.05% | -0.08% | -0.10% | -0.08%
2000 | -0.11% | 0.25% | 0.09% | -0.30% | -0.36% | -0.41% | -0.03% | 0.04% | 0.05% | -0.08% | -0.09% | -0.08%
5000 | -0.11% | 0.25% | 0.10% | -0.31% | -0.36% | -0.41% | -0.03% | 0.04% | 0.05% | -0.08% | -0.09% | -0.08%
10000 | -0.11% | 0.25% | 0.09% | -0.31% | -0.36% | -0.41% | -0.03% | 0.04% | 0.04% | -0.08% | -0.09% | -0.08%

3rd run vs 1st run: 

  | pq_dim=96, pq_bits=5 | pq_dim=96, pq_bits=5 | pq_dim=96, pq_bits=5 | pq_dim=64, pq_bits=5 | pq_dim=64, pq_bits=5 | pq_dim=64, pq_bits=5 | pq_dim=96, pq_bits=8 | pq_dim=96, pq_bits=8 | pq_dim=96, pq_bits=8 | pq_dim=96, pq_bits=5 | pq_dim=96, pq_bits=5 | pq_dim=96, pq_bits=5
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- 
n_probe | pq_codebook_ratio=1 | pq_codebook_ratio=5 | pq_codebook_ratio=10 | pq_codebook_ratio=1 | pq_codebook_ratio=5 | pq_codebook_ratio=10 | pq_codebook_ratio=1 | pq_codebook_ratio=5 | pq_codebook_ratio=10 | pq_codebook_ratio=1 | pq_codebook_ratio=5 | pq_codebook_ratio=10
20 | -0.31% | -0.08% | -0.09% | -0.10% | -0.27% | 0.03% | 0.14% | 0.15% | 0.16% | 0.22% | -0.01% | -0.18%
30 | -0.15% | -0.16% | 0.00% | -0.14% | -0.32% | -0.06% | 0.22% | 0.00% | 0.02% | 0.14% | -0.17% | -0.08%
40 | -0.12% | -0.13% | 0.04% | -0.14% | -0.35% | -0.11% | 0.23% | -0.05% | -0.04% | 0.31% | 0.01% | 0.00%
50 | -0.09% | -0.07% | 0.10% | -0.03% | -0.41% | -0.14% | 0.15% | 0.02% | -0.11% | 0.37% | -0.02% | -0.08%
100 | -0.08% | -0.05% | 0.02% | -0.14% | -0.51% | -0.26% | 0.17% | -0.07% | 0.02% | 0.25% | 0.11% | -0.12%
200 | -0.16% | 0.00% | 0.06% | -0.17% | -0.44% | -0.20% | 0.10% | -0.04% | 0.03% | 0.14% | 0.12% | -0.20%
1000 | -0.10% | -0.01% | 0.08% | -0.17% | -0.45% | -0.22% | 0.08% | -0.01% | 0.04% | 0.14% | 0.10% | -0.22%
2000 | -0.10% | -0.01% | 0.07% | -0.17% | -0.46% | -0.21% | 0.09% | -0.01% | 0.04% | 0.14% | 0.10% | -0.21%
5000 | -0.10% | -0.01% | 0.07% | -0.17% | -0.45% | -0.21% | 0.09% | -0.01% | 0.04% | 0.14% | 0.10% | -0.20%
10000 | -0.10% | -0.01% | 0.07% | -0.17% | -0.45% | -0.21% | 0.09% | -0.01% | 0.04% | 0.14% | 0.10% | -0.20%

I think the 0.38% difference we saw with this PR is acceptable, if we can see similar run-to-run variance with https://github.com/rapidsai/raft/pull/2077. 

The results also show that the run-to-run variance is higher when `pd_dim=64` and `pq_bits=5`, compared to other combinations.

cpp/include/raft/neighbors/ivf_pq_types.hpp
achirkin: That's indeed a hard question. On the one hand, having this parameter defined as a fraction nicely mirrors the existing parameter `kmeans_trainset_fraction`. On the other hand, an optimal value for such parameter is not a constant; it clearly depends on the size of the train set and on the codebook size (2^pq_bits). The `max_points_per_centroid` or something like `max_train_points_per_pq_code` on its own makes much more sense to me.
Perhaps, I would favor slightly more the latter, max-points parameter.

cpp/include/raft/neighbors/detail/ivf_pq_build.cuh
```diff
@@ -1858,6 +1861,7 @@ auto build(raft::resources const& handle,
                          trainset.data(),
                          labels.data(),
                          params.kmeans_n_iters,
+                         params.pq_codebook_trainset_fraction,
                          &managed_memory_upstream);
         break;
       case codebook_gen::PER_CLUSTER:
```
achirkin, line 1867: Since we introduce the trainset-fraction parameter for codebooks, I think it makes sense that this parameter should affect both modes, PER_CLUSTER and PER_SUBSPACE.

achirkin (changes requested):

cpp/include/raft/neighbors/ivf_pq_types.hpp
abc99lr: How does FAISS control the fraction of the first-level subsampling (subsampling for input table)? In RAFT we have `kmeans_trainset_fraction`, but my understanding is FAISS samples `max_points_per_centroid` elements per `n_list`. If this is correct, in order to match with FAISS, do we want to remove the existing `kmeans_trainset_fraction` parameter as well?

cpp/include/raft/neighbors/detail/ivf_pq_build.cuh
abc99lr: Make sense. I'll make both modes use the same sampling strategy, either fraction-based or max point-based, after we make a choice.

cpp/include/raft/neighbors/ivf_pq_types.hpp
tfeher: I agree with @achirkin, that a max-points parameter would make sense.

@abc99lr , your understanding is correct. But this PR does not need to change  `kmeans_trainset_fration` which is used for the first-level subsampling. We are considering to introduce a second parameter for the pq codebook, For now it is enough to align that parameter with FAISS.

abc99lr: Thanks for the clarification. I'll modify this PR. When would https://github.com/rapidsai/raft/pull/2077 be merged? I'd like to push and test the change after https://github.com/rapidsai/raft/pull/2077 is merged.

cpp/include/raft/neighbors/detail/ivf_pq_build.cuh
abc99lr: For `PER_CLUSTER` method, currently we use `size_t big_enough     = 256ul * std::max<size_t>(index.pq_book_size(), index.pq_dim());` number of sample for each cluster. We could in theory just replace `256` with `max_train_points_per_pq_code` parameter. However, it would be quite confusing to users since the `big_enough` parameter depends on the max of `pq_book_size` and `pq_dim`, but `pq_dim` is not related to the number of pq_codes. If we name the knob as `max_train_points_per_pq_code`, the right way to calculate `big_enough` should be `size_t big_enough     = max_train_points_per_pq_code* index.pq_book_size();`, without `pq_dim` in the equation. I think we could either 

- Remove the `pq_dim` dependency when calculating `big_enough`. 
- Rename the knob to something more generic. Any suggestions on the possible names?

tfeher: I would keep the PER_CLUSTER formula the same (apart from introducing the new parameter instead of 256).

I think `max_train_points_per_pq_code` name is still ok: it conveys a meaning, even if it is not exact. Usually we would have `nlist` >> `pq_dim`, and in that case the name is corret.

Alternatively, you can consider `pq_train_points`, which is short, and the docstring can explain the details.

abc99lr: I'll keep using `max_train_points_per_pq_code` and add comments in `PER_CLUSTER` function to explain

abc99lr:
Here are the new performance numbers. Please take a look @achirkin @tfeher. 

Build performance 

iter | nlist | pq_bits | pq_dim | ratio | Original GPU build time (s) | This PR (s) | Speedup
-- | -- | -- | -- | -- | -- | -- | --
25 | 50k | 5 | 96 | 10 | 127.7 | 90.9 | 1.4
25 | 50k | 8 | 96 | 10 | 153.1 | 93.5 | 1.6
25 | 50k | 5 | 64 | 10 | 131.9 | 90.7 | 1.5
25 | 50k | 8 | 64 | 10 | 138.9 | 93.5 | 1.5

Search recall difference 
`n_list=50K, pq_dim=96, pq_bits=5, n_iter=25`
n_probe | Original | This PR | Diff
-- | -- | -- | --
20 | 0.80789 | 0.80695 | -0.12%
30 | 0.8413 | 0.84034 | -0.11%
40 | 0.8589 | 0.85823 | -0.08%
50 | 0.86988 | 0.86898 | -0.10%
100 | 0.89062 | 0.8906 | 0.00%
200 | 0.89907 | 0.89987 | 0.09%
1000 | 0.90343 | 0.90405 | 0.07%
2000 | 0.90362 | 0.90421 | 0.07%
5000 | 0.90364 | 0.90423 | 0.07%
10000 | 0.90365 | 0.90424 | 0.07%

`n_list=50K, pq_dim=96, pq_bits=8, n_iter=25`
n_probe | Original | This PR | Diff
-- | -- | -- | --
20 | 0.84634 | 0.84262 | -0.44%
30 | 0.88776 | 0.88373 | -0.45%
40 | 0.91073 | 0.9063 | -0.49%
50 | 0.92531 | 0.92116 | -0.45%
100 | 0.95417 | 0.95117 | -0.31%
200 | 0.96776 | 0.96488 | -0.30%
1000 | 0.97425 | 0.97188 | -0.24%
2000 | 0.97453 | 0.97217 | -0.24%
5000 | 0.97458 | 0.9722 | -0.24%
10000 | 0.97456 | 0.9722 | -0.24%

`n_list=50K, pq_dim=64, pq_bits=5, n_iter=25`
n_probe | Original | This PR | Diff
-- | -- | -- | --
20 | 0.67624 | 0.67812 | 0.28%
30 | 0.69861 | 0.69986 | 0.18%
40 | 0.71009 | 0.71077 | 0.10%
50 | 0.71702 | 0.71796 | 0.13%
100 | 0.73109 | 0.73108 | 0.00%
200 | 0.73687 | 0.73699 | 0.02%
1000 | 0.73996 | 0.7395 | -0.06%
2000 | 0.74009 | 0.73958 | -0.07%
5000 | 0.74009 | 0.73962 | -0.06%
10000 | 0.74009 | 0.73962 | -0.06%


`n_list=50K, pq_dim=64, pq_bits=8, n_iter=25`
n_probe | Original | This PR | Diff
-- | -- | -- | --
20 | 0.79924 | 0.79591 | -0.42%
30 | 0.83194 | 0.82838 | -0.43%
40 | 0.84907 | 0.84623 | -0.33%
50 | 0.85932 | 0.8576 | -0.20%
100 | 0.88041 | 0.87919 | -0.14%
200 | 0.88923 | 0.8885 | -0.08%
1000 | 0.89364 | 0.89306 | -0.06%
2000 | 0.89383 | 0.89326 | -0.06%
5000 | 0.89388 | 0.89327 | -0.07%
10000 | 0.89388 | 0.89327 | -0.07%

The recall difference should be within the tolerance range based on the results previously collected (https://github.com/rapidsai/raft/pull/2052#issuecomment-1890191014). But also note that for `n_list=50K, pq_dim=96, pq_bits=8, n_iter=25`. User probably need to increase the number of max point to achieve better recall, although the difference is small.

python/pylibraft/pylibraft/neighbors/ivf_pq/ivf_pq.pyx
```diff
@@ -156,6 +156,14 @@ cdef class IndexParams:
         repeated calls to `extend` (extending the database).
         To disable this behavior and use as little GPU memory for the
         database as possible, set this flat to `True`.
+    max_train_points_per_pq_code : int, default = 256
+        The max number of data points to use per PQ code during PQ codebook
+        training. Using more data points per PQ code may increase the
+        quality of PQ codebook but may also increase the build time. The
+        parameter is applied to both PQ codebook generation methods, i.e.,
+        PER_SUBSPACE and PER_CLUSTER. For PER_CLUSTER,
+        max_train_points_per_pq_code per pq_dim numbers of points would be
```
tfeher, line 165: Remove or reformulate the last sentence (same as for C++ docstring).

tfeher (changes requested):
Thanks Rui for the update. The results look great. I have seen similar recall variations, and I think that looks good as well. Just a few small things.

cpp/bench/ann/src/raft/raft_ann_bench_param_parser.h
```diff
@@ -87,6 +87,9 @@ void parse_build_param(const nlohmann::json& conf,
                                "', should be either 'cluster' or 'subspace'");
     }
   }
+  if (conf.contains("max_train_points_per_pq_code")) {
+    param.max_train_points_per_pq_code = 1.0 / (double)conf.at("max_train_points_per_pq_code");
```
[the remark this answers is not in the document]
abc99lr: We don't need divide. This is from the old fraction-based subsampling. Fixed

python/pylibraft/pylibraft/neighbors/ivf_pq/ivf_pq.pyx
abc99lr: Fixed

tfeher (approved):
Thanks Rui for the update! The PR looks good to me!

cjnolet (approved):

abc99lr:
Hi @achirkin , I think the change you requested has been fixed. Could you please approve this PR?

achirkin:
Sorry for being late, but yes, LGTM! :)

PR #2057, opened 2023-12-12
Add the deterministic version of AIR Top-K ( the radix-based topk in …

ChristinaZ:
Hi,
1. This PR is about the radix-based top-k algorithm in RAFT, we call it AIR Top-K in our just published paper (third_party/raft/cpp/include/raft/matrix/detail/select_radix.cuh)

2. We have recieved several feedbacks about the deterministic of AIR Top-K. 
In detail, AIR Top-K will return the smallest or largest K elements.    
One thing to notice is that there might be more than one "Kth smallest/largest element" for the given dataset. For example, assuming K=100, the value of the Kth element is 58, while there are three element's value are 58 and there already 99 element whose value are larger than 58.  In this case, we might not output all the equaling element as we ensure the output number is K. In this example, for all three elements, we only choose one element to store it in the results.

Previously, we choose the element euqaling to the Kth value randomly. In the deterministic version, we always ensure that the ones with smaller indices will be the output. 

3. In this PR, we only added the code in kernel and add a template parameter stable_last_filter with default value false. 
 It means our previous code don't need to change anything. 

4. I think it's better to open this PR first and discuss about current implement. Then we can discuss about adding one more API to expose this API to customer.


cjnolet:
@ChristinaZ it would be great to have this feature in RAFT. I'm going to push the release to 24.08 since we're approaching code freeze for 24.06. 24.08 is in August, do you think we might be able to get this merged by then?

ChristinaZ:
> @ChristinaZ it would be great to have this feature in RAFT. I'm going to push the release to 24.08 since we're approaching code freeze for 24.06. 24.08 is in August, do you think we might be able to get this merged by then?

Got it. I will work on the integration as soon as possible.

Issue #2058, opened 2023-12-13
[FEA] Improve refine_host performance when batch size is small


anaruse:
The current implementation of `refine_host` supports thread parallelism in OpenMP, but the degree of parallelism is the batch size, i.e., the number of queries. Therefore, when the batch size is 1, the parallelism is 1, and only one CPU core can be used. Especially when the K value of Top-K is large and the amount of distance calculation is large, the current parallelization method may slow down the performance with `refine_host`.

By changing the parallelization method, it is possible to use multiple threads even when the batch size is 1. I plan to submit a PR for this.

PR #2059, opened 2023-12-13
Improve parallelism of refine host

anaruse:
This PR addresses https://github.com/rapidsai/raft/issues/2058 by changing the thread parallelism method.

In the first half of the `refine` process, the distance calculation is performed on all candidate vectors, i.e., the number of queries * the original top-k vectors. Since the distance calculations for each vector can be performed independently, this part is thread-parallelized assuming that maximum parallelism is the number of queries * original top-k. This means that even if the number of queries is 1, this part can be executed in thread parallel.

On the other hand, the second half of the `refine` process, the so-called top-k calculation, can be performed independently for each query, but it is difficult to thread parallelize the calculation for a given query, Therefore, this part is parallelized assuming the maximum parallelism is the number of queries, as in the current implementation.


cpp/include/raft/neighbors/detail/refine_host-inl.hpp
```diff
@@ -43,35 +43,43 @@ template <typename DC, typename IdxT, typename DataT, typename DistanceT, typena
...
 
-#pragma omp parallel num_threads(suggested_n_threads)
-  {
-    std::vector<std::tuple<DistanceT, IdxT>> refined_pairs(orig_k);
-    for (size_t i = omp_get_thread_num(); i < n_queries; i += omp_get_num_threads()) {
-      // Compute the refined distance using original dataset vectors
+  // For efficiency, each thread should read a certain amount of array elements.
+  // The number of threads for distance computation is determined taking this into account.
+  constexpr int n_elements = 512;
+  size_t max_n_threads = ((n_queries * orig_k * dim) + n_elements - 1) / n_elements;
```
achirkin, line 50: Nitpick (requires `#include <raft/util/integer_utils.hpp>`)
```suggestion
  auto max_n_threads = raft::div_rounding_up_safe<size_t>(n_queries * orig_k * dim, n_elements);
```

cpp/include/raft/neighbors/detail/refine_host-inl.hpp
```diff
@@ -43,35 +43,43 @@ template <typename DC, typename IdxT, typename DataT, typename DistanceT, typena
...
+  auto suggested_n_threads_for_distance = std::min(size_t(suggested_n_threads), max_n_threads);
+
+  // The max number of threads for topk computation is the number of queries.
+  auto suggested_n_threads_for_topk = std::min(size_t(suggested_n_threads), n_queries);
+
+  std::vector<std::vector<std::tuple<DistanceT, IdxT>>>
+      refined_pairs(n_queries, std::vector<std::tuple<DistanceT, IdxT>>(orig_k));
```
achirkin, line 57: With the PR we allocate a whole n_queries*orig_k array here rather than keeping one independent orig_k array per thread. Have you checked if this has any adverse effect on performance with large batches (e.g. n_queries 10K)?

anaruse: Yes, the larger the batch size, the larger the size of `refined_pairs`, so performance may be degraded if the batch size is very large. I'll check and add a workaround (blocking) if there is a problem.

anaruse: As per your concern, I found that using this implementation when the batch size is large results in lower performance, so I have changed it to use this finer-grained thread parallelism only when the number of queries is small.

cpp/include/raft/neighbors/detail/refine_host-inl.hpp
```diff
@@ -43,6 +43,58 @@ template <typename DC, typename IdxT, typename DataT, typename DistanceT, typena
...
+        indices(i, j) = std::get<1>(refined_pairs[i][j]);
+        if (distances.data_handle() != nullptr) {
+          distances(i, j) = DC::template postprocess(std::get<0>(refined_pairs[i][j]));
+        }
+      }
+    }
+    return;
+  }
+
   if (size_t(suggested_n_threads) > n_queries) { suggested_n_threads = n_queries; }
```
achirkin, line 99: Nitpick: the check is not redundant thank to the new code
```suggestion
```

achirkin (approved):
Thanks, apart from a small nitpick, LGTM.

cjnolet (approved):

cjnolet:
@anaruse it looks like there's a minor compile error in this PR, but otherwise the changes are ready to be merged.

PR #2061, opened 2023-12-13
[BUG] fix empty initialization of device_ndarray in pylibraft

mfoerste4:
The `device_ndarray.empty()` function can be used to allocate device memory without initialization. Previously, the memory has been allocated (uninitialized) on the host and then been copied to the device.

This PR fixes the behavior for 'empty()' by allowing the `device_ndarray` to be initialized by an `array_interface` instead of an `numpy.ndarray` instance, which conditionally allows to skip the initialization of the `DeviceBuffer`.

CC @tfeher, @cjnolet

### DIFF
--- a/python/pylibraft/pylibraft/common/device_ndarray.py
+++ b/python/pylibraft/pylibraft/common/device_ndarray.py
@@ -31,7 +31,9 @@ class device_ndarray:
 
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


cjnolet (approved):
LGTM. Thanks @mfoerste4

cjnolet (approved):
LGTM. Thanks @mfoerste4

PR #2063, opened 2023-12-13
Benchmark brute force knn

benfred:
Add our bfknn code to the raft-ann-bench project


cjnolet (changes requested):
Looks good, I think it just needs a static assertion so we can throw an error if an unsupported data type is used.

benfred:
>Looks good, I think it just needs a static assertion so we can throw an error if an unsupported data type is used.

@cjnolet I added the static assert here https://github.com/rapidsai/raft/pull/2063/commits/f71f6578dd39a1e9885177bd186832aa8f54cdf7

cjnolet (approved):

PR #2080, opened 2024-01-04
Remove hardcoded limit in `print_results` function

narangvivek10:
The `print_results` function here is currently hardcoded to print only 2 results irrespective of the number of queries. A better way here could be to replace the hardcoded limit and allow printing results for the actual number of queries.


cjnolet (approved):
LGTM. Thanks for the contribution @narangvivek10!

Issue #2072, opened 2023-12-26
[QST] CudaErrorInvalidConfiguration Error when query number is out of 6e4?


FakeEnd:
**What is your question?**
When I use cagre including 100000 number of queries and index to do computation, something wrong with cudaErrorInvalidConfiguration happens.

I try to adjust the parameters of SearchParams, but it still fails.
```
import cupy as cp
import torch

from pylibraft.common import DeviceResources
from pylibraft.neighbors import cagra

batch_size = 1024
n_samples = 200
n_features = 64
# n_queries = 1000
k = 3
dataset = torch.randn(batch_size, n_samples, n_features, dtype=torch.float32).cuda()

handle = DeviceResources()
build_params = cagra.IndexParams(metric="sqeuclidean")

dataset_2d = dataset.view(batch_size * n_samples, n_features).clone()
seperate_constant = torch.arange(batch_size).view(-1, 1).repeat(1, n_samples).view(-1, 1).cuda() * 100
dataset_2d += seperate_constant

index = cagra.build(build_params, dataset_2d, handle=handle)
distances, neighbors = cagra.search(cagra.SearchParams(),
                                     index, dataset_2d,
                                     k, handle=handle)
```

Error:
```
  File "/archive/bioinformatics/Zhou_lab/shared/jjin/project/ANN/ANN.py", line 60, in forward
    distances, neighbors = cagra.search(cagra.SearchParams(),
  File "handle.pyx", line 225, in pylibraft.common.handle.auto_sync_handle.wrapper
  File "/archive/bioinformatics/Zhou_lab/shared/jjin/software/anaconda3/envs/ANN/lib/python3.9/site-packages/pylibraft/common/outputs.py", line 83, in wrapper
    ret_value = f(*args, **kwargs)
  File "cagra.pyx", line 746, in pylibraft.neighbors.cagra.cagra.search
  File "cagra.pyx", line 747, in pylibraft.neighbors.cagra.cagra.search
RuntimeError: CUDA error encountered at: file=/__w/raft/raft/cpp/include/raft/neighbors/detail/cagra/search_single_cta_kernel-inl.cuh line=953: call='cudaPeekAtLastError()', Reason=cudaErrorInvalidConfiguration:invalid configuration argument
Obtained 64 stack frames
[...]

PR #2081, opened 2024-01-08
Fix `max_queries` for CAGRA

lowener:
Fix for #2072: CAGRA search is launching a thread per query in single-CTA. The maximum number of thread is 65535 so the `max_queries` auto selection should be bounded to this number.

### DIFF
--- a/cpp/include/raft/neighbors/detail/cagra/cagra_search.cuh
+++ b/cpp/include/raft/neighbors/detail/cagra/cagra_search.cuh
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023, NVIDIA CORPORATION.
+ * Copyright (c) 2023-2024, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


cpp/include/raft/neighbors/detail/cagra/cagra_search.cuh
```diff
@@ -118,7 +118,7 @@ void search_main(raft::resources const& res,
   RAFT_EXPECTS(queries.extent(1) == index.dim(), "Queries and index dim must match");
   const uint32_t topk = neighbors.extent(1);
 
-  if (params.max_queries == 0) { params.max_queries = queries.extent(0); }
+  if (params.max_queries == 0) { params.max_queries = std::min<size_t>(queries.extent(0), 65535); }
```
cjnolet, line 121: It would be more robust to compute the value, perhaps using std::limits for int 16, MAX_UNSIGNED_SHORT, or (1<<16)-1.

cjnolet (approved):
LGTM!

PR #2084, opened 2024-01-09
ANN bench fix latency measurement overhead

tfeher:
The CPU time stamp `start` is taken before the ANN algo is copied to all threads. This is fixed by initializing `start` a few lines later.


cjnolet (approved):

cpp/bench/ann/src/common/benchmark.hpp
```diff
@@ -287,15 +287,14 @@ void bench_search(::benchmark::State& state,
     std::make_shared<buf<std::size_t>>(current_algo_props->query_memory_type, k * query_set_size);
 
   cuda_timer gpu_timer;
-  auto start = std::chrono::high_resolution_clock::now();
   {
     nvtx_case nvtx{state.name()};
+    [[maybe_unused]] auto ntx_lap = nvtx.lap();
+    [[maybe_unused]] auto gpu_lap = gpu_timer.lap();
```
achirkin, line 293: Please don't move the `.lap()` functions out of the loop: as the name suggests, they should be invoked each iteration.
1. NVTX: this marks the benchmark iterations with the NVTX ranges in the nsys timeline; moving this out of the loop breaks the marks. If you really like to avoid this (minimal) overhead, just disable the NVTX reporting when calling the benchmark with nsys.
2. GPU: this serves not only as the timer but also as a synchronization point. Without that, the GPU work in the `gpu_timer.stream()` spills out of the iteration section. That is the lack of synchronization what may give you the observed speedup, not the GPU timing overheads. Removing the synchronization between iterations may in theory be fine if we really think we need it (which I'm not sure); however, I've been doing some testing - this may lead to occasional segfaults under load in throughput mode (with 16-32 threads or so). I haven't been able to identify the root of the problem so far.

cpp/bench/ann/src/common/benchmark.hpp
```diff
@@ -287,15 +287,14 @@ void bench_search(::benchmark::State& state,
...
+    auto start                    = std::chrono::high_resolution_clock::now();
```
achirkin, line 294: I guess you intended to move the initialization of the `start` past the copy of the `algo`?

achirkin (changes requested):
The move of `.lap()` functions breaks the functionality. I'd suggest to only leave change that moves `start` to initialize after the `algo` is copied.

tfeher:
@achirkin gbench `real_time` is not the value that we want to define as latency. See details here: https://github.com/rapidsai/raft/pull/1940#issuecomment-1786293407

cpp/bench/ann/src/common/benchmark.hpp
tfeher: Moving these outside the loop was meant as a quick check on how much overhead is added. Currently we measure `latency = GPU time + synch overhead`. Question is whether sync overhead shall be part of the equation or not. I would prefer to keep `gpu_timer.lap()` in the loop, and possibly make it optional.

achirkin:
Thanks, @tfeher , for clarification! Indeed, now I see that the total iterations counter sums the iterations from all threads, whereas the real_time counts the time only by the main thread; as a result, the `real_time` shows the proper average latency divided by the number of threads.

This means the `GPU` counter is probably correct and the `Latency` counter is probably only slightly off due to the wrapper copy overheads.

In that case, I modify my suggestion: to be on the safe side, let's replace the explicit `state.iterations()` division with the `kAvgIterations` flag; it divides the total time by the total number of iterations (in all threads), hence gives the proper average latency for a single thread:
```C++
...
    auto end      = std::chrono::high_resolution_clock::now();
    auto duration = std::chrono::duration_cast<std::chrono::duration<double>>(end - start).count();
    if (state.thread_index() == 0) { state.counters.insert({{"end_to_end", duration}}); }
    state.counters.insert({"Latency", {duration, benchmark::Counter::kAvgIterations}});
  }

  state.SetItemsProcessed(queries_processed);
  if (cudart.found()) {
    state.counters.insert({"GPU", {gpu_timer.total_time(), benchmark::Counter::kAvgIterations}});
  }
...
```

tfeher:
The PR is reduced to fixing the starting of the timer. Other overheads are not addressed. For reference, here is the other issue that was discussed in a previous version of the PR.

> The current latency measurement includes overhead in the order of 10 microseconds, which is visible when we compare the benchmark output columns "GPU" and "Latency". This could be significant for benchmarks with small batch size. 

>This overhead is mainly due to the `lap()` call within the benchmark loop.  It would be probably better to make it optional to measure GPU latency.

cjnolet (approved):

cpp/bench/ann/src/common/benchmark.hpp
tfeher: Removed this change.

cpp/bench/ann/src/common/benchmark.hpp
```diff
@@ -314,17 +314,16 @@ void bench_search(::benchmark::State& state,
 
       queries_processed += n_queries;
     }
+    auto end      = std::chrono::high_resolution_clock::now();
+    auto duration = std::chrono::duration_cast<std::chrono::duration<double>>(end - start).count();
+    if (state.thread_index() == 0) { state.counters.insert({{"end_to_end", duration}}); }
+    state.counters.insert({"Latency", {duration, benchmark::Counter::kAvgIterations}});
```
tfeher, line 320: I have also applied Artem's suggestion to store latency values with `benchmark::Counter::kAvgIterations` marker.

Earlier we manually divided by number of iterations and let gbench average over threads using `kAvgThreads`. Since iterations are counted as total iterations performed by all threads, using kAviIterations leads to the same results (without manual divisions by iterations).

cpp/bench/ann/src/common/benchmark.hpp
```diff
@@ -314,17 +314,16 @@ void bench_search(::benchmark::State& state,
...
   }
-  auto end      = std::chrono::high_resolution_clock::now();
-  auto duration = std::chrono::duration_cast<std::chrono::duration<double>>(end - start).count();
-  if (state.thread_index() == 0) { state.counters.insert({{"end_to_end", duration}}); }
-  state.counters.insert(
-    {"Latency", {duration / double(state.iterations()), benchmark::Counter::kAvgThreads}});
 
   state.SetItemsProcessed(queries_processed);
   if (cudart.found()) {
     double gpu_time_per_iteration = gpu_timer.total_time() / (double)state.iterations();
```
achirkin, line 325: Unused var
```suggestion
```

achirkin (approved):
LGTM!

PR #2085, opened 2024-01-10
Support for fp16 in CAGRA and IVF-PQ

achirkin:
Add fp16 (CUDA half) support to CAGRA and its dependencies.


achirkin:
NB: this does not add the fp16 capabilities to the `ann-bench` executable. I suggest we add that in a follow-on PR.

cpp/include/raft/util/device_loads_stores.cuh
```diff
@@ -148,6 +149,26 @@ DI void sts(int32_t* addr, const int32_t (&x)[4])
...
+{
+  auto s = __cvta_generic_to_shared(reinterpret_cast<uint16_t*>(addr));
+  asm volatile("st.shared.u16 [%0], {%1};" : : "l"(s), "h"(*reinterpret_cast<const uint16_t*>(&x)));
+}
+DI void sts(half* addr, const half (&x)[1])
+{
+  auto s = __cvta_generic_to_shared(reinterpret_cast<uint16_t*>(addr));
+  asm volatile("st.shared.u16 [%0], {%1};" : : "l"(s), "h"(*reinterpret_cast<const uint16_t*>(x)));
+}
+DI void sts(half* addr, const half (&x)[2])
```
tfeher, line 162: We do not need this with larger vector size?

tfeher (approved):
Thanks @achirkin for this PR!

We have here a large amount of boilerplate code. Fortunately the nontrivial changes are relatively small, and are confined to 
- mdspan_numpy_serializer.hpp
- device_load_stores.cuh
- test/neighbors/ann_cagra.cuh

The PR looks good to me!

cpp/test/neighbors/ann_cagra.cuh
```diff
@@ -110,42 +112,69 @@ testing::AssertionResult CheckOrder(raft::host_matrix_view<IdxT, int64_t> index_
...
-  const int32_t resolution =
-    1 << static_cast<unsigned>(std::floor((24 - std::log2(dim) - (diff_flag ? 1 : 0)) / 2));
-  raft::random::uniformInt(
-    handle, rng, reinterpret_cast<int32_t*>(ptr), size, -resolution, resolution - 1);
+  const int_type resolution = std::max<int_type>(
+    int_type{128},
+    int_type{1} << static_cast<unsigned>(
+      std::floor((mapper_type::kBitshiftBase - std::log2(dim) - (diff_flag ? 1 : 0)) / 2)));
+  raft::random::uniformInt<int_type>(
+    handle, rng, reinterpret_cast<int_type*>(ptr), size, -resolution, resolution - 1);
```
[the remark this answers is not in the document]
enp1s0: The transformation is not correct, so the sorting test will fail depending on the random numbers generated. However, this method of generating random numbers, which results in rounding errors, cannot be used for FP16 when the dimension of the vector is greater than 256, because there is no range of random numbers that allows us not to introduce rounding errors.

I have three possible ways to solve this problem.
1. Skip the sorting test for FP16 when the dimension is greater than 256.
2. Use std::max like the PR code. This does not guarantee that there will be no rounding error, so the test will fail depending on the random numbers generated.
3. Stop using the no rounding error dataset and compute the norm of a vector in the test code with the same order of addition as in the CAGRA implementation. I didn't use this method in the current test implementation because I thought the test and the CAGRA implementation should be independent.

cpp/test/neighbors/ann_cagra.cuh
```diff
@@ -110,42 +112,69 @@ testing::AssertionResult CheckOrder(raft::host_matrix_view<IdxT, int64_t> index_
   return testing::AssertionSuccess();
 }
 
+template <typename T>
+struct fpi_mapper {};
+
+template <>
+struct fpi_mapper<double> {
+  using type                         = int64_t;
+  static constexpr int kBitshiftBase = 50;
```
enp1s0, line 121: `kBitshiftBase` is the mantissa length, including the implicit 1 bit.

```suggestion
  static constexpr int kBitshiftBase = 53;
```

cpp/test/neighbors/ann_cagra.cuh
```diff
@@ -110,42 +112,69 @@ testing::AssertionResult CheckOrder(raft::host_matrix_view<IdxT, int64_t> index_
...
+};
+
+template <>
+struct fpi_mapper<float> {
+  using type                         = int32_t;
+  static constexpr int kBitshiftBase = 22;
```
enp1s0, line 127:
```suggestion
  static constexpr int kBitshiftBase = 24;
```

cpp/test/neighbors/ann_cagra.cuh
```diff
@@ -110,42 +112,69 @@ testing::AssertionResult CheckOrder(raft::host_matrix_view<IdxT, int64_t> index_
...
+};
+
+template <>
+struct fpi_mapper<half> {
+  using type                         = int16_t;
+  static constexpr int kBitshiftBase = 8;
```
enp1s0, line 133:
```suggestion
  static constexpr int kBitshiftBase = 11;
```

cpp/include/raft/util/device_loads_stores.cuh
```diff
@@ -148,6 +149,26 @@ DI void sts(int32_t* addr, const int32_t (&x)[4])
                : "l"(s4), "r"(x[0]), "r"(x[1]), "r"(x[2]), "r"(x[3]));
 }
 
+DI void sts(half* addr, const half& x)
```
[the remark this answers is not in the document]
mdoijade: The additions to `device_loads_stores.cuh` looks good to me, I agree it is good to have matching `sts` function call for `lds` for larger fp16 vector sizes.

cpp/include/raft/util/device_loads_stores.cuh
achirkin: I think we need all of these, up two 16-byte-long vectors. That's because of the automatic templated vectorization in the ivf-flat scan kernel.
If I remember correctly, every instance I added here was requested by the compiler complaining for the missing instances.

cpp/test/neighbors/ann_cagra.cuh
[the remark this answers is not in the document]
achirkin: Thanks for checking this! I think, the (1) is the most reasonable solution; I'll just add the GTEST_SKIP when the resolution is 1 bit or less.

cpp/include/raft/util/device_loads_stores.cuh
achirkin: Ah, I misunderstood the question on the first read.
Apparently, we don't need `sts` with `half` data type and vector sizes larger than two at the moment. Though since you both suggest we'd better have these for symmetry, I'll add them.

enp1s0 (approved):

PR #2090, opened 2024-01-11
Add `compile-library` by default on pylibraft build

lowener:
This will avoid confusion for users launching only `./build.sh pylibraft`.


build.sh
```diff
@@ -377,6 +377,10 @@ fi
 if hasArg --incl-cache-stats; then
     BUILD_REPORT_INCL_CACHE_STATS=ON
 fi
+if hasArg pylibraft; then
+    COMPILE_LIBRARY=ON
+    CMAKE_TARGET="${CMAKE_TARGET};raft_lib"
```
pentschev, line 382: Is it ok to append multiple times the same target? If one now specifies `pylibraft --compile-lib` then `raft_lib` will exist twice. Perhaps a safer thing to do is have the `hasArg pylibraft` in https://github.com/rapidsai/raft/blob/03d97fd50cd35969c32bf24b9025a8012ed8399e/build.sh#L308, e.g.:

```
if hasArg --compile-lib || hasArg pylibraft || (( ${NUMARGS} == 0 )); then
```

lowener: True, this proposed solution also avoid code duplication. Thanks!

pentschev (approved):
LGTM, thanks @lowener !

cjnolet (approved):

cjnolet:
Thanks for fixing this @lowener. It looks like CI has some complaints about `nvtx3` headers, but once those are fixed I think this is ready to go in.

cpp/CMakeLists.txt
```diff
@@ -251,7 +251,7 @@ endif()
 
 if(RAFT_NVTX)
   # This enables NVTX within the project with no option to disable it downstream.
-  target_link_libraries(raft INTERFACE CUDA::nvToolsExt)
+  target_link_libraries(raft INTERFACE CUDA::nvtx3)
```
lowener, line 254: This has been added to resolve the following deprecation warning:
```
The library that is being linked to, CUDA::nvToolsExt, is marked as being
deprecated by the owner.  The message provided by the developer is:

nvToolsExt has been superseded by nvtx3 since CUDA 10.0 and CMake 3.25.
Use CUDA::nvtx3 and include <nvtx3/nvToolsExt.h> instead.
```

cpp/template/cmake/thirdparty/get_raft.cmake
```diff
@@ -46,7 +46,7 @@ function(find_and_configure_raft)
             "BUILD_TESTS OFF"
             "BUILD_PRIMS_BENCH OFF"
             "BUILD_ANN_BENCH OFF"
-            "RAFT_NVTX   ${ENABLE_NVTX}"
+            "RAFT_NVTX ${PKG_ENABLE_NVTX}"
```
bdice, line 49: This doesn't seem to match the variable name `ENABLE_NVTX` on lines 20 and 62.

lowener: All the parameters of this function are used with `PKG_` before the name. Same goes with `PKG_COMPILE_LIBRARY` used one line below, it is declared without the `PKG_`, but used with `PKG_`

bdice: Hmm. Maybe that's what
```
cmake_parse_arguments(PKG "${options}" "${oneValueArgs}"
"${multiValueArgs}" ${ARGN} )
```
does.

Sorry, I missed that earlier. This might not work.

cpp/template/cmake/thirdparty/get_raft.cmake
```diff
@@ -46,7 +46,7 @@ function(find_and_configure_raft)
             "BUILD_TESTS OFF"
             "BUILD_PRIMS_BENCH OFF"
             "BUILD_ANN_BENCH OFF"
-            "RAFT_NVTX   ${ENABLE_NVTX}"
+            "RAFT_NVTX ${ENABLE_NVTX}"
```
robertmaynard, line 49:
```suggestion
            "RAFT_NVTX ${PKG_ENABLE_NVTX}"
```

robertmaynard (changes requested):

robertmaynard (approved):

raydouglass (approved):

PR #2092, opened 2024-01-11
add CUDA 12.2 support for conda packages and wheels

jameslamb:
## Description

* switches to CUDA 12.2.2 for building conda packages and wheels
* adds new tests running against CUDA 12.2.2

### Notes for Reviewers

This is part of ongoing work to build and test packages against CUDA 12.2.2 across all of RAPIDS.

For more details see:

* https://github.com/rapidsai/build-planning/issues/7
* https://github.com/rapidsai/shared-workflows/pull/166

Planning a second round of PRs to revert these references back to a proper `branch-24.{nn}` release branch of `shared-workflows` once https://github.com/rapidsai/shared-workflows/pull/166 is merged.

*(created with `rapids-reviser`)*

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -28,7 +28,7 @@ concurrency:
 jobs:
   cpp-build:
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@branch-24.02
+    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@test-cuda-12.2
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
@@ -37,7 +37,7 @@ jobs:
   python-build:
     needs: [cpp-build]
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-python-build.yaml@branch-24.02
+    uses: rapidsai/shared-workflows/.github/workflows/conda-python-build.yaml@test-cuda-12.2
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
@@ -46,7 +46,7 @@ jobs:
   upload-conda:
     needs: [cpp-build, python-build]
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-upload-packages.yaml@branch-24.02
+    uses: rapidsai/shared-workflows/.github/workflows/conda-upload-packages.yaml@test-cuda-12.2
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
@@ -57,7 +57,7 @@ jobs:
     if: github.ref_type == 'branch'
     needs: python-build
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@branch-24.02
+    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@test-cuda-12.2
     with:
       arch: "amd64"
       branch: ${{ inputs.branch }}
@@ -69,7 +69,7 @@ jobs:
       sha: ${{ inputs.sha }}
   wheel-build-pylibraft:
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@branch-24.02
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@test-cuda-12.2
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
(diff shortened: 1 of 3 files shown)


bdice:
https://github.com/rapidsai/raft/actions/runs/7494925981/job/20403970360?pr=2092#step:7:644
```
-- CPM: Using local package CCCL@2.2.0.0
```
This is causing issues due to buggy install rules which we've patched in rapids-cmake but aren't present in the CUDA 12.2 local copy that this build is finding.

We probably need to do something to prevent RAPIDS from using the CUDA 12.2 CCCL version 2.2.0, since it doesn't have the CMake patches we need. Always downloading is the only solution I can think of, unless there's a safe way for us to force patching a local package. I think this probably needs something like `CPM_DOWNLOAD_CCCL`? cc: @robertmaynard @vyasr

bdice:
While working on #2102, I noticed an issue. In the `libraft-headers-only` recipe, we currently rely on `cuda-cudart-dev` in `host` to add some desired run-exports for `cuda-cudart`. This is going to need some rethinking for CUDA 12.2, since the run-exports we've been using to our advantage in CUDA 12.0 are not going to be correct for CUDA Enhanced Compatibility for CUDA 12.2. You can see that here, where CI is pulling `cuda-version=12.2` for the 12.0 CI job. https://github.com/rapidsai/raft/actions/runs/7562776933/job/20636164208?pr=2092#step:7:345 We're not testing what we thought we were testing (it's really 12.2 libraries and not 12.0 libraries).

We will also need to ignore the strong run export of `cuda-version >={{ cuda_version }}` from `cuda-nvcc`, in my understanding. https://github.com/conda-forge/cuda-nvcc-feedstock/blob/717380326f9e9599c77fac378b19adf894380b3c/recipe/meta.yaml#L47

I'm not sure if we want to treat this as a CUDA packaging bug or as a bug in libraft's recipe (it applies to all of RAPIDS). We had discussed "assuming CUDA Enhanced Compatibility by default" and considered the possibility of loosening the CUDA compiler pin to `12.*` in conda-forge rather than migrating for each minor version. cc: @jakirkham

dependencies.yaml
```diff
@@ -482,19 +454,14 @@ dependencies:
...
-          - {matrix: {cuda: "11.4"}, packages: *run_raft_dask_packages_cu11}
-          - {matrix: {cuda: "11.2"}, packages: *run_raft_dask_packages_cu11}
+          - matrix: {cuda: "12.*"}
+            packages:
+              - &pylibraft_cu12 pylibraft-cu12==24.2.*
+              - &ucx_py_cu12 ucx-py-cu12==0.36.*
+          - matrix: {cuda: "11.*"}
+            packages:
+              - &pylibraft_cu11 pylibraft-cu11==24.2.*
+              - &ucx_py_cu11 ucx-py-cu11==0.36.*
```
trxcllnt, line 464: `ucx-py-cu11==0.37.*` changed to `ucx-py-cu11==0.36.*` here. Is this expected?

dependencies.yaml
```diff
@@ -482,19 +454,14 @@ dependencies:
...
-            packages: &run_raft_dask_packages_cu11
-              - &pylibraft_cu11 pylibraft-cu11==24.4.*
-              - &ucx_py_cu11 ucx-py-cu11==0.37.*
-          - {matrix: {cuda: "11.5"}, packages: *run_raft_dask_packages_cu11}
-          - {matrix: {cuda: "11.4"}, packages: *run_raft_dask_packages_cu11}
-          - {matrix: {cuda: "11.2"}, packages: *run_raft_dask_packages_cu11}
+          - matrix: {cuda: "12.*"}
+            packages:
+              - &pylibraft_cu12 pylibraft-cu12==24.2.*
+              - &ucx_py_cu12 ucx-py-cu12==0.36.*
```
trxcllnt, line 460: `ucx-py-cu12==0.37.*` changed to `ucx-py-cu12==0.36.*` here. Is this expected?

cpp/include/raft/sparse/linalg/spmm.hpp
```diff
@@ -69,6 +69,9 @@ void spmm(raft::resources const& handle,
 
   detail::spmm(handle, trans_x, trans_y, is_row_major, alpha, descr_x, descr_y, beta, descr_z);
 
+  raft::copy(
+    z.data_handle(), z_tmp.data_handle(), z_tmp.size(), raft::resource::get_cuda_stream(handle));
+
```
jakirkham: IIUC PR ( https://github.com/rapidsai/raft/pull/2124 ) ensured that this copy is no longer needed. So think we can drop this

```suggestion
```

jameslamb: ha already removed! you were too quick!

https://github.com/rapidsai/raft/pull/2092/commits/c28e84752d934da772c9a40e2017be752cdfb348

jameslamb: this got left behind because I just did this to get past the conflicts

```shell
git merge -X theirs branch-24.04
```

and this was the only line from that file that didn't have a conflict with this PR 🙈

jakirkham: No worries. Thanks James! 🙏 

Looks good! 😄

jakirkham:
Looks like this is now green! 🥳

jakirkham:
One of the jobs failed due to `git clone`. So needs a restart. All the other ones passed

bdice:
All CI is passing and CI logs look good. We need an ops review from @AyodeAwe / @raydouglass to proceed. I'll trigger the `/merge` for now.

bdice (approved):

AyodeAwe (approved):

PR #2093, opened 2024-01-11
[DOC] Documentation updates for release 24.02

cjnolet:


docs/source/vector_search_tutorial.md
```diff
@@ -216,6 +216,55 @@ While it may seem like unnecessary additional work to wrap the result in a `host
 
 The following sections present some advanced features that we have found can be useful for squeezing more utilization out of GPU hardware. As you've seen in this tutorial, RAFT provides several very useful tools and building blocks for developing accelerated applications beyond vector search capabilities.
 
+### Serialization
```
dantegd, line 219: The tutorial has become kinda large, I wonder if an index at the beginning of the sections would be helpful, what do you think?

dantegd (approved):
Changes look great, had a small suggestion

docs/source/vector_search_tutorial.md
cjnolet: Like a table of contents? sure we can do that

PR #2094, opened 2024-01-11
Properly taking ownership of nccl subcomm (and destroying it)

cjnolet:

### DIFF
--- a/cpp/include/raft/comms/detail/std_comms.hpp
+++ b/cpp/include/raft/comms/detail/std_comms.hpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020-2023, NVIDIA CORPORATION.
+ * Copyright (c) 2020-2024, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


cpp/include/raft/comms/detail/std_comms.hpp
```diff
@@ -116,6 +122,11 @@ class std_comms : public comms_iface {
   {
     requests_in_flight_.clear();
     free_requests_.clear();
+
+    if (own_nccl_comm_) {
+      ncclCommDestroy(nccl_comm_);
```
ChuckHastings, line 127: Should wrap in `RAFT_NCCL_TRY_NO_THROW`

ChuckHastings (commented):
Thanks!  Will test in my branch before approving.

cpp/include/raft/comms/detail/std_comms.hpp
cjnolet: Ah! Good catch.

ChuckHastings:
This clears up the customer problem I was chasing, Thanks!

ChuckHastings (approved):

PR #2095, opened 2024-01-12
Remove usages of rapids-env-update

KyleFromNVIDIA:
Reference: https://github.com/rapidsai/ops/issues/2766

Replace rapids-env-update with rapids-configure-conda-channels,
rapids-configure-sccache, and rapids-date-string.


ajschmidt8 (approved):

ajschmidt8:
looks like there's an unrelated failing test on this PR. might be a flake. we can rerun it after the workflow completes to check

PR #2097, opened 2024-01-16
Allow topk larger than 1024 in CAGRA

benfred:
This change allows CAGRA search to have an arbitrarily large top-k, instead of being limited to 1024 like in the previous code.

This works by using the multi-kernel search path, and replacing the _cuann_find_topk code with the matrix::select_k code - which can handle large K values.


cpp/include/raft/neighbors/detail/cagra/search_multi_kernel.cuh
```diff
@@ -597,6 +598,95 @@ void set_value_batch(T* const dev_ptr,
...
+                            workspace,
+                            sort,
+                            hints,
+                            stream);
+  }
+
+  rmm::device_uvector<float> input_keys_storage(0, stream);
+  rmm::device_uvector<float> output_keys_storage(0, stream);
+  rmm::device_uvector<ValT> input_values_storage(0, stream);
+  rmm::device_uvector<ValT> output_values_storage(0, stream);
```
tfeher, line 650: To avoid reconstructing these in every function call, I would recommend to make ` _find_topk` a member function of the `search` class, then we can keep these as member variables.

tfeher (changes requested):
Thanks @benfred for the PR! Overall it looks good: while the extra copies are not ideal (and we should add it to a tracker to improve that), it alleviates the max-k problem without affecting the perf where k< 1024. Just one comment below.

cpp/include/raft/neighbors/detail/cagra/search_multi_kernel.cuh
benfred: thanks for the suggestion! change is in latest commit

tfeher (approved):
Thanks @benfred for the update, LGTM!

PR #2099, opened 2024-01-17
Replace call to mr::get_mem_info()

harrism:
In the near future memory resources will no longer have `get_mem_info()`.

There is one place in RAFT that uses this. This PR replaces it with `rmm::available_device_memory()` which just calls `cudaMemGetInfo()`.


harrism:
Accidentally pushed some changes that need to happen after an RMM PR merges. Reverted.

achirkin (approved):
Thanks for taking care of this! LGTM

cjnolet (approved):

PR #2102, opened 2024-01-19
Remove extraneous host pinnings from libraft-headers-only.

bdice:
We have `run` pinnings for librmm, spdlog, and fmt, so we don't need to list them in `host` as well. Currently, we are allowing `fmt>=10.1.1,<11` in `host`, which is pulling a newer version 10.2.1 and adding a run-export that is `fmt >=10.2.1,<11.0a0`, which disallows `fmt 10.1.1` contrary to our intentions.

### DIFF
--- a/conda/recipes/libraft/meta.yaml
+++ b/conda/recipes/libraft/meta.yaml
@@ -1,4 +1,4 @@
-# Copyright (c) 2022-2023, NVIDIA CORPORATION.
+# Copyright (c) 2022-2024, NVIDIA CORPORATION.
 
 # Usage:
 #   conda build . -c conda-forge -c nvidia -c rapidsai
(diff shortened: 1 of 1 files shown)


bdice:
I confirmed that the host environment no longer includes `fmt`. [Current build log](https://github.com/rapidsai/raft/actions/runs/7588245013/job/20670394378?pr=2102#step:7:482), [previous recent build log](https://github.com/rapidsai/raft/actions/runs/7585289780/job/20660814318#step:7:482).

I'm still waiting for CI to complete to know about the run-exports being fixed for sure, but this should be a sufficient fix from what I can observe so far.

jakirkham (approved):

raydouglass (approved):

bdice:
Double-confirmation: I downloaded the artifacts and I see what I expected: depends on `fmt >=10.1.1,<11`. This is good to merge.

jakirkham:
Thanks all! 🙏

PR #2125, opened 2024-01-24
Add fused cosine 1-NN cutlass based kernel

mdoijade:
- Adds cosine 1-NN cutlass based kernel for SM 8.0 or higher using tensor cores.
- based on 3x TF32
- unifies the fusedDistanceNN kernels for L2/cosine.
- expose this API in pylibraft as `fused_distance_nn_arg_min` supporting cosine & L2 distance metrics.


cpp/include/raft/distance/detail/fused_distance_nn/cutlass_base.cuh
```diff
@@ -46,6 +47,14 @@ namespace raft {
 namespace distance {
 namespace detail {
 
+template <typename IdxT>
+__global__ void initBinMutexKernel(cuda::binary_semaphore<cuda::thread_scope_device>* mut, IdxT m)
```
bdice, line 51:
```suggestion
RAFT_KERNEL initBinMutexKernel(cuda::binary_semaphore<cuda::thread_scope_device>* mut, IdxT m)
```

cpp/include/raft/distance/detail/fused_distance_nn/simt_kernel.cuh
```diff
@@ -0,0 +1,186 @@
...
+
+template <typename DataT,
+          typename OutT,
+          typename IdxT,
+          typename P,
+          typename ReduceOpT,
+          typename KVPReduceOpT,
+          typename OpT,
+          typename FinalLambda>
+__global__ __launch_bounds__(P::Nthreads, 2) void fusedDistanceNNkernel(OutT* min,
```
bdice, line 69:
```suggestion
__launch_bounds__(P::Nthreads, 2) RAFT_KERNEL fusedDistanceNNkernel(OutT* min,
```

cpp/include/raft/distance/detail/fused_distance_nn/helper_structs.cuh
```diff
@@ -0,0 +1,145 @@
...
+  DI void operator()(LabelT rid, DataT* out, const KVP& other)
+  {
+    if (other.value < *out) { *out = other.value; }
+  }
+
+  DI void init(DataT* out, DataT maxVal) { *out = maxVal; }
+};
+
+template <typename DataT, typename OutT, typename IdxT, typename ReduceOpT>
+__global__ void initKernel(OutT* min, IdxT m, DataT maxVal, ReduceOpT redOp)
```
bdice, line 110:
```suggestion
RAFT_KERNEL initKernel(OutT* min, IdxT m, DataT maxVal, ReduceOpT redOp)
```

bdice (commented):
Kernels must have internal linkage. See #1898, #1935.

cpp/include/raft_runtime/distance/fused_distance_nn.hpp
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright (c) 2022-2024, NVIDIA CORPORATION.
```
tfeher, line 2:
```suggestion
 * Copyright (c) 2024, NVIDIA CORPORATION.
```

cpp/src/distance/fused_distance_nn.cu
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright (c) 2021-2024, NVIDIA CORPORATION.
```
tfeher, line 2:
```suggestion
 * Copyright (c) 2024, NVIDIA CORPORATION.
```

cpp/src/distance/fused_distance_nn.cu
```diff
@@ -0,0 +1,60 @@
...
+    void* workspace,                                                          \
+    bool sqrt,                                                                \
+    bool initOutBuffer,                                                       \
+    bool isRowMajor,                                                          \
+    raft::distance::DistanceType metric,                                      \
+    float metric_arg,                                                         \
+    cudaStream_t stream)
+
+instantiate_raft_distance_fusedDistanceNNMinReduce(double, double, int);
+instantiate_raft_distance_fusedDistanceNNMinReduce(double, double, int64_t);
```
tfeher, line 40: I am not sure that we need the instantiations for `double` type. The ANN methods are only instantiated in `float`, therefore I would recommend to only add `float` to `libraft`. 

Since we test in `double`, it does make sense to instantiate `double` for the tests. That can be done without including these in `libraft`. A similar thing is done for CAGRA: libraft only instantiate `uint32_t` index type, and we have extra `uint64_t` [instantiations](https://github.com/rapidsai/raft/blob/branch-24.04/cpp/test/neighbors/ann_cagra/search_kernel_uint64_t.cuh) for [the test](https://github.com/rapidsai/raft/blob/branch-24.04/cpp/test/neighbors/ann_cagra/test_float_int64_t.cu).

cpp/include/raft/distance/fused_distance_nn_helpers.cuh
```diff
@@ -17,7 +17,8 @@
 #pragma once
 
 #include <raft/core/resource/cuda_stream.hpp>
-#include <raft/distance/detail/fused_l2_nn.cuh>
+// #include <raft/distance/detail/fused_l2_nn.cuh>
```
tfeher, line 20: remove

cpp/include/raft/distance/fused_distance_nn.cuh
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright (c) 2021-2024, NVIDIA CORPORATION.
```
tfeher, line 2:
```suggestion
 * Copyright (c) 2024, NVIDIA CORPORATION.
```

cpp/include/raft/distance/detail/fused_distance_nn/fused_cosine_nn.cuh
```diff
@@ -0,0 +1,135 @@
+/*
+ * Copyright (c) 2021-2024, NVIDIA CORPORATION.
```
tfeher, line 2:
```suggestion
 * Copyright (c) 2024, NVIDIA CORPORATION.
```

cpp/include/raft/distance/detail/fused_distance_nn.cuh
```diff
@@ -0,0 +1,89 @@
...
+      <<<nblks, P::Nthreads, 0, stream>>>(min, m, maxVal, redOp);
+    RAFT_CUDA_TRY(cudaGetLastError());
+  }
+
+  switch (metric) {
+    case DistanceType::CosineExpanded:
+      fusedCosineNN<DataT, OutT, IdxT, P, ReduceOpT, KVPReduceOpT>(
+        min, x, y, xn, yn, m, n, k, workspace, redOp, pairRedOp, sqrt, stream);
+      break;
+    default: assert("only cosine metric is supported with fusedDistanceNN\n"); break;
```
tfeher, line 83: Why don't we dispatch L2 distance?

tfeher (changes requested):
Thanks @mdoijade for this PR, it is great to have the fused cosine distance kernels to further accelerate k-means clustering!

It would be nice if we could further improve code reuse between `L2` and `Cosine` variants of fusedNN (decrease duplication of boiler plate code), but I see that we have some discrepancy in the required parameters, and this could negatively affect readibility.

Otherwise the PR looks good in general. Here are my comments

cpp/include/raft/distance/detail/fused_distance_nn.cuh
mdoijade: only reason is to avoid multiple entry points for L2 distance and not break the existing fusedL2NN API for downstream RAFT users.
But I think it is better to add it here and later add deprecate the fusedL2NN API always allowing it to be called from fusedDistanceNN instead.

mdoijade: @cjnolet currently fused_l2_nn_min_arg supports fp32 & fp64 in `libraft/pylibraft`. 
This PR adds a new cosine fused nn kernel exposed via new API `fused distance nn`  which we plan expose only as fp32 in `libraft/pylibraft`, we want to deprecate/drop `fused_l2_min_arg` in future release in favor of `fused_distance_nn_min_arg` as it provides common API for any distance metric fused with 1-NN feature. For that to happen we shall have to remove its fp64 support, will it be fine to do that?
Or do you think `fused distance nn`  should also expose fp64 support in libraft/pylibraft.

Note that `fused distance nn` does support fp64 we test that it in raft c++ testsuite only we don't expose in libraft/pylibraft. And if dropping fp64 support from `fused_l2_min_arg` is fine that can allow to further unify its code with fused distance nn code base.

tfeher (approved):
Thanks @mdoijade for addressing the issues! The code looks good to me.

Please still have a look at the copyright start year in some of the new files.

mdoijade:
the pylibraft test passes but an unrelated cpp test is failing `LINALG_TEST` certainly seems not due to this PR.

cpp/include/raft/distance/detail/fused_distance_nn.cuh
mdoijade: @cjnolet  *ping* for response on comment above.

tfeher: I believe we can accept the current version of the PR. It is a non-braking change, the question is what data type to support in the newly added `fused_distance_nn` functions. "Header only" we support and test both fp32 and fp64. In the compiled lib (and therefore in the Python API) we only support fp32. If needed, we can still add fp64 instantiations later to the library.

python/pylibraft/pylibraft/distance/fused_distance_nn.pyx
```diff
@@ -0,0 +1,221 @@
...
+    "jaccard": DistanceType.JaccardExpanded,
+    "hellinger": DistanceType.HellingerExpanded,
+    "braycurtis": DistanceType.BrayCurtis,
+    "jensenshannon": DistanceType.JensenShannon,
+    "hamming": DistanceType.HammingUnexpanded,
+    "kl_divergence": DistanceType.KLDivergence,
+    "minkowski": DistanceType.LpUnexpanded,
+    "russellrao": DistanceType.RusselRaoExpanded,
+    "dice": DistanceType.DiceExpanded,
+}
```
benfred, line 78: We should use the `DISTANCE_TYPES` dict that is already defined

```suggestion
from pylibraft.distance.pairwise_distance import DISTANCE_TYPES
```

benfred (commented):
looks good! one minor comment:

benfred (approved):

PR #2126, opened 2024-01-24
Enable CAGRA index building without adding dataset to the index

tfeher:
We can build a CAGRA graph even for datasets that do not fit GPU mem. The IVF-PQ build method only requires that the temporary IVF-PQ index (that we use for creating the knn-graph) fits the GPU. But once the CAGRA graph is constructed, we try to initialize the CAGRA index which would copy the dataset to device memory. 

This PR adds a build flag, that would disable copying the dataset into the index. That enables building CAGRA graph for large datasets, and also allows users to customize allocator used for storing the data.

### DIFF
--- a/cpp/include/raft/neighbors/detail/cagra/cagra_build.cuh
+++ b/cpp/include/raft/neighbors/detail/cagra/cagra_build.cuh
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023, NVIDIA CORPORATION.
+ * Copyright (c) 2023-2024, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


cpp/include/raft/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -296,7 +296,8 @@ index<T, IdxT> build(
   std::optional<experimental::nn_descent::index_params> nn_descent_params = std::nullopt,
   std::optional<float> refine_rate                                        = std::nullopt,
   std::optional<ivf_pq::index_params> pq_build_params                     = std::nullopt,
-  std::optional<ivf_pq::search_params> search_params                      = std::nullopt)
+  std::optional<ivf_pq::search_params> search_params                      = std::nullopt,
+  bool construct_index_with_dataset                                       = true)
```
tfeher, line 300: Currently only `cagra::detail::build` is changed. This is enough to unblock benchmarking with RAFT-ANN-Bench, but ideally we should have such an option in the public API. Opinions?

cjnolet (approved):

cjnolet (approved):

PR #2145, opened 2024-02-01
[FIX] Ensure hnswlib can be found from RAFT's build dir

trxcllnt:
This PR fixes using RAFT from a build dir, i.e. `cmake -S cuml/cpp -B cuml/cpp/build -Draft_ROOT=raft/cpp/build`.

Without this fix, CMake errors when cuML does `find_package(RAFT)` with the following error:
```
CMake Error at raft/cpp/build/latest/raft-targets.cmake:56 (set_target_properties):
  The link interface of target "raft::raft" contains:

    hnswlib::hnswlib

    * An ALIAS target is missing.
```


cjnolet (approved):

bdice (approved):

cpp/cmake/thirdparty/get_hnswlib.cmake
```diff
@@ -64,6 +64,10 @@ function(find_and_configure_hnswlib)
         GLOBAL_TARGETS hnswlib
         NAMESPACE hnswlib::)
     endif()
+    include("${rapids-cmake-dir}/export/find_package_root.cmake")
+    rapids_export_find_package_root(
+      BUILD hnswlib [=[${CMAKE_CURRENT_LIST_DIR}]=] EXPORT_SET raft-exports
+    )
```
vyasr, line 70: Alternatively, we could include the build `rapids_export` outside of this conditional and then export the root unconditionally, perhaps that's more correct. WDYT?
```suggestion
      include("${rapids-cmake-dir}/export/find_package_root.cmake")
      rapids_export_find_package_root(
        BUILD hnswlib [=[${CMAKE_CURRENT_LIST_DIR}]=] EXPORT_SET raft-exports
      )
    endif()
```

vyasr (changes requested):

cpp/cmake/thirdparty/get_hnswlib.cmake
trxcllnt: @vyasr yes, I think you're right. The `rapids_export(BUILD hnswlib)` should be moved outside the `if(NOT PKG_EXCLUDE_FROM_ALL)` block.

vyasr (approved):

Issue #2141, opened 2024-01-31
[BUG] Large memory overhead while subsampling vectors for IVF methods


tfeher:
**Describe the bug**

The IVF methods use random subsampling to create a training set for k-means clustering.

The random sampling algorithm allocates several temporary buffers: 
https://github.com/rapidsai/raft/blob/d4ae27157720e5ea2d83c7474d2926710a08a0bc/cpp/include/raft/random/detail/rng_impl.cuh#L293-L296

On top of this `sortPairs `would also allocate temporary buffers, with size `O(len)`

Here `len` is the size (number of vectors) in the whole dataset. When we are indexing DEEP-1B, then `len=1e9`, and the temporary space becomes prohibitive.

**Steps/Code to reproduce bug**
Run DEEP-1B test with IVF-PQ, with subsample ratio 10. It will run OOM. 

**Expected behavior**
Subsample vectors with minimal memory overhead.

tfeher:
A quick fix for this problem is to revert the following PRs:
- https://github.com/rapidsai/raft/pull/2077 this introduced the memory overhead
- https://github.com/rapidsai/raft/pull/2052 relies on random sampling for correctness, therefore needs to be reverted if #2077 is reverted.

I am investigating whether we have an alternative quick fix to reduce the memory overhead for the subsampling.

PR #2155, opened 2024-02-05
IVF random sampling with improved memory utilization.

tfeher:
The random sampling of IVF methods was reverted (#2144) due to large memory utilization #2141.

This PR improves the memory consumption of subsamling: it is O(n_train) where n_train is the size of the subsampled dataset. This way we can guarantee that enough space is available, since for subsampling we have O(n_train*dims) space available.

Random subsampling is restored for the training set.

### DIFF
--- a/cpp/bench/prims/random/subsample.cu
+++ b/cpp/bench/prims/random/subsample.cu
@@ -1,0 +1,197 @@
+/*
+ * Copyright (c) 2022-2023, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <common/benchmark.hpp>
+#include <cub/cub.cuh>
+#include <raft/core/device_mdarray.hpp>
+#include <raft/core/device_resources.hpp>
+#include <raft/core/host_mdarray.hpp>
+#include <raft/core/operators.hpp>
+#include <raft/random/permute.cuh>
+#include <raft/random/rng.cuh>
+#include <raft/random/sample_without_replacement.cuh>
+#include <raft/spatial/knn/detail/ann_utils.cuh>
+#include <raft/util/cudart_utils.hpp>
+#include <rmm/device_scalar.hpp>
+
+namespace raft::bench::random {
+
+struct sample_inputs {
+  int n_samples;
+  int n_train;
+  int method;
+};  // struct sample_inputs
+
+template <typename IdxT>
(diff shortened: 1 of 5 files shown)


cpp/include/raft/matrix/detail/gather.cuh
```diff
@@ -336,6 +344,70 @@ void gather_if(const InputIteratorT in,
...
+
+  const size_t max_batch_size = 32768;
+  // Gather the vector on the host in tmp buffers. We use two buffers to overlap H2D sync
+  // and gathering the data.
+  raft::common::nvtx::push_range("gather::alloc_buffers");
+  auto out_tmp1 = raft::make_pinned_matrix<T, IdxT>(res, max_batch_size, n_dim);
+  auto out_tmp2 = raft::make_pinned_matrix<T, IdxT>(res, max_batch_size, n_dim);
+  auto view1    = out_tmp1.view();
+  auto view2    = out_tmp2.view();
+  raft::common::nvtx::pop_range();
```
achirkin, line 387: Do I understand it correctly that you use manual pop/push instead of the `nvtx::range` to keep the out_tmp1, out_tmp2 in scope outside the code block?
I see this is less verbose than having to type the matrix type in a separate declaration. Still, I'm not sure if this is worth losing the safety in case if `raft::make_pinned_matrix` throws an exception.

achirkin (changes requested):
Thanks, Tamas, for the PR! Looks good overall and I like the idea of the excess sampling. Just having a few minor questions.

cpp/test/random/excess_sampling.cu
```diff
@@ -0,0 +1,116 @@
...
+    }
+  }
+
+ protected:
+  inputs params;
+  raft::resources res;
+  cudaStream_t stream;
+  RngState state;
+  device_vector<T, int64_t> out, in;
+  host_vector<T, int64_t> h_out;
```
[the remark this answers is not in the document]
tfeher: You are right, it is simpler when they are local. Changed it.

tfeher (commented):
Thanks @achirkin for the review, I have addressed the issues.

achirkin (approved):
Thanks for updates, Tamas, LGTM!

cpp/include/raft/matrix/detail/gather.cuh
```diff
@@ -336,6 +346,83 @@ void gather_if(const InputIteratorT in,
...
+  // Gather the vector on the host in tmp buffers. We use two buffers to overlap H2D sync
+  // and gathering the data.
+  auto out_tmp1 = raft::make_pinned_matrix<T, MatIdxT>(res, max_batch_size, n_dim);
+  auto out_tmp2 = raft::make_pinned_matrix<T, MatIdxT>(res, max_batch_size, n_dim);
+
+  // Usually a limited number of threads provide sufficient bandwidth for gathering data.
+  int n_threads = std::min(omp_get_max_threads(), 32);
+
+  // The gather_buff function has a parallel for loop. We start the the omp parallel
+  // region here, to avoid repeated overhead within the device_offset loop.
```
achirkin, line 398: Does that mean the omp for loop in gather_buff automatically collapses/reuses the threads spawned here rather than creating `n_threads*n_threads` threads in total? Cool I didn't know that

tfeher: Yes, the `omp parallel` creates the threads, the `omp for` just reuses the existing threads.

benfred (approved):
lgtm!

PR #2165, opened 2024-02-07
Add environment-agnostic scripts for running ctests and pytests

trxcllnt:
This PR adds environment-agnostic `run_*_{ctests,pytests}.sh` scripts, and updates `test_*_{cpp,python}.sh` to call them.

The `test_*_{cpp,python}.sh` scripts assume they're running in our CI environment, and they do more than just run the tests.

This PR allows devs and downstream consumers to only run the tests, and skip the unrelated logic in `test_*_{cpp,python}.sh`.


AyodeAwe (approved):

cjnolet:
It looks like the c++ tests are running so yeah that's fine

PR #2167, opened 2024-02-07
Fix CAGRA filter gtests

benfred:
Commit https://github.com/rapidsai/raft/pull/2097/commits/fea490a4cbe674443e81a3d6df3a9a8b3e9ec31a attempted to fix a problem where the search_params.itopk_size wasn't being set in the CAGRA filter tests.

This change was required to enable testing k>1024, since the default itopk_size was too small. However, this broke the unittest for k < 1024 - and was causing illegal memory access errors.

Fix by reverting the filter tests to the previous behaviour, and disabling the filter tests for k>1024


cpp/test/neighbors/ann_cagra.cuh
```diff
@@ -497,9 +497,13 @@ class AnnCagraFilterTest : public ::testing::TestWithParam<AnnCagraInputs> {
         search_params.algo         = ps.algo;
         search_params.max_queries  = ps.max_queries;
         search_params.team_size    = ps.team_size;
-        search_params.itopk_size   = ps.itopk_size;
         search_params.hashmap_mode = cagra::hash_mode::HASH;
 
+        // TODO: setting search_params.itopk_size here breaks the filter tests, but is required for
+        // k>1024 skip these tests until fixed
+        if (ps.k >= 1024) return;
```
achirkin, line 504: Make gtest mark the current case as skipped rather than silently accepting it.
```suggestion
        if (ps.k >= 1024) { GTEST_SKIP(); }
```

cpp/test/neighbors/ann_cagra.cuh
```diff
@@ -613,9 +617,13 @@ class AnnCagraFilterTest : public ::testing::TestWithParam<AnnCagraInputs> {
...
```
achirkin, line 624: Make gtest mark the current case as skipped rather than silently accepting it.
```suggestion
        if (ps.k >= 1024) { GTEST_SKIP(); }
```

cpp/test/neighbors/ann_cagra.cuh
```diff
@@ -497,9 +497,13 @@ class AnnCagraFilterTest : public ::testing::TestWithParam<AnnCagraInputs> {
         search_params.algo         = ps.algo;
         search_params.max_queries  = ps.max_queries;
         search_params.team_size    = ps.team_size;
-        search_params.itopk_size   = ps.itopk_size;
```
achirkin, line 500: You've reverted the change here and at line 616, but not at line 262. Is that intended?

benfred: yeah, this was intentional.

When I was testing locally, I was only seeing a problem with the Cagra filter tests (line 616/line 600) - but not with the main unfiltered AnnCagraTest on line 262.  I wanted to maintain the test for k>1024 in the AnnCagraTest code, which requires setting itopk_size .

Btw, this change fixed the test failures locally for me (went from failing to all passing on my local machine), but I'm still seeing failures in CI - it looks like it passed on the a100/arm64 instance but failed on the amd64 machines =(.

achirkin: Yes, that's why I've asked. It looks like the remaining failing tests are those with `itopk_size` still set.

benfred:
Closing in favour of https://github.com/rapidsai/raft/pull/2181

Issue #1555, opened 2023-05-30
[BUG] IVF-Flat TopK operation is limited to `k=256`


lowener:
https://github.com/rapidsai/raft/blob/99655eadf34e28d923d690ff4560d5c48c9b92f9/cpp/include/raft/matrix/detail/select_warpsort.cuh#L100
https://github.com/rapidsai/raft/blob/99655eadf34e28d923d690ff4560d5c48c9b92f9/cpp/include/raft/neighbors/detail/ivf_flat_interleaved_scan-inl.cuh#L1003-L1008
Providing a fallback when `k>256` would be useful.

PR #2169, opened 2024-02-09
IVF-FLAT support k > 256

mfoerste4:
Add support for topk > 256 for ivf_flat (Issue [#1555](https://github.com/rapidsai/raft/issues/1555))

The PR adds a non-fused version of topk that is utilized if k > 256.

FYI, @tfeher


cpp/include/raft/neighbors/detail/ivf_flat_build.cuh
```diff
@@ -42,6 +42,64 @@ namespace raft::neighbors::ivf_flat::detail {
 
 using namespace raft::spatial::knn::detail;  // NOLINT
 
+/**
+ * Update the state of the dependent index members.
+ */
+template <typename T, typename IdxT>
+void recompute_internal_state(const raft::resources& res, index<T, IdxT>& index)
```
tfeher, line 49: Is there anything different here compared to what we need to do for IVF-PQ? Can't we add this to `ivf_common.cuh`?

cpp/include/raft/neighbors/detail/ivf_flat_search-inl.cuh
```diff
@@ -42,25 +43,35 @@ void search_impl(raft::resources const& handle,
...
-  rmm::device_uvector<float> coarse_distances_dev(n_queries * n_probes, stream, search_mr);
+  rmm::device_uvector<float> coarse_distances_dev(n_queries_probes, stream, search_mr);
   // The topk  index of cluster(list) and queries
-  rmm::device_uvector<uint32_t> coarse_indices_dev(n_queries * n_probes, stream, search_mr);
+  rmm::device_uvector<uint32_t> coarse_indices_dev(n_queries_probes, stream, search_mr);
+
+  // Optional structures if postprocessing is required
   // The topk distance value of candidate vectors from each cluster(list)
-  rmm::device_uvector<AccT> refined_distances_dev(n_queries * n_probes * k, stream, search_mr);
+  rmm::device_uvector<AccT> refined_distances_dev(0, stream, search_mr);
```
tfeher, line 68: Here these arrays just store temporary output values after IVF-Flat search. Since writing the IVF-Flat code, we have added a [refinement](https://github.com/rapidsai/raft/blob/c95bf6aecb48f8eb751e4b706b727e48eaf9757f/cpp/include/raft/neighbors/refine-inl.cuh#L70) operation which is applicable after vector search with lossy compression (like IVF-PQ), and it is unrelated to what we do here. That makes the variable names `refined_` somewhat confusing. May I ask you to rename these arrays to `distances_tmp_dev` or something similar without the `refined` term?

tfeher (changes requested):
Thanks Malte for the PR! It looks great overall, here are my comments

tfeher (approved):
Thanks Malte for the updates, it is great to see the increasing code reuse between IVF methods! The PR looks good to me.

As discussed offline, the perf of the non-fused top-k search is currently less then ideal due to large number of dummy values (as a result from cluster size imbalance). We discussed how to improve that in a follow up PR for radix top-k.

cjnolet (approved):

PR #2172, opened 2024-02-11
Reapply: Support for fp16 in CAGRA and IVF-PQ

achirkin:
1. Add fp16 (CUDA half) support to CAGRA and its dependencies (#2085).
2. Fix the shared memory size error in the ivf-flat that got exposed by new tests in #2085.

Regarding the point (2):
Warp-sort top-k queue uses shared memory; the module provides the required shmem size calculation function decoupled from the queue object itself. As a result, it's easy to plug-in wrong types and get the calculation incorrectly.
IVF-Flat scan kernel always kept the distances in the queue as floats, but we calculated the shmem size as if it used `AccT` (IVF-Flat's internal accumulation type). Hence, with adding the tests with fp16 inputs (and `AccT`), the allocated shmem became too small, which resulted in memory access violation errors.


benfred (approved):
lgtm! 

I didn't review the whole change - but did a diff between this branch and the original PR https://github.com/rapidsai/raft/pull/2085/files and found the only difference was the fix 

```patch
diff --git a/cpp/include/raft/neighbors/detail/ivf_flat_interleaved_scan-inl.cuh b/cpp/include/raft/neighbors/detail/ivf_flat_interleaved_scan-inl.cuh
index 51cd2876d..1cf042c6c 100644
--- a/cpp/include/raft/neighbors/detail/ivf_flat_interleaved_scan-inl.cuh
+++ b/cpp/include/raft/neighbors/detail/ivf_flat_interleaved_scan-inl.cuh
@@ -844,7 +844,7 @@ void launch_kernel(Lambda lambda,
   int smem_size              = query_smem_elems * sizeof(T);
   constexpr int kSubwarpSize = std::min<int>(Capacity, WarpSize);
   auto block_merge_mem =
-    raft::matrix::detail::select::warpsort::calc_smem_size_for_block_wide<float, IdxT>(
+    raft::matrix::detail::select::warpsort::calc_smem_size_for_block_wide<AccT, IdxT>(
       kThreadsPerBlock / kSubwarpSize, k);
   smem_size += std::max<int>(smem_size, block_merge_mem);
 ```

which should resolve the unittest failures that we were seeing before.

Thanks for fixing this @achirkin !

PR #2176, opened 2024-02-12
Remove libraft.a from base package

### LINKED ISSUE #2175: [BUG] conda libraft package includes static library
**Describe the bug**

**Expected behavior**

msarahan:
fix #2175


bdice:
@msarahan We probably need to re-evalute this recipe as a whole. Not only are we shipping too many outputs, we're also (re-)building too many objects, if I recall correctly from looking at past build outputs. I am wondering if we can modify the scripts to only build libraft code _once_ (with a top-level build script in the recipe) and then use CMake install scripts for each of the outputs (`libraft-headers-only` ships the headers, `libraft-headers` is a metapackage that adds CUDA dependencies, `libraft` ships the `.so`, `libraft-static` ships the `.a`, `libraft-tests` ships the test binaries). Compare to cudf: build everything ([script](https://github.com/rapidsai/cudf/blob/branch-24.04/conda/recipes/libcudf/build.sh), the recipe implicitly calls `build.sh` from the recipe directory), install libcudf ([script](https://github.com/rapidsai/cudf/blob/branch-24.04/conda/recipes/libcudf/install_libcudf.sh), [recipe line calling it](https://github.com/rapidsai/cudf/blob/49c2995b1b861b12d3b25ad997adec9c50ed872f/conda/recipes/libcudf/meta.yaml#L80)).

msarahan:
Makes sense. I'll see what I can do.

cjnolet:
@msarahan have you been able to figure this one out? This is definitely a problem for our users but to my knowledge we haven't found a solution for it yet on our end.

vyasr:
We've discussed moving to a single top-level libraft build in the past like for the rest of RAPIDS but have been unable to do so for \~reasons\~ that I confess I have no recollection of any longer. We are technically rebuilding now, but at minimum we've aimed to make it so that sccache has a 100% hit rate on all subsequent builds within the same run to at least minimize CI time. I'll try and find the issues that we had with a single build in the past, maybe they are no longer an issue. I'd be happy to see someone try to unify the build again.

bdice:
I raised an alternative proposal in #2209.

PR #2185, opened 2024-02-14
Fixing workaround for cuSPARSE bug with correct copy dimensions

mfoerste4:
This PR tries to fix the workaround that introduce a temporary copy of the Z matrix for SpMM. 
Note that the correct sizing for the copy is crucial - this is covered by the SPARSE_DIST_TEST.

However, I did NOT test whether this fixes the cuSparse bug. CC @cjnolet @trxcllnt


divyegala (approved):

jakirkham:
Thanks all! 🙏

Issue #1889, opened 2023-10-11
[FEA] CAGRA-Q 


cjnolet:
There has been research and exploration around some extensions to CAGRA that use product quantization to effectively compress the input dataset, thus shrinking the CAGRA footprint and providing support for larger data sizes. 

This is a placeholder issue to be taken up once CAGRA-Q research has gotten to a point where we feel it's ready to be ported to RAFT. Currently, this is showing great performance on GH using huge page pinned memory for the underlying optimized graph, allowing for the search times to be comparable to the graph stored fully in device memory.

PR #2206, opened 2024-02-29
CAGRA-Q search

enp1s0:
Rel: #1889

## Limitations
- Only 8-bit PQ is supported
- Sub-space size is only 2 supported


achirkin (commented):
Thanks, @enp1s0, for working on this!
I love your modular approach of using dataset descriptor to abstract away the distance computation code. I think we'll have to think a bit later whether it's possible to unify/backport this to IVF methods.

I'm trying now integrate the index building (compression) part, and I want to figure out how much of the functionality to implement at first.
As far as I see now, both the ported code and the original prototype support only `pq_bits = 8`, is that correct? There's probably some limitation on possible `pq_dim` values as well.
Could you please write down these and possibly other limitations of the prototype/initial port (search part only) in the description of the PR?

cpp/include/raft/neighbors/detail/cagra/search_plan.cuh
```diff
@@ -69,7 +69,11 @@ struct search_plan_impl_base : public search_params {
   }
 };
 
-template <class DATA_T, class INDEX_T, class DISTANCE_T, class SAMPLE_FILTER_T>
+template <class DATA_T,
+          class INDEX_T,
+          class DISTANCE_T,
+          class DATASET_DESCRIPTOR_T,
```
achirkin, line 75: Here and in other places parameterized by the descriptor type: would it make sense to remove INDEX_T, DISTANCE_T template parameters and get them from the descriptor (i.e. `DATASET_DESCRIPTOR_T::INDEX_T`)?

enp1s0: I have removed `INDEX_T`, `DATA_T`, and `DISTANCE_T` from the template params and use `DATASET_DESCRIPTOR_T::` instead.

cpp/include/raft/neighbors/cagra.cuh
```diff
@@ -323,16 +323,18 @@ void search(raft::resources const& res,
...
-                             internal_IdxT,
-                             decltype(raft::neighbors::filtering::none_cagra_sample_filter()),
-                             IdxT>(res,
-                                   params,
-                                   idx,
-                                   queries_internal,
-                                   neighbors_internal,
-                                   distances_internal,
-                                   raft::neighbors::filtering::none_cagra_sample_filter());
+  if (1) {  // standard CAGRA
```
achirkin, line 326: I've prepared a stub for the runtime dispatch of search variants, please have a look:
https://github.com/rapidsai/raft/pull/2213/files#diff-11ca764d4ee96d5f9fdb500a088130a8772ea8e135274d2493a1912e7d71380bR352-R375

cpp/include/raft/neighbors/detail/cagra/cagra_search.cuh
```diff
@@ -243,6 +294,9 @@ void search_main(raft::resources const& res,
     RAFT_FAIL("Unrecognized dataset format");
   }
 
+  // Refinement
+  if (use_vpq) {}
+
```
achirkin: Refinement in raft is a separate operation, no need to include it into cagra-q search or parameters.

cpp/include/raft/neighbors/detail/cagra/cagra_search.cuh
```diff
@@ -149,6 +149,71 @@ void search_main_core(
...
+                                                          0>;
+      dataset_desc_t dataset_desc(dataset->data.data_handle(),
+                                  dataset->pq_dim(),
+                                  dataset->vq_n_centers(),
+                                  dataset->vq_code_book.data_handle(),
+                                  0,
+                                  dataset->pq_code_book.data_handle(),
+                                  0,
+                                  dataset->n_rows(),
+                                  dataset->dim());
```
achirkin, line 206: It's probably lost in merging, please check if the mapping from vpq_dataset to dataset_desc_t is correct. I believe it should be:
```
dataset_desc_t dataset_desc(vpq_dset->data.data_handle(),	
                                vpq_dset->encoded_row_length(),	
                                vpq_dset->pq_dim(),	
                                vpq_dset->vq_code_book.data_handle(),	
                                1.0f,	
                                vpq_dset->pq_code_book.data_handle(),	
                                1.0f,	
                                size_t(vpq_dset->n_rows()),	
                                vpq_dset->dim());
```

enp1s0: Is it correct that the pq and vq scales are fixed as 1.f because only f16 is currently supported as a dataset type?

achirkin: Yes, as far as I understood it from the reference implementation, the scaling factors other than one are used only for (u)int8 types, which was also just an experiment there.

cpp/include/raft/neighbors/detail/cagra/compute_distance.hpp
```diff
@@ -36,152 +36,54 @@ _RAFT_DEVICE constexpr unsigned get_vlen()
...
-        }
-      }
-    }
-    for (uint32_t offset = TEAM_SIZE / 2; offset > 0; offset >>= 1) {
-      norm2 += __shfl_xor_sync(0xffffffff, norm2, offset);
-    }
-    return norm2;
-  }
+template <>
+struct code_book_load_t_core<1> {
```
tfeher, line 44: Could we use `TxN_t` https://github.com/rapidsai/raft/blob/3e67a6c3bc5c73cd4a77a71bb5c2d911f8a7bcd4/cpp/include/raft/util/vectorized.cuh#L238-L249

tfeher (changes requested):
Thanks @enp1s0 for the PR. I am still looking into the details, until than here are my first comments.

cpp/include/raft/neighbors/detail/cagra/search_multi_kernel.cuh
```diff
@@ -993,5 +983,71 @@ struct search : search_plan_impl<DATA_T, INDEX_T, DISTANCE_T, SAMPLE_FILTER_T> {
...
+                  INDEX_T* const topk_indices_ptr,       // [num_queries, topk]
+                  DISTANCE_T* const topk_distances_ptr,  // [num_queries, topk]
+                  const DATA_T* const queries_ptr,       // [num_queries, dataset_dim]
+                  const uint32_t num_queries,
+                  const INDEX_T* dev_seed_ptr,              // [num_queries, num_seeds]
+                  uint32_t* const num_executed_iterations,  // [num_queries,]
+                  uint32_t topk,
+                  SAMPLE_FILTER_T sample_filter)
+  {
+  }
```
tfeher, line 1049: This is the placeholder for the CAGRA-Q specialization of the multi_kernel search kernels, right? Since the quantized version is not yet implemented, we could probably throw a "not implemented" error, and this feature can be added in next release.

tfeher (changes requested):
Thanks @enp1s0 for the PR! It is nice how the dataset descriptor decides whether we have a compressed or a regular distance calculation function. Overall the PR looks good. The main potential issues that I currently see:
- clarify what is the upper limit of dim for CAGRA-Q
- binary size increase. I will quantify this below.

achirkin:
As per our discussion, I `--force-push`ed all commits signed to enable the CI.

cpp/include/raft/neighbors/detail/cagra/search_multi_kernel.cuh
enp1s0: Added error message output in search class constructor

cpp/include/raft/neighbors/detail/cagra/compute_distance.hpp
enp1s0: I changed to `TxN_t`

enp1s0:
@tfeher @achirkin Thank you for the review and help. There is no remaining task. Can you review the code again?

tfeher:
Many thanks for the great work @enp1s0 and @achirkin! Overall the PR looks good. With this PR, the binary size of libraft.so grows to 655 MB. Here is information on the size of the newly added components:

component | binary size (MB)
-- | --
cagra | 52.0
cagra_q_multi_cta | 13.5
cagra_q_search_single_cta_float_uint32 | 23.3
cagra_q_search_single_cta_float_uint64 | 21.7
cagra_q_search_single_cta_half_uint32 | 23.3
cagra_q_search_single_cta_half_uint64 | 21.7
cagra_q_search_single_cta_int8_uint32 | 23.3
cagra_q_search_single_cta_uint8_uint32 | 23.3
**total** | **202.3**

In our offline discussion @enp1s0 motivated adding 64bit index type because CAGRA-Q is expected to handle large graphs. But if we need to reduce size, then we could keep uint32 idx type in the initial release. @cjnolet please let us know what would you prefer.

cpp/include/raft/neighbors/detail/cagra/cagra_search.cuh
```diff
@@ -173,6 +147,145 @@ void search_main(raft::resources const& res,
...
   }
+}
+
+template <class T,
+          class DatasetT,
+          class DatasetIdxT,
+          class InternalIdxT,
+          class DistanceT,
+          class CagraSampleFilterT>
+void lauch_vpq_search_main_core(
```
tfeher, line 158: The max dataset dim for cagra-q search shall be documented. Also might make sense to refuse building for index combinations that will not work.

cpp/test/neighbors/ann_cagra_vpq.cuh
```diff
@@ -0,0 +1,337 @@
...
+  rmm::device_uvector<DataT> database;
+  rmm::device_uvector<DataT> search_queries;
+};
+
+const std::vector<AnnCagraVpqInputs> vpq_inputs = raft::util::itertools::product<AnnCagraVpqInputs>(
+  {100},                                              // n_queries
+  {1000, 10000},                                      // n_rows
+  {128, 256},                                         // dim
+  {8, 12},                                            // k
+  {2, 4},                                             // pq_len
```
tfeher, line 323: We should extend test coverage with different values of n_queries, n_rows, dim and k. Note that it should not be a tensor product. We can vary individual parameters one at a time

tfeher: At would also make sense to add refine_ratio as a test option here.

tfeher (approved):
Thanks @enp1s0 and @achirkin for your work. The PR looks good to me. There are a few additional issues that we can handle in a follow up PR.

Additionally, the test seem to compile too long, I will check whether there is any unintended template instantiation:

file | compile time | binary size
-- | -- | --
ann_cagra/test_half_uint32_t.cu.o | 31:37 min | 91.827 MB
ann_cagra/test_int8_t_uint32_t.cu.o | 31:11 min | 91.846 MB
ann_cagra/test_float_uint32_t.cu.o | 30:41 min | 91.709 MB
ann_cagra/test_uint8_t_uint32_t.cu.o | 30:22 min | 91.874 MB
ann_cagra_vpq/test_float_int64_t.cu.o | 17:40 min | 46.242 MB
ann_cagra/test_half_int64_t.cu.o | 17:27 min | 46.422 MB
ann_cagra/test_float_int64_t.cu.o | 15:31 min | 37.865 MB
bench/cagra_float_uint32_t.cu.o | 14:09 min | 47.566 MB
test_filter_float_int64_t.cu.o | 10:49 min | 19.376 MB
ann_cagra_vpq/test_float_uint32_t.cu.o | 7:52 min | 11.921 MB

achirkin (dismissed):

cpp/include/raft/neighbors/detail/cagra/cagra_search.cuh
```diff
@@ -173,6 +147,145 @@ void search_main(raft::resources const& res,
...
+  if (vpq_dset->pq_bits() == 8) {
+    if (vpq_dset->pq_len() == 2) {
+      using dataset_desc_t = cagra_q_dataset_descriptor_t<T,
+                                                          DatasetT,
+                                                          8 /*PQ bit*/,
+                                                          2 /* Subspace dimension*/,
+                                                          0,
+                                                          DistanceT,
+                                                          InternalIdxT,
+                                                          0>;
```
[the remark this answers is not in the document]
enp1s0: Yes, right. They are replaced with actual values just before the search kernel is launched. This is because we can't determine `TEAM_SIZE` and `BLOCK_DIM` when making the instance of dataset_desc. But I'm not sure this code design is good.
https://github.com/enp1s0/raft/blob/38ab2bd787699992a23296532b770859b1bf7512/cpp/include/raft/neighbors/detail/cagra/search_single_cta_kernel-inl.cuh#L913-L914

cpp/include/raft/neighbors/detail/cagra/cagra_search.cuh
enp1s0: I updated the VPQ similarity calculation to fix a bug. This fix also allows us to support dim >= MAX_BLOCK_DIM.

cpp/include/raft/neighbors/detail/cagra/cagra_search.cuh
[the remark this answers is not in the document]
enp1s0: An alternative design is to remove the `TEAM_SIZE' and `BLOCK_DIM' template parameters from the dataset_desc and add them to the `compute_similarity' function instead.

cpp/include/raft/neighbors/detail/cagra/compute_distance_vpq.cuh
```diff
@@ -88,6 +88,26 @@ struct cagra_q_dataset_descriptor_t : public dataset_descriptor_base_t<half, DIS
   {
   }
 
+  template <uint32_t DATASET_BLOCK_DIM>
+  __device__ void copy_query(const DATA_T* const dmem_query_ptr,
+                             QUERY_T* const smem_query_ptr,
+                             const std::uint32_t query_smem_buffer_length)
+  {
+    for (unsigned i = threadIdx.x * 2; i < dim; i += blockDim.x * 2) {
+      half2 buf2 = {CUDART_ZERO_FP16, CUDART_ZERO_FP16};
```
achirkin, line 97: I believe `CUDART_ZERO_FP16` is not available prior to cuda 12.2 and this will break CI.

cpp/include/raft/neighbors/detail/cagra/compute_distance_vpq.cuh
```diff
@@ -88,6 +88,26 @@ struct cagra_q_dataset_descriptor_t : public dataset_descriptor_base_t<half, DIS
...
+      if (i < dim) { buf2.x = static_cast<half>(static_cast<float>(dmem_query_ptr[i])); }
```
achirkin, line 98: Probably, we should replace the casts with `spatial::knn::detail::utils::mapping<float>{}` as well, right? In case if queries are 8-bit ints, the mapping will do normalization.

PR #2207, opened 2024-03-01
Replace usages of raw `get_upstream` with `get_upstream_resource()`

miscco:
We want to get rid of raw memory resources so move to the new interface instead


miscco:
There is still one usage within `cpp/test/core/device_resources_manager.cpp` but I dont know how to properly transfer that to a resource_ref

miscco:
@cjnolet I believe this should be ready for review

miscco:
Can we get this in the release? @cjnolet 

We want to deprecate `get_upstream()`soon

cjnolet (approved):

PR #2208, opened 2024-03-04
add top-level conda build to avoid rebuild in subs

### LINKED ISSUE #2175: [BUG] conda libraft package includes static library

msarahan:
Adjusts the conda recipe to do the build of all components once, and then uses the subpackage scripts to bundle the produced binaries appropriately into $PREFIX.

This is a draft. I've put it up for discussion and awarenesss, but it's not really ready yet. In particular, the script_env stuff is broken. I don't really understand the scoping rules in YAML, but it seems that declaring the anchor section in the top-level build section does not make it available in the outputs.

fix #2175

### DIFF
--- a/conda/recipes/libraft/build_base.sh
+++ b/conda/recipes/libraft/build_base.sh
@@ -1,0 +1,4 @@
+#!/usr/bin/env bash
+# Copyright (c) 2022-2024, NVIDIA CORPORATION.
+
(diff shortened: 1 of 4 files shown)


bdice:
@msarahan I think this can be closed now that #2209 was merged. Feel free to reopen if needed.

PR #2209, opened 2024-03-04
Only install necessary components in conda packages.

### LINKED ISSUE #2175: [BUG] conda libraft package includes static library

bdice:
Closes #2175.

**This PR cuts the binary size of `libraft` by about 50%.** In total, this PR shrinks the binary size of the entire C++ conda channel (the sum of all conda packages produced by `conda/recipes/libraft/meta.yaml`) by around 20%, from 1.5 GB to 1.2 GB. The main difference is dropping the extraneous copy of `libraft.a` (~300 MB) that was shipped in the `libraft` package.


conda/recipes/libraft/build_libraft.sh
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
-# Copyright (c) 2022-2023, NVIDIA CORPORATION.
+# Copyright (c) 2022-2024, NVIDIA CORPORATION.
 
-./build.sh libraft --allgpuarch --compile-lib --build-metrics=compile_lib --incl-cache-stats --no-nvtx
+./build.sh libraft --allgpuarch --compile-lib --build-metrics=compile_lib --incl-cache-stats --no-nvtx -n
+cmake --install cpp/build --component compiled
```
bdice, line 5: This now looks a lot like https://github.com/rapidsai/raft/blob/branch-24.04/conda/recipes/libraft/build_libraft_static.sh except with a different component being installed.

msarahan:
compared with https://github.com/rapidsai/raft/pull/2208, this approach avoids the top-level build. I think this PR is probably a better approach. Certainly fewer changes, and it *should* have the same effect. I'd be happy to close #2208, but I'll let you guys have the say on whether you want to.

bdice:
This is failing for some reason due to `hnswlib`. I haven't had a chance to debug this yet. Just copying it here for awareness, for the moment.

```
CMake Error at /opt/conda/conda-bld/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/lib/cmake/raft/raft-targets.cmake:61 (set_target_properties):
  The link interface of target "raft::raft" contains:

    hnswlib::hnswlib

    * An ALIAS target is missing.

Call Stack (most recent call first):
  /opt/conda/conda-bld/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/lib/cmake/raft/raft-config.cmake:84 (include)
  build/cmake/CPM_0.38.5.cmake:243 (find_package)
  build/cmake/CPM_0.38.5.cmake:303 (cpm_find_package)
  build/_deps/rapids-cmake-src/rapids-cmake/cpm/find.cmake:189 (CPMFindPackage)
  cmake/thirdparty/get_raft.cmake:36 (rapids_cpm_find)
  cmake/thirdparty/get_raft.cmake:57 (find_and_configure_raft)
  CMakeLists.txt:34 (include)
```

robertmaynard:
> This is failing to build libraft-template for some reason due to hnswlib

The `libraft-template` builds build against an installed version of libraft. The installed version of libraft is now just installing the `libraft` target which means that the hnswlib headers / library aren't being installed into the `libraft` conda package. Which causes the consumers to fail.

The hnswlib install rules in `get_hnswlib.cmake` need to be expanded with an explicit `COMPONENT` and that component needs to get added to `cmake --install` invocation.

bdice:
@robertmaynard It looks like I got the same error again. https://github.com/rapidsai/raft/actions/runs/8159304257/job/22303364726?pr=2209#step:7:3758

Can you look at the implementation (c101777bbf47aaf1a1d815920c004cea1d2e1c95) and offer guidance?

cpp/cmake/thirdparty/get_hnswlib.cmake
```diff
@@ -47,7 +47,11 @@ function(find_and_configure_hnswlib)
      "$<INSTALL_INTERFACE:include>")
 
     if(NOT PKG_EXCLUDE_FROM_ALL)
-      install(TARGETS hnswlib EXPORT hnswlib-exports)
+      install(
+        TARGETS hnswlib
+        COMPONENT hnswlib
+        EXPORT hnswlib-exports
+      )
       install(DIRECTORY "${hnswlib_SOURCE_DIR}/hnswlib/" DESTINATION include/hnswlib)
```
robertmaynard, line 55: You need to install the hnswlib headers as part of the `hnswlib` component

robertmaynard (changes requested):

cpp/CMakeLists.txt
```diff
@@ -179,8 +179,8 @@ rapids_cpm_init()
 
 if(NOT BUILD_CPU_ONLY)
   # CCCL before rmm/cuco so we get the right version of CCCL
```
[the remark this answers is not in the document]
robertmaynard: I was thinking the issue was due to `rmm` encoding where to search for the RAPIDS version of CCCL, but that isn't the case. Reverting the change.

robertmaynard (approved):

bdice:
Thanks @robertmaynard for the help in debugging!

bdice:
It looks like `libraft-tests` is packaging `libraft.a` now. I suspect this is because it is no longer provided by the `libraft` conda package, but is still being built by the tests? Investigating now.

bdice:
@robertmaynard I couldn't figure out how to make the `build_libraft_tests.sh` _not_ build/install the static library, so instead I added `libraft-static` as a `host` dependency of `libraft-tests`. Hopefully this will keep it from repackaging the static library and other files. See baed39a23fa92910e8b3de7de6ab3b58ac6d5820.

If we can figure out how to make the tests avoid building the static library, that would be a better solution.

bdice:
As of baed39a, we are packaging the right things. It may be possible to cut down the build process further (by not building `libraft.a` for the `libraft-tests` package, or reusing other parts of the build for `libraft-tests`), but at least the resulting packages contain only the relevant pieces.

In total, this PR shrinks the binary size of the entire C++ conda channel by around 20%, from 1.5 GB to 1.2 GB. The main difference is dropping the extraneous copy of `libraft.a` that was in the `libraft` package.

cjnolet (approved):

raydouglass (approved):

PR #2211, opened 2024-03-05
Improve RBC eps-neighborhood query performance

mfoerste4:
This PR significantly improves performance of epsilon neighborhood search via RBC.

Scope: 
* the rbc index was modified to contain a reordered dataset allowing for better access patterns
* all kernels for dense, sparse and hybrid have been improved

Optimizations include:
- improve pruning of complete landmarks by pre-fetching distances and minimizing overhead for skipping 
- specialized 2D and 3D kernels, allowing for register per-fetch of query points
- improve inner loop iterating landmark neighborhood
    - pruning of points within selected landmark neighborhood using triangle inequality
    - reverse iterate landmark neighborhood to allow complete processing stop once subsequent points cannot be reached
    - removal of shared memory atomics in favor of warp voting
- general prevention of branches within a warp (at least to some extent)

CC @cjnolet , @tfeher


cpp/include/raft/spatial/knn/detail/ball_cover/registers-inl.cuh
```diff
@@ -476,70 +477,112 @@ RAFT_KERNEL block_rbc_kernel_eps_dense(const value_t* X_index,
...
+      const uint32_t R_size          = R_indptr[cur_k + 1] - R_start_offset;
+
+      // we have precomputed the query<->landmark distance
+      const value_t cur_R_dist = raft::shfl(lane_R_dist, k_offset);
+
+      const uint32_t limit = Pow2<WarpSize>::roundDown(R_size);
+      int i                = limit + lid;
+
+      // look ahead for next k_offset
+      k_offset = lane_mask != 0 ? __ffs(lane_mask) - 1 : WarpSize;
```
Nyrio, line 531: To save a few instructions, I would remove the ternary expression here:

```c++
k_offset = __ffs(lane_mask) - 1;
```

And change the loop condition to:

```c++
while (lane_mask);
```

But then to avoid the unnecessary invocation when the mask is 0, the `__ffs` would be moved to the top of the loop instead of computing ahead for the next iteration. It would look like:

```c++
int lane_mask = raft::ballot(lane_check);

while(lane_mask) {
    uint32_t k_offset = __ffs(lane_mask) - 1;
    ...
}
```

Also, as discussed offline, there is an instruction to find the most significant bit but not the least, so you can change the order to avoid the bit reversal:

```c++
uint32_t k_offset = 31 - __clz(lane_mask);
```

Nyrio: As a side note, if you need to keep the previous order from the least significant 1, you can use `__brev` before the loop and then `__clz` inside the loop. I think that the current version has to do the bit reverse every iteration because the mask is changed.

cpp/include/raft/spatial/knn/detail/ball_cover/registers-inl.cuh
```diff
@@ -476,70 +477,112 @@ RAFT_KERNEL block_rbc_kernel_eps_dense(const value_t* X_index,
...
+    uint32_t k_offset = __ffs(lane_mask) - 1;
+    do {
+      const uint32_t cur_k = cur_k0 + k_offset;
+
+      // update lane_mask for next iteration - erase bits up to k_offset
+      lane_mask &= -(1 << k_offset + 1);
+
+      // The whole warp should iterate through the elements in the current R
+      const value_idx R_start_offset = R_indptr[cur_k];
+      const uint32_t R_size          = R_indptr[cur_k + 1] - R_start_offset;
```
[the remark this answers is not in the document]
Nyrio: It turns out that the while loop executes very sparsely (on average 1.4 iterations) so the preload is not helping.

cpp/include/raft/spatial/knn/detail/ball_cover/registers-inl.cuh
```diff
@@ -476,70 +483,114 @@ RAFT_KERNEL block_rbc_kernel_eps_dense(const value_t* X_index,
...
+      // look for next k_offset
+      const uint32_t k_offset = __clz(lane_mask);
+
+      const uint32_t cur_k = cur_k0 + k_offset;
+
+      // The whole warp should iterate through the elements in the current R
+      const value_idx R_start_offset = R_indptr[cur_k];
+
+      // update lane_mask for next iteration - erase bits up to k_offset
+      lane_mask &= (1 << max_lid - k_offset) - 1;
```
Nyrio, line 531: Nitpick:

```c++
lane_mask &= (0x7fffffff >> k_offset);
```

(2 instructions instead of 4)

tfeher (approved):
Thanks Malte for the PR,  it is great to have a large speedup for the eps-neighborhood kernels. Overall the PR looks good to me.

Please add a few more details about the optimizations into the description, eg:
- pruning of points within selected landmark neighborhood using triangle inequality
- specialized 2D and 3D kernels

Many thanks to @Nyrio for helping with the optimizations and the review!

cpp/include/raft/spatial/knn/detail/ball_cover/registers-inl.cuh
```diff
@@ -551,69 +601,275 @@ RAFT_KERNEL block_rbc_kernel_eps_csr_pass(const value_t* X_index,
...
+        }
+        // abort in case subsequent points cannot possibly be in reach
+        i *= (cur_R_dist - min_warp_dist <= eps);
+      }
 
-      const value_t* y_ptr = X_index + (n_cols * cur_candidate_ind);
-      if (dfunc(x_ptr, y_ptr, n_cols) <= eps) {
-        auto row_pos = atomicAdd(&column_index_smem, 1);
-        if (pass2) adj_ja[row_pos] = cur_candidate_ind;
+      while (i >= WarpSize) {
```
Nyrio, line 834: Oh, one more thing. I was trying to make a reproducer for the "branch on divergence" (`BRA.DIV`) instructions that we've been seeing, and it occurred to me that the loop condition of the inner-most loop depends on the lane id, because `i` is `limit + lid`, but iiuc the possible out-of-bounds are handled before the loop, and all lanes always execute that while loop together.

So one thing to try would be to help the compiler know that all lanes are active: for example broadcast `i0`, use it in the loop condition, and inside the loop use `i = i0 + lid`.

Nyrio: Erm, unfortunately that idea does not work well with the early exit:

```c++
// abort in case subsequent points cannot possibly be in reach
i0 *= (cur_R_dist - min_warp_dist <= eps);
```

Nyrio: You can probably implement that with a `VOTE`.

Replace:
```c++
// abort in case subsequent points cannot possibly be in reach
i *= (cur_R_dist - min_warp_dist <= eps);
```

with:
```
next_loop = (i >= WarpSize) && (cur_R_dist - min_warp_dist <= eps);
```

and change loop condition to:

```c++
while (raft::any(next_loop))
```

Issue #1738, opened 2023-08-15
[FEA] Primitives for pre-filtering


cjnolet:
The IVF methods currently accept pre-filtering functions that can be applied during scan but still need optimized primitives that can allow users to efficiently express and perform the filtering logic. 

So far we've discussed simple bitsets, roaring bitmaps, bloom filters, and hash table primitives to build filtering functions on top of. Initially, these APIs could probably accept an array of ids on device or host and produce a data structure from which a filter function can be produced which can be passed directly into `search_with_filter`. I think it will be important to consider the API design up front so we can provide a unified API experience as much as possible

cjnolet:
Taks list so far:

- [ ] biset
- [ ] bitmap
- [ ] bloom filter
- [ ] hash table
- [ ] hashmap

tfeher:
Linking discussion about template parameters for filters: https://github.com/rapidsai/raft/pull/2212#issuecomment-1979439358, and related task
- [ ] avoid recompiling `ivf:pq::search` kernels when index type changes

PR #2212, opened 2024-03-05
Add explicit instantiations for IVF-PQ search kernels used in tests

tfeher:
Compilation of IVF-PQ search kernels can be time consuming. In `libraft.so` the compilation is done in parallel for kernels without filtering and with `int64_t` index type.

We have test with `uint32_t` index type as well as tests for `bitset_filter` with both 32 and 64 bit index types. This PR adds explicit template instantiations for the test. This way we avoid repeated compilation of the kernels with filter and this also enables parallel compilation of the `compute_similarity` kernel for different template types. The kernels with these additional type parameters are not added to `libraft.so`, only linked together with the test executable. 

Note that this PR does not increase the number of compiled kernels, but it enables to compile them in parallel.


cjnolet:
> One idea would be to cast away the index type if none_ivf_sample_filter is selected while we [wrap it into the adapter](https://github.com/rapidsai/raft/blob/054731df309959f04889cd44f9c0b3c04f833136/cpp/include/raft/neighbors/detail/ivf_pq_search.cuh#L655). Could we have a way where the filter adapter is not templated on IdxT?

We had a discussion   a little while back of just having the `search_with_filter()` accept a `cuda::std::function` as an argument (thus removing the template for the filter type).  I wonder if it's finally time to make that change in the 24.04 version.

tfeher:
> I wonder if it's finally time to make that change in the 24.04 version.

As discussed offline, I recommend to push further changes to the filter function to 24.06, and keep this PR in its current, non-breaking form. I have added a reference to the tracking issue for filters https://github.com/rapidsai/raft/issues/1738

tfeher:
Overview of the changes in compile time. Test and bench with filtering plus test with uint32_t (highlighted in bold) is now compiled in parallel

file | build time before | after PR (mm:ss)
-- | -- | --
**ann_ivf_pq/test_float_uint32_t.cu** | 50:15 | 3:13
**ann_ivf_pq/test_filter_float_int64_t.cu** | 26:52 | 1:30
**ann_ivf_pq/test_filter_int8_t_int64_t.cu**| 26:48 | 1:50
**bench/ivf_pq_filter_float_int64_t.cu** | 21:05 | 1:48
ann_ivf_pq/test_int8_t_int64_t.cu | 3:17 | 3:14
ann_ivf_pq/test_float_int64_t.cu | 3:14 | 4:14
ann_ivf_pq/test_uint8_t_int64_t.cu | 3:11 | 4:11
bench/ivf_pq_int8_t_int64_t.cu | 1:55 | 1:45
bench/ivf_pq_uint8_t_int64_t.cu | 1:53 | 1:39
bench/ivf_pq_float_int64_t.cu | 1:53 | 1:56
ann_ivf_pq/ivf_pq_compute_similarity_half_fp8_true_bitset32 | 0 (new) | 6:06
ann_ivf_pq/ivf_pq_compute_similarity_half_fp8_true_filt32 | 1 (new) | 6:03
ann_ivf_pq/ivf_pq_compute_similarity_half_fp8_true_bitset64 | 2 (new) | 5:37
ann_ivf_pq/ivf_pq_compute_similarity_half_fp8_false_filt32 | 3 (new) | 5:15
ann_ivf_pq/ivf_pq_compute_similarity_half_fp8_false_bitset32 | 4 (new) | 5:15
ann_ivf_pq/ivf_pq_compute_similarity_half_half_bitset32 | 5 (new) | 5:06
ann_ivf_pq/ivf_pq_compute_similarity_float_fp8_true_bitset32 | 6 (new) | 5:04
ann_ivf_pq/ivf_pq_compute_similarity_half_fp8_false_bitset64 | 7 (new) | 5:03
ann_ivf_pq/ivf_pq_compute_similarity_half_half_filt32 | 8 (new) | 5:03
ann_ivf_pq/ivf_pq_compute_similarity_float_fp8_true_filt32 | 9 (new) | 5:02
ann_ivf_pq/ivf_pq_compute_similarity_float_fp8_true_bitset64 | 10 (new) | 4:51
ann_ivf_pq/ivf_pq_compute_similarity_float_fp8_false_bitset32 | 11 (new) | 4:46
ann_ivf_pq/ivf_pq_compute_similarity_half_half_bitset64 | 12 (new) | 4:44
ann_ivf_pq/ivf_pq_compute_similarity_float_fp8_false_filt32 | 13 (new) | 4:44
ann_ivf_pq/ivf_pq_compute_similarity_float_fp8_false_bitset64 | 14 (new) | 4:42
ann_ivf_pq/ivf_pq_compute_similarity_float_half_bitset32 | 15 (new) | 4:37
ann_ivf_pq/ivf_pq_compute_similarity_float_half_filt32 | 16 (new) | 4:37
ann_ivf_pq/ivf_pq_compute_similarity_float_half_bitset64 | 17 (new) | 4:35
ann_ivf_pq/ivf_pq_compute_similarity_float_float_bitset32 | 18 (new) | 4:34
ann_ivf_pq/ivf_pq_compute_similarity_float_float_filt32 | 19 (new) | 4:31
ann_ivf_pq/ivf_pq_compute_similarity_float_float_bitset64 | 20 (new) | 4:28

cpp/include/raft/neighbors/detail/ivf_pq_build.cuh
```diff
@@ -442,15 +442,16 @@ void train_per_subset(raft::resources const& handle,
                  stream);
 
     // train PQ codebook for this subspace
-    auto sub_trainset_view =
-      raft::make_device_matrix_view<const float, IdxT>(sub_trainset.data(), n_rows, index.pq_len());
-    auto centers_tmp_view = raft::make_device_matrix_view<float, IdxT>(
+    auto sub_trainset_view = raft::make_device_matrix_view<const float, int64_t>(
```
achirkin, line 445: Do I understand it right, that all of the changes in this file are replacing `IdxT` with `int64_t`, and the purpose of this is to reduce the number of instantiations of `kmeans_balanced`?
This file is in `detail` namespace. Would it make sense to define something like `using internal_extents_t = int64_t` somewhere above in this file and use that one instead of `int64_t`? I hope this would make the intent clear, in case if someone needs to update the file in future and wonders where to use `IdxT` and where to use `int64_t`.

achirkin (approved):
Thanks, @tfeher, for the PR! Looks good to me. A small nitpick about the internal indexing type below.

cpp/include/raft/neighbors/detail/ivf_pq_build.cuh
tfeher: Yes, this change only affects the mdpan extent type internally, and it is only made to reduce the number of template instantiations. Thanks for the good suggestion, I have made the change accordingly.

PR #2213, opened 2024-03-05
Add CAGRA-Q build (compression)

achirkin:
Add a `cagra::compress` function that implements CAGRA-Q (VQ + PQ) compression of a given dataset.
The result, `compressed_dataset`, is supposed to complement the CAGRA graph during `cagra::search` in place of a raw dataset.

### Current state:

  - The code runs and produces a meaningful output (tested internally by running the original prototype search with the generated compressed dataset); the recall levels are approximately the same as with the prototype implementation.
  - No test coverage yet (need to coordinate with the search PR https://github.com/rapidsai/raft/pull/2206)
  - Full `pq_bits` support ([4,5,6,7,8] - same as in IVF-PQ)
  - Any `pq_dim` values are accepted, but the dataset is not padded and thus `dim` must be a multiple of `pq_dim`.
  - The codebook math type is hardcoded to `half` to match the prototype implementation for now. This could be a runtime (build) parameter as well.
  - All common input data types should work (`uint8_t`, `int8_t`, `half`, and `float` compile), but I tested only `float`.


cpp/include/raft/neighbors/dataset.hpp
```diff
@@ -0,0 +1,296 @@
...
+  {
+    return data.extent(1);
+  }
+  /** The number of "coarse cluster centers" */
+  [[nodiscard]] constexpr inline auto vq_n_centers() const noexcept -> uint32_t
+  {
+    return vq_code_book.extent(0);
+  }
+  /** The bit length of an encoded vector element after compression by PQ. */
+  [[nodiscard]] constexpr inline auto pq_bits() const noexcept -> uint32_t
```
tfeher, line 265: This seem to be calculating  a (default?) value based on the number of centers. Could you clarify the intent in the docstring?

cpp/include/raft/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -344,6 +345,16 @@ index<T, IdxT> build(
   RAFT_LOG_INFO("Graph optimized, creating index");
   // Construct an index from dataset and optimized knn graph.
   if (construct_index_with_dataset) {
+    if (params.compression.has_value()) {
+      index<T, IdxT> idx(res, params.metric);
+      idx.update_graph(res, raft::make_const_mdspan(cagra_graph.view()));
+      idx.update_dataset(
+        res,
+        // TODO: ATM, only float math type is supported in kmeans training.
```
tfeher, line 353: Note that kmeans can take int8 input, but under the hood it maps to float. This will improve in the future.

cpp/include/raft/neighbors/dataset.hpp
tfeher: Maybe I am just got confused between `vq_n_centers` and `pq_n_centers` So instead of storing a field "pq_bits", we derive this value from the size the codebook table.

tfeher (changes requested):
Thanks @achirkin for this PR! Here is a first batch of my comments.

cpp/include/raft/neighbors/dataset.hpp
achirkin: You're right, I don't store the `pq_bits` here, because it's derived from `pq_n_centers` (PQ book size). I find it very nice that this way we don't have any redundant information (one cannot leave `vpq_dataset` in invalid state by replacing one of the components by accident).

cpp/include/raft/neighbors/dataset.hpp
```diff
@@ -0,0 +1,296 @@
...
+ *
+ */
+template <typename MathT, typename IdxT>
+struct vpq_dataset : public dataset<IdxT> {
+  /** Vector Quantization codebook - "coarse cluster centers". */
+  device_matrix<MathT, uint32_t, row_major> vq_code_book;
+  /** Product Quantization codebook - "fine cluster centers".  */
+  device_matrix<MathT, uint32_t, row_major> pq_code_book;
+  /** Compressed dataset.  */
+  device_matrix<uint8_t, IdxT, row_major> data;
```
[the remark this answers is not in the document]
achirkin: The `IdxT` here is used for consistency with the other dataset types (mdarrays) that are already in CAGRA and IVF-* methods, thus I'd prefer not to change it here.
I like the idea of removing the `IdxT` type though, it would simplify the whole hierarchy of dataset types (it's  only used for extents in this file). I'd suggest we could revise this (for all ANN algorithms together) in a follow-on PR or  in cuVS.

cpp/include/raft/neighbors/dataset.hpp
```diff
@@ -0,0 +1,296 @@
...
+ * @tparam MathT the type of elements in the codebooks
+ * @tparam IdxT type of the vector indices (represent dataset.extent(0))
+ *
+ */
+template <typename MathT, typename IdxT>
+struct vpq_dataset : public dataset<IdxT> {
+  /** Vector Quantization codebook - "coarse cluster centers". */
+  device_matrix<MathT, uint32_t, row_major> vq_code_book;
+  /** Product Quantization codebook - "fine cluster centers".  */
+  device_matrix<MathT, uint32_t, row_major> pq_code_book;
```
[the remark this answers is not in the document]
achirkin: We use the extent type inside extend (build) kernels and we have the same pattern in IVF-PQ and in the prototype CAGRA-Q, hence I left it `uint32_t` for consistency.
Again, I like the idea of hardcoding everything to `int64_t`, but I suggest we'd do this for all ANN algorithms at once in a dedicated PR.

cpp/include/raft/neighbors/dataset.hpp
achirkin: (commented the code)

cpp/include/raft/neighbors/dataset.hpp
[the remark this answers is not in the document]
tfeher: Since these are smaller arrays, having 32 bit extent type can be justified. I leave this to your judgement.

cpp/include/raft/neighbors/dataset.hpp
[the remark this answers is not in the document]
tfeher: I do not see how the "consistency" argument applies here. In CAGRA we have `int64_t` as mdarray extent type. https://github.com/rapidsai/raft/blob/2d714cbfda6de5a29858b423e4c8b37a75675014/cpp/include/raft/neighbors/cagra_types.hpp#L353-L356

If we use the mdspan index operator, then the offsets would be calculated using the extent type. For large arrays like `data`, I think it is important to use 64 bit types.

tfeher (changes requested):
Many thanks Artem for updating the PR, and also adding serialization methods. Overall it looks good, but I still have an issue with the index types.

cpp/include/raft/neighbors/dataset.hpp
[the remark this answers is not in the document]
achirkin: This hasn't changed in the PR, because I instantiate the `neighbors::dataset` type with `int64_t` as well. That is, in the current state, we only ever instantiate the dataset with one extents type.
At the same time, we're using `IdxT` in IVF-* methods. I think there's no harm to keep the `IdxT` in the dataset (aiming to converge IVF-* and CAGRA methods one day) as long as it does not change the behavior of CAGRA.

tfeher (commented):
Thanks Artem for integrating data compression into RAFT CAGRA! The PR looks good to me, just have two minor questions below.

Ideally we should go ahead and merge this, so that the follow up PR (#2206) is easier to  review.

cpp/include/raft/neighbors/cagra.cuh
```diff
@@ -279,62 +280,6 @@ index<T, IdxT> build(raft::resources const& res,
...
-  RAFT_EXPECTS(
-    queries.extent(0) == neighbors.extent(0) && queries.extent(0) == distances.extent(0),
-    "Number of rows in output neighbors and distances matrices must equal the number of queries.");
-
-  RAFT_EXPECTS(neighbors.extent(1) == distances.extent(1),
-               "Number of columns in output neighbors and distances matrices must equal k");
-  RAFT_EXPECTS(queries.extent(1) == idx.dim(),
-               "Number of query dimensions should equal number of dimensions in the index.");
-
-  using internal_IdxT   = typename std::make_unsigned<IdxT>::type;
```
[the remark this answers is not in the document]
achirkin: Supposedly, this PR does not change anything in that regard.
Actually, the change in this file is not related to the PR at all - it just removes a duplicated code by making `search` call the public `search_with_filtering` rather than doing identical checks and calling `cagra::detail::search_main` directly.

cpp/include/raft/neighbors/detail/cagra/cagra_serialize.cuh
```diff
@@ -23,6 +23,7 @@
 #include <raft/core/resource/cuda_stream.hpp>
 #include <raft/core/serialize.hpp>
 #include <raft/neighbors/cagra_types.hpp>
+#include <raft/neighbors/detail/dataset_serialize.hpp>
 
 #include <cstddef>
 #include <cstdint>
```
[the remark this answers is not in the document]
achirkin: Thanks for the reminder! done

tfeher (approved):
Thanks Artem, the PR looks good to me!

cpp/include/raft/neighbors/dataset.hpp
```diff
@@ -0,0 +1,334 @@
...
+ *
+ * @tparam SrcT the source mdarray or mdspan
+ *
+ * @param[in] res raft resources handle
+ * @param[in] src the source mdarray or mdspan
+ * @param[in] align_bytes the required byte alignment for the dataset rows.
+ * @return maybe owning current-device-accessible strided matrix
+ */
+template <typename SrcT>
+auto construct_aligned_dataset(const raft::resources& res,
```
cjnolet, line 201: Nitpick- we use `make` across the codebase when we're simply constructing an object out of a given input and we use `build` when we're building a complex object, which requires compute. Can we follow that same nomenclature here as well?

achirkin:
Latest update:

1. Renamed `dataset_view()` back to `dataset()` thus making the PR non-breaking.
2. Made a few other smaller renamings as per our discussion
3. Marked `dataset()` deprecated and added an `EXPERIMENTAL` note on the new index parameter.

cpp/include/raft/neighbors/cagra_types.hpp
```diff
@@ -145,25 +153,37 @@ struct index : ann::index {
...
+   *   dataset type at runtime.
+   */
+  [[nodiscard]] [[deprecated("Use data()")]] inline auto dataset() const noexcept
     -> device_matrix_view<const T, int64_t, layout_stride>
   {
-    return dataset_view_;
+    auto p = dynamic_cast<strided_dataset<T, int64_t>*>(dataset_.get());
+    if (p != nullptr) { return p->view(); }
+    auto d = dataset_->dim();
+    return make_device_strided_matrix_view<const T, int64_t>(nullptr, 0, d, d);
```
divyegala, line 180: Should this be reachable at all? I'm not sure about surprising a user with returning a view to `nullptr`

cpp/include/raft/neighbors/cagra_types.hpp
```diff
@@ -334,26 +360,10 @@ struct index : ann::index {
...
-                   static_cast<size_t>(dataset_view_.extent(1)),
-                   static_cast<size_t>(dataset_view_.stride(0)));
-  }
-
   raft::distance::DistanceType metric_;
-  raft::device_matrix<T, int64_t, row_major> dataset_;
   raft::device_matrix<IdxT, int64_t, row_major> graph_;
-  raft::device_matrix_view<const T, int64_t, layout_stride> dataset_view_;
   raft::device_matrix_view<const IdxT, int64_t, row_major> graph_view_;
+  std::unique_ptr<neighbors::dataset<int64_t>> dataset_;
```
divyegala, line 366: Why does this need to be a `unique_ptr`? If `dataset` is to be owned by the index then it can be a plain-old member

cpp/include/raft/neighbors/dataset.hpp
```diff
@@ -0,0 +1,330 @@
...
+  [[nodiscard]] virtual auto is_owning() const noexcept -> bool = 0;
+  virtual ~dataset() noexcept                                   = default;
+};
+
+template <typename IdxT>
+struct empty_dataset : public dataset<IdxT> {
+  using index_type = IdxT;
+  uint32_t suggested_dim;
+  explicit empty_dataset(uint32_t dim) noexcept : suggested_dim(dim) {}
+  [[nodiscard]] auto n_rows() const noexcept -> index_type final { return 0; }
```
divyegala, line 53:
```suggestion
  [[nodiscard]] constexpr auto n_rows() const noexcept -> index_type final { return 0; }
```

cpp/include/raft/neighbors/dataset.hpp
```diff
@@ -0,0 +1,330 @@
...
+  [[nodiscard]] auto dim() const noexcept -> uint32_t final { return suggested_dim; }
+  [[nodiscard]] auto is_owning() const noexcept -> bool final { return true; }
```
divyegala, line 55:
```suggestion
  [[nodiscard]] constexpr auto is_owning() const noexcept -> bool final { return true; }
```

cpp/include/raft/neighbors/dataset.hpp
```diff
@@ -0,0 +1,330 @@
...
+};
+
+template <typename DataT, typename IdxT>
+struct non_owning_dataset : public strided_dataset<DataT, IdxT> {
+  using index_type = IdxT;
+  using value_type = DataT;
+  using typename strided_dataset<value_type, index_type>::view_type;
+  view_type data;
+  explicit non_owning_dataset(view_type v) noexcept : data(v) {}
+  [[nodiscard]] auto is_owning() const noexcept -> bool final { return false; }
```
divyegala, line 85:
```suggestion
  [[nodiscard]] constexpr auto is_owning() const noexcept -> bool final { return false; }
```

cpp/include/raft/neighbors/dataset.hpp
```diff
@@ -0,0 +1,330 @@
...
+              device_matrix<uint8_t, IdxT, row_major>&& data)
+    : vq_code_book{std::move(vq_code_book)},
+      pq_code_book{std::move(pq_code_book)},
+      data{std::move(data)}
+  {
+  }
+
+  [[nodiscard]] auto n_rows() const noexcept -> IdxT final { return data.extent(0); }
+  [[nodiscard]] auto dim() const noexcept -> uint32_t final { return vq_code_book.extent(1); }
+  [[nodiscard]] auto is_owning() const noexcept -> bool final { return true; }
```
divyegala, line 278:
```suggestion
  [[nodiscard]] constexpr auto is_owning() const noexcept -> bool final { return true; }
```

cpp/include/raft/neighbors/dataset.hpp
```diff
@@ -0,0 +1,330 @@
...
+ * @tparam SrcT the source mdarray or mdspan
+ *
+ * @param[in] res raft resources handle
+ * @param[in] src the source mdarray or mdspan
+ * @param[in] required_stride the leading dimension (in elements)
+ * @return maybe owning current-device-accessible strided matrix
+ */
+template <typename SrcT>
+auto make_strided_dataset(const raft::resources& res, const SrcT& src, uint32_t required_stride)
+  -> std::unique_ptr<strided_dataset<typename SrcT::value_type, typename SrcT::index_type>>
```
divyegala, line 131: Same question about `unique_ptr`, I think it's unnecessary.

cpp/include/raft/neighbors/dataset.hpp
```diff
@@ -0,0 +1,330 @@
...
+namespace raft::neighbors {
+
+/** Two-dimensional dataset; maybe owning, maybe compressed, maybe strided. */
+template <typename IdxT>
+struct dataset {
+  using index_type = IdxT;
+  /**  Size of the dataset. */
+  [[nodiscard]] virtual auto n_rows() const noexcept -> index_type = 0;
+  /** Dimensionality of the dataset. */
+  [[nodiscard]] virtual auto dim() const noexcept -> uint32_t = 0;
```
divyegala, line 42: Can we save this in an `Extents` object? Then the behavior will at least be similar to `mdspan/mdarray`.

cpp/include/raft/neighbors/cagra_types.hpp
achirkin: Yes, it's reachable because it could be another child of `dataset` type -  `vpq_dataset`.
Also, even before this PR you could get `nullptr` as the `data_handle()` of the view - if you forget to set the dataset to cagra.

cpp/include/raft/neighbors/cagra_types.hpp
achirkin: It cannot be plain-old member, because dataset is an abstract class. Atm three different classes can be instantiated here, so we need a pointer.

cpp/include/raft/neighbors/dataset.hpp
achirkin: No, we cannot use the `constexpr` here, because this is a virtual (final) method (this is only possible in C++20 and above).

cpp/include/raft/neighbors/dataset.hpp
achirkin: Same as above - cannot use constexpr.

cpp/include/raft/neighbors/dataset.hpp
achirkin: Same as above - cannot use constexpr.

cpp/include/raft/neighbors/dataset.hpp
achirkin: Same as above - cannot use constexpr.

cpp/include/raft/neighbors/dataset.hpp
achirkin: Same as above: we have two different types hiding behind the `strided_dataset` - we need a pointer.

cpp/include/raft/neighbors/dataset.hpp
achirkin: I'm more inclined to mirror the interface of the `index` types here. You can also get an extents object of a strided dataset as `dataset.view().extents()`.

divyegala: It is closer to an mdspan/mdarray contextually as a type than it is to an index type. I don't feel too strongly about it either way, I'd just like the behavior to be as non-breaking as possible when returning this type publicly, despite the deprecation warning.

Issue #2205, opened 2024-02-28
[BUG] Sum inaccurate result


tfeher:
**Describe the bug**
The [sum kernel](https://github.com/rapidsai/raft/blob/67893676f3d9b90e572f78b969172f840115b22f/cpp/include/raft/stats/detail/sum.cuh#L65) does not handle underflows correctly, and that leads to inaccurate results.

**Steps/Code to reproduce bug**

As reported by @lijinf2:
> We also did an experiment of summing up a column of 0.1 across 130,382,757 rows. Observed raft returns 12,528,200.

**Expected behavior**
Compensate for underflow e.g. by [Kahan summation](https://en.wikipedia.org/wiki/Kahan_summation_algorithm) or a similar method.

Issue #2204, opened 2024-02-28
[BUG] Stats kernels invalid access


tfeher:
**Describe the bug**

Similar to https://github.com/rapidsai/raft/pull/2183, the mean and stdev kernels have also potential out of bounds access

https://github.com/rapidsai/raft/blob/branch-24.04/cpp/include/raft/stats/detail/stddev.cuh#L48
https://github.com/rapidsai/raft/blob/branch-24.04/cpp/include/raft/stats/detail/mean.cuh#L46

Additionally the [minmax](https://github.com/rapidsai/raft/blob/67893676f3d9b90e572f78b969172f840115b22f/cpp/include/raft/stats/detail/minmax.cuh#L159minmax) kernel should be also checked, whether it has a similar problem.

**Steps/Code to reproduce bug**
- Add a similar test that was added in #2183.
- Run `compute-sanitizer gtests/STATS_TEST`. Without compute-sanitizer the test could silently pass.

**Additional context**
- We have to add tests that cover edge cases so that such errors are not occuring

PR #2223, opened 2024-03-13
Fix illegal acces mean/stdev, sum add Kahan Summation

mfoerste4:
This PR addresses #2204 and #2205. 

* fixes illegal access / test coverage for mean row-wise kernel
* fixes illegal access / test coverage for stdev row-wise kernel
* modified sum kernels to utilize Kahan/Neumaier summation per thread, also increase load per thread to benefit from this

FYI, @tfeher


cpp/test/stats/mean.cu
```diff
@@ -95,39 +95,33 @@ class MeanTest : public ::testing::TestWithParam<MeanInputs<T>> {
...
+  {0.15f, -1.f, 1024, 32, false, false, 1234ULL},  {0.15f, -1.f, 1024, 64, false, false, 1234ULL},
+  {0.15f, -1.f, 1024, 128, false, false, 1234ULL}, {0.15f, -1.f, 1024, 256, false, false, 1234ULL},
+  {0.15f, 1.f, 1024, 32, true, true, 1234ULL},     {0.15f, 1.f, 1024, 64, true, true, 1234ULL},
+  {0.15f, 1.f, 1024, 128, true, true, 1234ULL},    {0.15f, 1.f, 1024, 256, true, true, 1234ULL},
+  {0.15f, -1.f, 1024, 32, false, true, 1234ULL},   {0.15f, -1.f, 1024, 64, false, true, 1234ULL},
+  {0.15f, -1.f, 1024, 128, false, true, 1234ULL},  {0.15f, -1.f, 1024, 256, false, true, 1234ULL},
+  {0.15f, -1.f, 1030, 1, false, false, 1234ULL},   {0.15f, -1.f, 1030, 60, true, false, 1234ULL},
+  {2.0f, -1.f, 31, 120, false, false, 1234ULL},    {2.0f, -1.f, 1, 130, true, false, 1234ULL},
+  {0.15f, -1.f, 1030, 1, false, true, 1234ULL},    {0.15f, -1.f, 1030, 60, true, true, 1234ULL},
+  {2.0f, -1.f, 31, 120, false, true, 1234ULL},     {2.0f, -1.f, 1, 130, true, true, 1234ULL}};
```
tfeher, line 110: Could you add a few more cases where both dims are below 32? E.g. 1x1, 7x23, 17x5.

cpp/test/stats/minmax.cu
```diff
@@ -145,45 +145,31 @@ class MinMaxTest : public ::testing::TestWithParam<MinMaxInputs<T>> {
...
+const std::vector<MinMaxInputs<float>> inputsf = {
+  {0.00001f, 1024, 32, 1234ULL},   {0.00001f, 1024, 64, 1234ULL},  {0.00001f, 1024, 128, 1234ULL},
+  {0.00001f, 1024, 256, 1234ULL},  {0.00001f, 1024, 512, 1234ULL}, {0.00001f, 1024, 1024, 1234ULL},
+  {0.00001f, 4096, 32, 1234ULL},   {0.00001f, 4096, 64, 1234ULL},  {0.00001f, 4096, 128, 1234ULL},
+  {0.00001f, 4096, 256, 1234ULL},  {0.00001f, 4096, 512, 1234ULL}, {0.00001f, 4096, 1024, 1234ULL},
+  {0.00001f, 8192, 32, 1234ULL},   {0.00001f, 8192, 64, 1234ULL},  {0.00001f, 8192, 128, 1234ULL},
+  {0.00001f, 8192, 256, 1234ULL},  {0.00001f, 8192, 512, 1234ULL}, {0.00001f, 8192, 1024, 1234ULL},
+  {0.00001f, 1024, 8192, 1234ULL}, {0.00001f, 1023, 5, 1234ULL},   {0.00001f, 1025, 30, 1234ULL},
+  {0.00001f, 2047, 65, 1234ULL},   {0.00001f, 2049, 22, 1234ULL},  {0.00001f, 31, 644, 1234ULL},
+  {0.00001f, 33, 999, 1234ULL}};
```
tfeher, line 157: Same here, please add edge cases where both dims are below 32.

cpp/test/stats/stddev.cu
```diff
@@ -159,13 +165,27 @@ const std::vector<StdDevInputs<double>> inputsd = {
   {0.1, -1.0, 2.0, 1024, 32, false, true, 1234ULL},
   {0.1, -1.0, 2.0, 1024, 64, false, true, 1234ULL},
   {0.1, -1.0, 2.0, 1024, 128, false, true, 1234ULL},
-  {0.1, -1.0, 2.0, 1024, 256, false, true, 1234ULL}};
+  {0.1, -1.0, 2.0, 1024, 256, false, true, 1234ULL},
+  {0.1, -1.0, 2.0, 1099, 97, false, false, 1234ULL},
+  {0.1, -1.0, 2.0, 1022, 694, true, false, 1234ULL},
+  {0.5, -1.0, 2.0, 31, 1, true, true, 1234ULL},
```
tfeher, line 171: Same here, please add edge cases where both dims are below 32.

cpp/test/stats/sum.cu
```diff
@@ -81,27 +96,37 @@ class SumTest : public ::testing::TestWithParam<SumInputs<T>> {
   rmm::device_uvector<T> data, sum_act;
 };
 
-const std::vector<SumInputs<float>> inputsf = {
-  {0.05f, 4, 5, 1234ULL}, {0.05f, 1024, 32, 1234ULL}, {0.05f, 1024, 256, 1234ULL}};
-
-const std::vector<SumInputs<double>> inputsd = {{0.05, 1024, 32, 1234ULL},
-                                                {0.05, 1024, 256, 1234ULL}};
+const std::vector<SumInputs<float>> inputsf = {{0.0001f, 4, 5, true, 1},
```
tfeher, line 99: Could we have one more test with more rows than columns where both are below 32?

cpp/include/raft/stats/detail/sum.cuh
```diff
@@ -34,30 +34,72 @@ RAFT_KERNEL sumKernelRowMajor(Type* mu, const Type* data, IdxType D, IdxType N)
...
+    }
+    thread_sum = t;
+  }
+  thread_sum += thread_c;
+  __shared__ Type smu[ColsPerBlk];
+  if (threadIdx.x < ColsPerBlk) smu[threadIdx.x] = Type(0);
+  __syncthreads();
+  raft::myAtomicAdd(smu + thisColId, thread_sum);
   __syncthreads();
   if (threadIdx.x < ColsPerBlk && colId < D) raft::myAtomicAdd(mu + colId, smu[thisColId]);
```
tfeher, line 78: As discussed offline, we are still loosing accuracy here, because we cannot do atomic compensated summation. In a follow up PR, we should strive to improve this. A few notes:
- Within block: instead of shared memory atomics, could we do hierarchical reduction and keep the compensation?
- Across blocks: one could consider using a mutex to guard access. That is done in [fusedl2NN](https://github.com/rapidsai/raft/blob/3e67a6c3bc5c73cd4a77a71bb5c2d911f8a7bcd4/cpp/include/raft/spatial/knn/detail/fused_l2_knn-inl.cuh#L240-L340) and it might make sense to sync with @mdoijade to discuss pros / cons. Alternatively, dump values per block to temp space, and run a second compensated reduction over them.

cpp/include/raft/stats/detail/sum.cuh
```diff
@@ -34,30 +34,72 @@ RAFT_KERNEL sumKernelRowMajor(Type* mu, const Type* data, IdxType D, IdxType N)
...
+    if (abs(thread_sum) >= abs(cur_value)) {
+      thread_c += (thread_sum - t) + cur_value;
+    } else {
+      thread_c += (cur_value - t) + thread_sum;
+    }
+    thread_sum = t;
   }
-  Type acc = BlockReduce(temp_storage).Sum(thread_data);
+  thread_sum += thread_c;
+  Type acc = BlockReduce(temp_storage).Sum(thread_sum);
```
tfeher, line 102: This is not compensated right?

tfeher (changes requested):
Thanks Malte for the PR. It looks good, overall. Could you please check perf impact, and share some details on achieved accuracy?

cpp/include/raft/stats/detail/sum.cuh
mdoijade: I believe you would need to make use of extra smem here `smu[ColsPerBlk * RowsPerBlkPerIter ]` then store the each outputs something like `smu[ thisColId * RowsPerBlkPerIter + thisRowId ] = thread_sum` , followed by per-thread working on summing up `RowsPerBlkPerIter ` from a single warp0 with kahan algo if `RowsPerBlkPerIter`  is small and for larger `RowsPerBlkPerIter` like 32 you can use shfl based reduction with kahan algo applied on each of its 5 iteration.

cpp/include/raft/stats/detail/sum.cuh
```diff
@@ -34,30 +34,72 @@ RAFT_KERNEL sumKernelRowMajor(Type* mu, const Type* data, IdxType D, IdxType N)
...
-  raft::myAtomicAdd(smu + thisColId, thread_data);
+  raft::myAtomicAdd(smu + thisColId, thread_sum);
+  __syncthreads();
+  if (threadIdx.x < ColsPerBlk && colId < D) raft::myAtomicAdd(mu + colId, smu[thisColId]);
+}
+
+template <typename Type, typename IdxType, int TPB, int ColsPerBlk = 32>
+RAFT_KERNEL sumKahanKernelRowMajor(Type* mu, const Type* data, IdxType D, IdxType N)
+{
+  const int RowsPerBlkPerIter = TPB / ColsPerBlk;
```
mdoijade, line 53: you can consider making this `constexpr` ?

tfeher (approved):
Thanks Malte for the updates! Also many thanks to @mdoijade for the suggestions for improvement. I think we shall add those in a follow up PR, and settle with the current state for this release, since it already brings significant accuracy improvement. LGTM!

PR #2225, opened 2024-03-14
Re enable IVF random sampling

tfeher:
Random sampling of training set for IVF methods was reverted in https://github.com/rapidsai/raft/pull/2144 due to the large memory usage of the subsample method.

PR #2155 implements a new random sampling method. Using that we can now enable random sampling of IVF methods (#2052 and #2077), therefore this PR reverts #2144, and adjust the code to utilize the new sampling method.


tfeher:
Note this will only compile once #2155 is merged.

achirkin (approved):
LGTM. Love to see the IVF build code shrinking!

cpp/include/raft/neighbors/detail/ivf_pq_build.cuh
```diff
@@ -1702,77 +1667,46 @@ auto build(raft::resources const& handle,
...
   {
+    raft::random::RngState random_state{137};
     auto trainset_ratio = std::max<size_t>(
       1,
       size_t(n_rows) / std::max<size_t>(params.kmeans_trainset_fraction * n_rows, index.n_lists()));
     size_t n_rows_train = n_rows / trainset_ratio;
 
-    auto* device_memory = resource::get_workspace_resource(handle);
-    rmm::mr::managed_memory_resource managed_memory_upstream;
+    auto* device_mr = resource::get_workspace_resource(handle);
```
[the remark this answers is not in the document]
tfeher: Thanks, good point. I have changed the allocator for the `trainset` array. For the other smaller allocations I have left the `workspace_resource`.

cpp/include/raft/neighbors/detail/ivf_pq_build.cuh
```diff
@@ -1702,77 +1667,45 @@ auto build(raft::resources const& handle,
...
-                    trainset_tmp,
-                    index.dim() * sizeof(float) / sizeof(T),
-                    index.dim(),
-                    n_rows_train,
-                    stream);
-      }
+      // TODO(tfeher): Enable codebook generation with any type T, and then remove trainset tmp.
+      auto trainset_tmp = make_device_mdarray<T>(
+        handle, &managed_mr, make_extents<internal_extents_t>(n_rows_train, dim));
+      raft::matrix::detail::sample_rows<T, int64_t>(
```
cjnolet, line 1690: If not already, can we please expose the needed API publicly so that we can use the public API here please? 

In general, we should be using public APIs outside of the immediate namespaces because calling internal APIs end up coupling the code to them and make it very hard to migrate the APIs, should anything change (this is happening in many places right now across the codebase and it's simple enough just to switch to using the public APIs).

cpp/include/raft/neighbors/detail/ivf_pq_build.cuh
```diff
@@ -1702,77 +1667,45 @@ auto build(raft::resources const& handle,
...
+    raft::random::RngState random_state{137};
     auto trainset_ratio = std::max<size_t>(
       1,
       size_t(n_rows) / std::max<size_t>(params.kmeans_trainset_fraction * n_rows, index.n_lists()));
     size_t n_rows_train = n_rows / trainset_ratio;
 
-    auto* device_memory = resource::get_workspace_resource(handle);
-    rmm::mr::managed_memory_resource managed_memory_upstream;
+    auto* device_mr = resource::get_workspace_resource(handle);
+    rmm::mr::managed_memory_resource managed_mr;
```
cjnolet, line 1677: Please don't create new memory resources outside of the user's control.

cpp/include/raft/neighbors/ivf_pq_types.hpp
```diff
@@ -104,6 +104,14 @@ struct index_params : ann::index_params {
...
    */
   bool conservative_memory_allocation = false;
+  /**
+   * The max number of data points to use per PQ code during PQ codebook training. Using more data
+   * points per PQ code may increase the quality of PQ codebook but may also increase the build
+   * time. The parameter is applied to both PQ codebook generation methods, i.e., PER_SUBSPACE and
+   * PER_CLUSTER. In both cases, we will use `pq_book_size * max_train_points_per_pq_code` training
+   * points to train each codebook.
+   */
+  uint32_t max_train_points_per_pq_code = 256;
```
cjnolet, line 114: Why 256 here? Have we tested this empirically across many datasets ti verify this is a good default?

cjnolet (changes requested):

cpp/include/raft/neighbors/detail/ivf_pq_build.cuh
tfeher: Public API is there, I will switch to that in the cuVS version of the PR.

cpp/include/raft/neighbors/ivf_pq_types.hpp
tfeher: The default value is inspired by FAISS, which also has 256 as default. We tested on  DEEP-100M here https://github.com/rapidsai/raft/pull/2052#issuecomment-1890191014. I will share results on other datasets.

tfeher (commented):
Closing this in favor of https://github.com/rapidsai/cuvs/pull/122

PR #2226, opened 2024-03-14
Fix bug in blockRankedReduce

akifcorduk:
There was a bug appearing for negative floating point numbers with a max reduce operation. The `std::numeric_limits<T>::min()` is greater than the negative floating point values whereas we want it to be smaller than all representable values.

This PR replaces the `min` with the `lowest`.


cpp/include/raft/util/reduction.cuh
```diff
@@ -157,11 +157,11 @@ DI std::pair<T, i_t> blockRankedReduce(T val,
     val = values[lane];
     idx = indices[lane];
   } else {
-    // get the min if it is a max op, get the max if it is a min op
-    val = reduce_op(std::numeric_limits<T>::min(), std::numeric_limits<T>::max()) ==
-              std::numeric_limits<T>::min()
+    // get the lowest if it is a max op, get the max if it is a lowest op
+    val = reduce_op(std::numeric_limits<T>::lowest(), std::numeric_limits<T>::max()) ==
```
bdice, line 161: These are using the largest magnitude positive/negative _finite_ values as the identity of the operation. However, the proper identity of max/min over the floating point domain is negative/positive infinity.

See how cudf handles this here: https://github.com/rapidsai/cudf/blob/85f41dfdb606664cd8bb256dff943bf81aa8816d/cpp/include/cudf/detail/utilities/device_operators.cuh#L134

You might want to check `std::numeric_limits<T>::has_infinity` and use infinities instead of max/lowest.

Also if this needs to support CUDA specific types like __int128 or half-precision, etc, then use cuda::std:: functions from libcudacxx instead of std:: functions.

bdice (changes requested):

cpp/include/raft/util/reduction.cuh
tfeher: We use these helpers at some other places:
https://github.com/rapidsai/raft/blob/56c0b3a2d3f886ec541e70ac964a60033762d5c0/cpp/include/raft/util/cudart_utils.hpp#L374-L387

akifcorduk: Makes sense, thanks Bradley! I will do the changes.

akifcorduk: Great Tamas, no need to reimplement.

akifcorduk:
@bdice @tfeher i addressed your comments. Please let me know if you have any other suggestions.

cpp/include/raft/util/reduction.cuh
```diff
@@ -157,11 +157,10 @@ DI std::pair<T, i_t> blockRankedReduce(T val,
...
     idx = indices[lane];
   } else {
-    // get the min if it is a max op, get the max if it is a min op
-    val = reduce_op(std::numeric_limits<T>::min(), std::numeric_limits<T>::max()) ==
-              std::numeric_limits<T>::min()
-            ? std::numeric_limits<T>::max()
-            : std::numeric_limits<T>::min();
+    // get the lower_bound of the type if it is a max op,
+    // get the upper bound of the type if it is a min op
+    val = reduce_op(lower_bound<T>(), upper_bound<T>()) == lower_bound<T>() ? upper_bound<T>()
```
bdice, line 162: It seems like the identity should be known at compile time rather than evaluating a max/min operation to determine the identity for the given operation.

bdice (approved):
One suggestion, but the bug is fixed so I’ll approve.

tfeher (approved):
Thanks Akif for the fix, LGTM. 

It looks like that invalid values are flagged with -1 idx and lower or upper bound value. It would be good to extend the docstring with this expected behavior. But let's not hold the fix with this.

PR #2228, opened 2024-03-14
MAINT: Simplify NCCL worker rank identification 

VibhuJawa:
This PR is based on @seberg work in  https://github.com/rapidsai/raft/pull/1928 .  

From the PR:

This is a follow up on https://github.com/rapidsai/raft/pull/1926, since the rank sorting seemed a bit hard to understand.

It does modify the logic in the sense that the host is now sorted by IP as a way to group based on it. But I don't really think that host sorting was ever a goal? 

If the goal is really about being deterministic, then this should be more (or at least clearer) deterministic about order of worker IPs.

OTOH, if the NVML device order doesn't matter, we could just sort the workers directly. 

The original https://github.com/rapidsai/raft/pull/1587 mentions:

NCCL>1.11 expects a process with rank r to be mapped to r % num_gpus_per_node
which is something that neither approach seems to quite assure, if such a requirement exists, I would want to do one of:

Ensure we can guarantee this, but this requires initializing workers that are not involved in the operation.
At least raise an error, because if NCCL will end up raising the error it will be very confusing.


VibhuJawa:
@cjnolet , Would it be possible to get an okay to test comment here please. 

Want to understand why it will break on CI .

jakirkham:
@VibhuJawa could you please squash the commits so there is one signed commit? That should fix the CI issues here

VibhuJawa:
@cjnolet ,  Would it be possible to get a review and merge on this .

Since we all-ready approved, https://github.com/rapidsai/raft/pull/1928 and the only delta b/w them is a test we should be good to go.

cjnolet (approved):

PR #2244, opened 2024-03-25
Add CAGRA-Q subspace dim = 4 support

enp1s0:
This PR adds the support for subspace dim (pq_dim) = 4 in CAGRA-Q

### DIFF
--- a/cpp/include/raft/neighbors/detail/cagra/compute_distance_vpq.cuh
+++ b/cpp/include/raft/neighbors/detail/cagra/compute_distance_vpq.cuh
@@ -33,6 +33,8 @@ struct cagra_q_dataset_descriptor_t : public dataset_descriptor_base_t<half, DIS
   using CODE_BOOK_T = CODE_BOOK_T_;
   using QUERY_T     = typename dataset_descriptor_base_t<half, DISTANCE_T, INDEX_T>::QUERY_T;
 
+  static_assert(std::is_same_v<CODE_BOOK_T, half>, "CODE_BOOK_T = `half` is only supported now");
+
   const std::uint8_t* encoded_dataset_ptr;
   const std::uint32_t encoded_dataset_dim;
   const std::uint32_t n_subspace;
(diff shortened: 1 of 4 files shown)


cpp/include/raft/neighbors/detail/cagra/compute_distance_vpq.cuh
```diff
@@ -56,9 +56,17 @@ struct cagra_q_dataset_descriptor_t : public dataset_descriptor_base_t<half, DIS
...
+        buf2.y = pq_code_book_ptr[i + 1];
+
+        // Change the order of PQ code book array to reduce the
+        // frequency of bank conflicts.
+        constexpr auto num_elements_per_bank  = 4 / utils::size_of<CODE_BOOK_T>();
+        constexpr auto num_banks_per_subspace = PQ_LEN / num_elements_per_bank;
+        const auto j                          = i / num_elements_per_bank;
+        const auto smem_index =
+          (j / num_banks_per_subspace) + (j % num_banks_per_subspace) * (1 << PQ_BITS);
+        reinterpret_cast<half2*>(smem_pq_code_book_ptr)[smem_index] = buf2;
```
achirkin, line 69: Do we need a similar adjustment to the logic in the other branch of the if (CODE_BOOK_T is float)?

enp1s0: No, we don't need it because the current `compute_similarity` only supports `half` as the `CODE_BOOK_T`. I removed the branch of the CODE_BOOK_T == float and added a static_assert to check the `CODE_BOOK_T` type. If it is better to support `float` as the `CODE_BOOK_T` type, I'll create a PR for that.

achirkin: Sounds good, thanks!

achirkin (approved):
LGTM!

cjnolet (approved):

PR #2246, opened 2024-03-26
Update nvtx3 link in cmake

lowener:
- Link to `nvtx3` instead of deprecated `nvToolsExt`
- Use `LAZY` cmake install message, to avoid flooding the console with all the headers name when they are left untouched.


robertmaynard:
The `CMAKE_INSTALL_MESSAGE` currently breaks `rapids-cmake` infrastructure to generate installable tests, so that needs to be removed.

robertmaynard (approved):

PR #2255, opened 2024-04-04
Adds missing files to `update-version.sh`

AyodeAwe:


ci/release/update-version.sh
```diff
@@ -64,6 +64,7 @@ for FILE in python/*/pyproject.toml; do
   for DEP in "${DEPENDENCIES[@]}"; do
     sed_runner "/\"${DEP}==/ s/==.*\"/==${NEXT_SHORT_TAG_PEP440}.*\"/g" ${FILE}
   done
+  sed_runner "s/version = \"[0-9][0-9].[0-9][0-9]/version = \"${NEXT_SHORT_TAG}/g" ${FILE}
```
raydouglass, line 67: I think this was added for this line https://github.com/rapidsai/raft/blob/branch-24.06/python/raft-ann-bench/pyproject.toml#L12

So it should be the full version. But should also open an issue to get this switched to use a dynamic version like other `pyproject.toml` files.

raydouglass (changes requested):

raydouglass (approved):

Issue #2033, opened 2023-11-30
[FEA] Enable direct cosine distance computation in nearest neighbors algorithms


cjnolet:
Currently, the angular distance can be computed in these methods using the inner product of the normalized vectors but this puts the burden of managing the norms and doing the normalization on the users. We should support cosine first-class right in the distance computation to alleviate these burdens. 

Tagging @liliu-z for awareness and tracking.

- [ ] brute-force
- [ ] nn-descent
- [ ] cagra
- [ ] ivf-pq
- [ ] ivf-flat
- [ ] random ball cover

Issue #1124, opened 2023-01-06
[FEA] Support more distance metrics in ivf_flat and ivf_pq


benfred:
**Is your feature request related to a problem? Please describe.**

Currently we only support L2 and InnerProduct distances in ivf_flat and ivf_pq .  We should expand
the supported distance metric to support at least L2SqrtExpanded and Cosine distances too.

**Additional context**

We're removing the faiss ANN code wrappers in https://github.com/rapidsai/raft/pull/1121/files - which used to handle these other distance metrics. This means that code that uses this knnIndex like `cuml.neighbors.NearestNeighbors` will only work with `sqeuclidean` distance now. This is slightly awkward because the default metric for  `cuml.neighbors.NearestNeighbors` is `euclidean` (L2SqrtExpanded) - which isn't support by the ivf_flat/ivf_pq ann indices right now.

benfred:
L2SqrtExpanded support was added to ivf_* in https://github.com/rapidsai/raft/pull/1133 and https://github.com/rapidsai/raft/pull/1138. 

The brute force api's use this code https://github.com/rapidsai/raft/blob/9a97b52880c5c97e2ff7f24b749d6bda87b6cb6f/cpp/include/raft/spatial/knn/detail/processing.cuh#L91 to pre-process/post-process the input distances for Cosine and Correlation inputs. We could do something similar for ivf_pq fairly easily

PR #2256, opened 2024-04-05
Add support for CosineExpanded in IVF-FLAT

lowener:
This work is still a draft.
Having a built-in support of Cosine distance can be useful instead of always relying on external code to normalize and post-process the distances.
In this proposal the norm of each vector is computed multiple times in the build/extend steps.  I am considering computing the norms only once to save time at the expanse of a higher memory cost.
The current proposal is not modifying data in-place.
Related: #2033, #1124


cjnolet:
@lowener this will need to be moved to cuVS.

PR #2259, opened 2024-04-09
Improve coalesced reduction performance for tall and thin matrices (up to 2.6x faster)

Nyrio:
This PR implements two optimizations to `coalescedReductionThinKernel` which is used for coalesced reductions of tall matrices (many rows) and/or thin (few columns):

1. Process multiple rows per warp to increase bytes in flight and amortize load latencies.
2. Use a vectorized reduction to avoid the LSU bottleneck and have fewer global stores (and at least partially coalesced).

The benchmark below shows the achieved SOL percentage on A30. I also measured that on H200, it achieved 84% SOL for 32 columns and up to 94% for 512 columns.

2024-04-09_coalesced_reduction_vec


cpp/include/raft/util/reduction.cuh
```diff
@@ -197,4 +197,104 @@ DI i_t binaryBlockReduce(i_t val, i_t* shmem)
...
+ *  lane 5 | f0 f1 f2 f3 | v2 - - -
+ *  lane 6 | g0 g1 g2 g3 | v3 - - -
+ *  lane 7 | h0 h1 h2 h3 | v3 - - -
+ *
+ * @tparam logicalWarpSize Sub-warp size. Must be 2, 4, 8, 16 or 32.
+ * @tparam vecWidth Vector width. Must be a power of two.
+ * @tparam T Vector element type.
+ * @tparam ReduceLambda Reduction operator type.
+ * @param[in,out] acc Pointer to a vector of size vecWidth or more in registers
+ * @param[in] lane_id Lane id between 0 and logicalWarpSize-1
```
tfeher, line 238: What is the role of this parameter? Isn't lane_id something derived from ` laneid % logicalWarpSize`?

cpp/include/raft/linalg/detail/coalesced_reduction-inl.cuh
```diff
@@ -53,19 +60,72 @@ RAFT_KERNEL __launch_bounds__(Policy::ThreadsPerBlock)
...
-  if (threadIdx.x == 0) {
+
+  if constexpr (Policy::NoSequentialReduce) {
+    IdxType j = threadIdx.x;
+    if (j < D) {
+#pragma unroll
+      for (IdxType k = 0; k < Policy::RowsPerLogicalWarp; k++) {
+        // Only the first row is known to be within bounds. Clamp to avoid out-of-mem read.
+        const IdxType i = raft::min(i0 + k * Policy::NumLogicalWarps, N - 1);
+        acc[k]          = reduce_op(acc[k], main_op(data[j + (D * i)], j));
```
tfeher, line 85: What happen to the values that we load using the clamped index? Are they ignored later when using the final reduce op?

tfeher (commented):
Thanks Louis for this updte! Overeall it looks good, I just have two questions.

cpp/include/raft/util/reduction.cuh
Nyrio: To compute the laneid generically, there are 2 possibilities:

- Using PTX to query the laneid: incurs an additional S2R (special register to register).
- Using blockDim and threadIdx: have to assume a 3D grid and also do a bunch of unnecessary S2R.

The coalesced reduction kernel uses `logicalWarpSize` as the block x dimension so the laneid can be computed at no extra cost as we need threadIdx.x anyway. It's micro-optimization, not a massive impact as long as we have multiple elements/thread so the `S2R`s are amortized, and we're not LSU bound.

cpp/include/raft/linalg/detail/coalesced_reduction-inl.cuh
Nyrio: The clamping only applies to the row id, i.e. the whole logical warp is either within bound or out of bounds. To minimize branching and divergence, it will participate in compute with that data we don't care about and then the store is guarded:

```c++
if (i < N) { dots[i] = final_op(acc[k]); }
```

If the column is out of bound, it is ignored in the initial sequential reduction, and the accumulator is initialized as `init` (a.g 0 for a sum reduction), so it doesn't impact the reduction output.

Nyrio: Because a logical warp has multiple rows, and a warp has multiple logical warps, it would be divergent if one logical warp executed a reduction for 7 rows and another one for 8.

tfeher (approved):
Thanks Louis for the explanation! The PR looks good to me.

cjnolet (approved):

PR #2264, opened 2024-04-11
Build C++ wheel

vyasr:
This PR changes wheel building in raft to create a separate C++ wheel that is then found from the Python wheel. The C++ wheel is now a hard dependency of the Python wheel. This allows Python packaging to more closely mirror the structure of our conda packaging, and the way we would normally wish to package these in any other manager. It also allows us to reduce package sizes by allowing better sharing of artifacts between different Python packages that rely on the same C++ components from other packages.

Contributes to rapidsai/build-planning#33


vyasr:
The pip devcontainer CI run won't pass until https://github.com/rapidsai/devcontainers/pull/287 is merged, and I plan to wait to merge this PR until https://github.com/rapidsai/raft/pull/1983 is merged and can be incorporated into this one, but this PR is ready for review now.

raydouglass (approved):

divyegala (approved):
Python approval

PR #2265, opened 2024-04-15
Always use a static gtest and gbench

robertmaynard:
Removes the need for us to install GTest or GBench in our raft testing CI


bdice (approved):
Could you cross-link to a central issue for these PRs to help with tracking?

PR #2269, opened 2024-04-17
Convert device_memory_resource* to device_async_resource_ref

### LINKED ISSUE #2261: Replace device_memory_resource* with device_async_resource_ref
Update occurences of `rmm::mr::device_memory_resource*` to
`rmm::device_async_resource_ref` in this repo. The rationale is explained
in [this document](https://docs.google.com/document/d/1QsLQn17qexczFGbzjLzTkIiGztfaQvLu-DpNu2EPw1E/edit#heading=h.wdftxmu1lxt).

harrism:
## Description
Closes #2261

For reviewers:
Many of changes are simple textual replace of `rmm::mr::device_memory_resource *` with `rmm::device_async_resource_ref`.  However there are several places where RAFT used a default value of `nullptr` for `device_memory_resource*` parameters. This is incompatible with a `resource_ref`, which is a lightweight non-owning reference class, not a pointer. In most places, I was able to either remove the default parameter value, or use `rmm::mr::get_current_device_resource()`. In the case of ivf_pq, I removed the deprecated versions of `search` that took an `mr` parameter.

I removed the unused old src/util/memory_pool.cpp and its headers.


cpp/bench/ann/src/raft/raft_ann_bench_utils.h
```diff
@@ -100,7 +101,7 @@ class shared_raft_resources {
   ~shared_raft_resources() noexcept { rmm::mr::set_current_device_resource(orig_resource_); }
```
achirkin, line 101: There are a few build failures in CI caused by `set_current_device_resource` still requiring `rmm::mr::device_memory_resource*`. Do we need to pin a spicific RMM branch to make this work?

achirkin (changes requested):
Thanks for the PR! There are plenty build failures atm (see below).

In the meanwhile, what's your opinion regarding [get_workspace_resource](https://github.com/rapidsai/raft/blob/39f3854fc0baf5837db19286d8461c2fd6296cdd/cpp/include/raft/core/resource/device_memory_resource.hpp#L158-L166), which didn't make it into this PR? From the RMM design point of view, shall we better change it to return the resource_ref (thus, losing the information about its type), or keep returning the pointer? Note, we provide extra helpers to get the limits from the limiting adapter.

harrism:
I will have to dig in when I'm back on Monday. It all built successfully on my local copy.

harrism:
@achirkin I didn't have those failures locally because apparently in the dev container `build-raft-cpp -DBUILD_TESTS=ON -DBUILD_BENCHMARKS=ON` does not build any TU that includes that header file (benchmarks). Can you tell me how to build RAFT benchmarks, and why RAFT uses a different approach to building them than all other RAPIDS repos?

> In the meanwhile, what's your opinion regarding [get_workspace_resource](https://github.com/rapidsai/raft/blob/39f3854fc0baf5837db19286d8461c2fd6296cdd/cpp/include/raft/core/resource/device_memory_resource.hpp#L158-L166), which didn't make it into this PR? From the RMM design point of view, shall we better change it to return the resource_ref (thus, losing the information about its type), or keep returning the pointer? Note, we provide extra helpers to get the limits from the limiting adapter.

I don't want to tackle this until we redesign `rmm::set_current_device_resource` / `rmm::get_current_device_resource`, because that redesign can inform this one. FWIW, this is another place where RAFT does things differently than every other RAPIDS library. Consistency would be good.

harrism:
Figured out how to build benchmarks. 

```
option(BUILD_PRIMS_BENCH "Build raft C++ benchmark tests" OFF)
option(BUILD_ANN_BENCH "Build raft ann benchmarks" OFF)
```

cpp/bench/ann/src/raft/raft_ann_bench_utils.h
harrism: Fixed.

cpp/bench/ann/src/raft/raft_ann_bench_utils.h
```diff
@@ -130,8 +131,8 @@ class configured_raft_resources {
   {
   }
 
-  configured_raft_resources(configured_raft_resources&&)            = default;
-  configured_raft_resources& operator=(configured_raft_resources&&) = default;
+  configured_raft_resources(configured_raft_resources&&)            = delete;
+  configured_raft_resources& operator=(configured_raft_resources&&) = delete;
```
achirkin, line 135: Did you mark this constructor as deleted only because it defaulted to deleted already due to `raft::device_resources` member, or there was a more specific/design issue?
If the former, I'd prefer to wrap the device_resources into a unique pointer to keep the move constructor (we move it in benchmarks when initializing algo wrappers). But I can do it later in a follow-on PR.

achirkin (approved):
I've checked that the PR doesn't change the current behavior w.r.t. usage of memory resources anywhere in raft, LGTM. Also thanks for fixing numerous mischiefs like unnecessary dynamic or c-style casts.

> I don't want to tackle this until we redesign rmm::set_current_device_resource / rmm::get_current_device_resource, because that redesign can inform this one. 

Could you please share, in which direction you plan to redesign these?

> FWIW, this is another place where RAFT does things differently than every other RAPIDS library. Consistency would be good.

It may be doing things even more differently pretty soon. Please, have a look at the problem we're trying to solve: https://github.com/rapidsai/raft/pull/2194 . If you have any suggestions on how we can solve the issues outlined in that PR more in line with other RAPIDS libraries, your feedback would be very appreciated!

cpp/include/raft/matrix/detail/select_radix.cuh
```diff
@@ -894,14 +894,12 @@ void radix_topk(const T* in,
                 unsigned grid_dim,
                 int sm_cnt,
                 rmm::cuda_stream_view stream,
-                rmm::mr::device_memory_resource* mr)
+                rmm::device_async_resource_ref mr = rmm::mr::get_current_device_resource())
```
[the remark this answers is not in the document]
harrism: Suggest opening an issue for all of these that need to change.

cpp/include/raft/neighbors/ivf_flat-inl.cuh
```diff
@@ -462,8 +462,8 @@ void search_with_filtering(raft::resources const& handle,
                            uint32_t k,
                            IdxT* neighbors,
                            float* distances,
-                           rmm::mr::device_memory_resource* mr = nullptr,
-                           IvfSampleFilterT sample_filter      = IvfSampleFilterT())
+                           rmm::device_async_resource_ref mr,
```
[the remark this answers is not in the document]
harrism: Please file an issue.

cpp/bench/ann/src/raft/raft_ann_bench_utils.h
harrism: Yes exactly. clang-tidy complains that this is explicitly defaulted even though it is implicitly deleted due to the deleted ctors in the `res_` member. 

If you are going to have these clang-tidy settings, why ignore the warnings?  Wrapping the device resources in a unique ptr is up to you -- it's outside of my scope and the scope of this PR.

I think I can revert all changes to this file. Turns out the resource_ref header is no longer used.

harrism: Actually it should include `device_memory_resource.hpp` in order to correctly ICWYU.

harrism:
> Could you please share, in which direction you plan to redesign these?

Once we have one, yes.

> Please, have a look at the problem we're trying to solve: https://github.com/rapidsai/raft/pull/2194 . If you have any suggestions

Thanks. I commented on #2194.

dantegd (approved):
lgtm

cjnolet (approved):

cpp/include/raft/matrix/detail/select_radix.cuh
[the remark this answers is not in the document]
achirkin: Fixed in #2194

cpp/include/raft/neighbors/detail/ivf_flat_search-inl.cuh
```diff
@@ -276,13 +276,12 @@ inline void search(raft::resources const& handle,
                    uint32_t k,
                    IdxT* neighbors,
                    float* distances,
-                   rmm::mr::device_memory_resource* mr = nullptr,
-                   IvfSampleFilterT sample_filter      = IvfSampleFilterT())
+                   rmm::device_async_resource_ref mr = rmm::mr::get_current_device_resource(),
```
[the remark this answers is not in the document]
achirkin: Fixed in #2194

PR #2288, opened 2024-05-02
Enable warnings as errors for Python tests

mroeschke:
As part of https://github.com/rapidsai/build-planning/issues/26, warnings in Python tests will be converted to test failures

`ignore:Unknown pytest.mark.ucx:PytestUnknownMarkWarning` could be removed once https://github.com/rapidsai/raft/pull/2281 is merged cc @jameslamb

### DIFF
--- a/python/raft-dask/pytest.ini
+++ b/python/raft-dask/pytest.ini
@@ -1,4 +1,7 @@
 [pytest]
+filterwarnings =
+    error
+    ignore:Unknown pytest.mark.ucx:PytestUnknownMarkWarning
 markers =
   unit: marks unit tests
   quality: marks quality tests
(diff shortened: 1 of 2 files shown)


jameslamb:
Thanks for the `@`.

I didn't even think about how this work you're doing would have caught the thing I just happened to notice in logs that led to #2281... nice!!

@rapidsai/raft-python-codeowners #2281 is small and I think non-controversial, it's ready for review.

python/raft-dask/pytest.ini
```diff
@@ -1,4 +1,7 @@
 [pytest]
+filterwarnings =
+    error
+    ignore:Unknown pytest.mark.ucx:PytestUnknownMarkWarning
```
mroeschke, line 4:
```suggestion
```

dantegd (approved):
lgtm

Issue #2106, opened 2024-01-22
[FEA] IVF index building with pinned H2D transfer


tfeher:
**Is your feature request related to a problem? Please describe.**
For IVF-Flat ad IVF-PQ index building, large datasets are provided in host memory or as `mmap`-ed file. After the cluster centers are trained, both method streams through the whole dataset twice. Currently there is no overlap between host to device copies and additional data processing on the GPU.

**Describe the solution you'd like**
Use pinned buffers to copy the data to the GPU and overlap it with GPU side computation.

**Additional context**
- Since the dataset can be larger than the physical (host) memory of the system, it is not possible to load the whole dataset into pinned memory.

- Index subsampling already use pinned buffers to overlap vector gathering and H2D copies https://github.com/rapidsai/raft/pull/2077/commits/548555766b9acb485000c24e32816d6d874f58b5
 
IVF-Flat and IVF-PQ streams through the whole dataset here:
- assign vectors to cluster centers (k-means predict): [IVF-Flat](https://github.com/rapidsai/raft/blob/d7eebc68a2582d8e1ca699df3816091d0224dcf7/cpp/include/raft/neighbors/detail/ivf_flat_build.cuh#L192-L210), [IVF-PQ](https://github.com/rapidsai/raft/blob/d7eebc68a2582d8e1ca699df3816091d0224dcf7/cpp/include/raft/neighbors/detail/ivf_pq_build.cuh#L1625-L1660) 
- copy vectors to their respective cluster (additionally encode vectros and map to a specific layout): [IVF-Flat](https://github.com/rapidsai/raft/blob/d7eebc68a2582d8e1ca699df3816091d0224dcf7/cpp/include/raft/neighbors/detail/ivf_flat_build.cuh#L275-L292), [IVF-PQ](https://github.com/rapidsai/raft/blob/d7eebc68a2582d8e1ca699df3816091d0224dcf7/cpp/include/raft/neighbors/detail/ivf_pq_build.cuh#L1698-L1711)

We use [batch_load_iterator](https://github.com/rapidsai/raft/blob/d7eebc68a2582d8e1ca699df3816091d0224dcf7/cpp/include/raft/spatial/knn/detail/ann_utils.cuh#L397) to copy the data to host. Ideally, we could improve the batch load-iterator to prefetch the data into a pinned buffer.

tfeher:
Tagging @abc99lr who plans to work on this, and @achirkin for visibility.

PR #2293, opened 2024-05-06
Add support for nosync thrust exec policy

abc99lr:
Currently, all the `thrust` calls used in RAFT are sync calls. There is another RMM execution policy that support async (or nosync) `thrust` calls. Supporting async calls is important. For example, we need the kmeans predict to be async in order to achieve kernel/copy overlapping in IVF-Flat index build (https://github.com/rapidsai/raft/issues/2106). 

This PR 

1. Add support for nosync thrust policy in `raft::resource` 
2. ~~Change the thrust calls in kmeans predict to nosync versions. This would enable us achieve memcpy and kernel overlapping in IVF-Flat index building.~~ Based on discussions, we should use `raft::linalg::map` instead. Will open another PR for this change.

### DIFF
--- a/cpp/include/raft/cluster/detail/kmeans_balanced.cuh
+++ b/cpp/include/raft/cluster/detail/kmeans_balanced.cuh
@@ -24,6 +24,7 @@
 #include <raft/core/operators.hpp>
 #include <raft/core/resource/cuda_stream.hpp>
 #include <raft/core/resource/device_memory_resource.hpp>
+#include <raft/core/resource/thrust_nosync_policy.hpp>
 #include <raft/core/resource/thrust_policy.hpp>
 #include <raft/distance/distance.cuh>
 #include <raft/distance/distance_types.hpp>
@@ -102,7 +103,7 @@ inline std::enable_if_t<std::is_floating_point_v<MathT>> predict_core(
       auto minClusterAndDistance = raft::make_device_mdarray<raft::KeyValuePair<IdxT, MathT>, IdxT>(
         handle, mr, make_extents<IdxT>(n_rows));
       raft::KeyValuePair<IdxT, MathT> initial_value(0, std::numeric_limits<MathT>::max());
-      thrust::fill(resource::get_thrust_policy(handle),
+      thrust::fill(resource::get_thrust_nosync_policy(handle),
                    minClusterAndDistance.data_handle(),
                    minClusterAndDistance.data_handle() + minClusterAndDistance.size(),
                    initial_value);
@@ -128,7 +129,7 @@ inline std::enable_if_t<std::is_floating_point_v<MathT>> predict_core(
 
       // todo(lsugy): use KVP + iterator in caller.
       // Copy keys to output labels
-      thrust::transform(resource::get_thrust_policy(handle),
+      thrust::transform(resource::get_thrust_nosync_policy(handle),
                         minClusterAndDistance.data_handle(),
                         minClusterAndDistance.data_handle() + n_rows,
                         labels,
--- a/cpp/include/raft/cluster/detail/kmeans_balanced.cuh
+++ b/cpp/include/raft/cluster/detail/kmeans_balanced.cuh
@@ -24,7 +24,6 @@
 #include <raft/core/operators.hpp>
 #include <raft/core/resource/cuda_stream.hpp>
 #include <raft/core/resource/device_memory_resource.hpp>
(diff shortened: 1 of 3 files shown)


tfeher (approved):
Thanks Rui for fixing this, it is indeed important to switch to async policy to improve vector search build time. Overall it looks good!

@achirkin could you also have a look at the changes in resource handling?

cpp/include/raft/cluster/detail/kmeans_balanced.cuh
```diff
@@ -102,7 +103,7 @@ inline std::enable_if_t<std::is_floating_point_v<MathT>> predict_core(
       auto minClusterAndDistance = raft::make_device_mdarray<raft::KeyValuePair<IdxT, MathT>, IdxT>(
         handle, mr, make_extents<IdxT>(n_rows));
       raft::KeyValuePair<IdxT, MathT> initial_value(0, std::numeric_limits<MathT>::max());
-      thrust::fill(resource::get_thrust_policy(handle),
+      thrust::fill(resource::get_thrust_nosync_policy(handle),
                    minClusterAndDistance.data_handle(),
                    minClusterAndDistance.data_handle() + minClusterAndDistance.size(),
                    initial_value);
```
achirkin, line 109: This could be done with
```
raft::linalg::map(handle, minClusterAndDistance, raft::const_op{initial_value});
```

cpp/include/raft/cluster/detail/kmeans_balanced.cuh
```diff
@@ -128,7 +129,7 @@ inline std::enable_if_t<std::is_floating_point_v<MathT>> predict_core(
 
       // todo(lsugy): use KVP + iterator in caller.
       // Copy keys to output labels
-      thrust::transform(resource::get_thrust_policy(handle),
+      thrust::transform(resource::get_thrust_nosync_policy(handle),
                         minClusterAndDistance.data_handle(),
                         minClusterAndDistance.data_handle() + n_rows,
                         labels,
```
achirkin, line 135: This could be done with
```
raft::linalg::map(handle, minClusterAndDistance, raft::compose_op<raft::cast_op<LabelT>, raft::key_op>{}, minClusterAndDistance);
```

achirkin (changes requested):
Thanks for the PR! I welcome introduction of the `thrust_nosync_policy`, it could be useful to us in theory. However, both places you changed in `kmean_balanced.cuh` could better be done with raft's own `linalg::map`.

cjnolet:
> However, both places you changed in kmean_balanced.cuh could better be done with raft's own linalg::map

@abc99lr, I just want to give an additional +1 to @achirkin's comment here. In general, we prefer to reuse raft functions where at all possible across the codebase, even when the raft function itself might be a trivial wrapper around thrust, cub, or one of the math libs. The reason for this is that it centralizes these calls so that we can properly instrument, improve, or even fix bugs as they arise in a single place, rather than having to scrape through the codebase and fix them in many places. 

The other reason for this is API consistency- over time, raft has improved to become quite a pleasant API experience by being able to compose larger algorithms out of a series of raft functions, which improves code readability.

abc99lr:
Thanks for the comments. I'll change the thrust calls in `kmeans_balanced::predict` to `raft::linalg::map` in a separate PR. For this one, I'll only add the nosync policy and remove the changes in `kmeans_balanced::predict`. Sound good?

cjnolet:
> For this one, I'll only add the nosync policy and remove the changes in kmeans_balanced::predict. Sound good?

If we aren't going to be using this nosync policy, I'd like to avoid merging changes just for the sake of merging changes.

abc99lr:
Closing.

cjnolet:
Most functions in RAFT are assumed to be async, so I suspect we could probably scrape through all of the places we use the `thrust_policy` and replace them with `nosync_policy`. @abc99lr I'm not against merging this just for that reason alone, but would you mind creating an issue for the above so that we don't lose sight of it? Just want to avoid this getting stale and remaining unused for the non-foreseeable future.

abc99lr:
> If we aren't going to be using this nosync policy, I'd like to avoid merging changes just for the sake of merging changes.

Understand. But I do think RAFT should provide this functionality to developers and tell people to use nosync policy when could. Otherwise, there would be many unnecessary syncs introduced.

abc99lr:
Talked with @cjnolet, we think it's worth to try using `nosync_policy` by default. Going to create another PR for that

PR #2295, opened 2024-05-07
[FEA] Split Bitset code

lowener:
Splitting the Bitset code into a `.hpp` and `.cuh` header allow the declaration of `ann::search_with_filters` with a `bitset_filter` to be made without using CUDA code in headers.
This is useful for cuVS where the headers `ivf_pq.hpp` need to declare `search_with_filter` and stay CUDA-free.


cjnolet (approved):
Very clean!

PR #2299, opened 2024-05-08
Add VERSION to raft-ann-bench package

KyleFromNVIDIA:
https://github.com/rapidsai/raft/pull/2285 added dynamic versioning to `pyproject.toml` for `raft-ann-bench`, but did not properly copy in the `VERSION` file. Copy it in with `tool.setuptools.package-data`.


divyegala (approved):

bdice (approved):
Do we need to add a test that ensures `raft-ann-bench` packages are importable? It seems the previous PR #2285 should've failed CI.

KyleFromNVIDIA:
Indeed, I was wondering why CI passed - simply building the wheel fails.

PR #2300, opened 2024-05-08
Revert "Add `compile-library` by default on pylibraft build"

vyasr:
Partial reversion of rapidsai/raft#2090

The unintended consequence of this PR was to cause conda builds of pylibraft to rebuild libraft. As a result, since mid-24.04 we've seen the following: pylibraft conda builds have gone from 15-30 mins to 75-90 mins; pylibraft conda packages are fully repackaging libraft; and therefore, pylibraft conda packages have gone from ~2-3MB to ~300MB. A future PR may attempt to improve the developer experience again, at which point we should keep an eye out for similar regressions.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -305,7 +305,7 @@ if hasArg --allgpuarch; then
     BUILD_ALL_GPU_ARCH=1
 fi
 
-if hasArg --compile-lib || hasArg pylibraft || (( ${NUMARGS} == 0 )); then
+if hasArg --compile-lib || (( ${NUMARGS} == 0 )); then
     COMPILE_LIBRARY=ON
     CMAKE_TARGET="${CMAKE_TARGET};raft_lib"
 fi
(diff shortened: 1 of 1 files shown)


cpp/template/cmake/thirdparty/get_raft.cmake
```diff
@@ -33,12 +33,6 @@ function(find_and_configure_raft)
     #-----------------------------------------------------
     # Invoke CPM find_package()
     #-----------------------------------------------------
-    # Since the RAFT_NVTX option is used by targets generated by
-    # find_package(RAFT_NVTX) and when building from source we want to
-    # make `RAFT_NVTX` a  cache variable so we get consistent
-    # behavior
-    #
-    set(RAFT_NVTX ${PKG_ENABLE_NVTX} CACHE BOOL "Enable raft nvtx logging" FORCE)
```
robertmaynard, line 41: I think we still want these changes around `nvtx` so that we keep consistent behavior. Can we unrevert this part

vyasr: Good call, done.

vyasr:
The builds show that conda Python package build times are back down to 15-30 mins and the Python conda tarballs have shrunk by a few hundred MB.

raydouglass (approved):

robertmaynard (approved):

PR #2301, opened 2024-05-09
add gpuarch arg to allow for setting specific gpu architectures

jperez999:
This PR aims to add a new argument to the build script. `--gpuarch` gives the user the ability to set the desired build architectures. Currently there are only 2 options available to build script users, NATIVE and RAPIDS. However, NATIVE builds can fail if users have multiple GPUs that are not all the same architecture on their systems. In order to prevent this causing a failure, I added this argument that will override and set RAFT_CMAKE_CUDA_ARCHITECTURES to the values set by the user. 

Examples of usage:
`--gpuarch="70;89"` 

Values are expected to be integers, separated by semicolons `;`.


robertmaynard:
Please file an issue against https://github.com/rapidsai/rapids-cmake that `NATIVE` doesn't work properly with multiple GPU output, since it is designed to do so.

This change isn't needed once we correct `NATIVE` to work as intended.

jperez999:
I am closing in lieu of a ticket filed against rapids-cmake directly https://github.com/rapidsai/rapids-cmake/issues/605 as @robertmaynard suggested.

PR #2302, opened 2024-05-09
Make thrust nosync execution policy the default thrust policy

abc99lr:
Testing. Do not merge. 

Based on the discussions from https://github.com/rapidsai/raft/pull/2293, it's a good idea to test if we could use nosync thrust calls by default. This PR changes the current `rmm::exec_policy` to its async version `rmm::exec_policy_nosync`.

### DIFF
--- a/cpp/include/raft/core/device_resources.hpp
+++ b/cpp/include/raft/core/device_resources.hpp
@@ -121,7 +121,7 @@ class device_resources : public resources {
 
   cusparseHandle_t get_cusparse_handle() const { return resource::get_cusparse_handle(*this); }
 
-  rmm::exec_policy& get_thrust_policy() const { return resource::get_thrust_policy(*this); }
+  rmm::exec_policy_nosync& get_thrust_policy() const { return resource::get_thrust_policy(*this); }
 
   /**
    * @brief synchronize a stream on the current container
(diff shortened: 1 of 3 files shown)


abc99lr:
Tagging @cjnolet to enable testing

cjnolet:
@abc99lr for future reference, you should be in the rapids org and the raft-triage group. The reason CI isn't automatically starting on this PR is because you haven't signed your commits. This is something you should be able to set up once and it'll "just work" on your dev system going forward. [Here's](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) the instructions to do this.

abc99lr:
It seems this PR fails `pr / conda-cpp-tests / tests (amd64, 3.11, 12.2.2, ubuntu22.04, v100, latest) (push)`. But I checked the log and  all the status show success. Should we rerun the CI? Or maybe I missed something? 

Here is the link to the log. 
https://github.com/rapidsai/raft/actions/runs/9024355165/job/24800696954?pr=2302

cjnolet:
@abc99lr yeah it's weird indeed. i"m rerunnig that test checker. It's possible it could have been a chance occurrence.

cjnolet (approved):

PR #2304, opened 2024-05-09
[REVIEW] Adjust UCX dependencies

pentschev:
Add `distributed-ucxx` as a required dependency to `raft-dask` and remove direct dependencies to `ucx`/`ucx-proc` in favor of transitive installation from `ucx-py`/`distributed-ucxx`.


vyasr (approved):

raydouglass (commented):
Looks like the conda recipe might need to be updated as well?

https://github.com/rapidsai/raft/blob/branch-24.06/conda/recipes/raft-dask/meta.yaml#L76-L77

raydouglass (approved):

PR #2310, opened 2024-05-14
Remove the shared state and the mutex from NVTX internals

achirkin:
Until now, raft has stored a map of NVTX colors (annotation -> color) to avoid using the same color for different annotations and keep using the same color for the same annotations. This map is a shared state.
During an extensive ANN_BENCH throughput testing it has turned out that the mutex guarding the map can sometimes become a bottleneck when the number of concurrent threads is really large (>~ 256). This PR replaces the unordered map and the mutex guarding it with a deterministic hash value of the annotation instead (which is stateless).

**Pros:**
  - No shared state, no mutexes.
  - Assigns the same colors to the same annotations across program runs.
 
**Cons:**
  - Sometimes different annotations can have the same color (hash collisions).

### DIFF
--- a/cpp/include/raft/core/detail/nvtx.hpp
+++ b/cpp/include/raft/core/detail/nvtx.hpp
@@ -105,14 +105,12 @@ inline auto hsv2rgb(float h, float s, float v) -> uint32_t
 /**
  * @brief Helper method to generate 'visually distinct' colors.
  * Inspired from https://martin.ankerl.com/2009/12/09/how-to-create-random-colors-programmatically/
- * However, if an associated tag is passed, it will look up in its history for
- * any generated color against this tag and if found, just returns it, else
- * generates a new color, assigns a tag to it and stores it for future usage.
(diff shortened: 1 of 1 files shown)


cpp/include/raft/core/detail/nvtx.hpp
```diff
@@ -121,20 +117,12 @@ inline auto hsv2rgb(float h, float s, float v) -> uint32_t
  */
```
tfeher, line 117: Could you update the docstring? There is no more lookup.

tfeher (changes requested):
Thanks Artem for the PR, looks good, just a small update in docstring is missing!

cpp/include/raft/core/detail/nvtx.hpp
achirkin: Sure, thanks for noting this!

tfeher (approved):
Thanks for the update, LGTM!

cjnolet (approved):

PR #2313, opened 2024-05-14
ANN_BENCH: split instances of RaftCagra into multiple files

achirkin:
Split instances of RaftCagra into multiple files to compile them in parallel and thus reduce the total benchmark compile time.

PR #2316, opened 2024-05-14
[WIP] Persistent CAGRA kernel

achirkin:
An experimental version of the single-cta CAGRA kernel that run persistently while allowing many CPU threads submit the queries in small batches very efficiently.

<p align="center">
  CAGRA throughput @ recall = 0 976
  CAGRA Latency @ recall = 0 976
</p>

## API

In the current implementation, the public API does not change. An extra parameter `persistent` is added to the `ann::cagra::search_params` (only valid when `algo == SINGLE_CTA`).
The persistent kernel is managed by a global runner object in a `shared_ptr`; the first CPU thread to call the kernel spawns the runner, subsequent calls/threads only update a global "heartbeat" atomic variable (`runner_base_t::last_touch`). When there's no hearbeat in the last few seconds (`kLiveInterval`), the runner shuts down the kernel and cleans up the associated resources.

An alternative solution would be to control the kernel explicitly, in a client-server style. This would be more controllable, but would require significant re-thinking of the RAFT/cuVS API.

### lightweight_uvector

RMM memory resources and device buffers are not zero-cost, even when the allocation size is zero (a common pattern for conditionally-used buffers). They do at least couple `cudaGetDevice` calls during initialization. Normally, the overhead of this is negligible. However, when the number of concurrent threads is large (hundreds of threads), any CUDA call can become a bottleneck due to a single mutex guarding a critical section somewhere in the driver.

To workaround this, I introduce a `lightweight_uvector` in `/detail/cagra/search_plan.cuh` for several buffers used in CAGRA kernels. This is a stripped "multi-device-unsafe" version of `rmm::uvector`: it does not check during resize/destruction whether the current device has changed since construction.
We may consider putting this in a common folder to use across other RAFT/cuVS algorithms.

### Shared resource queues / ring buffers

`resource_queue_t` is an atomic counter-based ring buffer used to distribute the worker resources (CTAs) and pre-allocated job descriptors across CPU I/O threads.
We may consider putting this in a common public namespace in raft if we envision more uses for it.

### Persistent runner structs

`launcher_t` and `persistent_runner_base_t` look like they could be abstracted from the cagra kernel and re-used in other algos. The code in its current state, however, is not ready for this.

### Other code changes (solved)

This depends on (includes all the changes from):
  - https://github.com/rapidsai/raft/pull/2310
  - https://github.com/rapidsai/raft/pull/2311
  - https://github.com/rapidsai/raft/pull/2312
  - https://github.com/rapidsai/raft/pull/2313
  - https://github.com/rapidsai/raft/pull/2314
  - https://github.com/rapidsai/raft/pull/2315


cpp/bench/ann/src/common/ann_types.hpp
```diff
@@ -73,7 +76,7 @@ struct AlgoProperty {
 
 class AnnBase {
  public:
-  using index_type = size_t;
+  using index_type = uint32_t;
```
tfeher, line 79: Do we want to change this for all benchmarks?

cpp/include/raft/neighbors/cagra_types.hpp
```diff
@@ -124,6 +124,8 @@ struct search_params : ann::search_params {
   uint32_t num_random_samplings = 1;
   /** Bit mask used for initial random seed node selection. */
   uint64_t rand_xor_mask = 0x128394;
+  /** Whether to use the persistent version of the kernel (only SINGLE_CTA is supported a.t.m.) */
+  bool persistent = false;
```
tfeher, line 128: Is this the only parameter that needs to be controlled? What about the temporary buffer size (queue for queries, or max number of threads)?

tfeher (commented):
Thanks Artem for this work! Just a few initial comments here. 

For the API discussion, one question is how to start / stop the kernel. The current approach is nice, just one bool flag in search params, and rest is done automatically. I believe this would not preclude adding explicit start/stop methods if we deem that necessary.

Other question is what parameters do we need to control for the persistent kernel (apart from whether or not to use persistent kernel)? Eg.: `max num threads`, `kLiveInterval`? If we want to provide control for these, then we could have additional fields for `search_params`, but we could also think of them as a property of the `index`, since the long running kernel is associated with the index.

cpp/bench/ann/src/common/ann_types.hpp
achirkin: That was a temporary change for testing, I've removed that in cuVS PR (cuVS doesn't support uint32_t indexes in the refinement step)

cpp/include/raft/neighbors/cagra_types.hpp
achirkin: The buffer sizes are compile-time constants in the current design. I'd like to move some of the new constants to the parameters, but that conflicts with the implicit mechanics of running the kernel. They are not search parameters, but the "runner" parameters and should not be changed across calls to the kernel.
So this brings up again the question whether we want to have implicit vs explicit kernel runner.

achirkin:
Closing this in favor of https://github.com/rapidsai/cuvs/pull/215

PR #2323, opened 2024-05-16
Support building faiss main statically

robertmaynard:
Cleans up a collection of anti-patterns in the raft CMake code while also enabling building faiss from latest `main`


cjnolet:
cc @tfeher and @achirkin for awareness. This will probably need to be merged forward into your migration of raft-ann-bench to cuVS.

PR #2332, opened 2024-05-22
Add workaround for syevd in CUDA 12.0

lowener:
Bug 4580093 of cusolver is causing an issue with `cusolverDnXsyevd`. This bug has been seen in https://github.com/rapidsai/cuml/issues/5555 and impact PCA and Linear Regression with CUDA 12.0+. Setting the stream to a different value than `cudaStreamPerThread` seems to solve it as a workaround.

### DIFF
--- a/cpp/include/raft/linalg/detail/eig.cuh
+++ b/cpp/include/raft/linalg/detail/eig.cuh
@@ -91,6 +91,13 @@ void eigDC(raft::resources const& handle,
 #if CUDART_VERSION < 11010
   eigDC_legacy(handle, in, n_rows, n_cols, eig_vectors, eig_vals, stream);
 #else
+
+  // Use a new stream instead of `cudaStreamPerThread` to avoid cusolver bug # 4580093.
+  cudaStream_t stream_new;
+  cudaEvent_t sync_event;
+  RAFT_CUDA_TRY(cudaStreamCreate(&stream_new));
+  RAFT_CUDA_TRY(cudaEventCreate(&sync_event));
+
   cusolverDnHandle_t cusolverH = resource::get_cusolver_dn_handle(handle);
 
   cusolverDnParams_t dn_params = nullptr;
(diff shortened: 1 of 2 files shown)


cpp/include/raft/linalg/detail/eig.cuh
```diff
@@ -91,6 +91,15 @@ void eigDC(raft::resources const& handle,
 #if CUDART_VERSION < 11010
   eigDC_legacy(handle, in, n_rows, n_cols, eig_vectors, eig_vals, stream);
 #else
+
+  // Use a new stream instead of `cudaStreamPerThread` to avoid cusolver bug # 4580093.
```
tfeher, line 95: IIUC, the cusolver bug is solved in cuda toolkit 12.4.1.003. It would be great if we apply this workaround only when we use an earlier cuda version.

tfeher (changes requested):
Thanks @lowener for the fix. While the fix looks good, I think we should only apply it for the affected CTX versions.

cpp/include/raft/linalg/detail/eig.cuh
```diff
@@ -91,6 +91,15 @@ void eigDC(raft::resources const& handle,
...
+  cudaStream_t stream_new;
+  cudaEvent_t sync_event;
+  RAFT_CUDA_TRY(cudaStreamCreate(&stream_new));
+  RAFT_CUDA_TRY(cudaEventCreate(&sync_event));
```
achirkin, line 99: I think, in this case it would be justified to use the [resource::detail::get_cuda_stream_sync_event](https://github.com/rapidsai/raft/blob/branch-24.08/cpp/include/raft/core/resource/detail/stream_sync_event.hpp#L43) instead of manually managing the resource.
This event resource is normally intended for synchronization between the streams in the resource pool, but as long as you don't use the stream pool resource at the same time, it's ok to reuse the event.

You can also use the stream from the stream pool resource, but there is a small problem with it, that raft/rmm would create 16 streams by default instead of one :)

cpp/include/raft/linalg/detail/eig.cuh
```diff
@@ -91,6 +91,15 @@ void eigDC(raft::resources const& handle,
 #if CUDART_VERSION < 11010
   eigDC_legacy(handle, in, n_rows, n_cols, eig_vectors, eig_vals, stream);
 #else
+
+  // Use a new stream instead of `cudaStreamPerThread` to avoid cusolver bug # 4580093.
+  cudaStream_t stream_new;
+  cudaEvent_t sync_event;
+  RAFT_CUDA_TRY(cudaStreamCreate(&stream_new));
```
achirkin, line 98: Maybe use the `rmm::cuda_stream` instead (or a raft pool see below)?
We'd better never manually create/destroy resources: in an unlikely event someone catches the exception raised between creation of the stream here and its destruction at the end of the function, they'd get the leaked resource (cuda stream handle).

achirkin (changes requested):
Couple notes on manually creating streams / events.

tfeher (approved):
Thanks @lowener for the update. LGTM.

achirkin (approved):
Thanks for the update, LGTM!

cpp/include/raft/linalg/detail/eig.cuh
cjnolet: Please note that we shouldn't be calling detail APIs in any namdespaces outside of the immediate namespace where those detail APIs reside. Please expose this function if it's going to be used publicly.

cjnolet (approved):

PR #2342, opened 2024-05-24
Change --config-setting to --config-settings

KyleFromNVIDIA:
pip install uses --config-settings as its argument.


jameslamb (approved):
Thanks!

I think it'd be helpful to link to the context for why we're doing this, so anyone else reviewing it (or finding it from search in the future) understands the purpose.

I'll add that here: https://github.com/rapidsai/cuml/pull/5804#discussion_r1614064596

PR #2345, opened 2024-05-30
[REVIEW] Enable distance return for NN Descent

jinsolp:
- Enable NN Descent to return the distances array as well (previously only returning indices array)
  - Added a `return_distances` flag in `index_params`. When set to 1 (true), allocates a distance array to return.
- Test for checking distances recall compared to naive knn


cpp/include/raft/neighbors/nn_descent_types.hpp
```diff
@@ -85,6 +87,7 @@ struct index : ann::index {
       res_{res},
       metric_{raft::distance::DistanceType::L2Expanded},
       graph_{raft::make_host_matrix<IdxT, int64_t, row_major>(n_rows, n_cols)},
+      distances_{raft::make_host_matrix<DistData_t, int64_t, row_major>(n_rows, n_cols)},
```
divyegala, line 90: This is an eager allocation. Instead, the allocation should only occur if `return_distances=true`

cpp/include/raft/neighbors/nn_descent_types.hpp
```diff
@@ -105,6 +108,7 @@ struct index : ann::index {
       res_{res},
       metric_{raft::distance::DistanceType::L2Expanded},
       graph_{raft::make_host_matrix<IdxT, int64_t, row_major>(0, 0)},
+      distances_{raft::make_host_matrix<DistData_t, int64_t, row_major>(0, 0)},
       graph_view_{graph_view}
```
divyegala, line 112: Modify this constructor to allow the user to provide a pre-allocated distance matrix, just like how the constructor allows for the graph. The signature can be of the form, `std::optional<raft::host_matrix_view<...>>`.

Also, you'll need a function called `allocate_distances` or something so that when `return_distances == true`, you can allocate memory for the distances.

cpp/include/raft/neighbors/nn_descent_types.hpp
```diff
@@ -51,6 +52,7 @@ struct index_params : ann::index_params {
   size_t intermediate_graph_degree = 128;     // Degree of input graph for pruning.
   size_t max_iterations            = 20;      // Number of nn-descent iterations.
   float termination_threshold      = 0.0001;  // Termination threshold of nn-descent.
+  int return_distances             = 0;       // return distances if 1
```
divyegala, line 55:
```suggestion
  bool return_distances             = false;       // return distances if true
```

divyegala (changes requested):

cpp/include/raft/neighbors/nn_descent_types.hpp
jinsolp: I'm not sure if I understood correctly..! Do you mean
- there should be a member function in `index` struct that allocates the `host_matrix` for `distances_` 
- `distances_` should be declared as a `std::optional` type

divyegala: Yes, that is indeed what I am suggesting.

jinsolp: Instead of making another function, I used the struct constructor to allocate the `distances_` (and `distance_view_`) depending on `return_distances` flag

cpp/include/raft/neighbors/nn_descent_types.hpp
```diff
@@ -133,6 +148,13 @@ struct index : ann::index {
     return graph_view_;
   }
 
+  [[nodiscard]] inline auto distances() noexcept -> host_matrix_view<DistData_t, int64_t, row_major>
+  {
+    assert(return_distances_);
+    assert(distances_view_.has_value());
+    return distances_view_.value();
```
jinsolp, line 155: It seems like the test is failing because this part is not using `value_or` to refer to `distances_view_`.
However, I'm not sure how I should use `value_or` here since there is no default value to return in this case (also cannot return `nullptr`).
Any help would be appreciated : )!

cpp/include/raft/neighbors/detail/nn_descent.cuh
```diff
@@ -1405,6 +1422,12 @@ void build(raft::resources const& res,
   auto int_graph = raft::make_host_matrix<int, int64_t, row_major>(
     dataset.extent(0), static_cast<int64_t>(extended_graph_degree));
 
+  auto distances_graph = raft::make_host_matrix<DistData_t, int64_t, row_major>(0, 0);
```
divyegala, line 1425: Why is this extra allocation needed? Just copy from `h_dists` directly to the distance array allocated in the index.

cpp/include/raft/neighbors/nn_descent_types.hpp
```diff
@@ -25,7 +25,10 @@
 #include <raft/core/resources.hpp>
 #include <raft/distance/distance_types.hpp>
 
+#include <optional>
+
 namespace raft::neighbors::experimental::nn_descent {
+using DistData_t = float;
```
divyegala, line 31: This should be a template parameter on the index class

cpp/include/raft/neighbors/nn_descent_types.hpp
```diff
@@ -100,12 +109,18 @@ struct index : ann::index {
    * @param graph_view raft::host_matrix_view<IdxT, int64_t, raft::row_major> for storing knn-graph
    */
   index(raft::resources const& res,
-        raft::host_matrix_view<IdxT, int64_t, raft::row_major> graph_view)
+        raft::host_matrix_view<IdxT, int64_t, raft::row_major> graph_view,
+        std::optional<raft::host_matrix_view<DistData_t, int64_t, raft::row_major>> distances_view =
```
divyegala, line 113: @cjnolet and I discussed this offline, let's keep the distance matrix as `raft::device_matrix_view`

cpp/include/raft/neighbors/nn_descent_types.hpp
divyegala: You could default initialize the member below and that should fix this:
```std::optional<raft::host_matrix_view<DistData_t, int64_t, row_major>> distances_view_; = std::nullopt```

File: cpp/include/raft/neighbors/detail/nn_descent.cuh [hunk 8]
```diff
@@ -1405,6 +1422,12 @@ void build(raft::resources const& res,
   auto int_graph = raft::make_host_matrix<int, int64_t, row_major>(
```
divyegala, line 1422: I just realized that `int_graph` is also an excessive allocation... you could just copy the values directly from `graph_shrink_buffer` to the indices array allocated in the index object. You would need an extra template parameter on `GNND::build()` function to make that work. The NN Descent impl can only support `int` but RAFT aims to support `uint32_t`, `int8` and `uint8_t`. Which is fine, let NN Descent solve the problem in `int` and then copy from `graph_shrink_buffer` to the output indices array from the index in the correct type format.

cpp/include/raft/neighbors/nn_descent_types.hpp
jinsolp: Thank you! ☺️

cpp/include/raft/neighbors/detail/nn_descent.cuh
jinsolp: After looking into the code a bit, I think `int_graph` is needed.

As soon as `GNND::build()` is called, the given `output_graph` is casted like 
`graph_.h_graph      = (InternalID_t<Index_t>*)output_graph` where `InternalID_t<Index_t>*` ends up just being `int*` type. 

`InternalID_t<Index_t>` has to be an `int`(or at least a signed value) because of the way they keep track of old and new values (using the msb).

So passing the `idx.graph().data_handle()` directly to `GNND::build()` without allocating `int_graph` will end up having to change the logic inside `InternalID_t<Index_t>` because `idx.graph().data_handle()` could have types other than `int` (e.g. `uint32_t`).

divyegala: Ah you are right, I see there is a comment for that as well https://github.com/rapidsai/raft/blob/8fbce933bff14deee582eab345fb768f35cf7f27/cpp/include/raft/neighbors/detail/nn_descent.cuh#L994-L995

cpp/include/raft/neighbors/nn_descent_types.hpp
```diff
@@ -98,15 +110,25 @@ struct index : ann::index {
...
       res_{res},
       metric_{raft::distance::DistanceType::L2Expanded},
       graph_{raft::make_host_matrix<IdxT, int64_t, row_major>(0, 0)},
-      graph_view_{graph_view}
+      distances_{raft::make_device_matrix<T, int64_t>(res_, 0, 0)},
+      graph_view_{graph_view},
+      distances_view_(distances_view),
+      return_distances_(return_distances)
   {
+    if (!distances_view.has_value()) { distances_view_ = distances_.value().view(); }
```
divyegala, line 131: This is redundant
```suggestion

```

cpp/include/raft/neighbors/nn_descent_types.hpp
```diff
@@ -133,6 +155,16 @@ struct index : ann::index {
...
 
+  /** neighborhood graph distances [size, graph-degree] */
+  [[nodiscard]] inline auto distances() noexcept -> device_matrix_view<T, int64_t, row_major>
+  {
+    if (distances_view_.has_value()) {
+      return distances_view_.value();
+    } else {
+      return raft::make_device_matrix<T, int64_t>(res_, 0, 0).view();
+    }
+  }
```
divyegala, line 166: Let's just return the optional directly here
```suggestion
  [[nodiscard]] inline auto distances() noexcept -> std::optional<device_matrix_view<T, int64_t, row_major>>
  {
    return distances_view_;
  }
```

jinsolp: Wouldn't this require callers to this `distances()` function to take care of the optional?? (Like calling `index.distances().value()` instead of just calling `index.distances()`)?

divyegala: Yes, it would. But IMO it is the correct approach because the distances are actually optional in this index, and returning an optional conveys that.

cpp/include/raft/neighbors/detail/nn_descent.cuh
```diff
@@ -1410,10 +1430,21 @@ void build(raft::resources const& res,
...
+
+  if (idx.distances().has_value()) {
+    nnd.build(dataset.data_handle(),
+              dataset.extent(0),
+              int_graph.data_handle(),
+              params.return_distances,
+              idx.distances().value().data_handle());
+  } else {
+    RAFT_EXPECTS(
+      false, "Distance view not allocated. Using NN Descent requires return_distances to be true");
```
divyegala, line 1446: The else condition should only fail if user wants us to return distances but has not allocated memory, right?
```suggestion
    RAFT_EXPECTS(params.return_distances, "Distance view not allocated. Using NN Descent requires return_distances to be true");
```

cpp/include/raft/neighbors/nn_descent.cuh
```diff
@@ -56,9 +56,9 @@ namespace raft::neighbors::experimental::nn_descent {
  * @return index<IdxT> index containing all-neighbors knn graph in host memory
  */
 template <typename T, typename IdxT = uint32_t>
-index<IdxT> build(raft::resources const& res,
-                  index_params const& params,
-                  raft::device_matrix_view<const T, int64_t, row_major> dataset)
+index<detail::DistData_t, IdxT> build(raft::resources const& res,
```
divyegala, line 59: The signature looks incorrect to me. If `T = double` then the index is also storing distances in `double`, right?
```suggestion
index<T, IdxT> build(raft::resources const& res,
```

cpp/include/raft/neighbors/nn_descent.cuh
```diff
@@ -97,7 +97,7 @@ template <typename T, typename IdxT = uint32_t>
 void build(raft::resources const& res,
            index_params const& params,
            raft::device_matrix_view<const T, int64_t, row_major> dataset,
-           index<IdxT>& idx)
+           index<detail::DistData_t, IdxT>& idx)
```
divyegala, line 100: Same comment as above

cpp/include/raft/neighbors/nn_descent.cuh
```diff
@@ -130,9 +130,9 @@ void build(raft::resources const& res,
  * @return index<IdxT> index containing all-neighbors knn graph in host memory
  */
 template <typename T, typename IdxT = uint32_t>
-index<IdxT> build(raft::resources const& res,
-                  index_params const& params,
-                  raft::host_matrix_view<const T, int64_t, row_major> dataset)
+index<detail::DistData_t, IdxT> build(raft::resources const& res,
```
divyegala, line 133: Same as above

cpp/include/raft/neighbors/nn_descent.cuh
```diff
@@ -171,7 +171,7 @@ template <typename T, typename IdxT = uint32_t>
 void build(raft::resources const& res,
            index_params const& params,
            raft::host_matrix_view<const T, int64_t, row_major> dataset,
-           index<IdxT>& idx)
+           index<detail::DistData_t, IdxT>& idx)
```
divyegala, line 174: Same as above

cpp/include/raft/neighbors/detail/nn_descent.cuh
jinsolp: you are right!

cpp/include/raft/neighbors/nn_descent.cuh
jinsolp: But in`nn_descent.cuh` , the build is declared as `index<DistData_t, IdxT> build`, where `DistData_t` is declared as `float`.
So changing it like this results in not being able to call `build` when the data type is something other that `float`.
Should I change `nn_descent.cuh` so that it can take the distance type as a template parameter?

divyegala: You can just copy the data from float to type T. Unfortunately, you'll not be able to use raft::copy anymore and you'll have to use raft::linalg::unary_op most likely. 

That way, you don't have to change anything in detail/nn_descent.cuh and you can still solve NN Descent in float, and then copy to the expected type T of the index.

divyegala (approved):
Pre-approving based on offline discussion

cpp/include/raft/neighbors/nn_descent.cuh
jinsolp: ==Talked about this online and decided to keep it as float==

cpp/include/raft/neighbors/detail/nn_descent.cuh
```diff
@@ -1435,12 +1435,19 @@ void build(raft::resources const& res,
...
+  if (idx.distances().has_value() || !params.return_distances) {
+    nnd.build(dataset.data_handle(),
+              dataset.extent(0),
+              int_graph.data_handle(),
+              params.return_distances,
+              idx.distances()
+                .value_or(raft::make_device_matrix<float, int64_t>(res, 0, 0).view())
+                .data_handle());
+  } else {
+    RAFT_EXPECTS(false,
```
divyegala, line 1447: `RAFT_EXPECTS(!params.return_distances, "...")`

jinsolp:
This is no longer a breaking change

PR #2346, opened 2024-05-30
[Opt] Expose the `detail::popc` as public API

rhdong:
- For resolving the issue of cuVS: https://github.com/rapidsai/cuvs/issues/158


cpp/include/raft/core/popc.hpp
```diff
@@ -0,0 +1,41 @@
...
+ * @tparam value_t the value type of the vector.
+ * @tparam index_t the index type of vector and scalar.
+ *
+ * @param[in] res RAFT handle for managing expensive resources
+ * @param[in] values Device vector view containing the values to be processed.
+ * @param[in] max_len Host scalar view to store the Maximum number of bits to count.
+ * @param[out] counter Device scalar view to store the number of bits that are set to 1.
+ */
+template <typename value_t, typename index_t>
+void popc(const raft::resources& res,
```
cjnolet, line 33: To be quite honest, I'm not sure this is something that belongs in `raft::core`. 

Also, I'm a little confused by the max_len- can you explain why this additional argument is needed? This API seems a little implementation-specific, which makes me wonder if it should be more strongly coupled to the `bitset`/`bitmap`.

rhdong: The original wrap of `popc` is here: https://github.com/rapidsai/raft/blob/branch-24.08/cpp/include/raft/core/detail/mdspan_util.cuh#L48. But it can only calc nonzero bits in one int32/int64, while this one can calculate on an integer array. Would you have some recommendations on the `namespace`? Maybe the `raft::linalg` is another choice.

The `max_len` should be required for indicating the possible bits should be processed, for example:  100 bits need be calculated, and at least 4 `uint32_t` to present them, in this situation, `max_len` or like parameter is needed for telling the popc to stop on the right bits..

rhdong: This API is mainly for clean code. The `detail::popc` includes at least 10 lines of code because it has to process the corner cases. If we were using the original code, we would have to copy that block of code anywhere, that is ugly.

cjnolet: Since this is a computational function, I don't think it belongs in `raft/core`. I question whether or not it actually belongs in `raft/matrix` or in `raft/utils`, though. Hmm.

rhdong: I prefer `raft/utils` as it seems more reasonable.

cjnolet (approved):

PR #2347, opened 2024-05-31
Fix raft-dask dependency on pylibraft.

bdice:
This PR fixes an issue where `raft-dask-cu12` wheels had a dependency on both `pylibraft-cu12` and `pylibraft`. The latter should not exist.


jameslamb (approved):

jameslamb:
I've merged these changes into #2349 (since that also changes CI to ensure it fails if we didn't catch all the dependency issues). Closing this.

PR #2351, opened 2024-06-03
Use default init in reduction

akifcorduk:
During reduction in device code (reduction.cuh), the value assigned in the residual threads during last stage are zero initilized. However, if we want to reduce some custom type, it might not have the appropriate constructor. Thus, this PR makes the change so that we call the default constructor for the residual values.


dantegd (approved):
change lgtm

tfeher (approved):
LGTM.

PR #2358, opened 2024-06-12
[DO NOT MERGE] Test CCCL version bump

sleeepyjack:
This PR tests an upcoming CCCL version bump (https://github.com/rapidsai/rapids-cmake/pull/631) and should not be merged.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -11,6 +11,9 @@
 # or implied. See the License for the specific language governing permissions and limitations under
 # the License.
 cmake_minimum_required(VERSION 3.26.4 FATAL_ERROR)
+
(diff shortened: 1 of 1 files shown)


sleeepyjack:
One of the tests is running into an error: https://github.com/rapidsai/raft/actions/runs/9475150045/job/26109283214?pr=2358#step:7:824

We ran into a similar issue yesterday with cuco when bumping the CCCL version to 2.5.0. Turns out that the culprit was a sticky CUDA error that resurfaced during a downstream Thrust/cub call.

I'm not very familiar with this codebase. Could someone help investigate this issue?

bdice:
I can reproduce the issue locally in a devcontainer by running `build-raft-cpp && raft/cpp/build/latest/gtests/CORE_TEST -V`. I won't have much time to look deeper today but at least we have a reproducer outside of CI.

sleeepyjack:
Closing this PR since all checks have passed

PR #2361, opened 2024-06-12
Fix a bug where the wrong API is used to free the memory

PointKernel:
This PR fixes a bug uncovered by CCCL version bump #2358


PointKernel:
The style check failure seems irrelevant to the current PR.

PR #2362, opened 2024-06-14
[FEA] add the support of `masked_matmul`

rhdong:
https://github.com/rapidsai/raft/issues/2336


cpp/include/raft/sparse/linalg/masked_matmul.hpp
```diff
@@ -0,0 +1,64 @@
...
+#pragma once
+
+#include <raft/sparse/linalg/detail/masked_matmul.cuh>
+
+namespace raft {
+namespace sparse {
+namespace linalg {
+
+/**
+ * @brief Performs a masked multiplication of dense matrices A and B, followed by an element-wise
```
cjnolet, line 25: PR looks great, thanks @rhdong! Only one thing I can see is missing- please add this to the docs.

rhdong: Done!

cjnolet (approved):

PR #2365, opened 2024-06-18
Fix cagra graph opt bug

enp1s0:
backport of https://github.com/rapidsai/cuvs/pull/192

There is a bug in the current CAGRA graph rank-based neighbor reordering process. A low recall or illegal memory access can occur if there are many detourable nodes from a node to its neighbors, e.g. there is a small subgraph in the initial kNN graph. This PR fixes this.


tfeher (approved):
LGTM, thanks @enp1s0!

PR #2370, opened 2024-06-28
Binarize Dice Distance for Dense Inputs

aamijar:
Instead of binarizing in `pairwise_distances()` in cuml, do it in the distance function in raft.


cjnolet (approved):
LGTM!

lowener (approved):

Issue #2368, opened 2024-06-26
[BUG] `bitmap_view::set` not work


rhdong:
When call bitmap_view::set(int row, int col, bool new_values);
```
/raft/cpp/include/raft/core/bitmap.cuh(42): error: argument of type "__nv_bool *" is incompatible with parameter of type "int"
    set(row * cols_ + col, &new_value);
                           ^
```
If you encounter this problem, you can use the [bitset API](https://github.com/rapidsai/raft/blob/branch-24.08/cpp/include/raft/core/bitset.hpp#L215) as a temporary alternative.

PR #2371, opened 2024-06-28
[Fix] bitmap set/test issue

rhdong:
This PR fixes a bug in the `bitmap_view`.

Closes https://github.com/rapidsai/raft/issues/2368

### DIFF
--- a/cpp/test/CMakeLists.txt
+++ b/cpp/test/CMakeLists.txt
@@ -113,6 +113,7 @@ if(BUILD_TESTS)
     NAME
     CORE_TEST
     PATH
+    core/bitmap.cu
     core/bitset.cu
     core/device_resources_manager.cpp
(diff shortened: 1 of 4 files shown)


cpp/test/CMakeLists.txt
```diff
@@ -308,24 +309,7 @@ if(BUILD_TESTS)
     NAME
     SPARSE_TEST
     PATH
-    sparse/add.cu
```
cjnolet, line 311: Hey James, why are we removing all of these tests?

cjnolet (changes requested):

cpp/test/CMakeLists.txt
rhdong: Just recover it( it's for temporary debugging).

cjnolet (approved):

PR #2375, opened 2024-07-05
skip CMake 3.30.0

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/80

Adds constraints to avoid pulling in CMake 3.30.0, for the reasons described in that issue.


jameslamb:
I'd pushed another commit because I thought that maybe using "skip" had led to CI not running.

But I think it's just because there's an ongoing incident with GitHub webhooks: https://www.githubstatus.com/incidents/5yx1d67vq9hg

> Webhooks is experiencing degraded availability.

bdice (approved):

PR #2377, opened 2024-07-11
Docstring and fix generate grountruth

Kh4L:
Fix docstring

### DIFF
--- a/python/raft-ann-bench/src/raft_ann_bench/generate_groundtruth/__main__.py
+++ b/python/raft-ann-bench/src/raft_ann_bench/generate_groundtruth/__main__.py
@@ -47,7 +47,7 @@ def choose_random_queries(dataset, n_queries):
     return dataset[query_idx, :]
 
 
-def calc_truth(dataset, queries, k, metric="sqeuclidean"):
+def calc_truth(dataset, queries, k, metric="euclidean"):
     handle = DeviceResources()
     n_samples = dataset.shape[0]
     n = 500000  # batch size for processing neighbors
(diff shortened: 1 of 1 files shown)


python/raft-ann-bench/src/raft_ann_bench/generate_groundtruth/__main__.py
```diff
@@ -96,18 +96,18 @@ def main():
         "The input and output files are in big-ann-benchmark's binary format.",
         epilog="""Example usage
     # With existing query file
-    python -m raft_ann_bench.generate_groundtruth --dataset /dataset/base.\
-fbin --output=groundtruth_dir --queries=/dataset/query.public.10K.fbin
+    python -m raft_ann_bench.generate_groundtruth --output=groundtruth_dir \
+--queries=/dataset/query.public.10K.fbin /dataset/base.fbin
```
divyegala, line 100: What has changed here? I think we would like to keep the --dataset argument

tfeher (changes requested):
Thanks @Kh4L for this PR! In the PR description you write that "sqeuclidean is not supported in raft-ann-bench". I think that is not correct, I have tested in 24.06 and 24.08, and it is supported in both cases. Therefore I would recommend to keep the default metric "sqeuclidean". The rest of the docstring change look good to me.

tfeher (approved):
Thanks Serge, LGTM!

cjnolet:
Benchmarking tool has been migrated over to cuVS. Closing

PR #2378, opened 2024-07-11
Build and test with CUDA 12.5.1

KyleFromNVIDIA:
This PR updates the latest CUDA build/test version 12.2.2 to 12.5.1.

Contributes to https://github.com/rapidsai/build-planning/issues/73


jameslamb (approved):
These places are technically not "missed", they are accurate ... but I do think we should update them to CUDA 12.5, to help encourage the use of newer CUDA versions.

https://github.com/rapidsai/raft/blob/ab5e12871e91803769c48c564fe009597de8386c/README.md#L283-L285

https://github.com/rapidsai/raft/blob/ab5e12871e91803769c48c564fe009597de8386c/README.md#L291-L293

jakirkham:
Do we want to update these?

https://github.com/rapidsai/raft/blob/c1ea4082b9b518b5593bb9c509fd254e2625d56b/conda/environments/bench_ann_cuda-120_arch-aarch64.yaml#L19

https://github.com/rapidsai/raft/blob/ab5e12871e91803769c48c564fe009597de8386c/conda/environments/bench_ann_cuda-120_arch-x86_64.yaml#L19

jameslamb:
> Edit: Noticed [this line in rapidsai/docker](https://github.com/rapidsai/docker/blob/a1e5d98caf1efa58c5886659ff950cfcd5b7dd7e/raft-ann-bench/gpu/Dockerfile#L3). Maybe we should leave these as-is for now

You can safely change that. That's setting a default for a build argument that's always overridden. I'm going to propose removing that default (precisely for this reason of not confusing people) once `rapidsai/docker` CI is unblocked.

jakirkham:
After discussing offline, we decided to punt on RAFT ANN for now. This is in part as we understand this to benchmark related and it may involve more than updating the versions (like rerunning the benchmark and looking for changes). Have raised an issue for follow up: https://github.com/rapidsai/raft/issues/2383

We can circle back once we know better how to proceed

jakirkham:
Thanks Kyle and James! 🙏

PR #2380, opened 2024-07-12
Use slicing kernel to copy distances inside NN Descent

jinsolp:
### Description
This make use of raft's slicing kernel within NN Descent build.
I found that my previous implementation was inefficient (merged in [this PR](https://github.com/rapidsai/raft/pull/2345)).

### Improvements
Time to call NN Descent `build()` with `return_distances=True` before and after using this kernel
| Dataset | Before |After|
| ------------- | ------------- |---|
| mnist (60000, 784)  | 1550ms  | 1020ms|
| sift (1M, 128)  | 11342ms  |5546ms|
| gist (1M, 960)  | 13508ms |9278ms|

### DIFF
--- a/cpp/include/raft/neighbors/detail/nn_descent.cuh
+++ b/cpp/include/raft/neighbors/detail/nn_descent.cuh
@@ -19,11 +19,15 @@
 #include "../nn_descent_types.hpp"
 
 #include <raft/core/device_mdarray.hpp>
+#include <raft/core/device_mdspan.hpp>
 #include <raft/core/error.hpp>
 #include <raft/core/host_mdarray.hpp>
+#include <raft/core/mdspan.hpp>
+#include <raft/core/mdspan_types.hpp>
 #include <raft/core/operators.hpp>
 #include <raft/core/resource/cuda_stream.hpp>
 #include <raft/core/resources.hpp>
+#include <raft/matrix/slice.cuh>
 #include <raft/neighbors/detail/cagra/device_common.hpp>
 #include <raft/spatial/knn/detail/ann_utils.cuh>
 #include <raft/util/arch.cuh>  // raft::util::arch::SM_*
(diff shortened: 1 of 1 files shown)


cpp/include/raft/neighbors/detail/nn_descent.cuh
```diff
@@ -1365,12 +1368,22 @@ void GNND<Data_t, Index_t, epilogue_op>::build(Data_t* data,
...
   if (return_distances) {
-    for (size_t i = 0; i < (size_t)nrow_; i++) {
-      raft::copy(output_distances + i * build_config_.output_graph_degree,
-                 graph_.h_dists.data_handle() + i * build_config_.node_degree,
-                 build_config_.output_graph_degree,
-                 raft::resource::get_cuda_stream(res));
-    }
+    auto graph_d_dists = raft::make_device_matrix<DistData_t, int64_t, raft::row_major>(
+      res, nrow_, build_config_.node_degree);
+    raft::copy(graph_d_dists.data_handle(),
```
lowener, line 1373: Can this `raft::copy` be used: https://github.com/rapidsai/raft/blob/branch-24.08/cpp/include/raft/core/copy.cuh#L57? It supports mdspans.

jinsolp: Thanks! didn't know we had that!

cjnolet (approved):
LGTM!

jinsolp:
Okay now the types are a bit messy, because `graph_.h_dists` has index type `size_t`, and the mdspan `raft::copy` complains if `graph_d_dists` index type is not the same.

But then `slice` complains about meaningless comparison with 0 if the index type is not signed, which makes me have to set it to `int64_t` when getting the const view of `graph_d_dists`.

lowener (approved):

Issue #2366, opened 2024-06-25
[FEA] Consolidate SUM reductions across raft


mfoerste4:
**Is your feature request related to a problem? Please describe.**
Brought up by https://github.com/rapidsai/raft/issues/2205 the [sum kernel](https://github.com/rapidsai/raft/blob/b863f1810fdb0679f7777b656486dbfc96470d2b/cpp/include/raft/stats/detail/sum.cuh#L51) has been adapted to account for underflow issues when adding large amounts of values. However, there are several other locations within the code where summations are done naively like [stdev/var](https://github.com/rapidsai/raft/blob/b863f1810fdb0679f7777b656486dbfc96470d2b/cpp/include/raft/stats/detail/stddev.cuh), and the more general [reduction abstraction](https://github.com/rapidsai/raft/blob/branch-24.08/cpp/include/raft/linalg/coalesced_reduction.cuh).

**Describe the solution you'd like**
I would prefer a more robust implementation (e.g. KahanBabushkaNeumaierSum as in the [sum kernel](https://github.com/rapidsai/raft/blob/b863f1810fdb0679f7777b656486dbfc96470d2b/cpp/include/raft/stats/detail/sum.cuh#L51) ) to be placed as a special implementation for the add-operator within the general reduction implementation (coalesced/strided). We already have a specialization for add reduction operator in the [strided reduction](https://github.com/rapidsai/raft/blob/b863f1810fdb0679f7777b656486dbfc96470d2b/cpp/include/raft/linalg/detail/strided_reduction.cuh#L142). APIs like `sum `and `stdev` should delegate to these abstractions.

**Additional context**
See https://github.com/rapidsai/raft/issues/2205 for the initial bug report.

PR #2381, opened 2024-07-16
Consolidate SUM reductions

mfoerste4:
This PR consists of multiple parts:

1. redirect custom reduction kernels within `stats `namespace to `linalg::reduce`
2. Specialize reduction kernels for addition utilizing the _Kahan-Babushka-Neumaier-Sum_ [link](https://en.wikipedia.org/wiki/Kahan_summation_algorithm)
3. Slightly adjust kernel heuristics for coalesced reductions

This should address #2366 and #2205. With the kernel heuristics adjusted the maximum performance drop is 4%.

FYI, @tfeher


cpp/include/raft/linalg/detail/strided_reduction.cuh
```diff
@@ -33,33 +33,54 @@ namespace detail {
 // for column major layout
 template <typename Type, typename MainLambda>
```
tfeher, line 34: Could you add a short note that describe that a compensated summation is performed? Do I see correctly that even block reduction is compensated, and only the final grid reduction is the place where we are not having additional compensation?

cpp/include/raft/linalg/detail/coalesced_reduction-inl.cuh
```diff
@@ -391,13 +574,8 @@ void coalescedReductionThickDispatcher(OutType* dots,
...
-      dots, data, D, N, init, stream, inplace, main_op, reduce_op, final_op);
-  } else {
-    coalescedReductionThick<ReductionThickPolicy<256, 64>, ReductionThinPolicy<32, 128, 1>>(
-      dots, data, D, N, init, stream, inplace, main_op, reduce_op, final_op);
-  }
+  coalescedReductionThick<ReductionThickPolicy<256, 64>, ReductionThinPolicy<32, 128, 1>>(
+    dots, data, D, N, init, stream, inplace, main_op, reduce_op, final_op);
 }
 
 // Primitive to perform reductions along the coalesced dimension of the matrix, i.e. reduce along
```
tfeher, line 581: Could you extend this note (and possibly the public docstring as well) that reduction with `add` operator uses a compensated summation? Also mention that the compensation is not equivalent with a sequential compensated sum, so that we have more efficient parallel reductions.

tfeher (changes requested):
Thanks Malte, this looks really good! I love how a new compensated reduction is reused among various stat functions! I just have a small request to improve documentation.

tfeher (approved):
Thanks @mfoerste4 for the update, LGTM!

tfeher (approved):
Approving latest changes.

tfeher (approved):
Thanks for catching these issues, LGTM.

cjnolet (approved):

PR #2384, opened 2024-07-18
Add cusparseSpMV_preprocess to cusparse wrapper

Kh4ster:
Add cusparseSpMV_preprocess to cusparse wrapper

### DIFF
--- a/cpp/include/raft/core/cusparse_macros.hpp
+++ b/cpp/include/raft/core/cusparse_macros.hpp
@@ -34,8 +34,8 @@
 //
 // (i.e., before including this header)
 //
-#define CUDA_VER_10_1_UP (CUDART_VERSION >= 10100)
-#define CUDA_VER_12_4_UP (CUDART_VERSION >= 12400)
+#define CUDA_VER_10_1_UP (CUDART_VERSION >= 10010)
+#define CUDA_VER_12_4_UP (CUDART_VERSION >= 12040)
 
 namespace raft {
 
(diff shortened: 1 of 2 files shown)


cpp/include/raft/core/cusparse_macros.hpp
```diff
@@ -35,6 +35,7 @@
 // (i.e., before including this header)
 //
 #define CUDA_VER_10_1_UP (CUDART_VERSION >= 10100)
+#define CUDA_VER_12_4_UP (CUDART_VERSION >= 12040)
```
dantegd, line 38: Shouldn't this be like this? 
```suggestion
#define CUDA_VER_12_4_UP (CUDART_VERSION >= 12400)
```

Kh4ster: Woops, sorry you're right, good catch.

I had tested with CUDA 12.0 which incorrectly showed that it was working. Should have tried with CUDA 12.2 or other

lowener: https://nvidia.github.io/cuda-python/module/cudart.html#cuda.cudart.cudaRuntimeGetVersion
> The version is returned as (1000 * major + 10 * minor). For example, CUDA 9.2 would be represented by 9020

So for CUDA 12.4, `CUDART_VERSION = 12040`.
The line just above should be corrected to: `#define CUDA_VER_10_1_UP (CUDART_VERSION >= 10010)` (as well as line 62 of this same file)

Kh4ster: @lowener Ahaha you are right. I had written 12040 then forgot where I had found it to then corrected again to 12400 after the above comment.

Let me fix all those lines

cpp/include/raft/core/cusparse_macros.hpp
```diff
@@ -59,8 +60,8 @@ namespace detail {
 
 inline const char* cusparse_error_to_string(cusparseStatus_t err)
 {
-#if defined(CUDART_VERSION) && CUDART_VERSION >= 10100
-  return cusparseGetErrorString(err);
+#if defined(CUDART_VERSION) && CUDART_VERSION >= 10010
+  return cusparseGetErrorString(err)
```
lowener, line 64:
```suggestion
  return cusparseGetErrorString(err);
```
Is there a reason for this `;` to disappear?

Kh4ster: Nope, I just saw that, sorry again

PR #2388, opened 2024-07-24
split up CUDA-suffixed dependencies in dependencies.yaml

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/31

In short, RAPIDS DLFW builds want to produce wheels with unsuffixed dependencies, e.g. `cudf` depending on `rmm`, not `rmm-cu12`.

This PR is part of a series across all of RAPIDS to try to support that type of build by setting up CUDA-suffixed and CUDA-unsuffixed dependency lists in `dependencies.yaml`.

For more details, see:
* https://github.com/rapidsai/build-planning/issues/31#issuecomment-2245815818
* https://github.com/rapidsai/cudf/pull/16183

### Why target 24.08?

This is targeting 24.08 because:

1. it should be very low-risk
2. getting these changes into 24.08 prevents the need to carry around patches for every library in DLFW builds using RAPIDS 24.08


jameslamb:
I've updated this based on the suggestions from https://github.com/rapidsai/cudf/pull/16183.

Ran the following to check `update-version.sh`.

```shell
git fetch upstream --tags
ci/release/update-version.sh '24.10.00'

git grep -E '24\.8|24\.08|0\.39'
```

No changes to that script were required.

dependencies.yaml
```diff
@@ -562,7 +578,7 @@ dependencies:
     common:
       - output_types: conda
         packages:
-          - &ucx_conda_run ucx>=1.15.0
```
vyasr, line 565: I guess this anchor was just unused previously?

dependencies.yaml
```diff
@@ -529,18 +537,22 @@ dependencies:
...
+              cuda_suffixed: "true"
             packages:
               - distributed-ucxx-cu11==0.39.*,>=0.0.0a0
-          - {matrix: null, packages: [*distributed_ucxx_conda]}
+          - {matrix: null, packages: [*distributed_ucxx_unsuffixed]}
   depends_on_ucx_build:
     common:
       - output_types: conda
         packages:
-          - &ucx_conda_build ucx==1.15.0
```
vyasr, line 543: Previously unused anchor?

vyasr (approved):

dependencies.yaml
jameslamb: yeah exactly. Maybe I should have left that out of the diff, but figured it was low-risk, non-controversial, and related-enough to include here.

dependencies.yaml
jameslamb: yep: https://github.com/rapidsai/raft/pull/2388#discussion_r1690518350

bdice (approved):

jameslamb:
queuing this up to auto-merge whenever that last test job is picked up. I'll check back in a bit to see if anything needs to be manually restarted.

PR #2389, opened 2024-07-24
Remove 'sample' parameter from stats::mean API

mfoerste4:
This PR removes the sample-parameter from the `raft::stats::mean` API to prevent people from using it by accident when for example computing the mean for a sampled variance computation.

This also invalidates some of the testcases. Within raft only test-code is affected by this change as the active usage of the sample parameter was already removed in #2381. 

This PR is based on #2381 but was separated for tracking purposes.

~~Note that this requires adaption of downstream libraries using the API. I am aware of at least one occurrence in `cuml`.~~
The old API remains in the code marked as deprecated which allows us to adapt downstream libraries at least for the duration of one release cycle.


cpp/include/raft/stats/detail/mean.cuh
```diff
@@ -27,10 +27,9 @@ namespace stats {
 namespace detail {
 
 template <typename Type, typename IdxType = int>
-void mean(
-  Type* mu, const Type* data, IdxType D, IdxType N, bool sample, bool rowMajor, cudaStream_t stream)
+void mean(Type* mu, const Type* data, IdxType D, IdxType N, bool rowMajor, cudaStream_t stream)
```
cjnolet, line 30: What's the reason for removing the ability to compute a sample-based mean vs a population-based mean? This feature seems useful.

tfeher: Both sample and population mean shall have 1/N factor. That gives the unbiased estimate. (This is unlike stdev or variance where we need to use different factor for sample and population mean).

tfeher (approved):
Thanks Malte for this PR. I agree that using `1/(N-1)` factor for sample `mean` is not correct, we should have a single option that has `1/N` factor. Therefore it makes sense to remove the `sample` parameter.

One should still consider that this is a breaking change, so we can consider to postpone this PR for 24.10, and issue a warning in 24.08 if the sample param is used.

cjnolet:
> One should still consider that this is a breaking change, so we can consider to postpone this PR for 24.10, and issue a warning in 24.08 if the sample param is used.

This was the primary reason I pointed this out. We need to be careful about updates to public apis specifically, because we can only assume they are being used downstream (and we need to check).

cjnolet:
@mfoerste4 do you plan for this PR to be ready for 25.02 or should we push to 25.04? I just want to make sure we're keeping this on the roadmap but being realistic about the timeline.

cjnolet:
@mfoerste4 that's a great idea. Let's try and get this in 25.02 if we can. Burndown starts on Thursday (Jan 30).

cjnolet (approved):

PR #2403, opened 2024-07-31
[FEA] Batching NN Descent

jinsolp:
### Description
This PR implements batching NN Descent. It will be helpful for reducing device memory usages for large datasets (specifically if the dataset is kept on host).
`index_params` now has...
- `n_clusters`: number of clusters to make. Larger clusters reduce device memory usage. Default is 1, in which case it doesn't do the batched NND.

### Notes
- The batching approach may have duplicate indices in the knn graph (in rare cases) because sometimes distances calculated for the same pair may be slightly different. This results in putting the same index far apart after sorting by distances, making it difficult to get unique indices (which is done by looking at 2 indices before the current one). 
  - handled by adding a `max_duplicates` for `check_unique_indices` in tests

### Benchmarks
- Dataset for NND (no batch and batch) is on host
- Dataset for brute force knn is on device (but still won't be able to run with large datasets even if the data is put on the host because it brings the entire dataset to device anyway)
- The dataset is just a slice of the wiki-all dataset (88M, 768) to test for different sizes
Screenshot 2024-08-02 at 8 35 58 AM


cpp/include/raft/neighbors/detail/nn_descent.cuh
```diff
@@ -1195,6 +1199,20 @@ GNND<Data_t, Index_t, epilogue_op>::GNND(raft::resources const& res,
...
+{
+  thrust::fill(thrust::device,
+               dists_buffer_.data_handle(),
+               dists_buffer_.data_handle() + dists_buffer_.size(),
+               std::numeric_limits<float>::max());
+  thrust::fill(thrust::device,
+               reinterpret_cast<Index_t*>(graph_buffer_.data_handle()),
+               reinterpret_cast<Index_t*>(graph_buffer_.data_handle()) + graph_buffer_.size(),
+               std::numeric_limits<Index_t>::max());
+  thrust::fill(thrust::device, d_locks_.data_handle(), d_locks_.data_handle() + d_locks_.size(), 0);
```
divyegala, line 1213:
```suggestion
  thrust::fill(rmm::exec_policy_async(stream),
               dists_buffer_.data_handle(),
               dists_buffer_.data_handle() + dists_buffer_.size(),
               std::numeric_limits<float>::max());
  thrust::fill(rmm::exec_policy_async(stream),
               reinterpret_cast<Index_t*>(graph_buffer_.data_handle()),
               reinterpret_cast<Index_t*>(graph_buffer_.data_handle()) + graph_buffer_.size(),
               std::numeric_limits<Index_t>::max());
  thrust::fill(rmm::exec_policy_async(stream), d_locks_.data_handle(), d_locks_.data_handle() + d_locks_.size(), 0);
```

cpp/include/raft/neighbors/detail/nn_descent_batch.cuh
```diff
@@ -0,0 +1,684 @@
...
+  size_t k        = global_nearest_cluster.extent(1);
+
+  auto local_offset = raft::make_host_vector<IdxT>(n_clusters);
+
+  max_cluster_size = 0;
+  min_cluster_size = std::numeric_limits<size_t>::max();
+
+  for (size_t i = 0; i < n_clusters; i++) {
+    cluster_size.data_handle()[i] = 0;
+    local_offset.data_handle()[i] = 0;
```
divyegala, line 196: You can use mdspan-based indexing directly. This suggestion applies to other places as well. That said, please avoid raw-loops where possible. `std::fill` does the same and is more readable.
```suggestion
    cluster_size(i) = 0;
    local_offset(i) = 0;
```

cpp/include/raft/neighbors/detail/nn_descent_batch.cuh
```diff
@@ -0,0 +1,684 @@
...
+                            IdxT* global_indices,
+                            IdxT* batch_indices)
+{
+  size_t batch_row = blockIdx.x;
+  typedef cub::BlockMergeSort<KeyValuePair<float, IdxT>, BLOCK_SIZE, ITEMS_PER_THREAD>
+    BlockMergeSortType;
+  __shared__ typename cub::BlockMergeSort<KeyValuePair<float, IdxT>, BLOCK_SIZE, ITEMS_PER_THREAD>::
+    TempStorage tmpSmem;
+
+  extern __shared__ float sharedMem[];
```
divyegala, line 261: Byte-based array as you reinterpret the shared memory as different types below. You might have to switch the shared memory calculation accordingly
```suggestion
  extern __shared__ char sharedMem[];
```

cpp/include/raft/neighbors/detail/nn_descent_batch.cuh
```diff
@@ -0,0 +1,684 @@
...
+  float* blockKeys    = (float*)sharedMem;
```
divyegala, line 262: This suggestion applies below as well
```suggestion
  float* blockKeys    = reinterpret_cast<float*>(sharedMem);
```

cpp/include/raft/neighbors/detail/nn_descent_batch.cuh
```diff
@@ -0,0 +1,684 @@
...
+#include <vector_types.h>
+
+#include <cstddef>
+#include <cstdint>
+#include <limits>
+#include <optional>
+#include <random>
+#include <type_traits>
+
+#define INST_MERGE_SUBGRAPH(blockSize, elemPT)                                                 \
```
divyegala, line 46: I can't figure out why this was needed? The kernel is already a templated function?

cpp/include/raft/neighbors/detail/nn_descent_batch.cuh
```diff
@@ -0,0 +1,684 @@
...
+    cluster_size.data_handle()[i] = 0;
+    local_offset.data_handle()[i] = 0;
+  }
+
+  for (size_t i = 0; i < num_rows; i++) {
+    for (size_t j = 0; j < k; j++) {
+      IdxT cluster_id = global_nearest_cluster.data_handle()[i * k + j];
+      cluster_size.data_handle()[cluster_id]++;
+    }
+  }
```
divyegala, line 204: This section may benefit from an omp parallel region. `cluster_size(cluster_id)++` will have to be an atomic though. Overall, I don't know how beneficial this speedup will be so you can just leave this idea as a comment in the code so we can try later on if necessary

divyegala (changes requested):

cpp/include/raft/neighbors/detail/nn_descent_batch.cuh
jinsolp: Oh it's because it's so long I didn't want to write it multiple times in the if-else below where we launch the kernel..!
referred to what is being done [here](https://github.com/rapidsai/raft/blob/32f3703b0d28e5e8934fc463cf3ec6bd6cc82a79/cpp/include/raft/matrix/detail/columnWiseSort.cuh#L216)

divyegala: Let's not use macros, please. It's fine to have the kernel invocation verbosity as it makes the code more readable.

cpp/include/raft/neighbors/detail/nn_descent.cuh
```diff
@@ -1195,6 +1199,22 @@ GNND<Data_t, Index_t, epilogue_op>::GNND(raft::resources const& res,
   thrust::fill(thrust::device, d_locks_.data_handle(), d_locks_.data_handle() + d_locks_.size(), 0);
 };
 
+template <typename Data_t, typename Index_t, typename epilogue_op>
+void GNND<Data_t, Index_t, epilogue_op>::reset(raft::resources const& res)
+{
+  auto stream = raft::resource::get_cuda_stream(res);
+  thrust::fill(rmm::exec_policy(stream),
```
divyegala, line 1206: Ah, it's `nosync` not `async`

```suggestion
  thrust::fill(rmm::exec_policy_nosync(stream),
```

cpp/include/raft/neighbors/detail/nn_descent_batch.cuh
```diff
@@ -0,0 +1,680 @@
...
+      inverted_indices(offset(cluster_id) + local_offset(cluster_id)) = i;
+      local_offset(cluster_id) += 1;
+    }
+  }
+
+  for (size_t i = 0; i < n_clusters; i++) {
+    if (max_cluster_size < cluster_size(i)) { max_cluster_size = cluster_size(i); }
+    if (min_cluster_size > cluster_size(i)) { min_cluster_size = cluster_size(i); }
+  }
+}
```
divyegala, line 224: `std::max_element` and `std::min_element`

cpp/include/raft/neighbors/detail/nn_descent_batch.cuh
jinsolp: Okay 🙂

divyegala (approved):

cpp/include/raft/neighbors/nn_descent_types.hpp
```diff
@@ -56,6 +56,8 @@ struct index_params : ann::index_params {
   size_t max_iterations            = 20;      // Number of nn-descent iterations.
   float termination_threshold      = 0.0001;  // Termination threshold of nn-descent.
   bool return_distances            = false;   // return distances if true
+  size_t n_clusters                = 2;       // number of clusters if using batching
+  bool do_batch                    = false;
```
cjnolet, line 60: I think a more simplified user experience would be to default n_clusters to 1 and then remove the `do_batch` option. This way, all a user needs to do to enable batching is increase the number of clusters.

cpp/include/raft/neighbors/nn_descent.cuh
```diff
@@ -178,16 +188,32 @@ index<IdxT> build(raft::resources const& res,
...
+template <typename T,
+          typename IdxT        = uint32_t,
+          typename epilogue_op = DistEpilogue<IdxT, T>,
+          typename Accessor =
+            host_device_accessor<std::experimental::default_accessor<T>, memory_type::host>>
+index<IdxT> build(raft::resources const& res,
+                  const index_params& params,
+                  mdspan<const T, matrix_extent<int64_t>, row_major, Accessor> dataset,
+                  epilogue_op distance_epilogue = DistEpilogue<IdxT, T>(),
+                  bool do_batch                 = false)
```
cjnolet, line 210: We should enable this through the params and not modify the build function signatures. We are trying to keep the build/extend/search signatures as consistent as possible across index types.

cpp/include/raft/neighbors/nn_descent_types.hpp
```diff
@@ -56,6 +56,8 @@ struct index_params : ann::index_params {
   size_t max_iterations            = 20;      // Number of nn-descent iterations.
   float termination_threshold      = 0.0001;  // Termination threshold of nn-descent.
   bool return_distances            = false;   // return distances if true
+  size_t n_clusters                = 2;       // number of clusters if using batching
```
cjnolet, line 59: We should explain to the user very clearly exaclty what this option means. A user new to RAFT (or cuVS) is not going to understand what this does, nor the implications of the clustering. I would suggest maybe writing up a paragraph or two here about what this option really does. Write out the steps of the algorithm and the benefits/drawbacks for different number of clusters. Why would a user use this option?

cjnolet (changes requested):
So sorry for being so late to review this @jinsolp. I think this is very close, but in order for this to be useful to users, we should make it as easy as possible to use and document very clearly what the algorithms does (and why a user would want to use it).

cpp/include/raft/neighbors/nn_descent.cuh
jinsolp: This is never used elsewhere, but it's just for the current tests that do not use batching.
I've added it as a comment too, but the `kmeans_balanced::fit` inside batch build stops compilation with current nnd tests that have uint8 and int8 data types.

cjnolet: If this is just for testing then I’d add it to the tests. We shouldn’t expose test code in public APIs.

cjnolet (approved):

PR #2408, opened 2024-08-07
Improve update-version.sh

bdice:
This PR improves `update-version.sh` by clarifying suffix handling and ucxx/ucx-py version handling.


bdice:
> Nice! I'm especially happy to see literal `-cu11` and `-cu12` replaces with regexes that won't have to be updated for future CUDA versions.

This was why I began this round of updates across RAPIDS!

PR #2416, opened 2024-08-20
Lanczos Solver

aamijar:
## Lanczos Solver for Sparse Eigen Decomposition
We propose a new lanczos solver in raft that fixes the issues present in the previous solver `raft::sparse::solver::detail::computeSmallestEigenvectors`.

Specifically we address the following issues:
1. Numerical Stability for both float32 and float64 datatypes
2. Efficiency and Speed of Convergence

This new implementation is taken from the cupy library `cupyx.scipy.sparse.linalg.eigsh` where the thick-restart and full reorthogonalzation methods are used.

Additionally this PR exposes a python api for raft lanczos solver with an interface similar to `scipy.sparse.linalg.eigsh` and `cupyx.scipy.sparse.linalg.eigsh`.

```py3
from pylibraft.solver import eigsh
```


cpp/include/raft/sparse/solver/detail/lanczos.cuh
```diff
@@ -27,12 +27,54 @@
 #include <raft/spectral/detail/warn_dbg.hpp>
 #include <raft/spectral/matrix_wrappers.hpp>
 #include <raft/util/cudart_utils.hpp>
+// #include <raft/sparse/solver/lanczos.cuh>
```
cjnolet, line 30: Please remove commented out code.

cpp/include/raft/sparse/solver/detail/lanczos.cuh
```diff
@@ -1396,4 +1438,658 @@ int computeLargestEigenvectors(
...
+
+  auto triangular_matrix_view =
+    raft::make_device_matrix_view<const value_type_t, uint32_t, raft::col_major>(
+      triangular_matrix.data_handle(), ncv, ncv);
+
+  raft::linalg::eig_dc(handle, triangular_matrix_view, eigenvectors, eigenvalues);
+}
+
+template <typename index_type_t, typename value_type_t>
+void cupy_aux(raft::resources const& handle,
```
cjnolet, line 1559: Please don't prefix code in raft with `cupy_`. That's just bad form. This code is meant for RAFT, and we should treat it as such.

cpp/include/raft/sparse/solver/detail/lanczos.cuh
```diff
@@ -1396,4 +1438,658 @@ int computeLargestEigenvectors(
...
+    auto cublas_h = resource::get_cublas_handle(handle);
+
+    value_type_t alpha_i_host = 0;
+    value_type_t b            = 0;
+    value_type_t mone         = -1;
+
+    raft::copy<value_type_t>(&b, &beta(0, (i - 1 + ncv) % ncv), 1, stream);
+    raft::copy<value_type_t>(&alpha_i_host, &(alpha(0, i)), 1, stream);
+
+    raft::linalg::detail::cublasaxpy(
```
cjnolet, line 1639: Please use the publicly available APIs and don't call into detail for these wrappers. The whole reason these wrappers exist in detail is that the private bits can be hidden and exposed through a single public interface. When you call into these directly, you are side-stepping that interface. You should be calling `raft::linalg::saxpy` (e.g. from your perspective, it should not matter that this is using cublas at all.. it should just be a raft functions. Please do the same with all of these calls into the `_wrapper` headers.

cpp/include/raft/sparse/solver/detail/lanczos.cuh
```diff
@@ -1396,4 +1438,658 @@ int computeLargestEigenvectors(
...
+                                     1,
+                                     stream);
+
+    auto uu_i = raft::make_device_scalar_view(&uu(0, i));
+    raft::linalg::add(handle, make_const_mdspan(alpha_i), make_const_mdspan(uu_i), alpha_i);
+
+    kernel_clamp_down<<<1, 1>>>(alpha_i.data_handle(), static_cast<value_type_t>(1e-9));
+
+    raft::linalg::detail::cublassetpointermode(cublas_h, CUBLAS_POINTER_MODE_DEVICE, stream);
+    raft::linalg::detail::cublasnrm2(cublas_h, n, u.data_handle(), 1, &beta(0, i), stream);
```
cjnolet, line 1682: Please expose any of these functions publicly if needed, but do not call into detail APIs.

cpp/src/raft_runtime/solver/lanczos_solver.cuh
```diff
@@ -0,0 +1,93 @@
...
+ */
+
+#include <raft/sparse/solver/lanczos.cuh>
+#include <raft/spectral/cluster_solvers.cuh>
+#include <raft/spectral/eigen_solvers.cuh>
+#include <raft/spectral/matrix_wrappers.hpp>
+
+#include <raft_runtime/solver/lanczos.hpp>
+
+template <typename IndexType, typename ValueType>
```
cjnolet, line 24: `raft_runtime` is explicitly for code that's already compiled into RAFT so that users can compile their cpp code against it without needing a CUDA compiler. `raft_runtime` is not for cuda headers. Please move this into `include/raft/solver/detail` and expose a public header just for the public API functions that invoke the solver. Then swap out the spcectral clustering primitive in RAFT to use the new lanczos solver from its public API. Please keep the code well groomed and it will be easy to maintain / update going forward.

cpp/include/raft_runtime/solver/lanczos.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+namespace raft::runtime::solver {
+
+/**
+ * @defgroup rmat_runtime RMAT Runtime API
+ * @{
+ */
+
+#define FUNC_DECL(IndexType, ValueType)              \
+  void lanczos_solver(const raft::resources& handle, \
```
cjnolet, line 30: Please don't add public APIs that accept raw pointers. These APIs are all deprecated in RAFT. Please accept device_matrix_view, etc.. instead. That's why they exist.

cjnolet: Take a look at the other pylibraft cython files for examples of how to do this. pylibraft cython is newer than cuml's cython so we've adopted better design haibits.

cjnolet (changes requested):
I think this is coming along great, @aamijar. Mostly it needs some cleanup and polishing, but otherwise should be ready to merge once my comments are resolved.

lowener (changes requested):

cpp/test/sparse/solver/lanczos.cu
```diff
@@ -0,0 +1,415 @@
...
+    raft::random::uniform<ValueType>(handle, rng, theta.view(), 0, 1);
+
+    raft::device_matrix<IndexType, uint32_t, raft::row_major> out =
+      raft::make_device_matrix<IndexType, uint32_t, raft::row_major>(handle, n_edges * 2, 2);
+    raft::device_vector<IndexType, uint32_t, raft::row_major> out_src =
+      raft::make_device_vector<IndexType, uint32_t, raft::row_major>(handle, n_edges);
+    raft::device_vector<IndexType, uint32_t, raft::row_major> out_dst =
+      raft::make_device_vector<IndexType, uint32_t, raft::row_major>(handle, n_edges);
+
+    raft::random::RngState rng1{params.seed};
```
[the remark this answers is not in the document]
aamijar: Yes, I found that reusing the same rng state causes different generated data on different hardware.

cpp/include/raft/sparse/solver/detail/lanczos.cuh
```diff
@@ -1396,4 +1438,658 @@ int computeLargestEigenvectors(
...
+    std::cout << "res " << res << " " << iter << std::endl;
+  }
+
+  raft::copy(eigVals_dev, eigenvalues_k.data_handle(), nEigVecs, stream);
+  raft::copy(eigVecs_dev, ritz_eigenvectors.data_handle(), n * nEigVecs, stream);
+
+  return 0;
+}
+
+template <typename IndexTypeT, typename ValueTypeT>
```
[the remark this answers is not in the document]
aamijar: What do you mean by unused?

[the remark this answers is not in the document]
cjnolet: Is it being used to invoke another function or establish the type for any variables within this function?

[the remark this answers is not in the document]
aamijar: Yes it is being used for `A` and `config` index type which are variables in this function

cpp/include/raft/sparse/solver/detail/lanczos.cuh
```diff
@@ -1396,4 +1438,658 @@ int computeLargestEigenvectors(
   return status;
 }
 
+template <typename T>
+RAFT_KERNEL kernel_subtract_and_scale(T* u, T* vec, T* scalar, int n)
```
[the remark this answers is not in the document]
aamijar: Can I access the out value though? I have to do `u = u - scalar * vec`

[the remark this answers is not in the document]
aamijar: I know how to do `u = scalar * vec` for example but not the one above.

cpp/include/raft/sparse/solver/detail/lanczos.cuh
```diff
@@ -1396,4 +1438,658 @@ int computeLargestEigenvectors(
...
+  auto cublas_h      = resource::get_cublas_handle(handle);
+  value_type_t v0nrm = 0;
+  raft::linalg::detail::cublasnrm2(cublas_h, n, v0_view.data_handle(), 1, &v0nrm, stream);
+
+  raft::device_scalar<value_type_t> v0nrm_scalar = raft::make_device_scalar(handle, v0nrm);
+
+  raft::device_vector_view<const value_type_t> v0_vector_const =
+    raft::make_device_vector_view<const value_type_t>(v0, n);
+
+  raft::linalg::unary_op(
```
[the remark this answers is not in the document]
aamijar: The reason `unary_op` was used is because the scalar is on the device. divide_scalar needs a view on the host.

cpp/include/raft/sparse/solver/detail/lanczos.cuh
```diff
@@ -1396,4 +1438,658 @@ int computeLargestEigenvectors(
...
+      raft::make_device_matrix_view<value_type_t>(&((V.view())(nEigVecs, 0)), 1, n);
+    value_type_t unrm = 0;
+    raft::linalg::detail::cublasnrm2(cublas_h, n, u.data_handle(), 1, &unrm, stream);
+
+    raft::device_scalar<value_type_t> unrm_scalar = raft::make_device_scalar(handle, unrm);
+
+    raft::device_vector_view<const value_type_t> u_vector_const =
+      raft::make_device_vector_view<const value_type_t>(u.data_handle(), n);
+
+    raft::linalg::unary_op(handle,
```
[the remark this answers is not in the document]
aamijar: Same reason as before: scalar is on the device

cpp/include/raft/linalg/detail/norm.cuh
```diff
@@ -138,6 +140,15 @@ void colNormCaller(OutType* dots,
   };
 }
 
+template <typename T, bool DevicePointerMode = false>
+void nrm2(
+  raft::resources const& handle, int n, const T* x, int incx, T* result, cudaStream_t stream)
+{
+  cublasHandle_t cublas_h = resource::get_cublas_handle(handle);
+  detail::cublas_device_pointer_mode<DevicePointerMode> pmode(cublas_h);
+  detail::cublasnrm2(cublas_h, n, x, incx, result, stream);
```
cjnolet, line 149: How does this differ from the other norm functions? Can we just invoke this underneath the existing norm functions? Is it doing anything significantly different? Even better- could we just use the existing norm functions instead of having to call this at all?

cpp/include/raft/sparse/solver/detail/lanczos.cuh
```diff
@@ -1396,4 +1438,658 @@ int computeLargestEigenvectors(
...
+              raft::device_matrix_view<value_type_t> uu,
+              raft::device_matrix_view<value_type_t> vv)
+{
+  auto stream = resource::get_cuda_stream(handle);
+
+  int n = A->nrows_;
+
+  raft::copy(v.data_handle(), &(V(start_idx, 0)), n, stream);
+
+  std::cout << start_idx << " " << end_idx << std::endl;
```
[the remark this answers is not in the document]
cjnolet: +1 we shouldn't have explicit printing in any code. Logger should always be used (and this should be at the trace level so it doesn't show up by default).

cpp/include/raft/linalg/detail/norm.cuh
aamijar: Hi @cjnolet, I have refactoring the code to use the existing `raft::linalg::norm` functions and have removed the cublasnrm2 wrapper.

cpp/include/raft/sparse/solver/detail/lanczos.cuh
```diff
@@ -1396,4 +1435,699 @@ int computeLargestEigenvectors(
...
+    raft::copy(&unrm, output.data_handle(), 1, stream);
+
+    raft::device_scalar<ValueTypeT> unrm_scalar = raft::make_device_scalar(handle, unrm);
+
+    auto u_vector_const = raft::make_device_vector_view<const ValueTypeT>(u.data_handle(), n);
+
+    raft::linalg::unary_op(handle,
+                           u_vector_const,
+                           V_0_view,
+                           [device_scalar = unrm_scalar.data_handle()] __device__(auto y) {
```
lowener, line 1896: `unrm` can be skipped
```suggestion
                           [device_scalar = output.data_handle()] __device__(auto y) {
```

cpp/include/raft/sparse/solver/detail/lanczos.cuh
```diff
@@ -1396,4 +1435,699 @@ int computeLargestEigenvectors(
...
+      raft::make_device_matrix_view<const ValueTypeT>(beta_k.data_handle(), 1, nEigVecs);
+    raft::linalg::norm(handle,
+                       input2,
+                       output2.view(),
+                       raft::linalg::L2Norm,
+                       raft::linalg::Apply::ALONG_ROWS,
+                       raft::sqrt_op());
+    raft::copy(&res, output2.data_handle(), 1, stream);
+
+    RAFT_LOG_TRACE("Iteration %f: residual (tolerance) %d", iter, res);
```
lowener, line 2091: To use a copied value on host, the stream would need to be synchronized before. Since that sync can slow this function down it would be better to check the log level, and only copy and sync if necessary.

lowener (changes requested):
Can a test for the python API of Lanczos also be added?

lowener:
I unfortunately don't have permissions to push on the branch of this PR so I kept the commits and continued on PR #2481.

PR #2417, opened 2024-08-22
Drop Python 3.9 support

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/88

Finishes the work of dropping Python 3.9 support.

This project stopped building / testing against Python 3.9 as of https://github.com/rapidsai/shared-workflows/pull/235.
This PR updates configuration and docs to reflect that.

### How I tested this

Checked that there were no remaining uses like this:

```shell
git grep -E '3\.9'
git grep '39'
git grep 'py39'
```

And similar for variations on Python 3.8 (to catch things that were missed the last time this was done).


jakirkham:
One of the C++ tests failed in [this CI job]( https://github.com/rapidsai/raft/actions/runs/10509784822/job/29119156426?pr=2417#step:7:705 )

That said, this is simply cleaning up Python bits (not touching C++). So it doesn't seem caused by these changes (nor related)

Filing issue ( https://github.com/rapidsai/raft/issues/2418 ) to track. Will also mention this offline

Since it only happens in that one job, it looks flaky. Will trying restarting to see if it goes away

jakirkham (approved):

jakirkham:
Thanks James! 🙏

PR #2424, opened 2024-08-28
Add RAFT wrappers around current_device_resource functions

### LINKED ISSUE #2423: [FEA] Add RAFT wrappers for RMM current_device_resource functions.
Is your feature request related to a problem? Please describe.
I'm about to undertake yet another change to every function that takes an MR parameter. To prevent doing this too many times in the future, I would like to provide a wrapper in RAFT to centralize calls to rmm::get_current_device_resource[_ref]() and aliases for RMM resource_ref types. hopefully this can reduce the number of files that need to change in the future when this functionality changes (e.g. if it moves out of RMM to CCCL).

Describe the solution you'd like
Either add the following to `raft/core/device_resource.hpp`, or add a new header, say `raft/core/memory_resource.hpp`. Please provide suggestions on the best place for this. Contents something like this:

namespace raft {

using device_async_resource_ref = rmm::device_async_resource_ref;

inline device_async_resource_ref get_current_device_resource_ref() {
  return rmm::get_current_device_resource_ref();
}

inline device_async_resource_ref set_current_device_resource_ref(device_async_resource_ref mr) {
  return rmm::set_current_device_resource_ref(mr);
}

inline device_async_resource_ref reset_current_device_resource_ref() {
  return rmm::reset_current_device_resource_ref();
}

Describe alternatives you've considered

Additional context
https://github.com/rapidsai/rmm/pull/1598

related cudf PR: https://github.com/rapidsai/cudf/pull/16679

harrism:
Fixes #2423

Adds local RAFT wrappers around `rmm::mr::get/set_current_device_resource` and `resource_ref` aliases. The idea is to localize changes needed to keep up with RMM refactoring.

The wrappers are added in cpp/include/raft/core/resource/device_memory_resource.hpp

I've marked this PR as breaking because it breaks the ABI, however the API is compatible.

### DIFF
--- a/docs/source/vector_search_tutorial.md
+++ b/docs/source/vector_search_tutorial.md
@@ -368,6 +368,7 @@ The RAPIDS software ecosystem makes heavy use of the [RAPIDS Memory Manager](htt
 
 As an example, the following code snippet creates a `pool_memory_resource` and sets it as the default memory resource, which means all other libraries that use RMM will now allocate their device memory from this same pool:
 ```c++
+#include <raft/core/resource/device_memory_resource.hpp>
 #include <rmm/mr/device/pool_memory_resource.hpp>
 
 rmm::mr::cuda_memory_resource cuda_mr;
--- a/docs/source/vector_search_tutorial.md
+++ b/docs/source/vector_search_tutorial.md
@@ -375,7 +375,7 @@ rmm::mr::cuda_memory_resource cuda_mr;
 // set the initial size to half of the free device memory
 auto init_size = rmm::percent_of_free_device_memory(50);
 rmm::mr::pool_memory_resource<rmm::mr::cuda_memory_resource> pool_mr{&cuda_mr, init_size};
-rmm::mr::set_current_device_resource(&pool_mr); // Updates the current device resource pointer to `pool_mr`
+raft::resource::set_current_device_resource(&pool_mr); // Updates the current device resource pointer to `pool_mr`
 ```
 
 The `raft::device_resources` object will now also use the `rmm::current_device_resource`.  This isn't limited to C++, however. Often a user will be interacting with PyTorch, RAPIDS, or Tensorflow through Python and so they can set and use RMM's `current_device_resource` [right in Python](https://github.com/rapidsai/rmm#using-rmm-in-python-code).
--- a/cpp/bench/ann/src/faiss/faiss_gpu_benchmark.cu
+++ b/cpp/bench/ann/src/faiss/faiss_gpu_benchmark.cu
@@ -183,10 +183,10 @@ int main(int argc, char** argv)
   rmm::mr::pool_memory_resource<rmm::mr::cuda_memory_resource> pool_mr{
     &cuda_mr, rmm::percent_of_free_device_memory(50)};
   // Updates the current device resource pointer to `pool_mr`
-  auto old_mr = rmm::mr::set_current_device_resource(&pool_mr);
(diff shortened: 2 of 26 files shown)


docs/source/vector_search_tutorial.md
```diff
@@ -375,7 +375,7 @@ rmm::mr::cuda_memory_resource cuda_mr;
 // set the initial size to half of the free device memory
 auto init_size = rmm::percent_of_free_device_memory(50);
 rmm::mr::pool_memory_resource<rmm::mr::cuda_memory_resource> pool_mr{&cuda_mr, init_size};
-rmm::mr::set_current_device_resource(&pool_mr); // Updates the current device resource pointer to `pool_mr`
+raft::resource::set_current_device_resource(&pool_mr); // Updates the current device resource pointer to `pool_mr`
```
lowener, line 378: Add `raft/core/resource/device_memory_resource.hpp` to includes of this code example

harrism: Done.

achirkin (changes requested):
I'd suggest to put the new resource helpers somewhere into a `detail` namespace to discourage the use of these outside of where it is strictly necessary within raft.
Otherwise it goes against raft principle of keeping the program state locally in a raft handle by introducing the global state functions in the api.
In the docs, where we refer to `rmm::mr::get_current_device_resource()`, I'd rather keep the rmm namespaces, possibly adjusting them to reflect the upcoming API changes in the rmm - again because raft is not supposed to provide this API.

harrism:
> I'd suggest to put the new resource helpers somewhere into a detail namespace

This limits the utility of this PR. I have a dependent PR in each RAPIDS repo that depends on RAFT (cuVS, cuML, cumlprims_mg, cuGraph) to use these RAFT wrappers rather than directly using RMM's functions.

achirkin:
I'm sorry that this limits the utility of the PR, but, in my opinion, this plainly goes against the whole design of raft resources.
Other reviewers may disagree.

harrism:
Perhaps you can point me to the design document?

cjnolet:
> Perhaps you can point me to the design document?

We do outline what @achirkin is talking about in our [developer guide](https://github.com/rapidsai/raft/blob/branch-24.10/docs/source/developer_guide.md#public-interface). It may sound like something that was done in haste but it was actually very intentional and done for a specific purpose. 

We've always relied on RMM's APIs to be used outside of RAFT (and outside of cuML, cuGraph, etc...) so that we could merely honor whatever the user has configured with RMM. More recently, we did add a capability for a user to set a different RMM memory resource for temporary memory allocations (and an additionl one for smaller vs larger allocations), but those features still continue to follow the constraint that RAFT maintains no global state of its own. 

It's dangerous for us to do that because the underlying computational functions now have potential interdependencies that we have to manage and make guarantees with respect to RAFT itself, which now means RAFT needs to be concerned with thread safety. 

This is a separation of concerns- RAFT, cuML, cuGraph, cuOpt, and others have intentionally avoided doing memory management things in libraries outside of RMM because RMM Is supposed to be responsible for that. If you put this wrapper inside RMM itself then we'd still get the benefits of defining it somewhere upstream (in this case within the RMM namespace) so everyone downstream can use it. Is there any reason we can't just alias this inside of RMM in the meantime?

harrism:
Aliasing within RMM would just be duplicating something that is already in RMM. The point was to localize changes in non-RMM libraries to a single function rather than have to change *every* function as we go through this churn. It is painful to modify every function in every RMM-dependent library when RMM changes.

PR #2425, opened 2024-08-28
Exclude any kernel symbol that uses cutlass

robertmaynard:


.github/workflows/pr.yaml
```diff
@@ -51,7 +51,7 @@ jobs:
     with:
       build_type: pull-request
       enable_check_symbols: true
-      symbol_exclusions: _ZN\d+raft_cutlass
+      symbol_exclusions: raft_cutlass
```
KyleFromNVIDIA, line 54: Can you explain the rationale of this change? It looks like you've moved from GCC-style name mangling to a simple match.

KyleFromNVIDIA (approved):

PR #2429, opened 2024-09-06
[BUG] Fix bitset function visibility

lowener:
`raft::ceildiv` is also being replaced with `raft::div_rounding_up_safe` to avoid including CUDA headers when not needed.


rhdong (approved):
LGTM

cjnolet (approved):

PR #2430, opened 2024-09-11
Use runtime check of cudart version for eig

lowener:


bdice (approved):
This looks right to me. @lowener Were you able to verify that this fixes the problem?

PR #2431, opened 2024-09-11
Adding NCCL clique to the RAFT handle

viclafargue:


cpp/include/raft/comms/nccl_clique.cuh
```diff
@@ -0,0 +1,118 @@
...
+
+namespace raft::comms {
+
+nccl_clique::nccl_clique(int percent_of_free_memory)
+  : root_rank_(0),
+    percent_of_free_memory_(percent_of_free_memory),
+    per_device_pools_(0),
+    device_resources_(0)
+{
+  cudaGetDeviceCount(&num_ranks_);
```
cjnolet, line 53: I don't see any device code bring used in this file. I think this should be an hpp extension because we're just interacting with the CUDA runtime APIs.

cpp/include/raft/comms/nccl_clique.cuh
```diff
@@ -0,0 +1,118 @@
...
+    }                                              \
+  } while (0);
+
+namespace raft::comms {
+void build_comms_nccl_only(raft::resources* handle, ncclComm_t nccl_comm, int num_ranks, int rank);
+}
+
+namespace raft::comms {
+
+nccl_clique::nccl_clique(int percent_of_free_memory)
```
cjnolet, line 47: These can be combined in the `hpp` in-place ,especially since RAFT is header-only so this isn't really hiding any implmentation details and this file is only calling CUDA runtime APIs, not any device code from what i can see.

cpp/include/raft/core/device_resources.hpp
```diff
@@ -123,6 +124,11 @@ class device_resources : public resources {
 
   rmm::exec_policy_nosync& get_thrust_policy() const { return resource::get_thrust_policy(*this); }
 
+  const raft::comms::nccl_clique& get_nccl_clique_handle() const
```
cjnolet, line 127: I don't think we need to call this a `nccl_clique_handle`. I'd prefer to just call it `nccl_clique` for consistency with the actual naming elsewhere.

cpp/include/raft/core/resource/nccl_clique_handle.hpp
```diff
@@ -0,0 +1,66 @@
+/*
+ * Copyright (c) 2024, NVIDIA CORPORATION.
```
cjnolet, line 2: Please just name this file `nccl_clique.hpp`. No need to call it a handle.

cjnolet (changes requested):
Thanks for making this change, Victor. The changes mostly look good, but I have very minor changes to naming (and filenames).

cjnolet (approved):

PR #2439, opened 2024-09-18
[Feat] add `repeat`, `sparsity`,  `eval_n_elements` APIs to `bitset`

rhdong:
- This PR is a part of the feature that applies the prefilter brute-force in Cagra.


cpp/include/raft/core/bitset.cuh
```diff
@@ -60,6 +62,102 @@ _RAFT_DEVICE void bitset_view<bitset_t, index_t>::set(const index_t sample_index
...
+
+  if (0 == size_h) { return static_cast<double>(1.0); }
+
+  rmm::device_scalar<index_t> nnz(0, stream);
+
+  auto vector_view = raft::make_device_vector_view<const bitset_t, index_t>(data(), n_elements());
+  auto nnz_view    = raft::make_device_scalar_view<index_t>(nnz.data());
+  auto size_view   = raft::make_host_scalar_view<index_t>(&size_h);
+
+  raft::popc(res, vector_view, size_view, nnz_view);
```
lowener, line 154: The `count()` method can also be implemented in `bitset_view`, it will be useful for users.

lowener: I am also noticing that if the number of bits is not a multiple of the bitset element size the result might be wrong.

cpp/test/core/bitset.cu
```diff
@@ -145,6 +194,37 @@ class BitsetTest : public testing::TestWithParam<test_spec_bitset> {
     resource::sync_stream(res, stream);
     ASSERT_TRUE(hostVecMatch(bitset_ref, bitset_result, raft::Compare<bitset_t>()));
 
+    // test sparsity, repeat and eval_n_elements
+    if constexpr (std::is_same_v<bitset_t, uint32_t> || std::is_same_v<bitset_t, uint64_t>) {
```
lowener, line 198: Why restrict the test to those data types?

lowener (changes requested):

cpp/include/raft/core/bitset.cuh
rhdong: > I am also noticing that if the number of bits is not a multiple of the bitset element size the result might be wrong.

Should not, because the size is the actual number of bits. Maybe I miss something?

rhdong:
Hi @cjnolet @lowener , this PR merging will cause CUVS CI fail, so the CUVS PR https://github.com/rapidsai/cuvs/pull/350 must be merged immediately after this.

lowener (approved):
LGTM

PR #2440, opened 2024-09-18
Add missing `cuda_suffixed: true`

trxcllnt:
This change ensures `rmm-cuXX` is not in the package dependencies when building the pylibraft wheel with `-C "rapidsai.matrix-entry=cuda_suffixed=false`.


jameslamb (approved):
Ah! Thank you for this.

jameslamb:
I tried re-running CI here multiple times... it's failed in the same way each time (and I see other PRs and nightlies failing the same way).

Opened #2450 documenting that.

jameslamb:
Updated to pull in #2453, which temporarily skips one failing test that was blocking this (and because this was now too many commits behind `branch-24.10` to pass the `Recently Updated` check).

PR #2441, opened 2024-09-19
Fix sed syntax in `update-version.sh`

raydouglass:
Fixes a tiny `sed` syntax error


raydouglass:
I used the wrong `[skip ci]` to skip CI, so I manually killed it since this file is not tested in CI.

jameslamb (approved):

PR #2443, opened 2024-09-20
WIP: bump nccl floor to 2.18.1.1

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/102

Some RAPIDS libraries are using `ncclCommSplit()`, which was introduced in `nccl==2.18.1.1`. This is part of a series of PRs across RAPIDS updating libraries' pins to `nccl>=2.18.1.1` to ensure they get a new-enough version that supports that.

PR #2458, opened 2024-09-26
bump NCCL floor to 2.19

jameslamb:
Follow-up to #2443

As part of the work to support NumPy 2 across RAPIDS, we found reason to upgrade some libraries like `cugraph` to slightly newer NCCL (`>=2.19`). Context: https://github.com/rapidsai/build-planning/issues/102#issuecomment-2375595743

This applies that same bump here, to keep the range of NCCL versions consistent across RAPIDS.


jakirkham (approved):
Thanks James! 🙏

jakirkham:
One build failed with a networking error. Will wait for the rest of the builds to complete and then restart

jameslamb:
@jakirkham I just cancelled CI and restarted the failed jobs.

`conda-cpp-build` is upstream of a lot of other jobs, including `conda-cpp-tests` which can take 2-3 hours to run 😬 

I know we're getting very close to code freeze... will leave it to @cjnolet 's discretion whether to admin-merge at some point or to wait for all those test jobs to run. I think we can be highly confident that this will work, given that the tests were all already using NCCL 2.22 on this previous successful run: https://github.com/rapidsai/raft/actions/runs/11040960166/job/30673327293

(and should continue to, given that this just moves the floor from `>=2.18` to `>=2.19`)

jakirkham:
Sounds good. Thanks James! 🙏

Have also mentioned this PR in the release thread offline. So it is being tracked there

cjnolet (approved):

jakirkham:
Thanks James and Corey! 🙏

PR #2464, opened 2024-10-02
Use raft::copy in the inner loop of the host-side raft::gather

achirkin:
Use `raft::copy` instead of an assignment operator loop in the host-side gather operation. This should allow to leverage the logic of `raft::copy` to use fast memcpy when possible (especially for high-dimensional datasets).

Make the NVTX annotations a little bit more descriptive along the way.

### DIFF
--- a/cpp/include/raft/matrix/detail/gather.cuh
+++ b/cpp/include/raft/matrix/detail/gather.cuh
@@ -17,6 +17,7 @@
 #pragma once
 
 #include <raft/common/nvtx.hpp>
+#include <raft/core/copy.hpp>
 #include <raft/core/device_mdarray.hpp>
 #include <raft/core/device_mdspan.hpp>
 #include <raft/core/host_mdarray.hpp>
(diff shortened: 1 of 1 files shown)


achirkin:
Converting this to a draft until we properly benchmark the change.

tarang-jain (commented):
Thanks for the change. LGTM! But it would be nice to compare the perf in your benchmarking results.

cpp/include/raft/matrix/detail/gather.cuh
```diff
@@ -350,20 +351,23 @@ void gather_if(const InputIteratorT in,
...
+  raft::common::nvtx::range<common::nvtx::domain::raft> fun_scope(
+    "gather_host_buff(%zu, %zu)", uint64_t(batch_size), uint64_t(dim));
 
 #pragma omp for
   for (IdxT i = 0; i < batch_size; i++) {
     IdxT in_idx = indices(offset + i);
-    for (IdxT k = 0; k < buff.extent(1); k++) {
-      buff(i, k) = dataset(in_idx, k);
-    }
+    raft::copy(res,
```
tarang-jain, line 368: So the resource is actually unused since the src and dest are both in host mem? I wonder if we can create a resource on the fly and supply it to `raft::copy()` rather than adding the resource to `gather_buff`'s signature.

cjnolet:
@achirkin what's the status of this PR? Is it something you are still hoping to get into 25.02? Is it stale? Should it be closed?

PR #2466, opened 2024-10-03
Prune workflows based on changed files

KyleFromNVIDIA:
Contributes to https://github.com/rapidsai/build-planning/issues/94


jameslamb (approved):
Could you add `.pre-commit-config.yaml` to these, similar to https://github.com/rapidsai/cuml/pull/6094#pullrequestreview-2345951965?

Otherwise, looks good to me, I don't need to re-review.

PR #2467, opened 2024-10-08
WIP: make package installations in CI stricter

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/106

Proposes specifying the RAPIDS version in `conda install` calls in CI that install CI artifacts, to reduce the risk of CI jobs picking up artifacts from other releases.

Also proposes combining together successive `pip install` calls. `pip install AB` is safer than `pip install A; pip install B` because `pip` doesn't take the current set of installed packages into consideration when it installs new packages.

### DIFF
--- a/ci/build_docs.sh
+++ b/ci/build_docs.sh
@@ -6,6 +6,9 @@ set -euo pipefail
 rapids-logger "Create test conda environment"
 . /opt/conda/etc/profile.d/conda.sh
 
+RAPIDS_VERSION_MAJOR_MINOR="$(rapids-version-major-minor)"
+export RAPIDS_VERSION_NUMBER="$RAPIDS_VERSION_MAJOR_MINOR"
+
 rapids-dependency-file-generator \
   --output conda \
   --file-key docs \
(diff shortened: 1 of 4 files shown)


ci/build_docs.sh
```diff
@@ -23,14 +26,11 @@ PYTHON_CHANNEL=$(rapids-download-conda-from-s3 python)
...
   --channel "${CPP_CHANNEL}" \
   --channel "${PYTHON_CHANNEL}" \
-  libraft \
-  libraft-headers \
-  pylibraft \
-  raft-dask
+  "libraft=${RAPIDS_VERSION_MAJOR_MINOR}" \
+  "libraft-headers=${RAPIDS_VERSION_MAJOR_MINOR}" \
+  "pylibraft=${RAPIDS_VERSION_MAJOR_MINOR}" \
+  "raft-dask=${RAPIDS_VERSION_MAJOR_MINOR}"
```
KyleFromNVIDIA, line 32: Would this be better suited for a DFG file key?

ci/build_docs.sh
```diff
@@ -23,14 +26,11 @@ PYTHON_CHANNEL=$(rapids-download-conda-from-s3 python)
...
 
-export RAPIDS_VERSION="$(rapids-version)"
```
KyleFromNVIDIA, line 31: Why has `RAPIDS_VERSION` been removed?

ci/build_docs.sh
jameslamb: That'd be a more invasive and therefore time-consuming change.

Across these PRs, I'm intentionally not involving `dependencies.yaml` in this because that introduces other complexity I don't want to invest the time in taking on right now. https://github.com/rapidsai/build-planning/issues/22 tracks the work of doing that.

ci/build_docs.sh
jameslamb: As far as I can tell, it's unnecessary here. It doesn't affect anything from `gha-tools` or anything else invoked in this script.

I suspected that it was copied from conda-building scripts.

KyleFromNVIDIA (approved):

ci/build_docs.sh
jameslamb: Hey @KyleFromNVIDIA , I just requested another review because I pushed more changes.... could you look at the new, even stricter form I'm proposing here now?

Based on https://github.com/rapidsai/build-planning/issues/106#issuecomment-2402728240

* using the output of `rapids-version` instead of `rapids-version-major-minor` in constraints, to avoid falling back to older nightlies in the same major release
* setting `RAPIDS_VERSION_NUMBER` as a shell variable when running the one command that needs it (`rapids-upload-docs`) to make it clearer how it's used

KyleFromNVIDIA (approved):

PR #2478, opened 2024-11-01
Do not initialize the pinned mdarray at construction time

achirkin:
`thrust::host_vector` initializes its elements at creation and requires the element type be default-constructible.
This translates to `raft::pinned_mdarray` and makes the mdarray unusable for non-default-constructible objects, like `cuda::atomic<>` (and many user-defined types). This is against all other mdarray types in raft, which are based on `rmm::device_uvector` and are not initialized at construction time.
The PR changes the underlying container to a plain pointer + cudaMallocHost/cudaFreeHost.

**Breaking change**: if anyone relies on the `pinned_mdarray` to initialize itself, the code will break (but mdarrays should not initialize at construction in raft anyway). The affected classes have different private members now, so the ABI changes as well.

### DIFF
--- a/cpp/include/raft/core/pinned_container_policy.hpp
+++ b/cpp/include/raft/core/pinned_container_policy.hpp
@@ -63,7 +63,11 @@ struct pinned_container {
   }
 
   pinned_container(pinned_container&& other) { std::swap(this->data_, other.data_); }
-  pinned_container& operator=(pinned_container&& other) { std::swap(this->data_, other.data_); }
+  pinned_container& operator=(pinned_container&& other)
+  {
+    std::swap(this->data_, other.data_);
+    return *this;
+  }
   pinned_container(pinned_container const&) = delete;  // Copying disallowed: one array one owner
   pinned_container& operator=(pinned_container const&) = delete;
 
--- a/cpp/include/raft/core/pinned_container_policy.hpp
+++ b/cpp/include/raft/core/pinned_container_policy.hpp
@@ -108,8 +108,8 @@ struct pinned_vector_policy {
   {
     return c[n];
   }
-  [[nodiscard]] constexpr auto access(container_type const& c,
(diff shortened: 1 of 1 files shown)


achirkin:
CC @cjnolet @benfred do you have any objections to this? I may be missing some context on why the previous implementation was based on thrust.

achirkin:
The problem is encountered in https://github.com/rapidsai/cuvs/pull/261, which uses a custom container as a workaround at the moment.

wphicks (commented):
I'm pretty sure I introduced pinned mdarrays with the mdbuffer PR. I was squeamish about this at the time but was trying to stick to "official" container types from Thrust/RMM. If we still don't have a good pinned uvector type, though, I'm definitely in support of this change.

Would we want to do the allocation through RMM's pinned memory resource, or are we okay with a raw cudaMallocHost here?

achirkin:
Thanks, that's a good point! I've looked at the RMM resources and there were two pinned memory resources, so I was not sure which not to use. But perhaps more importantly, they have the stream semantics in their API, which is a bit misleading in the context of pinned arrays, which are created on the host and do not offer any stream ordering.

achirkin:
It's also worth noting, that we don't need the `.resize()` functionality for the mdarrays, which makes the implementation extremely simple compared to RMM.

wphicks (approved):
Good question on which pinned memory resource to use. As near as I can tell, the only difference is that pinned_host_memory_resource uses `cudaHostAlloc` while pinned_memory_resource uses `cudaMallocHost`. For modern platforms, I believe we should prefer the `cudaHostAlloc` version. In terms of stream ordering, I'm not really sure why we pass a stream there. Maybe just for API compatibility with other memory resources?

Regardless, I'd definitely support merging as is. How we perform the allocation is an implementation detail we can revisit later unless we have easy answers right now.

Issue #2465, opened 2024-10-03
[FEA] Remove dependency of `build_comms_nccl_only` on UCP


viclafargue:
**Is your feature request related to a problem? Please describe.**
The `raft::comms::build_comms_nccl_only` function should be able to be built without requiring UCP as a dependency. Also, taking note that https://github.com/rapidsai/cuvs/pull/231 should be updated to revert the usage of raw NCCL calls and replace them by the RAFT comms.

Issue #2459, opened 2024-09-27
[FEA] Improve NCCL clique initialization and


viclafargue:
**Is your feature request related to a problem? Please describe.**
The unification of the API layers and removal of the `mg` namespace as described in https://github.com/rapidsai/cuvs/issues/357 require some changes on the RAFT end. Namely, the NCCL clique should now be a core type and its presence in the resource handle inform on the algorithm implementation to run. The PR resolving this issue on the RAFT side should :
- Set the NCCL clique as a core type
- Separate NCCL clique initialization from its access and improve the initialization process
- Leave a separate access function to be used internally by the cuVS library

**Describe the solution you'd like**
- The `nccl_clique.hpp` file should be placed in the `raft/core` directory and the nccl_clique struct should be placed in the `raft::core` namespace.
- A `raft::resource::initialize_nccl_clique()` function to initialize a NCCL clique and add it to a resource handler. This function would be called before calling an algorithm implementation. The presence of the NCCL clique resource on the resource handler would inform the willingness to run the algorithms in multi-GPU mode. The function could also allow the configuration of the GPUs to include during clique initialization and the percentage of device memory to pre-allocate as a memory pool on each.
- A `raft::resource::get_nccl_clique()` function to access the NCCL clique internally inside of implementations.

PR #2487, opened 2024-11-08
Introduction of the `raft::device_resources_snmg` type

viclafargue:
Introduces the `raft::device_resources_snmg` type to hold all resources required for the NCCL clique.

~Answers https://github.com/rapidsai/raft/issues/2459~
Removed call to `raft::comms::build_comms_nccl_only` (https://github.com/rapidsai/raft/issues/2465)


cpp/include/raft/core/nccl_clique.hpp
```diff
@@ -153,4 +145,4 @@ struct nccl_clique {
   std::vector<raft::device_resources> device_resources_;
```
cjnolet, line 145: This is strange- device_resources has a nccl clique but a nccl clique has device resources? Please avoid the circular dependency here and just have the nccl_clique use the actua resources it needs (I assume that's just a stream?)

cpp/include/raft/core/resource/nccl_clique.hpp
```diff
@@ -25,38 +25,102 @@ namespace raft::resource {
 
 class nccl_clique_resource : public resource {
  public:
-  nccl_clique_resource() : clique_(std::make_unique<raft::comms::nccl_clique>()) {}
+  nccl_clique_resource(std::optional<std::vector<int>>& device_ids, int percent_of_free_memory)
+  {
+    if (device_ids.has_value()) {
+      clique_ = std::make_unique<raft::core::nccl_clique>(*device_ids, percent_of_free_memory);
+    } else {
+      clique_ = std::make_unique<raft::core::nccl_clique>(percent_of_free_memory);
```
cjnolet, line 33: What happens when the percent of free memory is used up? Does it automatically try to allocate more? Please document this class and the constructor.

cpp/include/raft/core/resource/nccl_clique.hpp
```diff
@@ -25,38 +25,102 @@ namespace raft::resource {
...
+{
+  return build_nccl_clique(res, std::nullopt, percent_of_free_memory);
+};
+
+/**
+ * Initializes a NCCL clique and sets it into a raft resource instance
+ *
+ * @param[in] res the raft resources object
+ * @param[in] device_ids selection of GPUs initialize the clique on
+ * @param[in] percent_of_free_memory percentage of device memory to pre-allocate as a memory pool on
```
cjnolet, line 101: Can you expan upon this briefly? It would help future eyes and users when figuring out what to set if they know how the memory gets used (and whether it grows or throws an OOM when it's used up).

viclafargue: I removed the upfront allocation of memory and added a method to optionally allocate a pool on every GPUs instead. If the user prefers, he should still be able to access each `device_resources` individually to configure them in a custom way of his choice.

cpp/include/raft/core/device_resources_snmg.hpp
```diff
@@ -0,0 +1,148 @@
...
+      RAFT_CUDA_TRY(cudaSetDevice(get_device_id(rank)));
+      size_t limit =
+        rmm::percent_of_free_device_memory(percent_of_free_memory);  // check limit for each device
+      raft::resource::set_workspace_to_pool_resource(get_device_resources(rank), limit);
+    }
+  }
+
+  ~device_resources_snmg()
+  {
+#pragma omp parallel for  // necessary to avoid hangs
```
cjnolet, line 120: Please add proper documentation to each of the methods in this class (even private methods). It's important for users of this API to know that they'll need to have multiple threads available for this, otherwise it'll hang.

cpp/include/raft/core/device_resources_snmg.hpp
```diff
@@ -0,0 +1,148 @@
...
+  inline const raft::device_resources& set_current_device_to_rank(int rank) const
+  {
+    RAFT_CUDA_TRY(cudaSetDevice(get_device_id(rank)));
+    return get_device_resources(rank);
+  }
+
+  void set_memory_pool(int percent_of_free_memory) const
+  {
+    for (int rank = 0; rank < get_num_ranks(); rank++) {
+      RAFT_CUDA_TRY(cudaSetDevice(get_device_id(rank)));
```
cjnolet, line 111: Doesn't `cudaSetDevice` have an overhead?

cpp/include/raft/core/device_resources_snmg.hpp
```diff
@@ -0,0 +1,148 @@
...
+                    #call,                         \
+                    status,                        \
+                    ncclGetErrorString(status));   \
+      throw raft::logic_error(msg);                \
+    }                                              \
+  } while (0);
+
+namespace raft {
+
+class device_resources_snmg : public resources {
```
cjnolet, line 50: Since this is a `device_resources` instance, I think it'll be better to extend `device_resources`. That way you can override important methods like `get_stream` and `get_worker_stream` to return a main stream. In general, we should expect the `device_resources_snmg` to behave the same as `device_resources` when passed into an algorithm that only operates single gpu while being able to operate in `snmg` mode when passed into an algorithm that supports single-node multi-gpu. I think it should store off and use the id of the GPU that is selected when creating the snmg handle by default when being used in single-gpu mode.

cpp/include/raft/core/device_resources_snmg.hpp
viclafargue: This method is optionally called once. The `cudaSetDevice` function have a small overhead, but it is necessary at the very least to check available free memory.

cpp/include/raft/core/device_resources_snmg.hpp
```diff
@@ -0,0 +1,213 @@
...
+  } while (0);
+
+namespace raft {
+
+/**
+ * @brief SNMG resource container object that stores a NCCL clique and
+ * all necessary resources used for calling device functions, cuda kernels,
+ * libraries and/or NCCL communications on each GPU
+ *
+ * Note that the NCCL communications sometimes require
```
cjnolet, line 55: Are we able to expand upon this a little bit more here? Are there specific things the user should be aware of or avoid? Are the SNMG-copable algorithms responsible for this? Are there any threading resources needed in order to make this possible? Is it 1 thread per gpu? Do we need to provide a thread pool here too?

cpp/include/raft/core/device_resources_snmg.hpp
```diff
@@ -0,0 +1,213 @@
...
+ * the use of multiple threads to avoid hangs
+ */
+class device_resources_snmg : public device_resources {
+ public:
+  /**
+   * @brief Construct a SNMG resources instance with all available GPUs
+   */
+  device_resources_snmg() : device_resources(), root_rank_(0)
+  {
+    cudaGetDevice(&main_gpu_id_);
```
cjnolet, line 65: We should probably document this as the expected behavior or allow the user to set the id of the gpu that should be used as the "default" just in case they would like it to be different than one of the GPUs in the nccl clique for some reason.

docs/source/cpp_api/core_resources.rst
```diff
@@ -55,6 +55,20 @@ namespace *raft::core*
     :project: RAFT
     :members:
 
+SNMG Device Resources
+---------------------
+
+The `raft::device_resources_snmg` provides a convenient way
+to setup a NCCL clique and access device-related resources on each device.
```
cjnolet, line 62: Maybe a little more description here about what this means. What does SNMG mean? How do we expect it to be used? What are the imolications/caveats? I think it's safe to mention NCCL here, since after all these years we don't really have an alternative.

cjnolet (commented):
Took another look and I think the implementation itself is great. The only suggestions I have are related to the docs. I understand this is a lower-level detail for most users but we should strive to provide as much info and details as possible for both end-users AND for future implementors of SNMG algorithms.

cpp/include/raft/core/device_resources_snmg.hpp
viclafargue: Sure, I will expand a bit on this mentioning more specifically the blocking NCCL calls such as send&receive. NCCL clique destruction also require multiple threads, but this is managed in the destructor method. Yes, SNMG-capable algorithms will be responsible for this. OpenMP can be used for this, but other libs will work too. I think it is preferable to stay agnostic to the lib used. A user could therefore create his own thread pool or simply use OpenMP loops. This do not necessarily have to be 1 thread per GPU, and if exchanges are managed separately, they could even take place in a single thread thanks to the use of a NCCL group. But, since in most cases exchanges happen in the middle of other computations, and computations are for the most part similar for every GPU, it is convenient to have 1 thread per GPU. That's why I felt like the user should be made aware of his responsibility for managing NCCL comms accross threads.

cpp/include/raft/core/device_resources_snmg.hpp
viclafargue: Expanded the documentation on that point. The default GPU is the GPU currently in use at the time the constructor is called. It could theoretically not be part of the clique. However, I worry a bit that allowing the user to change the default GPU through a constructor parameter might cause the `device_resources` constructor to be first called on the wrong GPU and then be associated with an other GPU which might cause issues down the line. Also, in term of consistency the `device_resources` object is always associated to the GPU it was instantiated with.

cjnolet (approved):

bdice:
@viclafargue @cjnolet This `nccl_clique.hpp` header removal appears to have broken cuVS builds: https://github.com/rapidsai/cuvs/actions/runs/12819602008/job/35747622539?pr=567#step:8:991

I don't see any cuVS PRs with a fix yet.

cjnolet:
@bdice this is the pr with the fix, I'm just trying to get it through CI: https://github.com/rapidsai/cuvs/pull/454

PR #2490, opened 2024-11-13
enforce wheel size limits and README formatting in CI, put a ceiling on Cython dependency

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/110

Proposes adding 2 types of validation on wheels in CI, to ensure we continue to produce wheels that are suitable for PyPI.

* checks on wheel size (compressed),
  - *to be sure they're under PyPI limits*
  - *and to prompt discussion on PRs that significantly increase wheel sizes*
* checks on README formatting
  - *to ensure they'll render properly as the PyPI project homepages*
  - *e.g. like how https://github.com/scikit-learn/scikit-learn/blob/main/README.rst becomes https://pypi.org/project/scikit-learn/*

Also puts a ceiling on Cython to its latest stable release (`<=3.0.11`), to fix https://github.com/rapidsai/raft/pull/2490#issuecomment-2474375209. Work to relax that is tracked in (https://github.com/rapidsai/raft/issues/2491).


jameslamb:
Seeing build issues like this when building `pylibraft`:

```text
    Error compiling Cython file:
    ------------------------------------------------------------
    ...
  
        shape = (cai.shape[0], cai.shape[1] if len(cai.shape) == 2 else 1)
  
        return make_device_matrix_view[int64_t, int64_t, row_major](
  
            <int64_t*><uintptr_t>cai.data, shape[0], shape[1])
 
    cdef device_matrix_view[const_float, int64_t, row_major] \ 
    ^
 
    ------------------------------------------------------------
  
    /home/coder/raft/python/pylibraft/pylibraft/common/mdspan.pyx:203:0: Function signature does not match previous declaration
```

([build link](https://github.com/rapidsai/raft/actions/runs/11811852559/job/32906119739?pr=2490#step:7:5761))

vyasr:
Pasting my message from Slack:

> My guess is that that this is happening because the pip devcontainer [is using the Cython 3.1 alpha](https://github.com/rapidsai/raft/actions/runs/11811852559/job/32906119739?pr=2490#step:7:1186), while [the normal wheel builds are not](https://github.com/rapidsai/raft/actions/runs/11811852559/job/32906162870?pr=2490#step:13:113). The alpha [was released five days ago](https://pypi.org/project/Cython/3.1.0a1/#history). I'm guessing there are some bugs in there that need sorting out. The issue is that (if you recall) our devcontainer builds install with --pre because of https://github.com/rapidsai/build-planning/issues/85 (otherwise we could specify alpha versions explicitly for dask/distributed and not need the global --pre)

jameslamb:
Thanks @vyasr ! It looks like adding the ceiling on Cython worked 🎉 

Put up https://github.com/rapidsai/raft/issues/2491 to track the work of relaxing that ceiling in the future.

Some CI jobs here are failing for what look like temporary issues, but I don't want to stop the in-progress C++ test job to restart everything, because those take several hours to run 😬 

I'll keep checking back and hopefully we can get a clean CI run and then merge this.

dependencies.yaml
```diff
@@ -164,7 +164,7 @@ dependencies:
       - output_types: [conda, requirements, pyproject]
         packages:
           - &cmake_ver cmake>=3.26.4,!=3.30.0
-          - cython>=3.0.0
+          - cython>=3.0.0,<=3.1.0a0
```
bdice, line 167: Is this supposed to be strictly less than?
```suggestion
          - cython>=3.0.0,<3.1.0a0
```

python/pylibraft/pyproject.toml
```diff
@@ -123,13 +123,21 @@ build-backend = "scikit_build_core.build"
...
 dependencies-file = "../../dependencies.yaml"
 matrix-entry = "cuda_suffixed=true;use_cuda_wheels=true"
 
+[tool.pydistcheck]
+select = [
+    "distro-too-large-compressed",
+]
+
+# detect when package size grows significantly
+max_allowed_size_compressed = '825M'
```
bdice, line 139: FYI, this size should decrease a lot with #2488.

bdice (approved):
One question / one comment. Overall it should be fine.

dependencies.yaml
jameslamb: Ahhh yes!!! Thank you for catching this 🙈 

Updated all of them in https://github.com/rapidsai/raft/pull/2490/commits/42064859d662c67bf342f6b79633e63a52c66ce2

python/pylibraft/pyproject.toml
jameslamb: 🥳 hey awesome! We should decrease this in #2488 or some follow-up PR after.

PR #2494, opened 2024-11-14
Disable hnswlib if not compiling shared library components of RAFT

divyegala:
As seen in https://github.com/rapidsai/cuvs/pull/465 when building newer versions of hnswlib, version conflicts are found to be not resolvable. The effect of this PR will be that `libraft-headers` conda package will not be configured with `hnswlib`.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -602,6 +602,7 @@ if(RAFT_COMPILE_LIBRARY)
              ${RAFT_CTK_MATH_DEPENDENCIES} # TODO: Once `raft::resources` is used everywhere, this
                                            # will just be cublas
              $<TARGET_NAME_IF_EXISTS:OpenMP::OpenMP_CXX>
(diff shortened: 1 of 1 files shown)


cpp/CMakeLists.txt
```diff
@@ -605,6 +602,7 @@ if(RAFT_COMPILE_LIBRARY)
              ${RAFT_CTK_MATH_DEPENDENCIES} # TODO: Once `raft::resources` is used everywhere, this
                                            # will just be cublas
              $<TARGET_NAME_IF_EXISTS:OpenMP::OpenMP_CXX>
+             $<TARGET_NAME_IF_EXISTS:hnswlib::hnswlib>
```
lowener, line 605: If the user wants to build with HNSW but the target name does not exist will there be an error or will it fail silently? A bad scenario would be to compile the whole library only to find that hnswlib was not included in this compilation when it was explicitly requested.

bdice:
Is this PR still relevant after #2498?

PR #2498, opened 2024-11-15
2412 remove libraft vss instantiations

cjnolet:
We are keeping random ball cover headers in RAFT for 24.12, and random ball cover depends on distances and brute-force. Because of this, we're going to leave all of the VSS headers in RAFT for the time being, and will remove them all in a future PR once RBC is formally migrated to cuVS. The tests, benchmarks, and instantiations for all of these APIs will be removed, though, so while the actual headers can still be used, they are no longer being tested and could fail without warning. 

I've also included a note to users in the README about this, stating to use at their own risk.


benfred (approved):
lgtm! 

I'm excited to get these changes in - I'm imagining this will really improve the build times here

bdice (approved):
Thanks for all the effort here. I pushed a few small fixes, and have a few questions for follow up work:

1. Do we want to delete `cpp/template/cmake/thirdparty/get_raft.cmake`? It is the only file left in `cpp/template/*`. If we do remove it, we also need to clean up a reference in `ci/release/update-version.sh`.
2. Are we able to purge all remaining instances of `RAFT_EXPLICIT_INSTANTIATE_ONLY`? There are quite a few files still mentioning this, but I think all of these are deprecated and thus we should be able to remove `RAFT_EXPLICIT_INSTANTIATE_ONLY`? Also this section of the docs could be removed or greatly simplified, I think. https://github.com/rapidsai/raft/blob/branch-24.12/docs/source/developer_guide.md#header-organization-of-expensive-function-templates
3. There are quite a few remaining references to `#include <hnswlib/hnswlib.h>` but we are no longer including the CMake code to find hnswlib. Is that a problem? Are these features usable (despite being unsupported) in the current state? I would not expect them to work anymore. It might make more sense to just purge all of this, because there's no use in keeping headers that aren't going to work with the current build system / missing dependencies.

cjnolet:
1. Looks like that file was brought back, likely as a result of a bad merge. We can remove it in a follow on.
2. This is okay. We do still have some instantiations and the important headers are no longer using it.
3. This is okay. I'm not removing the headers in this PR, so these will be left around. Users have been warned about this for some time, so they'll need to supply their own hnswlib if they still want to use those headers.

bdice:
>> Do we want to delete cpp/template/cmake/thirdparty/get_raft.cmake? It is the only file left in cpp/template/*. If we do remove it, we also need to clean up a reference in ci/release/update-version.sh.
>
> Looks like that file was brought back, likely as a result of a bad merge. We can remove it in a follow on.

I filed https://github.com/rapidsai/raft/pull/2500 to address this.

PR #2508, opened 2024-11-25
Forward-merge branch-24.12 to branch-25.02

bdice:
Manual forward merge from 24.12 to 25.02. This PR should not be squashed.


bdice:
We should probably let CI run on this forward-merge PR because so much has changed. Then I'll request an admin non-squash merge.

PR #2524, opened 2024-12-09
Fix rnd bit generation in rmat_rectangular_kernel

tfeher:
For certain architectures, the compiler always generates zero destination bit in the following loop https://github.com/rapidsai/raft/blob/ee45ce786686b54d1972408b927d7fcd8ce0cf20/cpp/include/raft/random/detail/rmat_rectangular_generator.cuh#L160-L162 irrespective of the random value that shall determine which bit to use for `dst_id`.

This PR refactors the loop. This way the `dst_id` number has the desired random distribution for all bits.

### DIFF
--- a/cpp/include/raft/random/detail/rmat_rectangular_generator.cuh
+++ b/cpp/include/raft/random/detail/rmat_rectangular_generator.cuh
@@ -151,15 +151,16 @@ RAFT_KERNEL rmat_gen_kernel(IdxT* out,
   raft::random::PCGenerator gen{r.seed, r.base_subsequence + idx, 0};
   auto min_scale = min(r_scale, c_scale);
   IdxT i         = 0;
-  for (; i < min_scale; ++i) {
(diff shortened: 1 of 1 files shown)


tfeher:
LinewiseOp test goes OOM, restarted the test.
```
[----------] 9 tests from LinewiseOp/TenGigs_float_uint64_t
[ RUN      ] LinewiseOp/TenGigs_float_uint64_t.run/0
unknown file: Failure
C++ exception with description "std::bad_alloc: out_of_memory: CUDA error at: /opt/conda/envs/test/include/rmm/mr/device/cuda_memory_resource.hpp:62: cudaErrorMemoryAllocation out of memory" thrown in the test body.
[  FAILED  ] LinewiseOp/TenGigs_float_uint64_t.run/0, where GetParam() = (0, 0) (32 ms
```

cjnolet (approved):

PR #2497, opened 2024-11-15
Remove raft-ann-bench

cjnolet:
This PR removes raft-ann-bench from the conda packages, build system, and documentation. This removal was previously announced for the 24.12 release in #2448.

PR #2529, opened 2024-12-16
reduce duplication, removed unused things in dependencies.yaml

jameslamb:
Proposes some small cleanup for `dependencies.yaml`

* removes `rapids_build_setuptools` dependency group
  - *#2497 removed the last use of `setuptools` here*
* breaks `cuda-python` and `rmm` out into `depends_on_*` groups to reduce duplication, and for consistency with other RAPIDS projects ([docs explaining this](https://github.com/rapidsai/build-planning/blob/d9e3c606d95c835ee384ac6480a4af0ac6cb024a/docs/docs/packaging.md#L181))
* alphabetizes lists

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -6,22 +6,22 @@ files:
       cuda: ["11.8", "12.5"]
       arch: [x86_64, aarch64]
     includes:
-      - rapids_build
-      - build_pylibraft
+      - checks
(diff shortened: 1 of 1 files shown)


bdice (approved):
Thanks. I've been meaning to clean some of this up for a while.

PR #2531, opened 2024-12-17
introduce libraft wheels

jameslamb:
Replaces #2306, contributes to https://github.com/rapidsai/build-planning/issues/33.

Proposes packaging `libraft` as a wheel, which is then re-used by:

* `pylibraft-cu{11,12}` and `raft-cu{11,12}` (this PR)
* `libcugraph-cu{11,12}`, `pylibcugraph-cu{11,12}`, and `cugraph-cu{11,12}` in https://github.com/rapidsai/cugraph/pull/4804
* `libcuml-cu{11,12}` and `cuml-cu{11,12}` in https://github.com/rapidsai/cuml/pull/6199

As part of this, also proposes:

* introducing a new CMake option, `RAFT_COMPILE_DYNAMIC_ONLY`, to allow building/installing only the dynamic shared library (i.e. skipping the static library)
* enforcing `rapids-cmake`'s preferred CMake style (https://github.com/rapidsai/raft/pull/2531#discussion_r1917039870)
* making wheel-building CI jobs always depend on other wheel-building CI jobs, not tests or `*-publish` (to reduce end-to-end CI time)

### Benefits of these changes

* smaller wheels (see "Size Changes" below)
* faster compile times (no more re-compiling RAFT in cuGraph and cuML CI)
* other benefits mentioned in https://github.com/rapidsai/build-planning/issues/33

### Wheel contents

`libraft`:

* `libraft.so` (shared library)
* RAFT headers
* vendored dependencies (`fmt`, CCCL, `cuco`, `cute`, `cutlass`)

`pylibraft`:

* `pylibraft` Python / Cython code and compiled Cython extensions

`raft-dask`:

* `raft-dask` Python / Cython code and compiled Cython extension

### Dependency Flows

In short.... `libraft` contains a `libraft.so` dynamic library and the headers to link against it.

* Anything that needs to link against RAFT at build time pulls in `libraft` wheels as a build dependency.
* Anything that needs RAFT's symbols at runtime pulls it in as a runtime dependency, and calls `libraft.load_library()`.

For more details and some flowcharts, see https://github.com/rapidsai/build-planning/issues/33#issuecomment-2590129852

### Size changes (CUDA 12, Python 3.12, x86_64)

| wheel                | num files (before) | num files (these PRs) | size (before)  | size (these PRs) |
|:---------------:|------------------:|-----------------:|--------------:|-------------:|
| `libraft`.            |   ---                       |  3169                      | ---                   | 19M                |
[...]

PR #2532, opened 2024-12-17
remove unused 'joblib' and 'numba' dependencies, other packaging cleanup

jameslamb:
Proposes some cleanup of packaging details, noticed while I was working on #2531

* removes runtime dependencies on `joblib` and `numba` for `raft-dask`
   - *`raft-dask` doesn't directly import from these libraries, and the git blame didn't suggest any other reason that they were being pinned here*
   - *checked with `git grep -E 'joblib|numba'`
* removes `setup.cfg` files
   - *these are currently being ignored by tools, in favor of identical configuration in `pyproject.toml` and `.flake8` files*
   - e.g. https://github.com/rapidsai/raft/blob/bfd190687ee396374b7106d9ac26add73b57b22a/.pre-commit-config.yaml#L16-L19
* packages license files in conda packages
  - *think these were just missed in the round of PRs like this: https://github.com/rapidsai/cuml/pull/6061*
* removes some outdated / inaccurate comments in packaging configs


bdice (approved):
LGTM! It seems like `setup.cfg` is mostly duplicating the settings we already have in `pyproject.toml`, so I assume this was an oversight and not removed earlier. Either way the final state looks good to me.

cjnolet (approved):
LGTM no concerns from my end.

PR #2535, opened 2025-01-07
Add support for different data type of bitset

lowener:
This PR is useful for Milvus.
Previously the `bitset_view` object only supported the data type used to create the bitset. With the proposed modifications, a `bitset_view` object can support any data type used to create the bitset by specifying the `original_nbits` parameter in the class constructor.


cjnolet (approved):

cpp/include/raft/core/bitset.cuh
```diff
@@ -32,12 +32,41 @@
 
 namespace raft::core {
 
+template <typename index_t>
+_RAFT_HOST_DEVICE void inline compute_original_nbits_position(const index_t original_nbits,
```
rhdong, line 36: It looks like the original endianness is not considered, if we have potential restrictions on the application scenario, we should let the caller know via assert or doc.

rhdong:
Hi @lowener , it's a great job to support any type of original data. Just a little worried that we may consider more than the original request from customers. I recall they requested to support only one type(uint8_t), am I right? If that's true, it's unnecessary to consider the `original_nbits_ > nbits`, and simply `interpret_cast<uint8_t>` to the type we supported. That means we can remove `original_nbits`,  which the caller doesn't need to provide originally. In my opinion, the `original_nbits` looks a little bit unnatural, as the `bitset_t` is a template argument, while, we provide the type information for the original bits type via a function argument. --- Please feel free to ignore the comment, just my personal opinion instead of regular comments.

lowener:
The use-case for this is for customers that have already created their own bitset with the data type that they chose (in this specific case `uint8_t`) and want to query a cuVS index.
CuVS indexes all consider that the bitset given as input is using `uint32_t` (see [here](https://github.com/rapidsai/cuvs/blob/branch-25.02/cpp/src/neighbors/cagra.cuh#L349-L351))
```cpp
auto& sample_filter =
      dynamic_cast<const cuvs::neighbors::filtering::bitset_filter<uint32_t, int64_t>&>(
        sample_filter_ref);
```
So if the customer wants their bitset to work with cuVS they will have to re-order it on their side to fit a `uint32_t` order since on our side we are reading in a `uint32_t` order no matter what at the moment. With the proposed solution, a customer can come with their own bitset on host, transfer it to device and directly use that bitset without re-ordering, because we will read the bitset with the order requested.
We will also be able to handle any type originally used by any customer. So if there is at some point a customer using `uint64_t` bitset there won't be any code change needed on cuVS side.

I agree that it looks a bit unnatural but it is enabling a lot of flexibility for customers.

cpp/include/raft/core/bitset.cuh
lowener: I don't think endianness should be considered in this situation, because the bitset is getting created and used on the same system.

rhdong (approved):
LGTM

rhdong:
> The use-case for this is for customers that have already created their own bitset with the data type that they chose (in this specific case `uint8_t`) and want to query a cuVS index. CuVS indexes all consider that the bitset given as input is using `uint32_t` (see [here](https://github.com/rapidsai/cuvs/blob/branch-25.02/cpp/src/neighbors/cagra.cuh#L349-L351))
> 
> ```c++
> auto& sample_filter =
>       dynamic_cast<const cuvs::neighbors::filtering::bitset_filter<uint32_t, int64_t>&>(
>         sample_filter_ref);
> ```
> 
> So if the customer wants their bitset to work with cuVS they will have to re-order it on their side to fit a `uint32_t` order since on our side we are reading in a `uint32_t` order no matter what at the moment. With the proposed solution, a customer can come with their own bitset on host, transfer it to device and directly use that bitset without re-ordering, because we will read the bitset with the order requested. We will also be able to handle any type originally used by any customer. So if there is at some point a customer using `uint64_t` bitset there won't be any code change needed on cuVS side.
> 
> I agree that it looks a bit unnatural but it is enabling a lot of flexibility for customers.

That makes sense! Thank you for the clarification!

cjnolet:
> I agree that it looks a bit unnatural but it is enabling a lot of flexibility for customers.

@lowener are these assumptions / caveats documented in the cuVS APIs (and raft apis if needed) to make sure users are aware of this when building their bitsets?

rhdong (approved):

lowener:
I added this to the doc in this Raft PR, but it needs to be documented on cuVS side as well

PR #2537, opened 2025-01-10
Fix broken link to python doc

lowener:
Apply the same change as https://github.com/rapidsai/cuml/pull/6202 to fix Python links to source code

### DIFF
--- a/docs/source/sphinxext/github_link.py
+++ b/docs/source/sphinxext/github_link.py
@@ -116,13 +116,14 @@ def _linkcode_resolve(domain, info, package, url_fmt, revision):
         else:
             return
     else:
-        # Test if we are absolute or not (pyx are relative)
-        if (not os.path.isabs(fn)):
-            # Should be relative to docs right now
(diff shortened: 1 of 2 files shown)


docs/source/sphinxext/github_link.py
```diff
@@ -96,15 +111,14 @@ def _linkcode_resolve(domain, info, package, url_fmt, revision):
...
         else:
             return
     else:
-        # Test if we are absolute or not (pyx are relative)
-        if (not os.path.isabs(fn)):
-            # Should be relative to docs right now
-            fn = os.path.abspath(os.path.join("..", "python", fn))
+        if fn.endswith(".pyx"):
+            sp_path = next(x for x in sys.path if re.match(".*site-packages$", x))
+            fn = fn.replace("/opt/conda/conda-bld/work/python/cuvs", sp_path)
```
cjnolet, line 121: I think we need to make sure this builds for developers and not just by CI, right? Won't this break developers trying to build the docs locally?

cjnolet (approved):

PR #2542, opened 2025-01-14
Fix bit order of RMAT Rectangular Generator to match expectation

mfoerste4:
Previously, the decimal indices of rows/columns of the adjacency matrix did not align with the node-ids created by the algorithm. This PR fixes the bits set for each decision during the computation as described by the docstring.

FYI @tfeher


tfeher (changes requested):
Thanks Malte, this looks good! Just one additional request: it would be great to update the docstring of `thetha` to be explicit about which element of the array describes which bit level.

I would recommend to start with the main docstring, and replace 
https://github.com/rapidsai/raft/blob/596d4b7338e62a92652503cd76feaeaa187ad740/cpp/include/raft/random/rmat_rectangular_generator.cuh#L33-L34

with

```c++
* This function generates a random graph represented by a (sparse) adjacency matrix. As described
* in [1], to generate connections, we recursively subdivide the adjacency matrix into four
* equal-sized partitions, and distribute edges within these partitions with a unequal
* probabilities. The probabilities are described by numbers [a, b, c, d]. We chose the upper left
* partition with probability `a`. The chosen partition is again subdivided into four smaller
* partitions, and the procedure is repeated until we reach a single element (1 × 1 partition).
*
* We can prescribe different probability distribution at each iteariton. The `theta` array stores
* the probability values for each level.
*
* [1] "R-MAT: A Recursive Model for Graph Mining" Deepayan Chakrabarti, Yiping Zhan, Christos
* Faloutsos (2004) https://doi.org/10.1137/1.9781611972740.43

```

After this, we can clarify the layout of `theta` be replacing 

https://github.com/rapidsai/raft/blob/596d4b7338e62a92652503cd76feaeaa187ad740/cpp/include/raft/random/rmat_rectangular_generator.cuh#L52-L56

with 

```c++
* @param[in]  theta   array [on device] with the distribution of each quadrant at each level of
*                     resolution. theta = [a0, b0, c0, d0, a1, b1, c1, d1, ...], where 
*                     [a0, b0, c0, d0]  defines the probability at the finest level (2x2).
*                     The last four elements in the array describe the probability in the
*                     coarsest level (where matrix size = [2^r_scale, 2^c_scale]).
*                     Since these are probabilities, the four [a_i, b_i, c_i, d_i] values for
*                     each level of the RMAT must sum to one.
*                     [dim = max(r_scale, c_scale) x 2 x 2]. 
```

What do you think?

tfeher (approved):
Thanks Malte for the update, LGTM!

PR #2548, opened 2025-01-22
use dynamic CUDA wheels on CUDA 11

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/137

Follow-up to #2531 .

See the linked issue for many more details, but in short... using a dynamically-loaded libraft which has statically-linked cuBLAS causes issues for other libraries.

There are now aarch64 CUDA 11 wheels for cuBLAS and other CUDA libraries, so it's possible to have RAFT wheels dynamically link against them. This PR does that.

## Notes for Reviewers

This has other side benefits in addition to fixing runtime issues... it also simplifies the wheel-building scripts and CMake, and makes CUDA 11 wheels noticeably smaller 😊

### DIFF
--- a/python/libraft/CMakeLists.txt
+++ b/python/libraft/CMakeLists.txt
@@ -22,8 +22,6 @@ project(
   LANGUAGES CXX
 )
 
-option(USE_CUDA_MATH_WHEELS "Use the CUDA math wheels instead of the system libraries" OFF)
-
 # Check if raft is already available. If so, it is the user's responsibility to ensure that the
 # CMake package is also available at build time of the Python raft package.
 find_package(raft "${RAPIDS_VERSION}")
@@ -35,14 +33,8 @@ endif()
 unset(raft_FOUND)
 
 # --- CUDA --- #
-find_package(CUDAToolkit REQUIRED)
 set(CUDA_STATIC_RUNTIME ON)
-set(CUDA_STATIC_MATH_LIBRARIES ON)
-if(CUDAToolkit_VERSION VERSION_GREATER_EQUAL 12.0)
-  set(CUDA_STATIC_MATH_LIBRARIES OFF)
-elseif(USE_CUDA_MATH_WHEELS)
-  message(FATAL_ERROR "Cannot use CUDA math wheels with CUDA < 12.0")
-endif()
+set(CUDA_STATIC_MATH_LIBRARIES OFF)
 
 # --- RAFT ---#
 set(BUILD_TESTS OFF)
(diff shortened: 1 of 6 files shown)


python/libraft/CMakeLists.txt
```diff
@@ -52,14 +44,13 @@ set(RAFT_COMPILE_LIBRARY ON)
...
+# assumes libraft.so is installed 2 levels deep, e.g. site-packages/libraft/lib64/libraft.so
+set_property(
+  TARGET raft_lib
+  PROPERTY INSTALL_RPATH
+           "$ORIGIN/../../nvidia/cublas/lib"
+           "$ORIGIN/../../nvidia/curand/lib"
+           "$ORIGIN/../../nvidia/cusolver/lib"
+           "$ORIGIN/../../nvidia/cusparse/lib"
+           "$ORIGIN/../../nvidia/nvjitlink/lib"
+)
```
jameslamb, line 56: Notice I've added an additional `../` in here... these RPATHs were wrong in #2531 (confirmed by inspecting the paths more closely tonight).

Also confirmed these will work for both CUDA 11 and CUDA 12 wheels.

I suspect we didn't notice this in CI before because the corresponding system-installed libraries were being found instead.

bdice: Is there a way to ensure CI will fail in the future if these are incorrect?

bdice (approved):
All looks fine to me. If this works, I am happy to merge it sooner so we can unblock CUDA 11 wheel CI for other repos.

python/libraft/CMakeLists.txt
jameslamb: I've been thinking about that... there are options but I don't love any of them.

Ways I can think of to test that an RPATH actually ends up getting used:

* set `LD_DEBUG=libs LD_DEBUG_OUTPUT=/tmp/logs` or something to generate debug logs from `ld`, then `grep` those to figure out if libraries were loaded from the expected place
* ensure there aren't any more of these libraries on the system by deleting them ([we used to do something similar for UCX](https://github.com/rapidsai/ucx-py/pull/1041/files#diff-904c7127d3d586dca8c2bcf3055ec8acecc095a9e1d53ff81d94150698c0e784R9-R20))... if loading works, then it must be that the one remaining copy of e.g. `libcublas.so` was used
* hide the those system locations by manipulating `LD_LIBRARY_PATH`, forcing `ld` to either find the wheel-installed libraries or raise an error
* start a long-lived process that loads the library, then use `lsof` to check that the one from site-packages was loaded (https://superuser.com/a/310205/1798685

I definitely don't think we should hold up this PR over this question, but it would be helpful to find a way to catch things like this in CI if we could find a way that isn't too complex.

bdice:
I'm going to trigger a `/merge` based on @jameslamb's successful tests offline (https://github.com/rapidsai/build-planning/issues/137#issuecomment-2606235401), and the fact that CI is passing (aside from one job that is being rerun due to a network failure).

python/libraft/CMakeLists.txt
bdice: I think we might be able to try building CI images without any system CUDA libraries? If we migrate all of RAPIDS to use CUDA wheels, that could be a step in the right direction, but it might also depend on other non-RAPIDS packages like `cupy` to use CUDA wheels, too. 🤔

vyasr: When we first put together the wheels and the images, I actually pushed for us to base [the citestwheel images](https://github.com/rapidsai/ci-imgs/blob/main/citestwheel.Dockerfile) on a base image that did not contain the CUDA libraries. I don't know if that's easily discoverable in the history anywhere. In any case, @bdice's suspicion here is correct; this wound up not working because you couldn't use cupy (or numba) in those images since they require the system libraries.

PR #2549, opened 2025-01-22
Introduction of the `raft::device_resources_snmg` type

viclafargue:


cpp/include/raft/core/device_resources_snmg.hpp
```diff
@@ -0,0 +1,258 @@
...
+#include <numeric>
+#include <vector>
+
+/**
+ * @brief Error checking macro for NCCL runtime API functions.
+ *
+ * Invokes a NCCL runtime API function call, if the call does not return ncclSuccess, throws an
+ * exception detailing the NCCL error that occurred
+ */
+#define RAFT_NCCL_TRY(call)                        \
```
cjnolet, line 34: It seems odd to have this defined here. Isn't this also defined in the communicator? Can we consolidate this into a "nccl.hpp" or "nccl_macros.hpp" file and have `device_resources_snmg` and the communicator import that file?

cpp/include/raft/core/device_resources_snmg.hpp
```diff
@@ -0,0 +1,258 @@
...
+  /**
+   * @brief Construct a SNMG resources instance with all available GPUs
+   */
+  device_resources_snmg() : device_resources()
+  {
+    RAFT_CUDA_TRY(cudaGetDevice(&main_gpu_id_));
+
+    // set root rank
+    int root_rank = 0;
+    raft::resource::set_clique_root_rank(*this, root_rank);
```
cjnolet, line 74: Minor thing, but since there's no namespacing in the nccl clique resource, we might want to call this `set_nccl_clique_root_rank`.

cpp/include/raft/core/resource/nccl_clique.hpp
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024, NVIDIA CORPORATION.
```
cjnolet, line 2: This file looks great!

cpp/include/raft/core/resource/nccl_clique.hpp
```diff
@@ -51,14 +79,75 @@ class nccl_clique_resource_factory : public resource_factory {
...
+ */
+inline int get_num_ranks(resources const& res)
+{
+  return raft::resource::get_nccl_clique(res).size();
+}
+
+/**
+ * @brief Get rank's raft::resources object
+ */
+inline const raft::resources& get_device_resources(resources const& res, int rank)
```
cjnolet, line 101: Just to make this more generally clear, can we call this `get_device_resources_for_rank`?

cpp/include/raft/core/resource/nccl_clique.hpp
```diff
@@ -51,14 +79,75 @@ class nccl_clique_resource_factory : public resource_factory {
...
     res.add_resource_factory(std::make_shared<nccl_clique_resource_factory>());
   }
-  return *res.get_resource<raft::comms::nccl_clique>(resource_type::NCCL_CLIQUE);
+  return *res.get_resource<std::vector<raft::resources>>(resource_type::NCCL_CLIQUE);
+};
+
+/**
+ * @brief Get number of ranks in clique
+ */
+inline int get_num_ranks(resources const& res)
```
cjnolet, line 93: Just to make this more generally obvious (since there's no nested namespaces in the resources APIs), can we call this `get_nccl_num_ranks`? That would future proof this a tad in case we later add in some `mpi_resource` (I doubt it, but just using it as an example).

cpp/include/raft/core/resource/nccl_clique.hpp
```diff
@@ -51,14 +79,75 @@ class nccl_clique_resource_factory : public resource_factory {
...
+inline ncclComm_t& get_nccl_comm(resources const& res, int rank)
+{
+  const raft::resources& dev_res = raft::resource::get_device_resources(res, rank);
+  return raft::resource::get_nccl_comm(dev_res);
+}
+
+/**
+ * @brief Set clique root rank
+ */
+inline void set_clique_root_rank(resources const& res, int clique_root_rank)
```
cjnolet, line 140: Can we call this `set_nccl_clique_root_rank`?

cpp/include/raft/core/resource/nccl_clique.hpp
```diff
@@ -51,14 +79,75 @@ class nccl_clique_resource_factory : public resource_factory {
...
+{
+  int& clique_root_rank_ = _get_clique_root_rank(res);
+  clique_root_rank_      = clique_root_rank;
 };
 
+/**
+ * @brief Get clique root rank
+ */
+inline int get_clique_root_rank(resources const& res) { return _get_clique_root_rank(res); };
```
cjnolet, line 149: Same here- can we at least add `nccl` to the name here?

cpp/include/raft/core/device_resources_snmg.hpp
```diff
@@ -0,0 +1,258 @@
...
+ * communications on each GPU. Note the `device_resources_snmg` object can also be used as a classic
+ * `device_resources` object. The associated resources will be the ones of the GPU used during
+ * object instantiation and a GPU switch operation will be ordered during the retrieval of said
+ * resources.
+ *
+ * The `device_resources_snmg` class is intended to be used in a single process to manage several
+ * GPUs. Please note that NCCL communications are the responsibility of the user. Blocking NCCL
+ * calls will sometimes require the use of several threads to avoid hangs.
+ */
+class device_resources_snmg : public device_resources {
```
cjnolet, line 63: Thinking out loud here, but I think we should consider adding a small bit of state to the underlying `raft::resources` that can be accessed with `is_multi_gpu` and `is_multi_node`. For the base `device_resources`, we can return `false` for both. For the `device_resources_snmg`, we can return `true` for the former and `false` for the latter. What this will allow us to do is introspect these accessors on any instance of `raft::resources` without it having to be of a concrete type like `device_resources_snmg`. Essentially this means anyone can extend `raft::resources` with their own multi gpu and/or multi-node types and use it in our APIs, so long as they adhere to the proper contracts in their underlying state. It makes it truly extensible. We don't need to add it in this PR, but it would be nice to have. I'll create an issue for it as a follow-up at some point. I've been thinking through this more and more because we're also going to be using these APIs in downstream libraries like cuML in addition to cuVS.

cjnolet (approved):
I left some comments with some minor things to clean up (such as naming and whatnot) but i don't think those need to hold up this PR any further. I'm okay merging this as-is and doing those minor cleanup things as an immediate follow-up

PR #2558, opened 2025-01-30
Take argument by `const&` as the input range is const

miscco:
Found breaking CCCL ci


KyleFromNVIDIA (approved):
Removed extraneous reviewers and approved

dantegd (approved):

PR #2562, opened 2025-01-31
Fix docs builds

bdice:
This PR fixes two errors in docs builds:
1. a function with `void` return type had a `@return` parameter, which causes an error. The error was `error: found documented return type for raft::random::device::warp_random_sample that does not return anything`
2. a function with return type `std::vector<S>` was being misinterpreted as the beginning of an HTML tag `<S>`. This resulted in `error: end of comment block while expecting command </s>`.

### DIFF
--- a/cpp/include/raft/random/device/sample.cuh
+++ b/cpp/include/raft/random/device/sample.cuh
@@ -27,12 +27,14 @@ namespace raft::random::device {
 
 /**
  * @brief warp-level random sampling of an index.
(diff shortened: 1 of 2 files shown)


benfred (approved):

bdice:
Root cause: the environment was updated from `doxygen=1.12.0` (which passed) to `1.13.2` (which caused the failures fixed in this PR).

PR #2563, opened 2025-01-31
WIP: disallow fallback to Make in wheel builds

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/146

Proposes:

* setting `[tool.scikit-build].ninja.make-fallback = false`, so `scikit-build-core` will not silently fallback to using GNU Make if `ninja` is not available
* including all groups in `dependencies.yaml` with `output_types: pyproject` in `output_types: requirements` as well.
  - *So that, for example, they show up in the solved environment in RAPIDS devcontainers (https://github.com/rapidsai/devcontainers)*


dependencies.yaml
```diff
@@ -352,7 +352,7 @@ dependencies:
               - *libcusparse114
   cuda_wheels:
     specific:
-      - output_types: pyproject
+      - output_types: [pyproject, requirements]
```
jameslamb, line 355: This should be fine... any builds that don't want these (i.e. devcontainers) pass `use_cuda_wheelse="false"`, so get the empty matrix a few lines down.

ref: https://github.com/rapidsai/devcontainers/blob/a6c9100b88160bc62072a358613c28193038bc4f/features/src/rapids-build-utils/opt/rapids-build-utils/bin/make-pip-dependencies.sh#L86

bdice (approved):

PR #2566, opened 2025-02-03
Use new rapids-logger library

vyasr:
Contributes to https://github.com/rapidsai/build-planning/issues/104.


cjnolet (approved):

cpp/include/raft/core/logger.hpp
```diff
@@ -0,0 +1,78 @@
...
+ * If the environment variable `RAFT_DEBUG_LOG_FILE` is defined, the default sink is a sink to that
+ * file. Otherwise, the default is to dump to stderr.
+ *
+ * @return sink_ptr The sink to use
+ */
+inline rapids_logger::sink_ptr default_sink()
+{
+  auto* filename = std::getenv("RAFT_DEBUG_LOG_FILE");
+  return (filename == nullptr)
+           ? static_cast<rapids_logger::sink_ptr>(std::make_shared<rapids_logger::stderr_sink_mt>())
```
bdice, line 39: Same question as on cudf about `static_cast`.

bdice (approved):

python/libraft/pyproject.toml
```diff
@@ -31,6 +31,7 @@ authors = [
 license = { text = "Apache 2.0" }
 requires-python = ">=3.10"
 dependencies = [
+    "librmm==25.4.*,>=0.0.0a0",
```
jameslamb, line 34: I'd expected to see direct runtime and build-time dependencies on `rapids-logger` in `libraft` wheels and the conda recipe, as was done for `libcudf` in https://github.com/rapidsai/cudf/pull/17899, because RAFT is exposing its own logger.

Don't we need that?

jameslamb (approved):
Approving so I don't block merging, since we're now in the state where we need to start merging this sequence of PRs, assuming all the testing-specific stuff will be dropped before merging.

Please see my comment about declaring the dependency on `rapids-logger`... I think that is missing here.

python/libraft/pyproject.toml
vyasr: Yes, we do. That is a good call, I'll add those.

python/libraft/libraft/load.py
```diff
@@ -45,6 +45,14 @@ def _load_wheel_installation(soname: str):
 
 def load_library():
     """Dynamically load libraft.so and its dependencies"""
+    try:
+        import librmm  # noqa: F401
+        import rapids_logger
+
+        rapids_logger.load_library()
```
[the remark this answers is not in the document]
bdice: I think we want `rapids_logger` first, right? In this case it may not matter because librmm is header-only, but that may change -- and we otherwise need to load these libraries in dependency order, iirc.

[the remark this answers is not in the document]
jameslamb: Yeah you're right, the order here should be flipped (`rapids-logger` before `librmm`).

cpp/include/raft/core/logger.hpp
vyasr: This is fixed now.

vyasr: Fixed.

ci/build_wheel.sh
```diff
@@ -29,6 +29,7 @@ EXCLUDE_ARGS=(
   --exclude "libcusparse.so.*"
   --exclude "libnvJitLink.so.*"
   --exclude "libucp.so.*"
+  --exclude "librapids_logger.so"
```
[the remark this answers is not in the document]
vyasr: Done.

cpp/include/raft/core/logger.hpp
vyasr: Fixed in https://github.com/rapidsai/raft/pull/2566/commits/c817c0c586ad55f237e72bd278b6cfac5afb61ac

python/libraft/libraft/load.py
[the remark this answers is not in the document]
vyasr: Fixed in https://github.com/rapidsai/raft/pull/2566/commits/fd93b1640c88746b44a95a6ae9fe800316b463a3.

ci/build_wheel.sh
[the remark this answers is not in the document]
vyasr: Fixed in https://github.com/rapidsai/raft/pull/2566/commits/843ccb7a159d0eb10eefc6fba2fda508cf2ae191

python/libraft/pyproject.toml
vyasr: Wow I really skated by on transitive dependencies in this repo. Fixed in https://github.com/rapidsai/raft/pull/2566/commits/a0bb265be07f60a61c5259ddeeb3a100b9731575.

PR #2567, opened 2025-02-03
[Feat] add support for bm25 and tfidf

jperez999:
This PR will add support for tfidf and BM25 preprocessing of sparse matrix. This PR supports encoding values in raft device COO or CSR sparse matrices. It breaks up the statistical recording (fit) phase and the transformation phase. This allows for batch fitting data and then transforming a target input. This class also allows for exporting/loading, so you can fit in one place and transform in a separate environment. This builds on https://github.com/rapidsai/raft/pull/2353


cpp/include/raft/sparse/matrix/preprocessing.cuh
```diff
@@ -0,0 +1,490 @@
...
+  raft::copy(featIdCount_md.data_handle(), featIdCount, vocabSize, stream);
+  std::ofstream saveFile(save_path);
+  if (saveFile.is_open()) {
+    std::ostringstream oss;
+    saveFile << vocabSize << " ";
+    saveFile << numRows << " ";
+    saveFile << fullIdLen << " ";
+    // serialize_mdspan<IndexType>(handle, oss, featIdCount_md.view());
+    for (int i = 0; i < vocabSize; i++) {
+      saveFile << featIdCount[i] << " ";
```
jperez999, line 204: could not use the serialize_mdspan, it does not work well when reading from a file that has other information on it.

jperez999: would probably be better to use a map here to save only indexes/values that are not zero. This would save time and space. Thoughts?

cpp/include/raft/sparse/matrix/preprocessing.cuh
```diff
@@ -0,0 +1,562 @@
...
+    }
+    saveFile.close();
+  }
+}
+
+template <typename ValueType, typename IndexType>
+void SparseEncoder<ValueType, IndexType>::_fit_feats(IndexType* cols,
+                                                     IndexType* counts,
+                                                     IndexType nnz,
+                                                     IndexType* results)
```
benfred, line 228: This function should accept a raft handle - since its launching cuda kernels, it needs to use the cuda stream when creating these kernels

cpp/include/raft/sparse/matrix/preprocessing.cuh
```diff
@@ -0,0 +1,562 @@
...
+{
+  int blockSize  = (nnz < 256) ? nnz : 256;
+  int num_blocks = (nnz + blockSize - 1) / blockSize;
+  raft::sparse::matrix::detail::_scan<<<blockSize, num_blocks>>>(cols, nnz, counts);
+  raft::sparse::matrix::detail::_fit_compute_occurs<<<blockSize, num_blocks>>>(
+    cols, nnz, counts, results, numFeats);
```
benfred, line 234: I think we can probably simplify this - 

This seems like this is doing something like `np.bincount`  - counting up the document frequencies of each term so that we can compute the `idf` for tf-idf/bm25.

I think we can avoid having the _scan/_fit_compute_occurs kernels here entirely (and also remove the requirement for the `coo_sort` that is performed previously) by using cub.  For example - other code inside of raft is using cub and `HistogramEven` to peform the bincount operation  (like https://github.com/rapidsai/raft/blob/842afd7e5f6309a1094d2790f2ec50594306d490/cpp/include/raft/stats/detail/silhouette_score.cuh#L98-L146 etc)

cpp/tests/sparse/preprocess.cu
```diff
@@ -0,0 +1,299 @@
...
+TEST_P(SparsePreprocessBigTT, Result) { Run(true, true); }
+
+using SparsePreprocessBigFT = SparsePreprocessCSR<float, int>;
+TEST_P(SparsePreprocessBigFT, Result) { Run(false, true); }
+
+using SparsePreprocessBigTF = SparsePreprocessCSR<float, int>;
+TEST_P(SparsePreprocessBigTF, Result) { Run(true, false); }
+
+const std::vector<SparsePreprocessInputs<float, int>> sparse_preprocess_inputs = {
+  {
```
rhdong, line 258: Since "std::vector" is used and CUDA kernels are involved in the main codes, it is strongly recommended to add test cases for multiple settings, especially the corner cases.Since "std::vector" is used and CUDA kernels are involved in the main codes, it is strongly recommended to add test cases for multiple settings, especially the corner cases.

cpp/include/raft/sparse/matrix/preprocessing.cuh
```diff
@@ -0,0 +1,568 @@
...
+ * @tparam IndexType
+ *   Type of the indices associated with the values.
+ *
+ * @param[in] feats
+ *   Value that represents the number of features that exist for the matrices encoded.
+ */
+template <typename ValueType, typename IndexType>
+SparseEncoder<ValueType, IndexType>::SparseEncoder(int feats) : numFeats(feats)
+{
+  cudaMallocManaged(&featIdCount, feats * sizeof(int));
```
rhdong, line 140: May need `RAFT_CUDA_TRY` 
```suggestion
  RAFT_CUDA_TRY(cudaMallocManaged(&featIdCount, feats * sizeof(int)));
```

cpp/include/raft/sparse/matrix/preprocessing.cuh
jperez999: fixed this, thanks for the tip

cpp/include/raft/sparse/matrix/detail/preprocessing.cuh
```diff
@@ -0,0 +1,206 @@
...
+ *   The input rows.
+ * @param[in] columns
+ *   The input columns (features).
+ * @param[in] values
+ *   The input values.
+ * @param[in] feat_id_count
+ *   The array holding the feature(column) occurrence counts for all fitted inputs.
+ * @param[in] counts
+ *   The array representing value changes in rows input.
+ * @param[in] out_values
```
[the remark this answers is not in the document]
jperez999: fixed

cpp/include/raft/sparse/matrix/preprocessing.cuh
```diff
@@ -0,0 +1,562 @@
...
+ *   The value that represents the total number of words seen during the fit process.
+ * @param[out] numFeats
+ *   A value that represents the number of features that exist for the matrices encoded.
+ * @param[out] numRows
+ *   The number of rows observed during the fit process, accumulates over all fit calls.
+ */
+template <typename ValueType = float, typename IndexType = int>
+class SparseEncoder {
+ private:
+  int* featIdCount;
```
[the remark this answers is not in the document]
jperez999: fixed

cpp/include/raft/sparse/matrix/preprocessing.cuh
```diff
@@ -0,0 +1,562 @@
...
+ *   Type of the indices associated with the values.
+ *
+ * @param[in] handle
+ *   Container for managing reusable resources.
+ * @param[in] csr_in
+ *   Raft container housing a compressed sparse row matrix representation.
+
+ * */
+template <typename ValueType, typename IndexType>
+void SparseEncoder<ValueType, IndexType>::fit(raft::resources& handle,
```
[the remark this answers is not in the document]
jperez999: fixed

cpp/include/raft/sparse/matrix/preprocessing.cuh
jperez999: fixed, but my kernel don't use the handle. I wonder if I am missing something else.

cpp/include/raft/sparse/matrix/preprocessing.cuh
```diff
@@ -0,0 +1,568 @@
...
+  raft::copy(d_cols.data_handle(), columns.data_handle(), nnz, stream);
+  raft::copy(d_vals.data_handle(), values.data_handle(), nnz, stream);
+  raft::sparse::op::coo_sort(
+    nnz, nnz, nnz, d_cols.data_handle(), d_rows.data_handle(), d_vals.data_handle(), stream);
+  IndexType* counts;
+  cudaMallocManaged(&counts, nnz * sizeof(IndexType));
+  cudaMemset(counts, 0, nnz * sizeof(IndexType));
+  _fit_feats(handle, d_cols.data_handle(), counts, nnz, featIdCount);
+  cudaFree(counts);
+  cudaDeviceSynchronize();
```
rhdong, line 272: cudaFree will sync the device implicitly (before releasing a pointer), so there is no need to call `cudaDeviceSynchronize` separately. Furthermore, <only for suggestion>: I believe `cudaMallocManaged` is low performance, maybe you can try pinned host memory via `cudaMallocHost` when you're pretty sure the device memory is not enough.

cpp/include/raft/sparse/matrix/preprocessing.cuh
```diff
@@ -0,0 +1,568 @@
...
+void SparseEncoder<ValueType, IndexType>::transform(
+  raft::resources const& handle,
+  raft::device_csr_matrix<ValueType,
+                          IndexType,
+                          IndexType,
+                          IndexType,
+                          raft::device_uvector_policy,
+                          raft::PRESERVING> csr_in,
+  float* results,
+  bool bm25_on,
```
rhdong, line 438: if we have more options on algorithms, we may need an enum to avoid public API changing(after release)

cpp/include/raft/sparse/matrix/preprocessing.cuh
```diff
@@ -0,0 +1,568 @@
...
+  cudaMallocManaged(&counts, maxLimit * sizeof(IndexType));
+  cudaMemset(counts, 0, maxLimit * sizeof(IndexType));
+  raft::sparse::op::coo_sort(IndexType(rows.size()),
+                             IndexType(columns.size()),
+                             IndexType(values.size()),
+                             rows.data_handle(),
+                             columns.data_handle(),
+                             values.data_handle(),
+                             stream);
+  raft::sparse::matrix::detail::_scan<<<num_blocks, blockSize>>>(rows.data_handle(), nnz, counts);
```
rhdong, line 510: Since we have a `stream`, we'd better to apply it in the kernels
```suggestion
  raft::sparse::matrix::detail::_scan<<<num_blocks, blockSize, 0, stream>>>(rows.data_handle(), nnz, counts);
```

cpp/include/raft/sparse/matrix/preprocessing.cuh
```diff
@@ -0,0 +1,568 @@
...
+  raft::sparse::matrix::detail::_transform<<<num_blocks, blockSize>>>(rows.data_handle(),
```
rhdong, line 511:
```suggestion
  raft::sparse::matrix::detail::_transform<<<num_blocks, blockSize, 0, stream>>>(rows.data_handle(),
```

cpp/include/raft/sparse/matrix/preprocessing.cuh
```diff
@@ -0,0 +1,568 @@
...
+                                                                      counts,
+                                                                      results,
+                                                                      numRows,
+                                                                      avgIdLen,
+                                                                      k_param,
+                                                                      b_param,
+                                                                      nnz,
+                                                                      numFeats,
+                                                                      bm25_on);
+  cudaFree(counts);
```
rhdong, line 524: A concern is that we should take care to use these explicitly and implicitly device-sync APIs, which will break the parallel pipeline in the end-user environment and hurt performance.

cpp/include/raft/sparse/matrix/preprocessing.cuh
```diff
@@ -0,0 +1,568 @@
...
+  loadFile >> fullIdLen;
+  IndexType val;
+  std::vector<IndexType> vals;
+  while (loadFile >> val) {
+    vals.push_back(val);
+  }
+  auto featIdCount_h = raft::make_host_vector<IndexType, int64_t>(num_feats);
+  raft::copy(featIdCount_h.data_handle(), vals.data(), vals.size(), stream);
+  loadFile.close();
+  return new SparseEncoder<ValueType, IndexType>(
```
rhdong, line 553: Considering using smart pointer like `std::shared_ptr`

cpp/tests/util/preprocess_utils.cu
```diff
@@ -0,0 +1,250 @@
...
+#include <raft/sparse/convert/dense.cuh>
+#include <raft/sparse/matrix/preprocessing.cuh>
+#include <raft/sparse/neighbors/cross_component_nn.cuh>
+#include <raft/sparse/op/filter.cuh>
+#include <raft/sparse/op/sort.cuh>
+
+namespace raft::util {
+
+template <typename T>
+void print_vals(raft::resources& handle, const raft::device_vector_view<T, size_t>& out)
```
rhdong, line 32: It seems to be unused code, which should be removed.

cpp/tests/sparse/preprocess.cu
```diff
@@ -0,0 +1,299 @@
...
+    auto b_columns = raft::make_device_vector<Index_, int64_t>(handle, nnz);
+    auto b_values  = raft::make_device_vector<Type_f, int64_t>(handle, nnz);
+    raft::util::create_dataset<Index_, Type_f>(handle,
+                                               b_rows.view(),
+                                               b_columns.view(),
+                                               b_values.view(),
+                                               5,
+                                               params.n_rows,
+                                               params.n_cols,
+                                               67584);
```
rhdong, line 135: Fixed random number seeds will cause the generated test data to always be the same, which may not cover certain special input situations and thus miss potential bugs.

cpp/include/raft/sparse/matrix/detail/preprocessing.cuh
```diff
@@ -0,0 +1,225 @@
...
+                    raft::device_vector_view<IndexType, int64_t> rowFeatCnts,
+                    float k_param,
+                    float b_param,
+                    raft::device_vector_view<ValueType, int64_t> results)
+{
+  float avgIdLen = (ValueType)fullIdLen / num_rows;
+  raft::linalg::map_offset(handle, results, [=] __device__(IndexType idx) {
+    ValueType tf     = (ValueType)raft::log<ValueType>(values[idx]);
+    ValueType idf_in = (ValueType)num_rows / featIdCount.data_handle()[columns[idx]];
+    ValueType idf    = (ValueType)raft::log<ValueType>(idf_in + 1);
```
benfred, line 188: nitpick: We should stay away from C-style casts (like `(ValueType)` here) - and use the corresponding c++ casts (`static_cast<ValueType>(...)` etc) instead.

However - I don't think we need a cast in a couple of these cases, and it seems like we might be casting the same type to itself.  Maybe something like:

```suggestion
    ValueType tf     = raft::log<ValueType>(values[idx]);
    ValueType idf_in = static_cast<ValueType>(num_rows) / featIdCount.data_handle()[columns[idx]];
    ValueType idf    = raft::log<ValueType>(idf_in + 1);
```

cpp/include/raft/sparse/matrix/detail/preprocessing.cuh
```diff
@@ -0,0 +1,225 @@
...
+    values_mat.data_handle(), nnz, 0.0f, raft::identity_op(), raft::add_op(), stream, values);
+  raft::copy(batchIdLen.data_handle(), values_mat.data_handle(), values_mat.size(), stream);
+  fullFeatCount += (int)batchIdLen(0);
+}
+
+/**
+ * @brief This function counts the number of occurrences per feature(column) and records the total
+ * number of features and calculates the per row feature occurrences.
+ * @param handle: raft resource handle
+ * @param rows: Input COO rows, of size nnz.
```
benfred, line 115: nitpick: We should mark the parameters as either in/out/inout  like: 

```suggestion
 * @param rows[in]: Input COO rows, of size nnz.
```

rhdong (approved):
LGTM

PR #2541, opened 2025-01-14
Allow some of the sparse utility functions to handle larger matrices

viclafargue:
Answers https://github.com/rapidsai/cuml/issues/6204

PR #2570, opened 2025-02-04
Avoid limited memory adaptor issue in balanced KMeans

csadorf:
- Switch to the use of `get_large_workspace_resource` instead of `get_workspace_resource`
- Do not use explicit managed memory allocation.

Based on and merge after #2541 ([diff](https://github.com/viclafargue/raft/compare/fix-sparse-utilities...csadorf:raft:sadorf/address-limited-memory-adaptor-issue))

### DIFF
--- a/cpp/include/raft/cluster/detail/kmeans_balanced.cuh
+++ b/cpp/include/raft/cluster/detail/kmeans_balanced.cuh
@@ -967,7 +967,8 @@ void build_hierarchical(const raft::resources& handle,
   IdxT n_mesoclusters = std::min(n_clusters, static_cast<IdxT>(std::sqrt(n_clusters) + 0.5));
   RAFT_LOG_DEBUG("build_hierarchical: n_mesoclusters: %u", n_mesoclusters);
 
-  // TODO: Remove the explicit managed memory- we shouldn't be creating this on the user's behalf.
+  // Need to use explicit managed_memory here since corresponding allocations
+  // must be both host and device accessible.
   rmm::mr::managed_memory_resource managed_memory;
   rmm::device_async_resource_ref device_memory = resource::get_large_workspace_resource(handle);
   auto [max_minibatch_size, mem_per_row] =
--- a/cpp/include/raft/cluster/detail/kmeans_balanced.cuh
+++ b/cpp/include/raft/cluster/detail/kmeans_balanced.cuh
@@ -967,7 +967,8 @@ void build_hierarchical(const raft::resources& handle,
   IdxT n_mesoclusters = std::min(n_clusters, static_cast<IdxT>(std::sqrt(n_clusters) + 0.5));
   RAFT_LOG_DEBUG("build_hierarchical: n_mesoclusters: %u", n_mesoclusters);
 
-  rmm::device_async_resource_ref managed_memory = rmm::mr::get_current_device_resource();
+  // TODO: Remove the explicit managed memory- we shouldn't be creating this on the user's behalf.
+  rmm::mr::managed_memory_resource managed_memory;
   rmm::device_async_resource_ref device_memory = resource::get_large_workspace_resource(handle);
   auto [max_minibatch_size, mem_per_row] =
     calc_minibatch_size<MathT>(n_clusters, n_rows, dim, params.metric, std::is_same_v<T, MathT>);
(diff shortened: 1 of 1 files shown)


cjnolet:
@csadorf we will also want to make sure we port this over to cuVS, since the kmeans in raft will be getting ported shortly after GTC.

cpp/include/raft/cluster/detail/kmeans_balanced.cuh
```diff
@@ -968,15 +970,15 @@ void build_hierarchical(const raft::resources& handle,
   IdxT n_mesoclusters = std::min(n_clusters, static_cast<IdxT>(std::sqrt(n_clusters) + 0.5));
   RAFT_LOG_DEBUG("build_hierarchical: n_mesoclusters: %u", n_mesoclusters);
 
-  // TODO: Remove the explicit managed memory- we shouldn't be creating this on the user's behalf.
-  rmm::mr::managed_memory_resource managed_memory;
-  rmm::device_async_resource_ref device_memory = resource::get_workspace_resource(handle);
+  rmm::device_async_resource_ref current_device_resource = rmm::mr::get_current_device_resource();
+  rmm::device_async_resource_ref workspace_resource =
```
achirkin, line 974: This is one of those rare cases where we do indeed need to explicitly allocate `rmm::mr::managed_memory_resource` (the removed TODO comment is actually incorrect).
The need to use managed memory here has nothing to do with the memory limit and the user choice, but rather is a part of the algorithm. We use the `managed_memory` variable across this file for not-so-big allocations that are accessed by both device and host (see, for example, the `build_fine_clusters` function above). Hence, using the device-only memory simply breaks the algorithm here.

achirkin (changes requested):
We have couple problem with kmeans_balanced here

cpp/include/raft/cluster/detail/kmeans_balanced.cuh
```diff
@@ -968,15 +970,15 @@ void build_hierarchical(const raft::resources& handle,
   IdxT n_mesoclusters = std::min(n_clusters, static_cast<IdxT>(std::sqrt(n_clusters) + 0.5));
   RAFT_LOG_DEBUG("build_hierarchical: n_mesoclusters: %u", n_mesoclusters);
 
-  // TODO: Remove the explicit managed memory- we shouldn't be creating this on the user's behalf.
-  rmm::mr::managed_memory_resource managed_memory;
-  rmm::device_async_resource_ref device_memory = resource::get_workspace_resource(handle);
+  rmm::device_async_resource_ref current_device_resource = rmm::mr::get_current_device_resource();
```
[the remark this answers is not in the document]
csadorf: Addressed.

cpp/include/raft/cluster/detail/kmeans_balanced.cuh
csadorf: Adressed.

However, I'm wondering whether we should be generally using [`make_managed_vector`](https://docs.rapids.ai/api/raft/stable/cpp_api/mdspan_mdarray/#_CPPv4I000EN4raft19make_managed_vectorEDaRKN4raft9resourcesE9IndexType) instead then. @achirkin Was there a specific motivation for the use of `rmm::uvector`s instead?

achirkin: No, only historic reasons: the balanced kmeans code arrived earlier than these managed helpers.

csadorf: Thanks for clarifying.

@cjnolet Considering that this code has moved to cuVS anyways I assume there is no point in refactoring this, is there?

csadorf:
Keeping this in draft mode until https://github.com/rapidsai/raft/pull/2541 is merged.

csadorf:
> @csadorf we will also want to make sure we port this over to cuVS, since the kmeans in raft will be getting ported shortly after GTC.

Prepared in https://github.com/rapidsai/cuvs/pull/659 .

cjnolet (approved):
Approving, as this code is deprecated anyways (and will be removed once cuML is using cuVS for this).

cpp/include/raft/cluster/detail/kmeans_balanced.cuh
cjnolet: No reason to refactor, just need to fix any issues that is blocking cuML UMAP ATM

cjnolet:
> Prepared in https://github.com/rapidsai/cuvs/pull/659 .

Thanks so much @csadorf!

cjnolet:
@csadorf if you haven't already, can you please open an issue in cuVS just to outline the challenge here? I understand we aren't planning on merging this diff into cuVS, but I want to make sure we have an issue opened so that we don't lose sight of this change. The improvement is going to need to be made, and soon.

csadorf:
> @csadorf if you haven't already, can you please open an issue in cuVS just to outline the challenge here? I understand we aren't planning on merging this diff into cuVS, but I want to make sure we have an issue opened so that we don't lose sight of this change. The improvement is going to need to be made, and soon.

Created https://github.com/rapidsai/cuvs/issues/682 .

csadorf:
@cjnolet I believe this should be ready to merge.

cjnolet (approved):

divyegala (approved):

PR #2575, opened 2025-02-06
Add `shellcheck` to pre-commit and fix warnings

gforsyth:
`shellcheck` is a fast, static analysis tool for shell scripts. It's good at
flagging up unused variables, unintentional glob expansions, and other potential
execution and security headaches that arise from the wonders of `bash` (and
other shlangs).

This PR adds a `pre-commit` hook to run `shellcheck` on all of the `sh-lang`
files in the `ci/` directory, and the changes requested by `shellcheck` to make
the existing files pass the check.

xref: rapidsai/build-planning#135

### DIFF
--- a/ci/validate_wheel.sh
+++ b/ci/validate_wheel.sh
@@ -5,7 +5,6 @@ set -euo pipefail
 
 package_dir=$1
 wheel_dir_relative_path=$2
-package_name=$3
 
 
 cd "${package_dir}"
@@ -14,10 +13,10 @@ rapids-logger "validate packages with 'pydistcheck'"
 
 pydistcheck \
     --inspect \
-    "$(echo "${wheel_dir_relative_path}"/"${package_name}"*.whl)"
+    "$(echo "${wheel_dir_relative_path}"/*.whl)"
 
 rapids-logger "validate packages with 'twine'"
 
 twine check \
     --strict \
-    "$(echo "${wheel_dir_relative_path}"/"${package_name}"*.whl)"
+    "$(echo "${wheel_dir_relative_path}"/*.whl)"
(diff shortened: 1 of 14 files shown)


ci/validate_wheel.sh
```diff
@@ -7,18 +7,17 @@ package_dir=$1
...
 
-RAPIDS_CUDA_MAJOR="${RAPIDS_CUDA_VERSION%%.*}"
 
 cd "${package_dir}"
 
 rapids-logger "validate packages with 'pydistcheck'"
 
 pydistcheck \
     --inspect \
-    "$(echo ${wheel_dir_relative_path}/*.whl)"
```
gforsyth, line 18: Previously `package_name` was unused in this script

jameslamb: This change is why https://github.com/rapidsai/raft/actions/runs/13188022880/job/36815130064?pr=2575 is failing like this:

```text
Usage: pydistcheck [OPTIONS] [FILEPATHS]...
Try 'pydistcheck --help' for help.

Error: Invalid value for '[FILEPATHS]...': Path 'final_dist/raft-dask*.whl' does not exist.
```

`package_name` here is `raft-dask` (with a `-`), but the wheel filename begins with `raft_dask` (with a `_`). That's intentional behavior.

I think a better fix here would be to just remove `package_name` from the script (and the corresponding calls to it).

jameslamb (changes requested):
Left on requested change on `validate_wheel.sh`.

ci/validate_wheel.sh
gforsyth: yeah, stripped out `package_name`.  thanks for tracking that down, @jameslamb !

jameslamb: No prob! It's my fault that `package_name` was left behind there unused in anyway 😅

jameslamb (approved):

PR #2576, opened 2025-02-06
Use shared-workflows branch-25.04

bdice:
This completes the migration to NVKS runners now that all libraries have been tested and https://github.com/rapidsai/shared-workflows/pull/273 has been merged.

xref: https://github.com/rapidsai/build-infra/issues/184

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -28,7 +28,7 @@ concurrency:
 jobs:
   cpp-build:
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@nvks-runners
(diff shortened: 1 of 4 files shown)


jameslamb:
`Label Checker` was failing despite having `non-breaking` and `improvement` on here.

Screenshot 2025-02-07 at 11 15 16 AM

If I recall correctly (can't find the link right now), this can occasionally happen with `rapids-reviser` PRs because the PR and its labels are created simultaneously, and there's a race condition in when all the data gets reflected back in GitHub's API.

I'll open a `rapids-reviser` issue about that, I have an idea for how we could avoid that.

I know that modifying the labels causes the check to re-run, so I just removed `non-breaking` and added it back. And that seemed to fix this.

jameslamb (approved):

PR #2578, opened 2025-02-11
`#include <numeric>` for `std::iota`

benfred:
I'm seeing CI failures in cuvs over

```
/home/coder/.conda/envs/rapids/include/raft/comms/nccl_clique.hpp(68): error: namespace "std" has no member "iota"
      std::iota(device_ids_.begin(), device_ids_.end(), 0);
```

This seems to be be cause we're using `std::iota` without including the header. Fix.

### DIFF
--- a/cpp/bench/prims/core/copy.cu
+++ b/cpp/bench/prims/core/copy.cu
@@ -13,7 +13,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 #include <common/benchmark.hpp>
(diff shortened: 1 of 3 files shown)


bdice (approved):

bdice:
Thanks @benfred! I recently removed some extraneous header inclusions from RMM in https://github.com/rapidsai/rmm/pull/1821, so this is expected.

wphicks (approved):

PR #2582, opened 2025-02-13
Fix COO symmetrization

viclafargue:


lowener (approved):

cjnolet (changes requested):
Before this is merged, I’d like a comparison done and a small test put in place. This is an important operation on the UMAP side right now and I want to make sure this is actually improving the correctness. Even an MRE in a unit test would be helpful.

jcrist:
What's the status of this fix? Is this something we'll want to get in for 25.04 (and if so, does anyone have capacity to do the investigation Corey requested above)?

cjnolet:
Until someone can demonstrate (with an appropriate gtest) that this behavior is causing a problem, I would treat it as a relatively low priority.

viclafargue:
I will write some tests to check if it actually improves the COO symmetrization.

cjnolet:
Any updates on this one @viclafargue?

viclafargue:
Yes, I wrote a test. This test would fail without the change in this PR.

cjnolet (approved):

Issue #2450, opened 2024-09-24
[BUG] C++ testing failing on arm64: `NEIGHBORS_ANN_NN_DESCENT_TEST`


jameslamb:
**Describe the bug**

Over at least the last day I've seen the `NEIGHBORS_ANN_NN_DESCENT_TEST` failing reproducibly and consistently in this CI job:

```text
conda-cpp-tests / tests (arm64, 3.11, 12.0.1, ubuntu20.04, a100, latest, latest)
```

Like this:

```text
/opt/conda/conda-bld/work/cpp/test/neighbors/ann_nn_descent/../ann_nn_descent.cuh:274: Failure
Value of: eval_neighbours(indices_naive, indices_NNDescent, distances_naive, distances_NNDescent, ps.n_rows, ps.graph_degree, 0.01, min_recall, true, static_cast<size_t>(ps.graph_degree * 0.1))
  Actual: false (Duplicated index 1780 at k 30 for query 194! )
Expected: true
[  FAILED  ] AnnNNDescentBatchTest/AnnNNDescentBatchTestF_U32.AnnNNDescentBatch/5, where GetParam() = dataset shape=4000x512, graph_degree=32, metric=0, host, clusters=3
 (477 ms)
...
[----------] Global test environment tear-down
[==========] 344 tests from 4 test suites ran. (153504 ms total)
[  PASSED  ] 343 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] AnnNNDescentBatchTest/AnnNNDescentBatchTestF_U32.AnnNNDescentBatch/5, where GetParam() = dataset shape=4000x512, graph_degree=32, metric=0, host, clusters=3

 1 FAILED TEST
CMake Error at run_gpu_test.cmake:34 (execute_process):
  execute_process failed command indexes:

    1: "Child return code: 1"

96% tests passed, 1 tests failed out of 24

Total Test time (real) = 4868.12 sec

The following tests FAILED:
	 22 - NEIGHBORS_ANN_NN_DESCENT_TEST (Failed)
```

All C++ tests appear to pass in other `conda-cpp-tests` jobs (which are all x86_64).

At https://github.com/rapidsai/raft/actions/workflows/pr.yaml, it looks like the most recent fully-passing run of the `pr` workflow that included the `conda-cpp-tests` jobs was 19 hours ago ([build link](https://github.com/rapidsai/raft/actions/runs/11002806214/job/30604900635)).

[...]

PR #2586, opened 2025-02-14
Fix duplicate indices in batch NN Descent

jinsolp:
### Purpose of this PR

Handling duplicate indices in batch NN Descent graph
- `bloom_filter`'s `nrow` and the `update_counter_` needs to be initialized at the beginning of the `build` function, considering that `build` is called multiple times with batch NN Descent

Resolves the following issue
- https://github.com/rapidsai/raft/issues/2450

### Notes
Testing added for batch NN Descent. Noticed that most tests were migrated to cuVS, so brought a few util code snippets.
- `ann_utils.cuh` from [here in cuVS](https://github.com/jinsolp/cuvs/blob/branch-24.10/cpp/test/neighbors/ann_utils.cuh)
- `naive_knn.cuh` from [here in cuVS](https://github.com/jinsolp/cuvs/blob/branch-24.10/cpp/test/neighbors/naive_knn.cuh)


jcrist:
Checking in here - what's the status of this PR?

divyegala:
@jinsolp can you verify if this PR resolves the issue of having to set a `graph_degree` > `n_neighbors` in UMAP and having to slice back the result indices/distances? The resultant graph should always be equivalent such that `graph_degree = n_neighbors`. The only thing that we should be needing to change is `intermediate_graph_degree`.

jcrist:
If this fixes a known bug in `raft`, then I'd say we should merge as is. Testing how this pans out in `cuml` is slightly easier using the main branch of raft, and if this improves result we'd want to refactor the parameter handling and remove the copies to make full use of these changes (a more substantial PR to cuml). If you'd rather wait until we have feedback on how this affects cuml, then I'd be happy to test things out beforehand, but if possible I'd prefer to merge here first and then iterate in cuml.

divyegala:
@jcrist as of now, that slicing bug in UMAP is not triaged. I wish to know whether this PR fixes it before merging. If it does fix, we can merge this and we can refactor whatever we need in cuml. 

This PR does not need to be held up by everything being ready on the cuml side, but a quick test of whether it fixes the slicing bug would be helpful.

jcrist:
> Any churn in merging PRs that are not thoroughly tested means that we have to propagate changes again from RAFT to cuVS or vice versa.

Sure, I can empathize with that. I think what I'm saying is:
- This fix here seems like it genuinely fixes some bugs, regardless of effect on `cuml` itself
- The performance issues we're seeing in `cuml.UMAP` when reducing graph degree _may_ be fixed by this, or may require additional fixes. But if this doesn't fully fix things, the remaining issues could be fully unrelated to the work here.

Anyway, I can test things out and report back here.

cjnolet:
> This fix here seems like it genuinely fixes some bugs, regardless of effect on cuml itself

@jcrist as far as we are concerned, cuml is the only remaining user of this code. We kept this code in RAFT only until cuML is able to migrate, but it's deprecated (as noted in RAFT's readme). I don't think we have to worry about downstream other than cuml, but it would be nice to know if it does genuinely fix the issue at hand.

jinsolp:
I did check on my local and the tests for cuml UMAP was working fine with `graph_degree=n_neighbors`. 
I can open up a PR for changes in the cuml python test file and check against the two branches one last time before merging this

divyegala:
@jinsolp thanks for testing! A quick branch in cuML is fine and then any refactors can follow later on.

@jcrist I am trying to emphasize that we have a feeling that this is what is causing the graph_degree != n_neighbors bug in cuML UMAP currently, and like @cjnolet said cuML is the only downstream user of this code.

To verify our intuition that the bug is fixed, we need to test against cuML. If we eagerly merge this PR and then find out that this solved the bug partially but not fully, we'll need to go back and make the changes again to RAFT and mirror to cuVS.

As Jinsol commented earlier today, looks like our intuition that this PR solves the UMAP bug is correct. Performance gains are orthogonal to correctness. NN Descent is an important algorithm, and will be used for more applications in cuML. We must test and iron out bugs as fully as possible.

jcrist:
Sorry, by "performance" above I meant performance in the trustworthiness score sense not speed (I can see how that was a confusing choice of words). If this PR allows us to reduce the graph degree without a decrease in `trustworthiness` (what the umap tests Jinsol is referencing compare) then I think that's a good indicator that this PR improves things.

All I was arguing above was that it seems clear that this PR does fix a bug in nndescent in raft (the new tests added plus clear bug in the prior code being addressed). It seemed fine to me to merge as is and address any additional needs (if any) from cuml in a followup. I'm just trying to ensure all these PRs get in by the upcoming code freeze.

divyegala:
@jinsolp could you also verify that the tests for UMAP with the current version of NN Descent (discounting this PR) actually fail with graph_degree != n_neighbors?

divyegala:
@jinsolp could you also verify that the tests for UMAP with the current version of NN Descent (discounting this PR) actually fail with graph_degree != n_neighbors?

@jcrist if the above tests fail without changes in this PR and pass with the changes then there's a coupling between cuML and the changes of this PR, and both unit tests and integration tests must be checked.

As for the unit tests, they have been proven unreliable before. This PR went in with all the unit tests passing in PR and nightlies for a few weeks during and after Jinsol's internship. Then suddenly, they started failing on ARM. Later on, I was able to reproduce them reliably on all architectures, which is when I disable them.

Considering the history, I do not consider passing unit tests presently as sufficient evidence that the bug is solved.

jinsolp:
@divyegala I'll check that too!

jinsolp:
Hello Divye, I checked with&without changes of this PR.
- without this PR + current cuML umap tests => works fine
- without this PR + setting `n_neighbors=graph_degree` in cuML umap tests => **FAILS for batch nnd tests (as expected)**
- with this PR + current cuML umap tests => works fine
- with this PR + setting `n_neighbors=graph_degree` in cuML umap tests => **works fine**

divyegala (approved):
LGTM, thanks for testing against UMAP!

jinsolp:
I've opened up a PR for cuML [here](https://github.com/rapidsai/cuml/pull/6413). It builds on this branch to check that the changes in `test_umap.py` works properly with this fix.

jinsolp:
@divyegala Should I re-run the CI (it seems like 2 unrelated checks are failing). If that's okay I think this one is good to go 👍

divyegala:
@jinsolp I reran the failing jobs

jinsolp:
@divyegala I don't think they reran?

PR #2591, opened 2025-02-21
Fixed pytest marker warnings by removing unused pytest.ini

TomAugspurger:
The CI for raft-dask includes warnings about unregistered pytest markers: https://github.com/rapidsai/raft/actions/runs/13394517078/job/37411078661#step:10:2656

```
../../../pyenv/versions/3.12.9/lib/python3.12/importlib/__init__.py:90
  /pyenv/versions/3.12.9/lib/python3.12/importlib/__init__.py:90: PytestUnknownMarkWarning: Unknown pytest.mark.mg - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
    return _bootstrap._gcd_import(name[level:], package, level)
```

Those markers are registered [here](https://github.com/rapidsai/raft/blob/3e53bf0bac7e40bf1ccdd09c09c1ef6f6a49aa26/python/raft-dask/pytest.ini#L4) at the root of raft-dask, but a second `pytest.ini` in `raft_dask/tests/pytest.ini` shadowed that `pytest.ini`. Without the second pytest.ini, I don't see these warnings locally.


benfred (approved):
lgtm - thanks for fixing @TomAugspurger !

PR #2592, opened 2025-02-24
[BUG] Fix illegal memory access in linalg::reduction

enp1s0:
Illegal memory access occurs when calling `mean` for a large matrix, specifically, when `extent(0) * extent(1)` exceeds INT_MAX. This PR fixes it.

Although I think it is safer to use `uint64_t` or `size_t` than `IdxType` as the index type in the kernel, I follow the [coalescedReduction kernels](https://github.com/rapidsai/raft/blob/branch-25.04/cpp/include/raft/linalg/detail/coalesced_reduction-inl.cuh).


lowener (approved):
lgtm

cjnolet (approved):

PR #2594, opened 2025-02-25
Replace `cub::Sum` and `cub::Max` with `cuda::std::plus` and `cuda::maximum`

miscco:
The former is deprecated and will be removed in a future release of CCCL


cjnolet (approved):
LGTM pending CI

miscco:
This is blocked on rapids moving to 2.8 because that was the release we introduced `cuda::minimum`

lowener (approved):

miscco:
@cjnolet I believe I have fixed the issues

bdice (approved):

PR #2595, opened 2025-02-25
Use conda-build instead of conda-mambabuild

bdice:
This changes from `conda mambabuild` to `conda build`. Conda now uses the mamba solver so no performance regressions are expected.

This is a temporary change as we plan to migrate to `rattler-build` in the near future. However, this is needed sooner to drop `boa` and unblock Python 3.13 migrations.

xref: https://github.com/rapidsai/build-infra/issues/149

### DIFF
--- a/ci/build_cpp.sh
+++ b/ci/build_cpp.sh
@@ -17,7 +17,7 @@ rapids-logger "Begin cpp build"
 
 sccache --zero-stats
 
-RAPIDS_PACKAGE_VERSION=$(rapids-generate-version) rapids-conda-retry mambabuild \
+RAPIDS_PACKAGE_VERSION=$(rapids-generate-version) rapids-conda-retry build \
   conda/recipes/libraft
 
 sccache --show-adv-stats
--- a/ci/build_python.sh
+++ b/ci/build_python.sh
@@ -25,7 +25,7 @@ sccache --zero-stats
 
 # TODO: Remove `--no-test` flags once importing on a CPU
 # node works correctly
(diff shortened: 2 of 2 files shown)


jameslamb (approved):
Saw some warnings I haven't seen before (might have just not noticed them in previous builds):

> WARNING: Adding .* to spec 'libraft-headers 25.04.00a38' to ensure satisfiability.  Please consider putting {{{{ var_name }}}}.* or some relational operator (>/</>=/<=) on this spec in meta.yaml, or if req is also a build req, using {{{{ pin_compatible() }}}} jinja2 function instead.  See https://conda.io/docs/user-guide/tasks/build-packages/variants.html#pinning-at-the-variant-level
WARNING: Adding .* to spec 'libraft 25.04.00a38' to ensure satisfiability.  Please consider putting {{{{ var_name }}}}.* or some relational operator (>/</>=/<=) on this spec in meta.yaml, or if req is also a build req, using {{{{ pin_compatible() }}}} jinja2 function instead.  See https://conda.io/docs/user-guide/tasks/build-packages/variants.html#pinning-at-the-variant-level

That's probably from stuff like this:

https://github.com/rapidsai/raft/blob/cb6fe7c777fa586c0ad6e76fb2fdfd81953d2308/conda/recipes/libraft/meta.yaml#L170

Which looks... fine to me? I'd consider it a bug that `conda build` raises a warning in this case. If that's a bug, then `pin_subpackage(.., exact=True)` should generate an `===` or something.

I think it's safe to ignore that, but just sharing here for searchability.

PR #2596, opened 2025-02-27
Fix skipped elements processing in vectorized_process

qieqieplus:
**Changes:**

-   Replaced the `if` condition for processing skipped elements with a `for` loop that correctly distributes the work. 

This ensures all elements are processed.

### DIFF
--- a/cpp/include/raft/matrix/detail/select_radix.cuh
+++ b/cpp/include/raft/matrix/detail/select_radix.cuh
@@ -175,11 +175,15 @@ _RAFT_DEVICE void vectorized_process(
 
     static_assert(WarpSize >= wide_t::Ratio);
     // Processes the skipped elements on the left
-    if (thread_rank < skip_cnt_left) { f(in[thread_rank], thread_rank); }
+    for (IdxT i = thread_rank; i < skip_cnt_left; i += num_threads) {
+      f(in[i], i);
+    }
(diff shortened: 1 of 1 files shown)


achirkin:
Thanks for submitting this pull request! Do you have a reproducer for this? Normally, the presumption is that the number of skipped elements (which is smaller than the alignment requirements) is always smaller than the grid size.

qieqieplus:
> Thanks for submitting this pull request! Do you have a reproducer for this? Normally, the presumption is that the number of skipped elements (which is smaller than the alignment requirements) is always smaller than the grid size.

Yes, this is only a theoretical bug (I found this function very useful and used this in another kernel, and my simple unittest broke). In practice, there are always enough threads.

achirkin:
Ok, although I see the point in your suggestion, I'm inclined to close this PR.
Having `if` there is rather intentional: (1) it tells the compiler, that it doesn't need those precious registers for the extra variable, and (2) it tells the human reader that we only access a handful of elements on both ends and do that in a single parallel read.

PS: if you need this coalesced processing functionality, you may also be interested in a more generic version of the same idea at https://github.com/rapidsai/raft/blob/branch-25.04/cpp/include/raft/matrix/detail/linewise_op.cuh

qieqieplus:
Thanks for your reply and suggestion!

PR #2600, opened 2025-03-11
[Fix] the bug of `spectral` illegal memory access

rhdong:
- When `β == 0`, the `y` in [cublas::gemv](https://docs.nvidia.com/cuda/cublas/#cublas-t-gemv) shouldn't be read, but older cuBLAS versions (e.g., CUDA 11.4) may still access it.
- I reproduced the issue and fixed it locally with this PR, but I'm not fully sure it works in the nightly environment. It's best to add CUDA 11.4 testing to the PR CI.

### DIFF
--- a/cpp/include/raft/sparse/solver/detail/lanczos.cuh
+++ b/cpp/include/raft/sparse/solver/detail/lanczos.cuh
@@ -1815,7 +1815,7 @@ auto lanczos_smallest(
   raft::copy(&res, output.data_handle(), 1, stream);
   resource::sync_stream(handle, stream);
 
-  auto uu  = raft::make_device_matrix<ValueTypeT>(handle, 0, nEigVecs);
+  auto uu  = raft::make_device_matrix<ValueTypeT>(handle, 1, nEigVecs);
   int iter = ncv;
   while (res > tol && iter < maxIter) {
     auto beta_view = raft::make_device_matrix_view<ValueTypeT, uint32_t, raft::row_major>(


cpp/include/raft/sparse/solver/detail/lanczos.cuh
```diff
@@ -1815,7 +1815,7 @@ auto lanczos_smallest(
   raft::copy(&res, output.data_handle(), 1, stream);
   resource::sync_stream(handle, stream);
 
-  auto uu  = raft::make_device_matrix<ValueTypeT>(handle, 0, nEigVecs);
+  auto uu  = raft::make_device_matrix<ValueTypeT>(handle, 1, nEigVecs);
```
divyegala, line 1818: Could you please add a comment here explaining why it should be 0 and not 1 and how setting it to 1 fixes the bug in CUDA 11 without changing the behavior in CUDA 12?

rhdong: > Could you please add a comment here explaining why it should be 0 and not 1 and how setting it to 1 fixes the bug in CUDA 11 without changing the behavior in CUDA 12?

Basically, it should be `1` according to the cuBLAS documentation, which doesn’t mention `y` being `nullptr`. So, a separate comment explaining potential misusage isn’t necessary, right? :grimacing:

divyegala: Just for hindsight, a comment with a link to the docs is appreciated :)

PR #2607, opened 2025-03-15
[Fix] Lanczos solver gemv fix

### LINKED ISSUE #2484: [BUG] `raft::linalg::gemv` calls can lead to `CUBLAS_STATUS_INVALID_VALUE`
Encountered during Lanczos PR: Using the public API of `raft::linalg::gemv` led to `CUBLAS_STATUS_INVALID_VALUE`, while using the cublas wrapper lead to the correct result.
https://github.com/rapidsai/raft/pull/2481/files#diff-54438d58f6ed96dc83cce94608f1c44adc787261ff00013cc1476f62d3b9af39R1838

### LINKED ISSUE #2601: Temporarily allow for increase in RAFT nightly failures
There have been certain spectral partitioning bugs that have popped up in CUDA 11.4 nightlies. While they are being investigated at the moment in https://github.com/rapidsai/raft/pull/2600, we are approaching burndown quickly and need certain PRs to go in to unblock downstream work. 

Specifically in cuML (mostly blocked on https://github.com/rapidsai/raft/pull/2586): 
1. Follow-up PR to simplify (and make more robust) the parameter handling in cuml UMAP cpp. This would be removing duplicate parameters, and making graph_degree set to n_neighbors.
2. Removing the now unnecessary slicing
3. Some customer-facing testing for performance and memory usage.
I4. mproving cuml UMAP's default behavior to better dispatch build algo based on data size and input type. This isn't really blocked on this upstream work, but would be nice to test in concert with it.

aamijar:
Related to #2600
Resolves #2484, Resolves #2605, Resolves #2601

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -104,9 +104,6 @@ jobs:
     if: fromJSON(needs.changed-files.outputs.changed_file_groups).test_cpp
     with:
       build_type: pull-request
-      # Filtering out 11.4 runs until 11.4 issues are resolved
-      # See https://github.com/rapidsai/build-planning/issues/164
-      matrix_filter: map(select(.CUDA_VER != "11.4.3"))
   conda-cpp-checks:
     needs: conda-cpp-build
     secrets: inherit
@@ -127,9 +124,6 @@ jobs:
     if: fromJSON(needs.changed-files.outputs.changed_file_groups).test_python
     with:
       build_type: pull-request
-      # Filtering out 11.4 runs until 11.4 issues are resolved
-      # See https://github.com/rapidsai/build-planning/issues/164
-      matrix_filter: map(select(.CUDA_VER != "11.4.3"))
   docs-build:
     needs: conda-python-build
     secrets: inherit
--- a/cpp/include/raft/sparse/solver/detail/lanczos.cuh
+++ b/cpp/include/raft/sparse/solver/detail/lanczos.cuh
@@ -1815,7 +1815,7 @@ auto lanczos_smallest(
   raft::copy(&res, output.data_handle(), 1, stream);
   resource::sync_stream(handle, stream);
 
-  auto uu  = raft::make_device_matrix<ValueTypeT>(handle, 0, nEigVecs);
+  auto uu  = raft::make_device_matrix<ValueTypeT>(handle, 1, nEigVecs);
   int iter = ncv;
   while (res > tol && iter < maxIter) {
     auto beta_view = raft::make_device_matrix_view<ValueTypeT, uint32_t, raft::row_major>(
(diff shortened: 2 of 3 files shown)


rhdong:
Hi @aamijar, great job! The current C++ test relies on magic numbers, making it less reliable. Could you optimize the C++ testing by introducing cases where the CSR matrix is randomly generated? You can reuse these functions: [create_sparse_matrix](https://github.com/rapidsai/raft/blob/branch-25.04/cpp/tests/sparse/select_k_csr.cu#L89) and [convert_to_csr](https://github.com/rapidsai/raft/blob/branch-25.04/cpp/tests/sparse/select_k_csr.cu#L113). Additionally, implementing a cpu_solver to generate expected results is required. Let me know your thoughts!

cjnolet:
+1 to @rhdong's suggestion here. We really need to start tightening up our testing, improving coverage, and making sure we're reducing potential for leakage wherever possible.

aamijar:
Hi @rhdong, instead of writing our own cpp cpu solver, can we use the scipy cpu solver as ground truth? We can test it through python with the random inputs.

cjnolet:
Sorry @rhdong and @aamijar, I admit in the craziness of GTC ramp-up, I missed the comment about implementing a CPU version of the solver. Unfortunately, this is very error prone and I've found in the past we end up spending more time maintaining and fixin bugs in the CPU solver such that it really negates the benefits. 

I think for the Gtest side we should do what we can to make sure the result is acceptable (e.g. making sure the eigenvectors are, in fact, orthogonal, etc...) and do a comparison against scipy / cupy in the Python layer.

In the meantime, though, for the interest of time, we need to get this change merged if it fixes CI, because it's currently blocking downstream libraries and we are amidst burndown for RAFT. Let's get this fixed ASAP and then @aamijar I'd like you to focus on improving the tests here.

rhdong (approved):
LGTM

divyegala (approved):
Thanks Anupam, LGTM!

divyegala:
@cjnolet will wait for your approval here before merging

cjnolet:
Yeah we can merge this if it fixes the nightlies. Has that been verified yet?

aamijar:
How do I run the nightly CI to check?

cjnolet:
I think verifying it works with your change in cuda 11.4 (and doesn’t work without it) is enough

aamijar:
@rhdong You had a cuda 11.4 environment setup already right? If you could run it for me that would be amazing! Otherwise, if someone knows how to create a 11.4 environment then I can do it.

rhdong:
> @rhdong You had a cuda 11.4 environment setup already right? If you could run it for me that would be amazing! Otherwise, if someone knows how to create a 11.4 environment then I can do it.

I will slack you the reproduce method.

viclafargue (approved):
LGTM. I also took a second look at the cuBLAS documentation to verify the validity of all input matrices shapes and it turns out that it is actually correct. Thanks @aamijar!

gforsyth (approved):

PR #2608, opened 2025-03-17
ci: pre-filter 11.4 jobs before they are enabled in shared workflows

gforsyth:
This PR adds a filter to skip CUDA 11.4 jobs on PRs as a precursor to enabling them in shared-workflows.
Once the 11.4 issues are fixed, this matrix filter should be removed so 11.4 gets tested on PRs.

xref: rapidsai/build-planning#164

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -104,6 +104,9 @@ jobs:
     if: fromJSON(needs.changed-files.outputs.changed_file_groups).test_cpp
     with:
       build_type: pull-request
+      # Filtering out 11.4 runs until 11.4 issues are resolved
+      # See https://github.com/rapidsai/build-planning/issues/164
+      matrix_filter: map(select(.CUDA_VER != "11.4.3"))
   conda-cpp-checks:
     needs: conda-cpp-build
     secrets: inherit
@@ -124,6 +127,9 @@ jobs:
     if: fromJSON(needs.changed-files.outputs.changed_file_groups).test_python
     with:
       build_type: pull-request
+      # Filtering out 11.4 runs until 11.4 issues are resolved
+      # See https://github.com/rapidsai/build-planning/issues/164
+      matrix_filter: map(select(.CUDA_VER != "11.4.3"))
   docs-build:
     needs: conda-python-build
     secrets: inherit


.github/workflows/pr.yaml
```diff
@@ -104,6 +104,9 @@ jobs:
     if: fromJSON(needs.changed-files.outputs.changed_file_groups).test_cpp
     with:
       build_type: pull-request
+      # Filtering out 11.4 runs until 11.4 issues are resolved
+      # See https://github.com/rapidsai/build-planning/issues/164
+      matrix_filter: map(select(.CUDA_VER != "11.4.3"))
```
jameslamb, line 109:
```suggestion
      matrix_filter: map(select(! .CUDA_VER.startswith("11.4")))
```

Would you consider something like this that changes the condition from "not identical to 11.4.3" to "does not being with 11.4"?

That'd avoid this breaking in the (I know, probably unlikely) instance where we switch to a different patch version of CUDA 11.4 or where CUDA_VER is changed from `11.4.3` to just `11.4`.

`jq` definitely does have `startswith()` (https://jqlang.org/manual/#startswith), but not sure if I have the syntax exactly right here.

.github/workflows/pr.yaml
```diff
@@ -124,6 +127,9 @@ jobs:
     if: fromJSON(needs.changed-files.outputs.changed_file_groups).test_python
     with:
       build_type: pull-request
+      # Filtering out 11.4 runs until 11.4 issues are resolved
+      # See https://github.com/rapidsai/build-planning/issues/164
+      matrix_filter: map(select(.CUDA_VER != "11.4.3"))
```
jameslamb, line 132:
```suggestion
      matrix_filter: map(select(! .CUDA_VER.startswith("11.4")))
```

jameslamb (approved):
Left a small suggestion for your consideration, but do whatever you want with it.

.github/workflows/pr.yaml
gforsyth: I can come back and update it to something like

```
(.CUDA_VER | startswith("11.4")) | not)
```

if this is going to continue in-place, but my hope is that this is a very short-term filter

jameslamb: ok no prob

PR #2629, opened 2025-04-08
Use NCCL wheels from PyPI for CUDA 12 builds

divyegala:
This PR attempts to de-vendor NCCL DSOs that we bundle along with our wheel builds.


conda/recipes/libraft/recipe.yaml
```diff
@@ -78,6 +78,8 @@ cache:
           - libcusolver-dev
           - libcusparse-dev
       - librmm =${{ minor_version }}
+      - nccl ${{ nccl_version }}
+      - ucxx ${{ ucxx_version }}
```
divyegala, line 82: `libraft` recipe installs the `distributed` component https://github.com/rapidsai/raft/blob/902ed56969f588cf504bd9e85ee8ee7419812c93/conda/recipes/libraft/recipe.yaml#L94 which has a build and runtime dependency on targets from `NCCL` and `ucxx` as seen here https://github.com/rapidsai/raft/blob/902ed56969f588cf504bd9e85ee8ee7419812c93/cpp/CMakeLists.txt#L392-L406

I think these changes are needed and I believe the reason it did not error out until now is because `distributed` is an `INTERFACE` target so if consumer libraries brought in NCCL themselves it would be okay. 

Happy to revert the changes, however, if this is considered an improper way of working with header only packages.

gforsyth: @divyegala I think these changes to the rattler-build recipe are reasonable, but can we break them out into a separate PR so we've got a cleaner history of when and why this was added?

bdice: Agreed, this seems fine but doesn't make sense to bundle with the NCCL wheels work.

divyegala: Reverted the commit, opened new PR https://github.com/rapidsai/raft/pull/2636

ci/build_wheel.sh
```diff
@@ -38,6 +38,14 @@ if [[ ${package_name} != "libraft" ]]; then
     )
 fi
 
+RAPIDS_CUDA_MAJOR="${RAPIDS_CUDA_VERSION%%.*}"
+if [[ ${RAPIDS_CUDA_MAJOR} == "12" ]]; then
```
bdice, line 42: Is there a reason we do this only for CUDA 12?

dependencies.yaml
```diff
@@ -700,3 +704,18 @@ dependencies:
           - matrix: null
             packages:
               - libucx>=1.15.0
+  depends_on_nccl:
+    common:
+      - output_types: conda
+        packages:
+          - &nccl_unsuffixed nccl>=2.19
```
bdice, line 711: Is 2.19 a high enough lower bound? The latest is 2.26. I'm pretty sure we rely on newer versions than 2.19 for Blackwell support and other features. I don't know the exact bound to use here but we should do some validation with the oldest NCCL release we claim to support.

ci/build_wheel.sh
divyegala: Yes, there's no NCCL wheel distributions available for CUDA 11 since April 2024 and even then, no arm64 binaries. I think we'll have to continue to vendor NCCL binaries with CUDA 11.

dependencies.yaml
divyegala: That's a good question. I think we could bump the version of NCCL as a follow-on to this PR depending on the . I just used the version constraint that was already present. I just checked our `ci-wheel` images and the system version on those is also `2.26`.

ci/build_wheel.sh
bdice: Great. That is fine! I just didn't know the constraints off the top of my head.

dependencies.yaml
bdice: Alrighty! If this matches other bounds already being used in RAFT, we can merge in the current state.

gforsyth (approved):
:shipit:

PR #2633, opened 2025-04-15
Remove CUDA whole compilation ODR violations

divyegala:
Reference https://github.com/rapidsai/build-infra/issues/38

It turns out there were no ODR violations. There might have been some possibly when the original issue was created, but it's likely those kernels ended up moving to cuVS between then and now.

### DIFF
--- a/cpp/include/raft/core/detail/macros.hpp
+++ b/cpp/include/raft/core/detail/macros.hpp
@@ -86,33 +86,7 @@
 // as a weak symbol rather than a global."
 #define RAFT_WEAK_FUNCTION __attribute__((weak))
 
-// The RAFT_HIDDEN_FUNCTION specificies that the function will be hidden
-// and therefore not callable by consumers of raft when compiled as
-// a shared library.
-//
-// Hidden visibility also ensures that the linker doesn't de-duplicate the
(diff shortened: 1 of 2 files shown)


KyleFromNVIDIA (approved):
Approved CMake changes

cpp/include/raft/core/detail/macros.hpp
```diff
@@ -86,33 +86,7 @@
 // as a weak symbol rather than a global."
 #define RAFT_WEAK_FUNCTION __attribute__((weak))
 
-// The RAFT_HIDDEN_FUNCTION specificies that the function will be hidden
```
robertmaynard, line 89: This can't be removed without changes to cuvs first.

Since cuvs uses the `RAFT_KERNEL` macro and builds some components with RDC enabled we need to keep this logic. CUVS explicitly decided to need re-implement the macro logic and instead leverage the RAFT versions

robertmaynard (changes requested):

cpp/include/raft/core/detail/macros.hpp
divyegala: Thanks for pointing this out! I re-added the macros.

robertmaynard (approved):

PR #2634, opened 2025-04-15
[FEA] Matrix shift rows and columns

jinsolp:
Implementation for in-place row or column shifting for matrices.

Supports the following shift-then-fill
- shifting then filling in with a single value
- shifting then filling in with a given matrix
- shifting then filling in with the rowid / colid

Has the following options
- `ShiftDirection`: can shift towards the beginning (left for columns, top for rows) or the end (right for columns, bottom for rows).
- `ShiftType`: either row or column 

Currently needed for the case where we need to add a self loop in a knn graph.


cpp/include/raft/matrix/shift.cuh
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * @brief col_shift: in-place shifts all columns by k columns to the right and fills the first k
+ * columns in with "val"
+ * @param[in] handle: raft handle
+ * @param[in out] in_out: input matrix of size (n_rows, n_cols)
+ * @param[in] val: value to fill in the first k columns (same for all rows)
+ * @param[in] k: shift size
+ */
+template <typename math_t, typename matrix_idx_t>
+void col_right_shift(raft::resources const& handle,
```
cjnolet, line 34: This naming is a bit confusing. Could we please instead have a "shift" function with an enum for "Direction::Left", "Direction::Right" and maybe use another enum for ShiftType::Row/ShiftType::Column? That would clean this up and minimize the number of free functions the user has to worry about invoking.

jinsolp: Thank you for the suggestion, totally agree.
We have `towards_end` and `towards_zero` as the shift direction because left/right doesn't work with row shifting.

cjnolet: Maybe `towards_start` and `towards_end`? `towards_zero` seems confusing (since we don't explicitly state that we're using indexing to determine the direction).

cpp/include/raft/matrix/detail/shift.cuh
```diff
@@ -0,0 +1,215 @@
...
+ */
+
+#pragma once
+
+#include <raft/core/detail/macros.hpp>
+#include <raft/core/device_mdspan.hpp>
+#include <raft/core/resources.hpp>
+
+namespace raft::matrix {
+enum ShiftDirection { TOWARDS_BEGINNING, TOWARDS_END };
```
cjnolet, line 24: Please put these in a file called `shift_types.hpp` (following the standard RAFT conventions here) and please don't bury them in implementation details. Users need to be able to find these types.

cpp/include/raft/matrix/shift.cuh
```diff
@@ -0,0 +1,146 @@
...
+                 "Shift size k should be smaller than the number of columns in matrix.");
+  } else {
+    RAFT_EXPECTS(static_cast<size_t>(in_out.extent(0)) > k,
+                 "Shift size k should be smaller than the number of rows in matrix.");
+  }
+  detail::shift(handle, in_out, val, k, shift_direction, shift_type);
+}
+
+/**
+ * @brief In-place operation. Shifts all rows or columns to shift_direction by k, and fills the
```
cjnolet, line 63: > Shifts all rows or columns to shift_direction by k

`k` doesn't seem to be an argument in this function.

cpp/include/raft/matrix/shift.cuh
```diff
@@ -0,0 +1,146 @@
...
+ *
+ * @param[in] handle: raft handle
+ * @param[in out] in_out: input matrix of size (n_rows, n_cols)
+ * @param[in] k: shift size
+ * @param[in] shift_direction: ShiftDirection::TOWARDS_BEGINNING shifts towards the 0th row/col
+ * direction, and ShiftDirection::TOWARDS_END shifts towards the (nrow-1)th row/col direction
+ * @param[in] shift_type: ShiftType::ROW shifts rows and ShiftType::COL shift columns
+ */
+template <typename math_t, typename matrix_idx_t>
+void shift_self(raft::resources const& handle,
```
cjnolet, line 130: Can we incorporate this as a specialized input into the prior functions so that we can avoid having a separate function just for this? Every template function we expose becomes another compiled kernel into the user's binary. The more we can consolidate, the better.

jinsolp: val is not a nullopt. If not specified, we fill with row ids and column ids.

cpp/include/raft/matrix/shift.hpp
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION.
```
cjnolet, line 2: RAFT is header-only, so it can quickly get confusing for users which file they need to include. We use the general convention of adding "_types" to the file which contain vocabulary types (and are safe to include in downstream APIs, because they, in general, don't require CUDA or any other complex types). Please rename this file to `shift_types.hpp`.

Issue #83, opened 2020-10-14
[FEA] Deploy raft globally


afender:
Currently, each repo carries its own RAFT under its own include directory. The versions of the different copies are not necessarily the same. 

Ideally, we would adopt RMM style for global deployment of RAFT.

That mostly requires

- always supporting the latest raft version
- shipping a new conda package

divyegala:
Now that cuML and cuGraph both point to HEAD in RAFT, is this still a valid issue? cc @dantegd

PR #2643, opened 2025-04-23
Fix nccl_comm.hpp warning: #83-D: type qualifier specified more than once

achirkin:
Warning treated as error downstream in cuVS
https://github.com/rapidsai/cuvs/actions/runs/14609667434/job/40985237074?pr=454#step:9:2675


tfeher (approved):
LGTM.

PR #2647, opened 2025-05-02
Decoupling multi gpu resources from nccl usage

jinsolp:
### Description
The existing `nccl_clique.hpp` assumes usage of nccl when using multiple gpus. However, this is not always the case, and we want to generalize the use of multi-gpu resource by decoupling it from nccl usage.
This PR is mostly a refactoring of existing code to lazy-load nccl for `device_resources_snmg` resource.

### Dependencies
**The merge of this PR will break cuVS**. Thankfully (and hopefully) it will not break anything else since `nccl_clique.hpp` was recently introduced and is only used for snmg neighbors index related code. 
[This PR in cuVS ](https://github.com/rapidsai/cuvs/pull/869) makes corresponding changes (changing include header file name etc...) to make this code work with cuVS. There are no functional differences.


cpp/include/raft/core/device_resources_snmg_nccl.hpp
```diff
@@ -0,0 +1,201 @@
...
+
+    RAFT_CUDA_TRY(cudaSetDevice(main_gpu_id_));
+  }
+
+  /**
+   * @brief Construct a SNMG resources instance with a subset of available GPUs
+   *
+   * @param[in] device_ids List of device IDs to be used by the NCCL clique
+   */
+  device_resources_snmg_nccl(const std::vector<int>& device_ids) : device_resources_snmg()
```
viclafargue, line 65:
```suggestion
  device_resources_snmg_nccl(const std::vector<int>& device_ids) : device_resources_snmg(device_ids)
```

cpp/include/raft/core/device_resources_snmg_nccl.hpp
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  bool has_resource_factory(resource::resource_type resource_type) const override
+  {
+    if (resource_type != raft::resource::MULTI_GPU && resource_type != raft::resource::NCCL_COMM &&
+        resource_type != raft::resource::CLIQUE_ROOT_RANK)
+      // for resources unrelated to SNMG switch current GPU to main GPU ID
+      cudaSetDevice(this->main_gpu_id_);
+    return raft::resources::has_resource_factory(resource_type);
+  }
```
viclafargue, line 178: The intent behind inheriting from `device_resources` and storing `main_gpu_id_` was to allow users to interact with multi-GPU resources as if they were single-GPU ones. This function handles automatic switching to the main GPU before providing any resources for single-GPU use. It might be more appropriate to move this logic into the `device_resource_snmg` class so both classes can benefit from it. Alternatively, if this feature is not essential, we should consider removing it entirely. cc @cjnolet

cpp/include/raft/core/device_resources_snmg_nccl.hpp
```diff
@@ -0,0 +1,201 @@
...
+  void set_memory_pool(int percent_of_free_memory) const
+  {
+    for (int rank = 0; rank < get_nccl_num_ranks(); rank++) {
+      const raft::resources& dev_res = set_current_device_to_rank(rank);
+      // check limit for each device
+      size_t limit = rmm::percent_of_free_device_memory(percent_of_free_memory);
+      raft::resource::set_workspace_to_pool_resource(dev_res, limit);
+    }
+    cudaSetDevice(this->main_gpu_id_);
+  }
```
viclafargue, line 169: This function could be useful for accelerating multi-GPU algorithms, regardless of NCCL usage. I recommend adding it to the `device_resource_snmg` class.

docs/source/cpp_api/core_resources.rst
```diff
@@ -58,20 +58,25 @@ namespace *raft::core*
 SNMG Device Resources
 ---------------------
 
-The `raft::device_resources_snmg` provides a convenient way to configure
+The `raft::device_resources_snmg` and `raft::device_resources_snmg_nccl` provides a convenient way to configure
```
viclafargue, line 61:
```suggestion
The `raft::device_resources_snmg` and `raft::device_resources_snmg_nccl` provide a convenient way to configure
```

viclafargue (changes requested):
Thanks @jinsolp! Just a few recommendations.

cpp/include/raft/core/device_resources_snmg.hpp
```diff
@@ -16,226 +16,99 @@
...
-
-    // initialize the NCCL clique
-    std::vector<raft::resources>& clique = raft::resource::get_nccl_clique(*this);
-    _init_clique(clique, device_ids);
-    _init_nccl_comms(clique);
+    raft::resource::set_root_rank(*this, root_rank);
 
-    RAFT_CUDA_TRY(cudaSetDevice(main_gpu_id_));
+    // initialize resources for the given device ids
+    std::vector<raft::resources>& clique = raft::resource::get_multi_gpu_resource(*this);
```
cjnolet, line 65: Maybe instead of clique we should call it something like "world_resources"? NCCL creates a clique, but we do not when we're just using "cudaSetDevice" manually.

cpp/include/raft/core/device_resources_snmg.hpp
```diff
@@ -16,226 +16,99 @@
...
-  bool has_resource_factory(resource::resource_type resource_type) const override
-  {
-    if (resource_type != raft::resource::NCCL_CLIQUE &&
-        resource_type != raft::resource::NCCL_COMM &&
-        resource_type != raft::resource::CLIQUE_ROOT_RANK)
-      // for resources unrelated to SNMG switch current GPU to main GPU ID
-      cudaSetDevice(this->main_gpu_id_);
-    return raft::resources::has_resource_factory(resource_type);
-  }
+  int main_gpu_id_;
```
cjnolet, line 84: Perhaps we should change 'root_rank' to this above. I like "main_gpu"id" better.

cpp/include/raft/core/device_resources_snmg_nccl.hpp
cjnolet: Agreed. This should exist in the base `device_resources_snmg`.

cpp/include/raft/core/device_resources_snmg_nccl.hpp
cjnolet: Agreed. Let's move this.

cpp/include/raft/core/resource/multi_gpu.hpp
```diff
@@ -0,0 +1,131 @@
...
+
+  ~root_rank_resource() override {}
+
+ private:
+  int root_rank_;
+};
+
+class root_rank_resource_factory : public resource_factory {
+ public:
+  resource_type get_resource_type() override { return resource_type::CLIQUE_ROOT_RANK; }
```
cjnolet, line 56: Can we rename "root_rank" to "main_gpu"? It's more general. Also, we can drop the CLIQUE, unless we're talking about a resource which is more specific to NCCL (NCCL and MPI have the notion of "cliques" but that's a very specific term for collective communications. Standard multi-gpu like manually switching devices to run things concurrently with multiple gpus is not a "clique".

cpp/include/raft/core/resource/multi_gpu.hpp
```diff
@@ -0,0 +1,131 @@
...
+  if (!res.has_resource_factory(resource_type::MULTI_GPU)) {
+    res.add_resource_factory(std::make_shared<multi_gpu_resource_factory>());
+  }
+  return *res.get_resource<std::vector<raft::resources>>(resource_type::MULTI_GPU);
+};
+
+/**
+ * @brief Get number of ranks
+ */
+inline int get_num_ranks(resources const& res)
```
cjnolet, line 86: Rank is really an MPI/NCCL term (specifically a collective comms term). Let's use "devices" here instead.

cpp/include/raft/core/device_resources_snmg.hpp
jinsolp: Thinking of it, why do we need the `main_gpu_id_` within `device_resources_snmg` class? Can we just use the `get_root_rank` (or after changing it, the `get_main_gpu`) function to retrieve the main gpu id instead? @viclafargue

cjnolet: Yes, ideally there's no internal state (e.g. member variables) being stored directly on the `device_resources` instances and all state is coming from the underlying `raft::resources` (and being accessed and mutated with `raft::resource::get_` and `raft::resources::set_`).

cjnolet: `raft::resources` (and `raft::resource` are the underlying contracts between the user and the lower-level APIs in RAFT/cuVS/cuML/cuGraph/cuOpt), `raft::device_resources_...` is just a convenience API so users don't have to concern themselves with worrying about calling `raft::resource::set_xxxx` before they can call any of our functions.

jinsolp: I've made it so that we call the `raft::resource::set_main_gpu_id` in the constructor of `device_resources_snmg`

jinsolp:
Changes; 
- use the term "main_gpu". Avoid usage of "rank", and "clique".
- remove `main_gpu_id_` from `device_resources_snmg` class and keep the same functionality by using `set_main_gpu_id` and `get_main_gpu_id`

cpp/include/raft/core/device_resources_snmg.hpp
cjnolet: Perfect!

cpp/include/raft/core/resource/multi_gpu.hpp
```diff
@@ -0,0 +1,134 @@
...
+  return raft::resource::get_multi_gpu_resource(res).size();
+}
+
+/**
+ * @brief Get specific GPU's raft::resources object
+ */
+inline const raft::resources& get_device_resources_for_gpu_id(resources const& res, int gpu_id)
+{
+  std::vector<raft::resources>& world_resources = raft::resource::get_multi_gpu_resource(res);
+  return world_resources[gpu_id];
```
viclafargue, line 100: It is very important to make a distinction between a rank and a device ID. Here `world_resources[gpu_id]` would effectively access the `device_resource` for the rank number `gpu_id`.

cpp/include/raft/core/device_resources_snmg.hpp
viclafargue: The root rank (not device ID) is part of the clique (unlike the main GPU ID which may lie outside of it) and serves a specific role in the MG algorithm. For example, operations such as ANN merging can be performed on the root rank. In contrast, the main GPU is primarily used to enable the `device_resources_snmg` class to function as a single-GPU resource. It is tied to this utility class and is not directly involved in MG operations.

viclafargue (changes requested):

cpp/include/raft/core/resource/multi_gpu.hpp
cjnolet: Depending on how we are using the term "gpu_id" here. Are we saying the id for a specific GPU? Those are allowed to skip numbers. A `rank` (or what I also call `canonical id` is not allowed to skip numbers. I think this is the differentiation that @viclafargue is referring to.

jinsolp: It is being used more like a rank. (e.g. 0,1,2,3 when we actually use GPUs 2,4,6,7) but I agree that it might be confusing.

viclafargue: What was previously referred to as the root rank has to be part of the set of device IDs. The function formerly named `set_current_device_to_root_rank` (now renamed to `set_current_device_to_main_gpu`) searched for the root rank to return its `device_resource`. For example, consider a clique with GPU IDs {2, 4, 6, 7} Previously the root rank defaulted to 0, and the merge operation was performed on GPU 2. Now, if the `device_resource_mnmg_nccl` class is instantiated on GPU 0, the ANN merge operation will be executed on GPU 2. However, if it is instantiated on say GPU 8, a segmentation fault will occur.

jinsolp:
After discussing offline with @viclafargue , decided to bring back `main_gpu_id_` and bring back ranks to differentiate from gpu ids.

viclafargue (approved):
This is ready to go. Thanks for the excellent work!

jinsolp:
@viclafargue Sorry for making changes to this PR Victor. After discussing with @cjnolet offline, it seems like having separate `devices_resources_snmg` `devices_resources_snmg_nccl` and forcing users to know which one they should use may not be a good idea. I've made additional changes so that we decouple nccl but keep one `devices_resources_snmg` by initializing nccl communication as part of the nccl resource.

cpp/include/raft/core/device_resources_snmg.hpp
```diff
@@ -47,195 +41,100 @@ class device_resources_snmg : public device_resources {
...
-    raft::resource::set_nccl_clique_root_rank(*this, root_rank);
-
-    // initialize the NCCL clique
-    std::vector<raft::resources>& clique = raft::resource::get_nccl_clique(*this);
-    _init_clique(clique, device_ids);
-    _init_nccl_comms(clique);
-
-    RAFT_CUDA_TRY(cudaSetDevice(main_gpu_id_));
+    int main_gpu_id;
+    RAFT_CUDA_TRY(cudaGetDevice(&main_gpu_id));
```
viclafargue, line 65: The main GPU would need to be set in the resource here.

viclafargue (approved):
Thanks @jinsolp, LGTM apart from one small thing

cpp/include/raft/core/device_resources_snmg.hpp
jinsolp: thanks for pointing out!

cpp/include/raft/core/device_resources_snmg.hpp
```diff
@@ -48,193 +42,94 @@ class device_resources_snmg : public device_resources {
...
    * @brief SNMG resources instance copy constructor
    *
-   * @param[in] clique A SNMG resources instance
+   * @param[in] world A SNMG resources instance
    */
-  device_resources_snmg(const device_resources_snmg& clique)
-    : device_resources(clique), main_gpu_id_(clique.main_gpu_id_)
-  {
-  }
+  device_resources_snmg(const device_resources_snmg& world) : device_resources(world) {}
```
achirkin, line 75: The copy constructor doesn't copy the main_gpu_id_, which causes [cudaErrorInvalidDevice: invalid device ordinal](https://github.com/rapidsai/cuvs/pull/869#issuecomment-2876235941) at a later point in time.
```suggestion
  device_resources_snmg(const device_resources_snmg& world)
    : device_resources(world), main_gpu_id_(world.main_gpu_id_)
  {
  }
```

achirkin (changes requested):
I'm not sure what is the intention behind the copy constructor of the device resources. Could you please write down in the docstring how it is to be used? Shall the user copy a single snmg resource into multiple threads?
I assume the copies are supposed to share the resources under the hood (via the shared pointers). Are all of the relevant resources initialized by the time copy constructor is called (because if not, they are not shared)?
Also why the copy constructor allowed and the move constructor is deleted?

cpp/include/raft/core/device_resources_snmg.hpp
jinsolp: thank you for pointing this out, I was unable to test yesterday because of devcontainer issues.

jinsolp: @achirkin As for the move constructor, I was sticking to the existing code, and it looks like other raft resources are set to be that way ([link](https://github.com/rapidsai/raft/blob/217189d67dab43cacb1b7f5076a6485502ef6010/cpp/include/raft/core/device_resources.hpp#L74-L76)). I'm thinking it's because we don't want to transfer ownership of resources. ccing @cjnolet for verification.

achirkin: Ah, right, that might be related to the mutex member in the parent resource. We had some thread-safety discussions related to that (each copy has its own mutex)

achirkin (approved):
Thanks for fixing the copy constructor!

jinsolp:
@divyegala ready to go!

PR #2650, opened 2025-05-06
Adding host COO utility

viclafargue:


cjnolet:
@viclafargue the `raft::sparse::COO` is deprecated and is really old. We put a lot of thought into the new COO API (`raft::sparse::host_coordinate_matrix`). It would be better to use that instead at this point.

PR #2657, opened 2025-05-12
CUDA 12.9 use updated compression flags

robertmaynard:
Set more fatbin compression flags to maintain existing binary sizes


bdice (approved):

jakirkham:
Thanks Rob and Bradley! 🙏

PR #2666, opened 2025-05-14
Python wrapper for `device_resources_snmg`

jinsolp:
Python wrapper for the `device_resources_snmg` introduced in https://github.com/rapidsai/raft/pull/2647

Need this in for 25.06 (needed for; cuML python level usage for building knn graph in multi-gpu in UMAP)


python/pylibraft/pylibraft/common/handle.pxd
```diff
@@ -51,3 +52,13 @@ cdef class DeviceResources:
...
     cdef int n_streams
+
+cdef extern from "raft/core/device_resources_snmg.hpp" namespace "raft":
+    cdef cppclass device_resources_snmg(device_resources):
+        device_resources_snmg() except +
+        device_resources_snmg(const vector[int]& device_ids) except +
+        device_resources_snmg(const device_resources_snmg&) except +
+
+cdef class DeviceResourcesSNMG:
+    cdef unique_ptr[device_resources] c_obj
```
divyegala, line 63: Shouldn't it be:
```suggestion
    cdef unique_ptr[device_resources_snmg] c_obj
```

python/pylibraft/pylibraft/common/handle.pyx
```diff
@@ -236,3 +236,71 @@ def auto_sync_handle(f):
...
+                ids.push_back(id)
+            snmg_ptr = new device_resources_snmg(ids)
+
+        self.c_obj.reset(snmg_ptr)
+
+    def sync(self):
+        """
+        Issues a sync on the stream set for this instance.
+        """
+        self.c_obj.get()[0].sync_stream()
```
divyegala, line 281: What does this mean in the SNMG context?

jinsolp: I tried to keep it consistent with the `DeviceResources`. This will sync stream on the main GPU (gpu for `main_gpu_id_` in the cpp layer).

divyegala: Let's remove it then, for now. For me, the expected functionality would be that it would sync all streams for all devices.

cjnolet: I think we want to follow the same semantics as all the other `raft::resources`. There's a notion of "main_stream" and "worker_stream". We want to make sure we differentiate between them.

divyegala: From my understanding, it seems like the main GPU will have two streams. The main stream will be synced by this function, however no work will be scheduled on that stream.

The stream on which real work is scheduled for the main GPU will be part of the world resources.

jinsolp: @divyegala you are right, and I believe for now the users of this class won't ever really have to do the sync at the python level.

divyegala: Okay cool. So actually in UMAP's case everything but the KNN graph construction will actually run on the main stream, so we should keep this.

divyegala (approved):

jinsolp:
@divyegala I think this is good to go!

PR #2671, opened 2025-05-15
Prefer host gather when dataset is available both on host and device

tfeher:
The `raft::matrix::sample_rows` function can select rows both from host and device datasets. This PR changes how the case is handled when the dataset is accessible both from host and device.

When the dataset pointer is accessible both on host and device, then in theory we can subsample it both with our CPU and GPU implementation. In practice,  it is likely that the data is physically located in host memory, therefore it is preferred to process it on the CPU side.

This has significance when the dataset is larger than GPU memory. In that case subsampling using the GPU kernels would lead to page trashing and low performance. In some HMM implementations this can also lead to application crash.

This PR fixes the problem by preferring CPU side subsampling when the dataset is accessible on the CPU.


achirkin (approved):
Sounds reasonable and the code looks good!

tfeher:
I could reproduce the CI error locally, investigating.

Issue #2491, opened 2024-11-13
[BUG] pylibraft compilation errors with Cython 3.1 nightlies


jameslamb:
**Describe the bug**

I recently observed pylibraft compilation errors in this project's pip devcontainer and not in any other wheel builds, like this:

```text
Error compiling Cython file:
------------------------------------------------------------

cdef device_matrix_view[const_float, int64_t, row_major]
^

/home/coder/raft/python/pylibraft/pylibraft/common/mdspan.pyx:203:0: Function signature does not match previous declaration
```

ref: https://github.com/rapidsai/raft/pull/2490#issuecomment-2474255976

The root cause, per https://github.com/rapidsai/raft/pull/2490#issuecomment-2474375209, is that those builds were pulling in a pre-release of Cython 3.1.

```text
Downloading Cython-3.1.0a1-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (3.2 kB)
```

That was "fixed" in #2490 by pinning to `cython<=3.0.11`, but we should work to remove that ceiling.

Specifically, we should:

1. create a smaller, not-RAFT-specific minimum reproducible example
2. modify RAFT source code and/or submit a bug report upstream with Cython
3. remove the ceiling

**Steps/Code to reproduce bug**

Modify this project's Cython dependency to something like `cython>=3.1.0a0' and try to build in the pip devcontainer.

**Expected behavior**

Successful builds of `pylibraft` with Cython 3.1.

**Environment details (please complete the following information):**

N/A

**Additional context**

N/A

vyasr:
The problem here is https://github.com/cython/cython/issues/6922

PR #2686, opened 2025-05-27
Work around Cython ctypedef bug

vyasr:
This is a workaround for https://github.com/cython/cython/issues/6922 that resolves https://github.com/rapidsai/raft/issues/2491. I expect that Cython will have a proper fix for that issue in an upcoming release with https://github.com/cython/cython/pull/6357, but in the meantime we can get raft working today with this minor change.


vyasr:
> Looks good. The devcontainer failure is from a typo in one of the forward mergers -- I can fix that in a separate PR if it won't block this getting merged.

Sounds good, thanks for fixing! Assuming everything else passes I'll just request an admin merge so that it won't block. No point in wasting CI cycles.

PR #2687, opened 2025-05-27
fix(devcontainers): typo in container name

gforsyth:
In #2686 the devcontainer job failed and I noticed it's just a typo probably resulting from a slightly broken forward-merger.  Fixing that up here.


jakirkham:
Thanks Gil! 🙏

PR #2698, opened 2025-06-10
Fix several issues that breaks LLVM

vitor1001:
Fix several issues that breaks LLVM:
- do not declare a function _RAFT_HOST_DEVICE if it has no host implementation;
- use std::min() instead of min();
- avoid usign non-existent methods on templated class;
- add missing include.


csadorf:
> Hi! Should I assign a reviewer manually?

@divyegala Can you vet the change and assign a reviewer?

divyegala:
@lowener @aamijar can you both take a look at this?

vitor1001:
Hi @lowener @aamijar!

The tests passed, is there anything you need from my side?

lowener (approved):
LGTM

PR #2703, opened 2025-06-12
refactor(shellcheck): fix all remaining warnings/errors

gforsyth:
Finishes up the shellcheck work tracked in rapidsai/build-planning#135


.shellcheckrc
```diff
@@ -0,0 +1,2 @@
+# Disable file checks (otherwise every use of `gha-tools` will get flagged)
+disable=SC1091
```
jameslamb, line 2: I think this should be added to the `ci-codeowners` block in CODEOWNERS:

https://github.com/rapidsai/raft/blob/23c942052bf17f8b116d7c6f79b37042296b93c3/.github/CODEOWNERS#L15-L17

and to the list of files that shouldn't trigger CI that needs GPUs:

https://github.com/rapidsai/raft/blob/23c942052bf17f8b116d7c6f79b37042296b93c3/.github/workflows/pr.yaml#L61-L82

build.sh
```diff
@@ -453,39 +451,41 @@ if (( ${NUMARGS} == 0 )) || hasArg libraft || hasArg docs || hasArg tests || has
       MSG="${MSG}<br/>parallel setting: $PARALLEL_LEVEL"
       MSG="${MSG}<br/>parallel build time: $compile_total seconds"
       if [[ -f "${LIBRAFT_BUILD_DIR}/libraft.so" ]]; then
-          LIBRAFT_FS=$(ls -lh ${LIBRAFT_BUILD_DIR}/libraft.so | awk '{print $5}')
+          LIBRAFT_FS=$(find "${LIBRAFT_BUILD_DIR}" -name libraft.so -printf '%s')
```
jameslamb, line 454: I recall seeing @divyegala change a similar thing in `cuml` from `ls -lh` to `stat -c`: https://github.com/rapidsai/cuml/pull/6780#discussion_r2138347882

and adding in the file size.

Would you consider doing that here too, since this is touching this line anyway?

jameslamb (approved):
Looks good to me! Left some small suggestions.

build.sh
gforsyth: Yeah, `find` is doing the same thing here, but printing out the number in bytes -- I can add the `awk` magic to format it as `%.2f MB`

jameslamb: ok cool, thanks for letting me expand the scope of this a tiny bit 😬

divyegala (approved):

PR #2707, opened 2025-06-17
Fix MST bug for graph with identical edge weights

### LINKED ISSUE #2706: [BUG]
**Describe the bug**
The current MST implementation leverages the Boruvka's algorithm. It starts from an empty set where every (super)vertex is a single component and for each iteration, (safe)edges with the lowest weight connecting components to other components are added. Distinct weights guarantee that the MST of the graph is unique and in case the weights are not unique, a `consistent` mechanism to break ties needs is required. This implementation internally alters weights with the function [alteration](https://github.com/rapidsai/raft/blob/branch-25.08/cpp/include/raft/sparse/solver/detail/mst_solver_inl.cuh#L126:L130) by pre-compute a set of [random variables](https://github.com/rapidsai/raft/blob/branch-25.08/cpp/include/raft/sparse/solver/detail/mst_solver_inl.cuh#L235:L236) which will be multiplied by [max](https://github.com/rapidsai/raft/blob/branch-25.08/cpp/include/raft/sparse/solver/detail/mst_kernels.cuh#L313) before being added to the original edge weights making them unique. In fact, `max` is the maximum alteration value that does not change the relative weights order however, this value is guaranteed to be zero if all weights are identical (which makes sense and serves its purpose) but doesn't guarantee distinct weights which is assumed by this implementation.

```
RAFT_EXPECTS(curr_mst_edge_count <= max_mst_edges,
                 "Number of edges found by MST is invalid. This may be due to "
                 "loss in precision. Try increasing precision of weights.");
```
As a result, more edges are selected at each iteration of the MST hence failing the above assertion

Below is a cuGraph reproducer but if needed I can provide a C++ raft only reproducer with the karate datasets
```
import cudf
import cugraph
from cugraph.datasets import karate

G = karate.get_graph(
    download=True,
    create_using=cugraph.Graph(directed=False),
    ignore_weights=False,
)

mst = cugraph.minimum_spanning_tree(G) // which is calling `raft::sparse::solver::mst`.

```

**Expected behavior**
MST should produce valid result irrespective of weights being identical or distinct.

 - Environment location: [Bare-metal, Docker]
 - Method of RAFT install: [conda, Docker, from source]

**Additional context**
Set the `max` value to 1 if the computed he maximum alteration value is 0 thereby guaranteeing the edge weights are unique. However, the MST implementation may become non deterministic for graphs with edge weights that are not distinct.

jnke2016:
This PR addresses this [issue](https://github.com/rapidsai/raft/issues/2706) by enforcing distinct edge weights if initial edge weights are identical.

closes #2706 
closes https://github.com/rapidsai/cugraph/issues/4458


divyegala (approved):

divyegala (changes requested):
Blocking review pending unit test case addition

cpp/include/raft/sparse/solver/detail/mst_solver_inl.cuh
```diff
@@ -210,7 +210,9 @@ alteration_t MST_solver<vertex_t, edge_t, weight_t, alteration_t>::alteration_ma
   auto init  = tmp.element(1, stream) - tmp.element(0, stream);
   auto max   = thrust::transform_reduce(
     policy, begin, end, alteration_functor<weight_t>(), init, thrust::minimum<weight_t>());
-  return max / static_cast<alteration_t>(2);
+  // Enforce distinct weights if initial edge weights are identical by returning
+  // a value of 1
+  return max > 0 ? max / static_cast<alteration_t>(2) : 1;
```
wolfram77, line 215: @jnke2016 Consider the case where all edge weights are `0.1`, and not `0.1 | 0.2`. Here, I think, the tests would fail - as code is assuming that if alternation_max is `0` (all weights are the same), they must be `1`. For this case, you could pick `max` as any edge weight - which here should be `0.1`.

jnke2016: @wolfram77 I just pushed a test with all edge weights being 0.1 and it passes.

jnke2016: > hey must be 1. For this case, you could pick max as any edge weight - which here should be 0.1. For this case, you could pick max as any edge weight - which here should be 0.1

That value doesn't really matter. Only the relatively order of the edges does where in this cases there is none.

wolfram77: @jnke2016 Oh yeah, you are right, my bad.

jnke2016: no worries @wolfram77, if you have any other concerns please feel free to reach out.

PR #2712, opened 2025-06-21
Fix laplacian self-loops

### LINKED ISSUE #2711: [BUG] `compute_graph_laplacian_kernel` does not handle self-loops correctly
According to [wikipedia](https://en.wikipedia.org/wiki/Laplacian_matrix#:~:text=Graph%20self%2Dloops%2C%20manifesting%20themselves%20by%20non%2Dzero%20entries%20on%20the%20main%20diagonal%20of%20the%20adjacency%20matrix%2C%20are%20allowed%20but%20do%20not%20affect%20the%20graph%20Laplacian%20values.) and the [scipy implementation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csgraph.laplacian.html) self-loops should not affect the laplacian matrix. However, the current raft compute_graph_laplacian_kernel does utilize self-loops (diagonal values in adj matrix) in the final output.

aamijar:
Resolves #2711


divyegala (commented):
Can you add a test for this?

aamijar:
> Can you add a test for this?

Yes, test added

PR #2713, opened 2025-06-21
Fix unsafe cast `coo_remove_scalar`

### LINKED ISSUE #2663: [BUG] unsafe cast in `coo_remove_scalar`
This seems wrong, it should be using `nnz_t*` instead of `unsigned long long int*` I found this bug when trying to work on an overloaded function https://github.com/rapidsai/raft/issues/2659.

https://github.com/rapidsai/raft/blob/c2dc3124ce3fbcb5ff2ccabd88d7f57570b6aea9/cpp/include/raft/sparse/op/detail/filter.cuh#L166-L166

aamijar:
Resolves #2663


cpp/include/raft/sparse/op/detail/filter.cuh
```diff
@@ -162,8 +162,16 @@ void coo_remove_scalar(COO<T, idx_t, nnz_t>* in,
   linalg::coo_degree(in->rows(), in->nnz, row_count.data(), stream);
   RAFT_CUDA_TRY(cudaPeekAtLastError());
 
-  linalg::coo_degree_scalar(
-    in->rows(), in->vals(), in->nnz, scalar, (unsigned long long int*)row_count_nz.data(), stream);
+  if constexpr (std::is_same_v<nnz_t, uint64_t>) {
```
lowener, line 165: Can you explain why casting is necessary in just that case?

aamijar: Without that cast, the code won't compile because the atomicAdd function doesn't support uint64_t

divyegala: `uint64_t` is not the same as `unsigned long long`?

aamijar: I thought it is, but it does not compile with uint64_t, which is why someone put a cast for this in the current code.

aamijar: Are we good to go on this one?

viclafargue (approved):

lowener (approved):

Issue #1616, opened 2023-06-26
[FEA] Use mdspan from CCCL


cjnolet:
This feature has been a long time coming, but when we initially set out to start using mdspan, the CCCL didn’t yet have support for it, which caused us to ship out own copy of the reference implementation. 

Now that CCCL has support for mdspan, we should include `libcu++` as a dependency, update our infrastructure to use that, and remove the copy of the Kokkos version from our `include/thirdparty` directory.

PR #2714, opened 2025-06-23
Avoid using internal method std::experimental::details::alignTo().

vitor1001:
There is a equivalent method on the raft:: namespace and it is much cleaner to use that instead.


divyegala:
@vitor1001 in that case, I wanted to point out that we have a long standing [issue](https://github.com/rapidsai/raft/issues/1616) to use the mdspan implementation maintained in `libcudacxx` in the `cuda::std::` namespace ([documentation](https://nvidia.github.io/cccl/libcudacxx/standard_api/container_library/mdspan.html)), and we would prefer to migrate our implementation to that version.

Would it be possible for you to redirect your effort into updating to this implementation instead?

vitor1001:
> Would it be possible for you to redirect your effort into updating to this implementation instead?

@divyegala: sounds good to me, I'll try. But in any case this PR is a good thing. If we want to migrate to another implementation, the first thing to do is to stop relying on internal implementation details. And in the case of this PR, it's quite trivial.

divyegala (approved):

divyegala:
@vitor1001 you will need to run pre-commit on this PR.

vitor1001:
> @vitor1001 you will need to run pre-commit on this PR.

@divyegala done!

divyegala:
@vitor1001 could you please address the build failures in CI?

```
tests/CMakeFiles/CORE_TEST.dir/core/mdspan_copy.cpp.o -c $SRC_DIR/cpp/tests/core/mdspan_copy.cpp
 │ │ In file included from $SRC_DIR/cpp/include/raft/core/device_mdspan.hpp:22,
 │ │                  from $SRC_DIR/cpp/include/raft/core/detail/copy.hpp:19,
 │ │                  from $SRC_DIR/cpp/include/raft/core/copy.hpp:18,
 │ │                  from $SRC_DIR/cpp/tests/core/mdspan_copy.cpp:19:
 │ │ $SRC_DIR/cpp/include/raft/util/cuda_dev_essentials.cuh: In function 'void raft::forEach(int, L)':
 │ │ $SRC_DIR/cpp/include/raft/util/cuda_dev_essentials.cuh:97:27: error: 'blockDim' was not declared in this scope
 │ │    97 |   int idx              = (blockDim.x * blockIdx.x) + threadIdx.x;
 │ │       |                           ^~~~~~~~
 │ │ $SRC_DIR/cpp/include/raft/util/cuda_dev_essentials.cuh:97:40: error: 'blockIdx' was not declared in this scope
 │ │    97 |   int idx              = (blockDim.x * blockIdx.x) + threadIdx.x;
 │ │       |                                        ^~~~~~~~
 │ │ $SRC_DIR/cpp/include/raft/util/cuda_dev_essentials.cuh:97:54: error: 'threadIdx' was not declared in this scope
 │ │    97 |   int idx              = (blockDim.x * blockIdx.x) + threadIdx.x;
 │ │       |                                                      ^~~~~~~~~
 │ │ $SRC_DIR/cpp/include/raft/util/cuda_dev_essentials.cuh:98:39: error: 'gridDim' was not declared in this scope
 │ │    98 |   const int numThreads = blockDim.x * gridDim.x;
 ```

vitor1001:
> I think it should be possible to compile this function, it is tagged as __device__ inline. I don't think we can remove this function without checking if any other downstream library is not using it.

Good catch, I was needing a coffee. It doesn't compile because I was including a `.cuh` from a `.hpp`.

BTW, I confirmed that cuda::std::mdspan doesn't provide an details::alignTo().

What do you suggest? I can move `raft::alignTo()` to a `.hpp` header (do you suggest one?) or duplicating it on mdpan.hpp.

vitor1001:
> @vitor1001 a duplicate should suffice for now, I don't want to move the function and break someone downstream by mistake. Thanks!

Agreed, done!

PR #2715, opened 2025-06-24
[REVIEW] Fix build of `math_device.cu` for `__CUDA_ARCH__ < 530` and `__CUDA_A…

legrosbuffle:
…RCH__ < 800`.

 - There is a missing `else` after the first `if constexpr` which means that `raft::exp` is instantiated when `__CUDA_ARCH__ < 530` and `Type` is `__half`.
 - In addition, the `if constexpr` branch for `nv_bfloat16` must be applied when `__CUDA_ARCH__ < 530` too.

Details:

When `__CUDA_ARCH__ < 530`, the code becomes (after preprocessing):

```
    if constexpr (std::is_same_v<Type, __half>) {
      return __float2half(raft::cos(__half2float(in)));
    }
    return raft::cos(in);
```

This is incorrect for two reasons:
 - It always tries to instantiate `raft::cos<Type>` (note that there is no constexpr `else`). When `Type` is `__half`, this fails with a static assert in `raft::cost<__half>`, which rightfully complains that __half is unsupported. The fix is to add an `else`.
 - When  `Type` is `nv_bfloat16`, the `if constexpr` handling `nv_bfloat16` has been preprocessed out, so we're again trying to instantiate a failing `raft::cos<nv_bfloat16>`.


achirkin:
Thanks for working on this! Could you please add to the description what are the symptoms of the problem this tries to solve? On the first look, to me everything seems fine with the original code: the compiler should be able to discard everything after the return statements in the first branch anyway, so it's not clear why {else/#elif}s  are necessary.

legrosbuffle:
> Thanks for working on this! Could you please add to the description what are the symptoms of the problem this tries to solve? On the first look, to me everything seems fine with the original code: the compiler should be able to discard everything after the return statements in the first branch anyway, so it's not clear why {else/#elif}s are necessary.

Done, hopefully the issue is clearer.

> the compiler should be able to discard everything after the return statements in the first branch anyway

If there is no `else` to the `if constexpr`, what comes after the returning statements will still be instantantiated, so it needs to compile. Without the `else` it does not.

legrosbuffle:
I think this might be the same issue as https://github.com/rapidsai/raft/pull/2718, so I'm putting this on hold for now.

aamijar:
Hi @legrosbuffle did this end up being a duplicate of #2718?

legrosbuffle:
> Hi @legrosbuffle did this end up being a duplicate of #2718?

Yes, it was. I'll close this.

PR #2717, opened 2025-06-24
Fix the constructor for `coordinate_structure` for non-zero `nnz`.

legrosbuffle:
The class fails to allocate the right storage for `c_rows_` and `c_cols_` when `nnz` is non-zero.

```
resources handle;
auto m = make_device_coordinate_structure(handlei, 5, 5, /*nnz*/5);
m.view();  // crash: creates a span on `c_rows_.data()` (`nullptr` with
`size` = 5).
```

To reproduce, run `cpp/tests/core/sparse_matrix.cu` with assertions enabled.


divyegala:
> To reproduce, run cpp/tests/core/sparse_matrix.cu with assertions enabled.

@legrosbuffle can you comment on this? I don't see the assertions disabled in our tests, I must be missing something.

divyegala:
@legrosbuffle I believe you will need to run pre-commit on this PR.

legrosbuffle:
> @legrosbuffle I believe you will need to run pre-commit on this PR.

Done.

PR #2718, opened 2025-06-24
[REVIEW] Fix compile error of `abs_op` when compiling with `clang` 

legrosbuffle:
```
cpp/include/raft/core/math.hpp:73:17: error: static assertion failed due to requirement 'sizeof(__half) != sizeof(__half)': __half is only supported on __CUDA_ARCH__ >= 530
   73 |   static_assert(sizeof(T) != sizeof(T), "__half is only supported on __CUDA_ARCH__ >= 530");
      |                 ^~~~~~~~~~~~~~~~~~~~~~
cpp/include/raft/core/operators.hpp:99:18: note: in instantiation of function template specialization 'raft::abs<__half>' requested here
   99 |     return raft::abs(in);
      |                  ^
```


legrosbuffle:
I think I understand what's happening (filed https://github.com/rapidsai/raft/issues/2719 with the details). WDYT about this fix ?

cpp/include/raft/core/math.hpp
```diff
@@ -66,26 +66,34 @@ constexpr RAFT_INLINE_FUNCTION auto abs(T x)
...
+#ifdef __CUDA_ARCH__
 #if (__CUDA_ARCH__ >= 530)
   return ::__habs(x);
 #else
   // Fail during template instantiation if the compute capability doesn't support this operation
   static_assert(sizeof(T) != sizeof(T), "__half is only supported on __CUDA_ARCH__ >= 530");
   return T{};
 #endif
+#else
+  return T{};
```
vitor1001, line 78: As this should never happen, maybe we should assert() here?

cpp/include/raft/core/math.hpp
```diff
@@ -66,26 +66,34 @@ constexpr RAFT_INLINE_FUNCTION auto abs(T x)
 template <typename T>
 RAFT_DEVICE_INLINE_FUNCTION typename std::enable_if_t<std::is_same_v<T, __half>, __half> abs(T x)
 {
+#ifdef __CUDA_ARCH__
```
divyegala, line 69: You could move these preprocessor macro conditionals outside the function? That would force substitution-failure error by making sure no function definition for `abs` exists if `CUDA_ARCH` is undefined or `CUDA_ARCH < 530`. I think it would simplify the structure of these functions quite a bit.

legrosbuffle: I was doing that in a previous version, but then there is no matching `abs(__half)` because we are explicitly removing `__half` from the list of supported types above:

```
template <typename T>
constexpr RAFT_INLINE_FUNCTION auto abs(T x)
  -> std::enable_if_t<!std::is_same_v<float, T> && !std::is_same_v<double, T> &&
#if defined(_RAFT_HAS_CUDA)
                        !std::is_same_v<__half, T> && !std::is_same_v<nv_bfloat16, T> &&
#endif
                        !std::is_same_v<int, T> && !std::is_same_v<long int, T> &&
                        !std::is_same_v<long long int, T>,
                      T>
{ ...
```

So the `#if defined(_RAFT_HAS_CUDA)` has to become `#if defined(_RAFT_HAS_CUDA) && defined(__CUDA_ARCH__)` there.

For other functions:

```
template <typename T,
          std::enable_if_t<CUDA_CONDITION_ELSE_TRUE(((!std::is_same_v<T, __half> &&
                                                      (!std::is_same_v<T, nv_bfloat16>)))),
                           int> = 0>
RAFT_INLINE_FUNCTION auto cos(T x)
```

this becomes:

```
/** Cosine */
template <typename T,
          std::enable_if_t<
#ifdef defined(__CUDA_ARCH__)
              CUDA_CONDITION_ELSE_TRUE(((!std::is_same_v<T, __half> &&
                                                      (!std::is_same_v<T, nv_bfloat16>))))
#else
true
#endif
,
                           int> = 0>
RAFT_INLINE_FUNCTION auto cos(T x)
```

So in the end if felt worse. @divyegala  WDYT ?

legrosbuffle: I think the fact that this code: 

```
#if defined(_RAFT_HAS_CUDA)
  !std::is_same_v<__half, T> && !std::is_same_v<nv_bfloat16, T> &&
#endif
```

exists at all makes it clear that the intent was to have `abs(half_)` exist when `_RAFT_HAS_CUDA` is false (when it's true, it's provided below) ?

divyegala: I think this shows that the intent is to provide overloads for `__half` and `nvbfloat_16` when `_RAFT_HAS_CUDA = true` and to not provide any overloads when ``_RAFT_HAS_CUDA = false`. The more I think on it, I feel that `_RAFT_HAS_CUDA` and `__CUDA_ARCH__` are completely dependent variables in the context of a header-only library like RAFT and will always have their values in tandem, i.e `_RAFT_HAS_CUDA = true` when `__CUDA_ARCH__` is defined. I cannot think of a situation to provide an overload of `abs(__half)` when `_RAFT_HAS_CUDA = false`, unless I am missing some context, but is this a use-case for you?

Let's look at this from negations, when `_RAFT_HAS_CUDA = false` and `__CUDA_ARCH__` is undefined, the code evaluates to (without any overloads of `__half` or `nvbfloat_16` present):
```
template <typename T>
RAFT_INLINE_FUNCTION auto abs(T x)
  -> std::enable_if_t<std::is_same_v<float, T> || std::is_same_v<double, T> ||
                        std::is_same_v<int, T> || std::is_same_v<long int, T> ||
                        std::is_same_v<long long int, T>,
                      T>
{
  return std::abs(x);
}

template <typename T>
constexpr RAFT_INLINE_FUNCTION auto abs(T x)
  -> std::enable_if_t<!std::is_same_v<float, T> && !std::is_same_v<double, T> &&
                        !std::is_same_v<int, T> && !std::is_same_v<long int, T> &&
                        !std::is_same_v<long long int, T>,
                      T>
{
  if constexpr (std::is_unsigned_v<T>) {
    return x;
  } else {
    return x < T{0} ? -x : x;
  }
}
```

And when `_RAFT_HAS_CUDA = true` and `__CUDA_ARCH__` is defined, and we use preprocessor macro guards like I mentioned in my first comment, we just get the extra function overload of `abs(__half)`:
```
template <typename T>
RAFT_INLINE_FUNCTION auto abs(T x)
  -> std::enable_if_t<std::is_same_v<float, T> || std::is_same_v<double, T> ||
                        std::is_same_v<int, T> || std::is_same_v<long int, T> ||
                        std::is_same_v<long long int, T>,
                      T>
{
#ifdef __CUDA_ARCH__
  return ::abs(x);
#else
  return std::abs(x);
#endif
}

template <typename T>
constexpr RAFT_INLINE_FUNCTION auto abs(T x)
  -> std::enable_if_t<!std::is_same_v<float, T> && !std::is_same_v<double, T> &&
                        !std::is_same_v<__half, T> && !std::is_same_v<nv_bfloat16, T> &&
                        !std::is_same_v<int, T> && !std::is_same_v<long int, T> &&
                        !std::is_same_v<long long int, T>,
                      T>
{
  if constexpr (std::is_unsigned_v<T>) {
    return x;
  } else {
    return x < T{0} ? -x : x;
  }
}

template <typename T>
RAFT_DEVICE_INLINE_FUNCTION typename std::enable_if_t<std::is_same_v<T, __half>, __half> abs(T x)
{
  return ::__habs(x);
}
```

cpp/include/raft/core/math.hpp
legrosbuffle: Done in the host function.

cpp/include/raft/core/math.hpp
legrosbuffle: > Okay, I think I understood the crux of the problem after going through your issue and PR again. We need some host-side overload of `__half` present no matter what for clang, and we can't hide that function completely because clang will aggressively attempt to instantiate template functions.

Exactly.

> That said, I do believe `_RAFT_HAS_CUDA` is throwing us off a bit here. [...] I think `_RAFT_HAS_CUDA` may be a vestigial macro borne from the intention of being able to compile RAFT without a CUDA-enabled compiler, but that ship has probably sailed.

I see, `__half` & co are only defined for cuda compilers I guess. We should probably keep it in case we're breaking someone ?

> I am not quite happy with the proposed solution, because the overload for `abs(__half)` is a device function, so `__CUDA_ARCH__` should always be defined IMO, 

Yes, I think that makes sense.

> and adding an undefined else for this macro in a device-only function does not seem like the correct approach to me.
> 
> How about this?
> 
> ```
> #ifdef __CUDA_ARCH__
> #if (__CUDA_ARCH__ >= 530)
> template <typename T>
> RAFT_DEVICE_INLINE_FUNCTION typename std::enable_if_t<std::is_same_v<T, __half>, __half> abs(T x)
> {
>   return ::__habs(x);
> }
> #endif
> 
> template <typename T>
> __host__ typename std::enable_if_t<std::is_same_v<T, __half>, __half> abs(T x)
> {
>   return T{};
> }
> ```

Looks good, done. Thanks !

legrosbuffle: > Maybe we convert the functions to be __host__ __device__ always? This is essentially what you proposed as the solution originally, with the exception of converting the functions from only __device__ tagged to adding a __host__ tag as well. Sorry for the long discussion and the rabbit chase, it's a bit hard to make all the compilers happy.

That's fine, thanks for taking the time to look at those PRs !

Issue #2483, opened 2024-11-05
[FEA] Add largest eigenvalues computation to Lanczos


lowener:
Follow-up of the Lanczos PR.
https://github.com/rapidsai/raft/pull/2481#discussion_r1828183184

PR #2722, opened 2025-06-27
Fix docs lanczos solver

aamijar:
#2483 closed, remove `todo` docs


lowener (commented):
Update the `@brief` doxygen as well since it can be used not only for the smallest eigenpairs. And the name of the function should also reflect that. It is a more impactful API change but I think it should be done before this function get used everywhere in the codebase.

lowener (approved):
LGTM

PR #2733, opened 2025-07-03
[REVIEW] Fix static initialization order fiasco in `lanczos.cu`

legrosbuffle:
`rows<>`/`cols<>`/`vals<>` are variable templates that are not explicitly specialized. As such, for their initialization, the rule that applies is "unordered dynamic initialization". This means that their initialization is "indeterminately sequenced with respect to all other dynamic initialization".

In particular, it can happen that `rows<int>` is initialized after its use to initialize `inputsd`, which makes `lanczos_inputs.rows` empty, with breaks tests (`rows.size()-1` underflows and results in `n==-1`, leading to the test trying to allocate a large amount of memory. Same for `cols<>` and `vals<>`.

This situation actually happens when building with clang. Fix it by turning `rows`/`cols`/`vals` into a template function.


aamijar (approved):
Thanks for the fix @legrosbuffle!

legrosbuffle:
> Hi @legrosbuffle, can you make sure to copy the vals array exactly. There is a small difference in some of the elements which is causing the tests to fail. Thank you!

Weird, I don't think I changed anything, but I might be wrong. I've split the change in two commits: one that introduces the function so we can check that there are no diffs, and one format-only commit.

aamijar:
@divyegala Can you rerun CI?

divyegala (approved):

PR #2734, opened 2025-07-03
Add missing header

vitor1001:


cpp/tests/core/mdarray.cu
```diff
@@ -26,6 +26,7 @@
 #include <raft/core/resources.hpp>
 #include <raft/util/cuda_utils.cuh>
 #include <raft/util/cudart_utils.hpp>
+#include <raft/thirdparty/mdspan/include/experimental/mdspan>
```
divyegala, line 29: We should not include this header directly. Please include [raft/core/mdspan.hpp](https://github.com/rapidsai/raft/blob/branch-25.08/cpp/include/raft/core/mdspan.hpp)

vitor1001: Done!

aamijar:
@divyegala merge?

divyegala (approved):

PR #2757, opened 2025-07-23
Removing CUDA 11 from docs and code

cjnolet:
Closes https://github.com/rapidsai/raft/issues/2741. 

Creating a new PR w/ verified commits.


jakirkham (commented):
Thanks Corey! 🙏

Think this is looking pretty good

Noticed one small thing and commented on that below. Nothing else stuck out

cpp/include/raft/core/cusparse_macros.hpp
```diff
@@ -23,18 +23,6 @@
...
-//(1.) CUDA_VER_10_1_UP aggregates all the CUDA version selection logic;
-//(2.) to enforce a lower version,
-//
-//`#define CUDA_ENFORCE_LOWER
-// #include <raft/sparse/detail/cusparse_wrappers.h>`
-//
-// (i.e., before including this header)
-//
-#define CUDA_VER_10_1_UP (CUDART_VERSION >= 10010)
-#define CUDA_VER_12_4_UP (CUDART_VERSION >= 12040)
```
jakirkham, line 36: Missed this before. Though noticed on CI. Should we keep the `CUDA_VER_12_4_UP` checks given we support CUDA 12.0-12.9?

cjnolet: I need to redo the removal of the checks anyways, but many of them have been there since CUDA 9 (and lots of files are about to be removed from RAFT) so I'm less concerned with those. Given code freeze is looming, let me get the docs and direct code mentions cleaned up at least.

docs/source/build.md
```diff
@@ -57,12 +52,6 @@ mamba install -c rapidsai -c conda-forge -c nvidia libraft-headers cuda-version=
 
 `pylibraft` and `raft-dask` both have packages that can be [installed through pip](https://rapids.ai/pip.html#install).
 
-For CUDA 11 packages:
-```bash
-pip install pylibraft-cu11 --extra-index-url=https://pypi.nvidia.com
-pip install raft-dask-cu11 --extra-index-url=https://pypi.nvidia.com
-```
-
 And CUDA 12 packages:
```
benfred, line 55: nit: the 'And' in this line is dangling - maybe we should just delete this line

benfred (approved):

docs/source/build.md
cjnolet: Let me fix that in 25.10.

PR #2761, opened 2025-07-25
Fix some invalid nodiscard warnings

vyasr:
This is the same issue as https://github.com/rapidsai/cuml/pull/7048 where the compiler in some environments is reporting new nodiscard warnings. In the case of raft the warnings are entirely erroneous, and the root cause is likely the same as with cuml (something to do with gtest I would guess). I don't think it merits much further investigation though since simply using a more tightly scoped variable is enough to make the compiler happy here.


vyasr:
Problems seem to have gone away on updated builds. Closing for now, can reopen if needed.

PR #2774, opened 2025-08-05
Update cutlass to a version that supports CUDA 13

robertmaynard:
Update the cutlass version we use to one that supports CUDA 13


bdice:
There are some errors in devcontainers like this (I removed some whitespace):

```
CMake Error in pylibraft/common/CMakeLists.txt:
    Imported target "raft::raft" includes non-existent path
      "/home/coder/.conda/envs/rapids/targets/x86_64-linux/include/cccl"
    in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:
    * The path was deleted, renamed, or moved to another location.
    * An install or uninstall procedure did not complete successfully.
    * The installation package was faulty and references files it does not
    provide
```

PR #2783, opened 2025-08-18
Updates to add CUDA 13 devcontainers

robertmaynard:
Updates required to get CUDA 13 devcontainers to be built in CI


robertmaynard:
Closing as https://github.com/rapidsai/raft/pull/2787/ is better

PR #2787, opened 2025-08-21
Build and test with CUDA 13.0.0

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/208

* uses CUDA 13.0.0 to build and test
* adds CUDA 13 devcontainers
* moves some dependency pins:
  - `cuda-python`: `>=12.9.2` (CUDA 12), `>=13.0.1` (CUDA 13)
  - `cupy`: `>=13.6.0`

Contributes to https://github.com/rapidsai/build-planning/issues/68

* updates to CUDA 13 dependencies in fallback entries in `dependencies.yaml` matrices (i.e., the ones that get written to `pyproject.toml` in source control)

## Notes for Reviewers

This switches GitHub Actions workflows to the `cuda13.0` branch from here: https://github.com/rapidsai/shared-workflows/pull/413

A future round of PRs will revert that back to `branch-25.10`, once all of RAPIDS supports CUDA 13.

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,9 +245,13 @@ mamba install -c rapidsai -c conda-forge -c nvidia raft-dask pylibraft cuda-vers
 Note that the above commands will also install `libraft-headers` and `libraft`.
 
 You can also install the conda packages individually using the `mamba` command above. For example, if you'd like to install RAFT's headers and pre-compiled shared library to use in your project:
+
 ```bash
-# for CUDA 12
+# CUDA 13
 mamba install -c rapidsai -c conda-forge -c nvidia libraft libraft-headers cuda-version=13.0
+
+# CUDA 12
+mamba install -c rapidsai -c conda-forge -c nvidia libraft libraft-headers cuda-version=12.9
 ```
 
 ### Installing Python through Pip


README.md
```diff
@@ -243,13 +247,19 @@ Note that the above commands will also install `libraft-headers` and `libraft`.
 You can also install the conda packages individually using the `mamba` command above. For example, if you'd like to install RAFT's headers and pre-compiled shared library to use in your project:
 ```bash
 # for CUDA 12
-mamba install -c rapidsai -c conda-forge -c nvidia libraft libraft-headers cuda-version=12.9
+mamba install -c rapidsai -c conda-forge -c nvidia libraft libraft-headers cuda-version=13.0
```
KyleFromNVIDIA, line 250: The comment doesn't match the command. Also, should both 12 and 13 be here?

conda/environments/all_cuda-129_arch-aarch64.yaml
```diff
@@ -14,9 +14,9 @@ dependencies:
 - cuda-nvcc
 - cuda-nvtx-dev
 - cuda-profiler-api
-- cuda-python>=12.6.2,<13.0a0
+- cuda-python>=12.9.2,<13.0a0
 - cuda-version=12.9
-- cupy>=12.0.0
+- cupy>=13.6.0
```
KyleFromNVIDIA, line 19: Note for any other reviewers: the cupy version is separate from the CUDA version. There is a cupy-cu12x package, version 13.6.0, that's being used here. So this line is correct.

KyleFromNVIDIA (changes requested):

conda/environments/all_cuda-129_arch-aarch64.yaml
jameslamb: ha yep, we just happen to be in one of those periods where the major versions sort of look like they line up

KyleFromNVIDIA (approved):

PR #2790, opened 2025-08-26
Update rapids-dependency-file-generator

KyleFromNVIDIA:
This PR updates the rapids-dependency-file-generator hook.


jameslamb (changes requested):
This needs the changes to `args:`

PR #2791, opened 2025-08-26
Remove UCX-Py

pentschev:
UCX-Py is deprecated and being archived. This change removes all UCX-Py tests, leaving only UCXX tests, and all mentions in favor of UCXX. See https://github.com/rapidsai/ucx-py/issues/1160 for details. There is some more code that can be removed from Cython but that will be better addressed in a follow-up PR as described in https://github.com/rapidsai/raft/issues/2792.


cjnolet (approved):

pentschev:
> @pentschev The devcontainers failure looks like a Github / CI issue? https://github.com/rapidsai/raft/actions/runs/17408304998/job/49418632293?pr=2791

Yes, it is looking like that to me as well. We'll need to rerun but I don't have permission to do so, would you mind triggering a rerun and then `/merge` (which I also don't have permissions for)? I think we need to wait for other tests to complete before retriggering though.

jameslamb:
Ah yeah, the issue is further up in the log:

```text
CMake Error at build/conda/cuda-13.0/release/CMakeFiles/fc-stamp/rapids-cmake/download-rapids-cmake.cmake:163 (message):
  Each download failed!

    error: downloading 'https://github.com/rapidsai/rapids-cmake/archive/refs/heads/branch-25.10.zip' failed
          status_code: 22
          status_string: "HTTP response code said error"
          log:
          --- LOG BEGIN ---
          Host github.com:443 was resolved.
```

([build link](https://github.com/rapidsai/raft/actions/runs/17408304998/job/49418632293?pr=2791#step:11:8740))

This looks like it was a network problem that was resolved by a re-run.

PR #2793, opened 2025-08-28
Fix index types for coo kernels

aamijar:
Resolves https://github.com/rapidsai/raft/issues/2794

There are some kernels where the index types from the public apis are not propagated correctly. Here are the ones that need fixes:
`coo_degree_nz_kernel`
`coo_symmetrize_kernel`
`coo_remove_scalar_kernel`

The api for `coo_degree_scalar` also needs an additional template type.


cjnolet (approved):

cjnolet:
@aamijar can you please double check to verify that this PR is actually non-breaking? Just wantt o make sure we don't surprise any downstream projects when we merge it.

PR #2802, opened 2025-09-09
WIP: update dependencies: use cuda-toolkit wheels

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/213

* switches uses of `nvidia-cuda-*-cu{12,13}` wheels to https://pypi.org/project/cuda-toolkit/

### Benefits of these changes

Using the `cuda-toolkit` metapackage means we don't need to maintain our own pins for mutually compatible versions of CTK sub-components (which each follow their own different versioning conventions).


python/libraft/CMakeLists.txt
```diff
@@ -45,12 +45,12 @@ set(RAFT_COMPILE_LIBRARY ON)
...
-           "$ORIGIN/../../nvidia/cusparse/lib"
-           "$ORIGIN/../../nvidia/nvjitlink/lib"
+set(rpaths
+    "$ORIGIN/../../nvidia/cublas/lib"
+    "$ORIGIN/../../nvidia/curand/lib"
+    "$ORIGIN/../../nvidia/cusolver/lib"
+    "$ORIGIN/../../nvidia/cusparse/lib"
+    "$ORIGIN/../../nvidia/nvjitlink/lib"
+    # starting with CTK 13 wheels, all libraries are grouped in this 'nvidia/cu13/lib' directory
+    "$ORIGIN/../../nvidia/cu13/lib"
```
jameslamb, line 55: I'd initially made this list of RPATHs conditional on detected CUDA version (from `nvcc`), like https://github.com/rapidsai/cuml/pull/7212, because I forgot this project intentionally does not set `CUDA` in its languages:

```text
  CMake Error at CMakeLists.txt:48 (if):
    if given arguments:

      "VERSION_GREATER_EQUAL" "13.0"

    Unknown arguments specified
```

([wheel-build-libraft link](https://github.com/rapidsai/raft/actions/runs/17596457720/job/49989885684?pr=2802#step:12:244))

RAFT selectively enables the CUDA language later:

https://github.com/rapidsai/raft/blob/720f60b38510514c19c4e3d37a3b66ef82d6da05/cpp/CMakeLists.txt#L504-L512

That's fine, I think it should be ok to have these RPATHs co-mingled like this, as `cuopt` did in https://github.com/NVIDIA/cuopt/pull/379

python/libraft/CMakeLists.txt
```diff
@@ -45,12 +45,12 @@ set(RAFT_COMPILE_LIBRARY ON)
...
 # assumes libraft.so is installed 2 levels deep, e.g. site-packages/libraft/lib64/libraft.so
-set_property(
-  TARGET raft_lib
-  PROPERTY INSTALL_RPATH
-           "$ORIGIN/../../nvidia/cublas/lib"
-           "$ORIGIN/../../nvidia/curand/lib"
-           "$ORIGIN/../../nvidia/cusolver/lib"
-           "$ORIGIN/../../nvidia/cusparse/lib"
-           "$ORIGIN/../../nvidia/nvjitlink/lib"
+set(rpaths
```
KyleFromNVIDIA, line 48: Where does the `rpaths` variable actually get used? Before, this was calling `set_property()`.

jameslamb: Thank you for catching that... you're right, this doesn't make sense.

Switched it back to `set_property()` in https://github.com/rapidsai/raft/pull/2802/commits/e20bef79e484e28dd138bcfc5e6d19b3bc962ee7

PR #2804, opened 2025-09-16
Fix UCX-Py mention to UCXX in docstring

pentschev:
UCX-Py is deprecated and is being archived, all other functional changes were already updated to UCXX, but there's still one reference to UCX-Py in a docstring that needs updating.


cjnolet (approved):

pentschev:
Thanks Corey!

PR #2836, opened 2025-10-14
Use CCCL's mdspan implementation

### LINKED ISSUE #1616: [FEA] Use mdspan from CCCL

bdice:
This PR adopts CCCL's `mdspan` implementation and deletes the vendored implementation.

Closes #1616.

We also introduce a breaking change by removing the default type for template parameter `ExtentType = std::uint32_t` in all mdarray and mdspan related types, aliases, and functions.

### DIFF
--- a/cpp/include/raft/neighbors/ivf_pq_types.hpp
+++ b/cpp/include/raft/neighbors/ivf_pq_types.hpp
@@ -395,7 +395,7 @@ struct index : ann::index {
   }
 
   using pq_centers_extents =
-    std::experimental::extents<uint32_t, dynamic_extent, dynamic_extent, dynamic_extent>;
+    raft::extents<uint32_t, dynamic_extent, dynamic_extent, dynamic_extent>;
   /**
    * PQ cluster centers
    *


cpp/tests/sparse/preprocess.cu
```diff
@@ -29,9 +29,9 @@ namespace sparse {
 
 template <typename T2, typename T1>
 void get_clean_coo(raft::resources& handle,
-                   raft::device_vector_view<T1> rows,
-                   raft::device_vector_view<T1> columns,
-                   raft::device_vector_view<T2> values,
+                   raft::device_vector_view<T1, int64_t> rows,
```
bdice: Need to check if this change is correct.

cpp/tests/util/preprocess_utils.cu
```diff
@@ -36,10 +36,10 @@ struct check_zeroes {
   }
 };
 
-template <typename T1, typename T2>
+template <typename T1, typename T2, typename DenseVectorView, typename ResultVectorView>
```
bdice: Need to check if this change is correct.

cpp/tests/core/mdarray.cu
```diff
@@ -57,15 +56,15 @@ void test_mdspan()
   auto stream = rmm::cuda_stream_default;
   rmm::device_uvector<float> a{16ul, stream};
   thrust::sequence(rmm::exec_policy(stream), a.begin(), a.end());
-  stdex::mdspan<float, stdex::extents<int, raft::dynamic_extent, raft::dynamic_extent>> span{
-    a.data(), 4, 4};
+  cuda::std::mdspan<float, cuda::std::extents<int, raft::dynamic_extent, raft::dynamic_extent>>
```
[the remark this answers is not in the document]
mhoemmen: The preferred option, though, would be to use CTAD.

```c++
  cuda::std::mdspan{a.data(), cuda::std::dims<2, int>{4, 4}}
```

cpp/tests/core/mdarray.cu
```diff
@@ -268,7 +267,7 @@ void test_mdarray_copy_move(ThrustPolicy exec, PolicyFn make_policy)
 
 TEST(MDArray, CopyMove)
 {
-  using matrix_extent = stdex::extents<size_t, dynamic_extent, dynamic_extent>;
+  using matrix_extent = cuda::std::extents<size_t, dynamic_extent, dynamic_extent>;
```
mhoemmen:
```suggestion
  using matrix_extent = cuda::std::dims<2>;
```

cpp/include/raft/core/mdspan.hpp
```diff
@@ -58,12 +64,119 @@ struct alignment {
 
 }  // namespace detail
 
+// Padded layout shims compatible with CCCL mdspan mappings
+template <size_t PaddingStride>
+struct layout_left_padded_impl {
```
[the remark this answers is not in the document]
divyegala: Yes, that's a good idea.

cpp/tests/core/mdarray.cu
```diff
@@ -515,7 +515,7 @@ void test_mdarray_padding()
      * padded device array
      */
     using padded_layout_row_major =
-      stdex::layout_right_padded<detail::padding<float, alignment_bytes>::value>;
+      raft::layout_right_padded_impl<detail::padding<float, alignment_bytes>::value>;
```
divyegala: Should this use `raft::layout_right_padded<float>`?

cpp/tests/sparse/preprocess.cu
divyegala: It is explicit but unnecessary.

cpp/include/raft/core/mdspan_types.hpp
```diff
@@ -16,39 +16,39 @@
...
 
-#include <raft/thirdparty/mdspan/include/experimental/mdspan>
+#include <cuda/std/mdspan>
 
 namespace raft {
 
-using std::experimental::dynamic_extent;
-using std::experimental::extents;
+using cuda::std::dynamic_extent;
+using cuda::std::extents;
```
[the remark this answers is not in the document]
divyegala: Actually, I'm re-reading Mark's comment and I think what he is saying is a good idea. We using `raft::` namespace outside of this header and bring `cuda::std::` symbols in `raft::` namespace in this header.

PR #2845, opened 2025-10-22
Use SPDX for all copyright headers

KyleFromNVIDIA:
Issue: https://github.com/rapidsai/build-infra/issues/297

### DIFF
--- a/cpp/include/raft/core/device_container_policy.hpp
+++ b/cpp/include/raft/core/device_container_policy.hpp
@@ -84,15 +84,10 @@ class device_uvector {
   using const_iterator = const_pointer;
 
  public:
-  ~device_uvector()                         = default;
-  device_uvector(device_uvector&&) noexcept = default;
-  device_uvector(device_uvector const& that) : data_{that.data_, that.data_.stream()} {}
-
-  auto operator=(device_uvector<T> const& that) -> device_uvector<T>&
-  {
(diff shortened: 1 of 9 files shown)


robertmaynard (changes requested):
Needs to be Apache+MIT licensed  ( Copyright hold of Leon Merten Lohse / numpy as well ):
```
cpp/include/raft/core/detail/mdspan_numpy_serializer.hpp
```

robertmaynard (changes requested):
I missed a file on my first review:

```
cpp/include/raft/util/integer_utils.hpp -> BlazingDB, Inc
```

Otherwise it looks great, thanks Kyle

cmake/RAPIDS.cmake
```diff
@@ -1,15 +1,8 @@
...
-#
-# Unless required by applicable law or agreed to in writing, software distributed under the License
-# is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
-# or implied. See the License for the specific language governing permissions and limitations under
-# the License.
+# cmake-format: off
+# SPDX-FileCopyrightText: Copyright (c) 2021-2025, NVIDIA CORPORATION.
+# SPDX-License-Identifier: Apache-2.0
+# cmake-format: on
 # =============================================================================
```
jakirkham: Was it intentional to drop the long form? Noticing this occurred in a few files

PR #2852, opened 2025-10-30
Fix for STD comm waitall function

viclafargue:
Closes [#7341](https://github.com/rapidsai/cuml/issues/7341)

When UCXX progress thread was active, the waitall() implementation used CallbackNotifier with registerGenericPre/Post which would block indefinitely waiting for callbacks that the progress thread couldn't service in this context. The solution is to simply yield the CPU with std::this_thread::yield() and let the background progress thread complete requests while the main thread polls isCompleted().

### DIFF
--- a/cpp/include/raft/comms/detail/std_comms.hpp
+++ b/cpp/include/raft/comms/detail/std_comms.hpp
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2020-2024, NVIDIA CORPORATION.
+ * SPDX-FileCopyrightText: Copyright (c) 2020-2025, NVIDIA CORPORATION.
  * SPDX-License-Identifier: Apache-2.0
  */
 


viclafargue:
> Assuming the problem is only in the callback notifier [...] If so, I'd be interested in that to debug that further.

Thanks @pentschev. Yes, the issue might actually be located in the callback notifier. Removing the interaction with it solves the deadlock (even though this prevents us from checking that the work is actually progressing).

> Do you have a consistent reproducer for the issue @viclafargue ?

```
import sys
import numpy as np


if __name__ == '__main__':
    np.random.seed(1989)
    X_train = np.random.rand(100_000, 64)

    import dask
    
    from dask_cuda import LocalCUDACluster
    from dask.distributed import Client
    
    cluster = LocalCUDACluster()
    client = Client(cluster)
    
    from cuml.dask.cluster import DBSCAN
    dbscan = DBSCAN(metric='cosine', min_samples=4, eps=0.1)
    
    clusters_gpu = dbscan.fit_predict(X_train, 'int32')

    print(f"\nmin: {clusters_gpu.min()} max: {clusters_gpu.max()}\n")

    cluster.close()
    client.close()

    sys.exit(0)
```

Causes a hang in the [waitall call in the final tree reduction of Dask DBSCAN](https://github.com/rapidsai/cuml/blob/f986a4083cca152999685521f0b203692ec91473/cpp/src/dbscan/mergelabels/tree_reduction.cuh#L61) when running with 8 GPUs.

PR #2860, opened 2025-11-04
Restore isort for Cython files

KyleFromNVIDIA:
Ruff does not yet support Cython, so restore isort only for Cython.

Issue: https://github.com/rapidsai/build-planning/issues/130

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,6 +13,15 @@ repos:
               - id: ruff-check
                 args: [--fix]
               - id: ruff-format
+      - repo: https://github.com/PyCQA/isort
+        rev: 5.12.0
+        hooks:
+              - id: isort
+                # Use the config file specific to each subproject so that each
+                # project can specify its own first/third-party packages.
+                args: ["--config-root=python/", "--resolve-all-configs"]
+                files: python/.*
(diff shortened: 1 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -13,6 +13,15 @@ repos:
               - id: ruff-check
                 args: [--fix]
               - id: ruff-format
+      - repo: https://github.com/PyCQA/isort
+        rev: 5.12.0
```
bdice: Why this version? Latest is 7.0.0. We used a different version in the RMM PR.

python/libraft/pyproject.toml
```diff
@@ -45,6 +45,41 @@ known-first-party = [
...
+skip = [
+    "thirdparty",
+    ".eggs",
+    ".git",
+    ".hg",
+    ".mypy_cache",
+    ".tox",
+    ".venv",
+    "_build",
+    "buck-out",
```
bdice: Some of this stuff might be extremely dated, so copy-paste might not be ideal. I don't know if `buck-out` is relevant. Similarly, we don't use Mercurial (`.hg`). `.eggs` and `.tox` are probably not needed either.

Issue #2830, opened 2025-10-06
Support COO input in `diagonal`, `set_diagonal`, `scale_by_diagonal_symmetric`


aamijar:
There are some utilities in the `raft::sparse::matrix` namespace that are used during **laplacian normalization** which need support for COO inputs.

PR #2861, opened 2025-11-05
COO support in sparse matrix utilities

aamijar:
Resolves https://github.com/rapidsai/raft/issues/2830

This PR adds COO input support to `diagonal`, `set_diagonal`, and `scale_by_diagonal_symmetric` in the `raft::sparse::matrix` namespace.

This PR also adds proper template type support for `IndptrType`, `IndexType` and `NNZType` for the existing CSR input functions.

This PR also uses the `raft::linalg::map_offset` instead of `thrust::for_each`.

### DIFF
--- a/cpp/cmake/modules/ConfigureCUDA.cmake
+++ b/cpp/cmake/modules/ConfigureCUDA.cmake
@@ -16,12 +16,9 @@ endif()
 # clang)
 if(CMAKE_COMPILER_IS_GNUCXX)
   list(APPEND RAFT_CXX_FLAGS -Wall -Werror -Wno-unknown-pragmas -Wno-error=deprecated-declarations)
-  list(APPEND RAFT_CUDA_FLAGS -Xcompiler=-Wall,-Werror,-Wno-error=deprecated-declarations)
-
-  # set warnings as errors
-  if(CMAKE_CUDA_COMPILER_VERSION VERSION_GREATER_EQUAL 11.2.0)
-    list(APPEND RAFT_CUDA_FLAGS -Werror=all-warnings)
-  endif()
+  list(APPEND RAFT_CUDA_FLAGS -Xcompiler=-Wall,-Werror,-Wno-error=deprecated-declarations
+       -Werror=all-warnings
+  )
 
 endif()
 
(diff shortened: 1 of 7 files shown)


cpp/include/raft/sparse/matrix/detail/diagonal.cuh
```diff
@@ -151,4 +154,126 @@ void set_diagonal(raft::resources const& res,
...
+  auto values   = coo_matrix_view.get_elements().data();
+  auto rows     = structure.get_rows().data();
+  auto cols     = structure.get_cols().data();
+  auto diag_ptr = diagonal.data_handle();
+
+  raft::matrix::fill(res, diagonal, T(0));
+
+  auto policy = raft::resource::get_thrust_policy(res);
+
+  thrust::for_each(policy,
```
tarang-jain: Wherever possible, please avoid `thrust::for_each` and use `raft::linalg::unary_op` / `map_offset` or similar.

tarang-jain (commented):
Just a few smaller comments -- raft::linalg is typically smaller in binary size in comparison to thrust. Lets use raft wherever we can. Things like for_each are easily replaceable by their raft::linalg counterparts.

cpp/include/raft/sparse/matrix/detail/diagonal.cuh
```diff
@@ -151,4 +154,126 @@ void set_diagonal(raft::resources const& res,
...
+                   thrust::counting_iterator<NNZType>(0),
+                   thrust::counting_iterator<NNZType>(nnz),
+                   [values, rows, cols, diag_ptr] __device__(NNZType idx) {
+                     auto row    = rows[idx];
+                     auto col    = cols[idx];
+                     T row_scale = 1.0f / diag_ptr[row];  // Scale factor for this row
+                     T col_scale = 1.0f / diag_ptr[col];  // Scale factor for the column
```
tarang-jain: Dont we need safeguards here to check if the denominator is zero?

cpp/include/raft/sparse/matrix/detail/diagonal.cuh
```diff
@@ -5,10 +5,12 @@
 
 #pragma once
 
+#include <raft/core/device_coo_matrix.hpp>
 #include <raft/core/device_csr_matrix.hpp>
 #include <raft/core/device_mdspan.hpp>
 #include <raft/core/resource/cuda_stream.hpp>
 #include <raft/core/resource/thrust_policy.hpp>
```
tarang-jain: we dont need the thrust policy, right?

PR #2866, opened 2025-11-11
Set memory pool through RMM

viclafargue:
This PR aims at updating the `device_resources_snmg` utility to make sure that the method in charge of setting up the RMM pools on the GPUs effectively work in multi-GPU context. This method used to set a pool workspace on the device resource to no effect. We instead now set the per device memory resource of each GPU to a pool through RMM.

Benchmark of search throughput for 100 rows queries on 2 GPUs CAGRA indexes built with NN Descent :
benchmark


achirkin (dismissed):
Hi @viclafargue , could you please update the description of the PR to let the reviewers know what is the problem it addresses?
I suspect, this is about perf bottlenecks related to locking the cuda context in the multi-gpu setup. If so, please attach the benchmarks. I'm also concerned a little about the lifetime of the objects allocated with the per-device pools. Could there be a situation when an object outlives its memory pool and segfaults the program?

tarang-jain (approved):
LGTM. I like the idea of using
`std::vector<std::unique_ptr<rmm::mr::pool_memory_resource<rmm::mr::device_memory_resource>>>`
This gives a clear message that the snmg resource owns these new RMM resources.

Issue #2872, opened 2025-11-14
Remove legacy lanczos solver


aamijar:
We should remove the legacy lanczos solver as that is not being used anymore.

Issue #2737, opened 2025-07-09
[FEA] Remove all ANN, KNN, and distance related things from RAFT


cjnolet:
Initially, we will remove everything except for the `eps neighborhood` APIs (and dependent kernels).

PR #2813, opened 2025-09-22
Remove Deprecated API

jnke2016:
This PR removes deprecated API from `raft` `neighbors`, `spacial`, `sparse/neighbors`, and `cluster`

PR #2885, opened 2025-12-03
Remove `neighbors/`, `cluster/`, `distance/`, `spatial/`, `sparse/neighbors/` apis

aamijar:
Supersedes #2813 and #2878
Resolves https://github.com/rapidsai/raft/issues/2737 and Resolves https://github.com/rapidsai/raft/issues/2872

Marking as **breaking** to get more eyes on this and mitigate risk.
This PR should not break downstream libraries as long as we merge the updates to them first: https://github.com/rapidsai/cuvs/pull/1610, https://github.com/rapidsai/cuml/pull/7561. 
I've found a usage of breaking api in FAISS here:
https://github.com/facebookresearch/faiss/blob/1721ebff6de6ed5a8481302123479be9d85059a2/faiss/gpu/GpuDistance.cu#L46.
https://github.com/facebookresearch/faiss/blob/5b19fca3f057b837ac898af52a8eb801c4744892/faiss/gpu/impl/CuvsFlatIndex.cu#L34

What does this PR do?
1. Removes `cluster/`, `distance/`, `neighbors/` (except `detail/faiss_select/`), `sparse/neighbors/`, `spatial/`
2. Removes unused includes that will be deprecated such as `#include <raft/distance/distance.cuh>`, `#include <raft/spatial/knn/knn.cuh>`, etc.
3. Removes legacy lanczos solver (`linalg/lanczos`, `sparse/linalg/lanczos` old functions in `sparse/solver/lanczos`) and removes legacy spectral apis (`spectral/ ` except modularity_maximization and partition which are metrics used by cugraph)
4. Removes corresponding gtests, raft_runtime, and bench files.

### DIFF
--- a/.github/ops-bot.yaml
+++ b/.github/ops-bot.yaml
@@ -7,3 +7,4 @@ label_checker: true
 release_drafter: true
 recently_updated: true
 forward_merger: true
+merge_barriers: true
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,42 @@
+# raft 25.12.00 (10 Dec 2025)
+
+### 🚨 Breaking Changes
+* More consistent container policies & host memory resource by @achirkin in https://github.com/rapidsai/raft/pull/2835
+* Require CUDA 12.2+ by @jakirkham in https://github.com/rapidsai/raft/pull/2850
+### 🐛 Bug Fixes
+* Correct tagging in the `irecv` function of the STD communicator by @viclafargue in https://github.com/rapidsai/raft/pull/2829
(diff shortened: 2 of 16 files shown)


cjnolet:
@aamijar do we still have any parts of raft using the `faiss_select` code? I thought only agglomerative clustering was using that code, but I could be wrong. If that is the case, however, I'd much prefer to move the faiss_select code to cuvs into the `src/` directory for agglomerative clustering than to keep it in RAFT. We also shouldn't be referencing internal `detail` namespaces from outside cuvs (though I do admit there are still a couple places we do this).

bdice (approved):
Thanks for the hard work on this. Packaging + CMake looks good to me.

PR #2888, opened 2025-12-09
Fix is_exhaustive, no longer constexpr

bdice:
This PR fixes an error observed with CCCL 3.2, where `is_exhaustive` is no longer `constexpr` for mdspans with dynamic extents. We must use `EXPECT_TRUE` to evaluate this at runtime, rather than `static_assert` at compile time.


vyasr:
Same comment as https://github.com/rapidsai/raft/pull/2909#issuecomment-3706511507

Issue #2875, opened 2025-11-19
Update public header for lanczos solver to support `nnz_t`


aamijar:
Looks like we need to make a small change to the header to add a new template parameter as a follow up to https://github.com/rapidsai/raft/issues/2832

PR #2891, opened 2025-12-10
Laplacian Kernel for COO inputs

aamijar:
Resolves https://github.com/rapidsai/raft/issues/2831 and Resolves https://github.com/rapidsai/raft/issues/2875.

The laplacian kernel implementation needs to be reimagined to work with COO input since the previous CSR version isn't compatible.
The implementation in this PR also crucially addresses the issue of allocating the correct output size.
To calculate the laplacian `L = D - A` for a COO input the following steps are used:
1. Count the number of extra nnz elements required to go from `A` to `L` by counting the number of nnz elements missing from diagonal of `A`. 
2. Keep track of which diagonal nnz elements are missing from `A` with binary array `marked_diagonal`. Compute exclusive_scan in an array `scan_diagonal`. This is used to place the extra missing elements in `L`.
3. Allocate and initialize the output `L`. Copy the indexes and values from `A` into `L`. Initialize the extra nnz elements required for the diagonal if any at the end.
4. Calculate `D` which is an array of degrees for each row of `A`
5. Edit `L` in place so that diagonal elements are subtracted from D, and off-diagonal elements are negated.


cpp/include/raft/sparse/linalg/laplacian.cuh
```diff
@@ -49,7 +60,41 @@ auto laplacian_normalized(
...
+/**
+ * @brief Given a COO adjacency matrix, return the normalized graph Laplacian
+ *
+ * Return the normalized Laplacian matrix, which is defined as D^(-1/2) * L * D^(-1/2),
+ * where D is the diagonal degree matrix and L is the graph Laplacian.
+ * Also returns the scaled diagonal degree matrix.
+ *
+ *
+ * @tparam ElementType The data type of the matrix elements
+ * @tparam IndptrType The data type of the row pointers
```
dantegd:
```suggestion
 * @tparam RowType The data type of the row pointers
```

cpp/include/raft/sparse/linalg/laplacian.cuh
```diff
@@ -49,7 +60,41 @@ auto laplacian_normalized(
...
+ * @tparam IndicesType The data type of the column indices
```
dantegd:
```suggestion
 * @tparam ColType The data type of the column indices
```

cpp/include/raft/sparse/linalg/detail/laplacian.cuh
```diff
@@ -108,6 +112,114 @@ auto compute_graph_laplacian(
...
+  auto degrees = raft::make_device_vector<ElementType, RowType>(res, dim);
+  raft::matrix::fill(res, degrees.view(), ElementType(0));
+  auto degrees_ptr = degrees.data_handle();
+
+  // D
+  auto result_nnz = result.structure_view().get_nnz();
+  raft::linalg::map_offset(res,
+                           raft::make_device_vector_view(result_values_ptr, result_nnz),
+                           [result_rows_ptr, result_values_ptr, degrees_ptr] __device__(auto idx) {
+                             atomicAdd(&degrees_ptr[result_rows_ptr[idx]], result_values_ptr[idx]);
```
dantegd: I think this could be a bit of a performance bottleneck I wonder if you've profiled/benchmarked it?

My though comes from the fact that if you have a  row with many connections (common in power-law graphs for example), many threads will contend on the same degrees_ptr[row] location simultaneously. Probably not a huge deal, but if it's a bottleneck something we could try (maybe a follow up) is either of:

- Segmented reduction: If COO is sorted by row, use thrust::reduce_by_key instead of atomics
- Histogram-style approaches: Use shared memory atomics first, then global atomics

dantegd (changes requested):
Code looks good to me, very good test quality as well. Just had mainly comments about typos and a docstring and a perf question.

PR #2894, opened 2025-12-11
Fix Lanczos Determinism

aamijar:
Resolves https://github.com/rapidsai/cuml/issues/7595

This PR switches to the deterministic cusparseSPMV implementation `CUSPARSE_SPMV_CSR_ALG2` or `CUSPARSE_SPMV_COO_ALG2` if the user sets a random seed.

Perf impact:
```
Performance: deterministic (seed=42) vs non-deterministic (seed=None)
================================================================================

=== n_samples = 10,000 ===
  Dataset      | Determ (ms)       | Non-det (ms)      | Diff   
  ------------ | ----------------- | ----------------- | -------
  moons        |    1756 +/-     3 |    1557 +/-   421 |  +12.8%
  circles      |     180 +/-     1 |     177 +/-     6 |   +1.7%
  blobs        |     324 +/-     1 |     283 +/-    19 |  +14.7%
  swiss_roll   |     109 +/-     0 |     110 +/-     8 |   -0.5%
  random       |      50 +/-     0 |      47 +/-     1 |   +6.1%

=== n_samples = 50,000 ===
  Dataset      | Determ (ms)       | Non-det (ms)      | Diff   
  ------------ | ----------------- | ----------------- | -------
  moons        |    2000 +/-     1 |    2071 +/-   104 |   -3.4%
  circles      |     267 +/-     1 |     273 +/-    28 |   -2.2%
  blobs        |     598 +/-     1 |     674 +/-   380 |  -11.3%
  swiss_roll   |     415 +/-     1 |     342 +/-    20 |  +21.4%
  random       |     110 +/-     0 |     151 +/-    67 |  -27.1%

=== n_samples = 100,000 ===
  Dataset      | Determ (ms)       | Non-det (ms)      | Diff   
  ------------ | ----------------- | ----------------- | -------
  moons        |    2929 +/-     1 |    3469 +/-  1555 |  -15.6%
  circles      |     301 +/-     1 |     305 +/-    34 |   -1.2%
  blobs        |     337 +/-     1 |    1432 +/-   519 |  -76.5%
  swiss_roll   |    1138 +/-     1 |     887 +/-   225 |  +28.3%
  random       |      88 +/-     0 |      88 +/-     7 |   +0.2%

=== n_samples = 1,000,000 ===
  Dataset      | Determ (ms)       | Non-det (ms)      | Diff   
  ------------ | ----------------- | ----------------- | -------
  moons        |    7828 +/-     3 |    7663 +/-   677 |   +2.2%
[...]

PR #2899, opened 2025-12-17
`eigsh` optional seed

aamijar:
Follow up to https://github.com/rapidsai/raft/pull/2894

This changes the python layer to pass in nullopt if the user does not select a seed.


divyegala (commented):
Tests?

PR #2917, opened 2026-01-08
Store any_resource in device_uvector_policy

bdice:
Part of https://github.com/rapidsai/rmm/issues/2011.

Depends on https://github.com/rapidsai/rmm/pull/2200.

Change `device_uvector_policy::mr_` from `device_async_resource_ref` to `cuda::mr::any_resource<cuda::mr::device_accessible>`.

This ensures the container policy owns its memory resource, preventing use-after-free if the original resource is destroyed while the policy still exists.


bdice:
Closing — superseded by #2996, which includes this change as part of the full RMM CCCL memory resource migration (rapidsai/rmm#2011).

PR #2961, opened 2026-02-20
Dry Run Protocol

achirkin:
The dry run protocol defines a mechanism to simulate the execution of algorithms to get a precise estimate of the memory requirements for a real execution with the same parameters.

```C++
#include <raft/util/dry_run_memory_resource.hpp>

raft::resources res;
// auto my_function(const raft::resources& res, my_args...);
auto stats = raft::util::dry_run_execute(res, my_function, my_args...);
// stats.device_global  – peak device memory (bytes)
```

This PR:
 - Introduces new infrastructure: `raft::util::dry_run_execute`, `raft::dry_run_resources`, and `resource::get_dry_run_flag` to let callers estimate peak memory usage of any RAFT algorithm without executing GPU work.
- Makes all public functions across all raft namespaces dry-run compliant: allocations are always visible to the tracker; CUDA work is skipped.
- Adds a small user guide (docs/source/dry_run_protocol.md)

**Note for reviewers**
The PR contains a lot of small tedious changes to cover all of raft library and the tests components.
Please start reading at `docs/source` updates to learn more about the topic and the principles guiding these changes.


cpp/include/raft/core/detail/copy.hpp
```diff
@@ -398,6 +399,10 @@ mdspan_copyable_t<DstType, SrcType> copy(resources const& res, DstType&& dst, Sr
     RAFT_EXPECTS(src.extent(i) == dst.extent(i), "Must copy between mdspans of the same shape");
   }
 
+  // Dry-run guard: raft::copy is a pure data-movement utility with no
+  // allocations that callers would need tracked.
+  if (resource::get_dry_run_flag(res)) { return; }
```
cjnolet: Oh, I didn't realize we were adding this as a new resource. This would make it hard to use the dry-run for pre-initializing resources.

achirkin: Yes, but that's fine! We can still push the initialized resources back to the original resources handle on destruction of the dry run resources

achirkin: NB: with https://github.com/rapidsai/raft/pull/3052 , resources initialized in dry run mode will be automatically shared back with the input resources.

achirkin:
@coderabbitai review

cpp/include/raft/linalg/reduce_rows_by_key.cuh
```diff
@@ -148,6 +149,7 @@ void reduce_rows_by_key(
   std::optional<raft::device_vector_view<const WeightType, IndexType>> d_weights = std::nullopt,
   bool reset_sums                                                                = true)
 {
+  if (resource::get_dry_run_flag(handle)) { return; }
   RAFT_EXPECTS(d_A.extent(0) == d_A.extent(0) && d_sums.extent(1) == n_unique_keys,
                "Output is not of size ncols * n_unique_keys");
```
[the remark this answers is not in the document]
achirkin: unrelated

cpp/include/raft/random/detail/rng_impl_deprecated.cuh
```diff
@@ -272,13 +274,37 @@ class RngImpl {
...
+    cub::DeviceRadixSort::SortPairs(workspace.data(),
+                                    workspace_size,
+                                    expWts.data(),
+                                    sortedWts.data(),
+                                    inIdxPtr,
+                                    outIdxPtr,
+                                    (int)len,
+                                    0,
+                                    sizeof(WeightsT) * 8,
+                                    stream);
```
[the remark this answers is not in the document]
achirkin: Keeping this as before for 1-1 correspondence with the state before the PR (sortPairs doesn't do this check).

cpp/include/raft/random/detail/rng_impl.cuh
```diff
@@ -301,13 +306,37 @@ void sampleWithoutReplacement(RngState& rng_state,
...
+  cub::DeviceRadixSort::SortPairs(workspace.data(),
+                                  workspace_size,
+                                  expWts.data(),
+                                  sortedWts.data(),
+                                  inIdxPtr,
+                                  outIdxPtr,
+                                  (int)len,
+                                  0,
+                                  sizeof(WeightsT) * 8,
+                                  stream);
```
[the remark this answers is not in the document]
achirkin: Keeping this as before for 1-1 correspondence with the state before the PR (sortPairs doesn't do this check).

cpp/include/raft/core/resource/cuda_stream.hpp
```diff
@@ -84,13 +85,18 @@ inline void set_cuda_stream(resources const& res, rmm::cuda_stream_view stream_v
  */
 inline void sync_stream(const resources& res, rmm::cuda_stream_view stream)
 {
+  if (raft::resource::get_dry_run_flag(res)) { return; }
```
huuanhhuyn, line 94: What is the motivation for guarding `sync_stream`? 

Your dry_run_protocol.md suggests the opposite that `sync_stream` is safe without guard.

docs/source/dry_run_protocol.md
```diff
@@ -0,0 +1,92 @@
...
+
+`dry_run_execute` swaps the memory resources, sets the flag, runs the callable, restores everything, and returns a `raft::memory_stats` snapshot of peak allocation usage.
+
+You can also construct `raft::dry_run_resources` directly for finer control (e.g. reading `get_bytes_current()` in addition to `get_bytes_peak()`).
+
+## Three Rules
+
+1. **Allocations must not be guarded.** Every `rmm::device_uvector`, `rmm::device_scalar`, `rmm::device_buffer`, `raft::make_(device|host|pinned|managed)_(mdarray|matrix|vector|scalar)` allocation must execute in both modes so the tracker sees it.
+
+2. **CUDA work must be guarded.** Kernel launches, Thrust algorithms, cuBLAS/cuSOLVER/cuSPARSE compute calls, `cudaMemcpyAsync`, `cudaMemsetAsync`, and `raft::interruptible::synchronize` must not run in dry-run mode.
```
huuanhhuyn: if all CUDA work is guarded / skipped in dry-run. How do we measure allocations whose allocation sizes are produced by a kernel?

If that kernel is dependent on earlier kernels and so on, dry-run must then execute all of them without guarding?

cpp/include/raft/core/resource/dry_run_flag.hpp
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * @brief Get the dry-run flag from a resources handle.
+ *
+ * @param res raft resources object
+ * @return true if dry-run mode is active
+ */
+inline auto get_dry_run_flag(resources const& res) -> bool
+{
+  if (!res.has_resource_factory(resource_type::DRY_RUN_FLAG)) {
+    res.add_resource_factory(std::make_shared<dry_run_flag_resource_factory>());
```
huuanhhuyn: How about just return False here? I think we don't need to create the factory at all in dry-run-free mode.

cpp/include/raft/core/device_container_policy.hpp
```diff
@@ -127,6 +127,29 @@ class device_uvector {
...
+   *
+   * Unlike resize(), this never copies old data.
+   * Thus, unlike in resize(), there's no point in time where the old and the new buffers are both
+   * alive, and the peak memory usage is lower.
+   *
+   * Unlike resize(), this deallocates the old buffer even if the new size is smaller.
+   * This ensures the memory is released promptly.
+   */
+  void reallocate(size_type size)
+  {
```
huuanhhuyn, line 141: Although your implementation avoids peak data (old + new) when the new size is larger than the current size, it has two issues:
- when the new size is the same -> it does NOT reallocate while its name is "reallocate"
- when the new size is smaller -> it DOES allocate while it could more efficiently shrink the current buffer (similar to what `resize()` already does)

I would suggest to do the current implementation only when the new size is larger. When the new size is less or equal, simply call `resize()` and clear the buffer.

I am not sure about clearing the buffer. It adds some overhead, but it sounds safer to me.

cpp/include/raft/core/device_mdarray.hpp
```diff
@@ -164,7 +165,7 @@ auto make_device_scalar(raft::resources const& handle, ElementType const& v)
   using policy_t = typename device_scalar<ElementType, IndexType>::container_policy_type;
   policy_t policy{};
   auto scalar = device_scalar<ElementType, IndexType>{handle, extents, policy};
-  scalar(0)   = v;
+  if (!resource::get_dry_run_flag(handle)) { scalar(0) = v; }
```
huuanhhuyn, line 170: I guess we don't need to guard this scalar operation (same for other resources). It is lightweight to run and keep the code a little bit simpler.

docs/source/dry_run_protocol.md
```diff
@@ -0,0 +1,92 @@
+# Dry Run Protocol
+
+The dry run protocol lets callers estimate an algorithm's memory footprint without executing it. When enabled, the runtime swaps memory resources for lightweight trackers that record every allocation and deallocation, producing peak-usage statistics at the end.
```
huuanhhuyn, line 3: I suppose we are not able to track non-raft allocations with dry-run. It is probably fine but we should be aware and document it somewhere.

Are we using third-party libraries which do allocations without RAFT somewhere? I am not aware about it.

huuanhhuyn (dismissed):
Thanks @achirkin for the heavy-lifting! I have went through this PR halfway.

cpp/include/raft/core/resource/cuda_stream.hpp
achirkin: The guide says it's safe to call this function, because it is dry-run compliant (takes the raft::resources as an argument) - hence it guards from any CUDA work inside.
`cudaStreamSynchronize` is cuda CUDA and it takes time, so we must guard it.

docs/source/dry_run_protocol.md
achirkin: The case you're describing falls under "allocation size depends on values/data". This is indeed generally a problem and we resort to providing safe upper bounds on allocations. Have a look at the sparse namespace - we have a lot cases like this.
In tests, this is covered by `DATA_DRIVEN` provenance: we compare that dry run produced stats are not smaller than the real produced stats.

cpp/include/raft/core/resource/dry_run_flag.hpp
achirkin: The argument `res` here can be instantiated by `raft::dry_run_resources`, i.e. the function `get_dry_run_flag` really is used in both dry run and regular mode.
You may suggest to use dynamic_cast here to check if this is dry run resources handle, but that would also not work: any algorithm is free to copy the resources handle (to make it non-const and override one or more resources); so we may get a regular resources handle here that is still in dry run mode.

cpp/include/raft/core/device_container_policy.hpp
achirkin: Funny enough, both of these are pretty much intentional features. Here's a bit more context for this.

By design, mdarrays and mdspans are not supposed to be resizeable at all. However, at some point we decided to implement sparse matrices using the same container policies as mdarrays. For sparse data, one needs the resize functionality to allow changing sparsity pattern. So the `resize` function is exclusively used for sparse data and only to allocate a new sparse matrix storage of the required size (`initialize_sparsity`).

Back to `reallocate` addressing your concerns. The intention here is to always free up the memory as soon as requested: the released buffer may be huge (e.g. half of GPU memory); the user may want to use the memory as soon as possible, so we want it to be available.

Essentially `reallocate` is almost the same as running `resize` + `shrink_to_fit` on an underlying RMM container. But there's a catch: `reallocate` guarantees to NOT copy the data. This catch is important for dry run: we fake the memory allocations, so any memory copy would cause OOM / break the program.

cpp/include/raft/core/device_mdarray.hpp
achirkin: The scalar is on device - in the default per-device memory resource, which is faked in dry run mode. Therefore we have to guard it (and it falls under "GPU work" category anyway).

docs/source/dry_run_protocol.md
achirkin: Plenty of rapids org libraries use the rmm memory resources directly. They are tracked though.

cpp/include/raft/core/resource/dry_run_flag.hpp
huuanhhuyn: Hmm, when we instantiate a `raft::dry_run_resources` from a non-dryrun existing resource, we init the dry run flag with `resource::set_dry_run_flag(*this, true);`. That would create a dry run resource factory already.  Afterwards `get_dry_run_flag()` will not execute line 65 because the factory is there. Even when you copy a dry-run resource into another resource, the factories are copied over to the new resource.

Therefore, line 65 is only executed if the input resource hasn't been added with a dry run factory before.

I may overlook somewhere though. Even if I am right, this is a minor thing IMO.

cpp/include/raft/core/resource/cuda_stream.hpp
huuanhhuyn: If I understand dry-run correctly, when res is in dry-run mode, the stream contains no CUDA work and contains only light-weight fake memory operations. Therefore it would be fast to sync. 

If user intentionally adds a CUDA/memory work to the stream without guarding, he wants the `sync_stream` to work I suppose.

I mean this guarding on sync stream wouldn't bring any benefit but limit its usage a little.

cpp/include/raft/core/resource/dry_run_flag.hpp
achirkin: Ah, you mean specifically remove the `add_resource_factory` on line 65? Probably this could work. But I think it's ok to keep the line there for homogeneity with the other resource types.

cpp/include/raft/core/resource/cuda_stream.hpp
achirkin: It is still CUDA work and a call to cuda context, which may briefly lock. In dry run mode, nobody should ever call CUDA, that is a hard constraint.

cpp/include/raft/core/device_mdarray.hpp
huuanhhuyn: Acknowledged.

docs/source/dry_run_protocol.md
huuanhhuyn: And non-rapidsai libraries like Cutlass, CUB, Thrust, etc.? Are their allocations tracked?

cpp/include/raft/core/device_container_policy.hpp
huuanhhuyn: Thank you for sharing the context with sparse matrix.

I still don't understand why when the new size is less than or equal to the current size, we don't simply free the redundant memory and keep the remaining `new size` part. This will give back immediately `current_size - new_size` to the user. Or do you mean we have to free everything at the current address and then allocate the new size some where else? If so, why?

huuanhhuyn: The current implementation makes sense to me only when the new size is greater than the current size

docs/source/dry_run_protocol.md
huuanhhuyn: Is finding the upper bound always possible? And what if the upper bound is 3x, 10x?

cpp/include/raft/core/detail/copy.hpp
achirkin: Marking this resolved via https://github.com/NVIDIA/raft/pull/3052 : running code in dry run mode will force-initialize lazy resources (such as cublas handles) for a following normal run unless explicitly prohibited (e.g. thrust policy gets discarded because it references rmm memory resources).

docs/source/dry_run_protocol.md
achirkin: This is covered in the same file below in the section "What Needs Guarding". In general, I don't think it's sustainable to maintain a list of all possible libraries to tell whether they use rmm or not. But it's normally clear from the API or documentation of those libraries.

Please also refer to the presentation for the relevant overview https://drive.google.com/file/d/1uMkM-Xzhi-2ZAKe_z3zvCfHoefn0bAAx/view?usp=drive_link

cpp/include/raft/core/resource/dry_run_flag.hpp
huuanhhuyn: Do you want to resolve this comment. I somehow cannot "resolve comment" myself

cpp/include/raft/core/device_mdarray.hpp
huuanhhuyn: -> resolve comment

divyegala (commented):
Would you consider guarding the kernel launch behind a dispatch function/macro where the dry run flag is checked, like we talked in your offline brown bag session?

achirkin:
Hi @divyegala , I definitely want to do that, but maybe in a separate PR in parallel, where we could focus on discussing the pros and cons of different approaches to do that? The main issue I see is that for being useful for dry run, the dispatcher should take `raft::resources` as an argument, which is not always available in legacy code in detail namespace.

divyegala:
@achirkin thank you, that makes sense to me. I will start reviewing the PR in its current form. 👍

cpp/include/raft/mr/dry_run_resource.hpp
```diff
@@ -0,0 +1,230 @@
...
+      // never read or written (see kDryRunProbeSize / dry_run_resource), and it
+      // may be held (via shared_ptr in an escaped dry-run mdarray) beyond the
+      // lifetime of the original allocation stream. Storing and reusing that
+      // stream_ref would risk dangling-stream UB; the per-thread default stream
+      // is always valid and safe for a buffer whose contents are irrelevant.
+      mr_.deallocate(cuda::stream_ref{cudaStreamPerThread}, ptr_, size_, alignment_);
+    } else {
+      mr_.deallocate_sync(ptr_, size_, alignment_);
+    }
+  }
```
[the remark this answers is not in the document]
achirkin: A good one! fixed

cpp/include/raft/sparse/linalg/detail/sddmm.hpp
```diff
@@ -66,6 +67,8 @@ void sddmm(raft::resources const& handle,
 
   rmm::device_uvector<uint8_t> tmp(bufferSize, resource::get_cuda_stream(handle));
 
+  if (resource::get_dry_run_flag(handle)) { return; }
+
```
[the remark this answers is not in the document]
achirkin: This is false positive. `resource::get_cuda_stream(handle)` is dry run compliant, since it has `raft::resources` in its arguments. Therefore, it doesn't perform any CUDA work in dry run mode - the synchronization is already guarded. Moreover, you suggestion is harmful, because leads to guarding the allocation of the `tmp` buffer.
Please suggest how to improve the review guidelines so that you don't make such mistakes again.

divyegala (approved):
LGTM, thanks!

cpp/include/raft/label/detail/classlabels.cuh
```diff
@@ -31,15 +31,17 @@ namespace detail {
...
- * \param [out] n_unique number of unique labels
+ * \param [in] y device array of labels, size [n]
+ * \param [in] n number of labels
  * \param [in] stream cuda stream
+ * \return number of unique labels (upper bound when dry_run is true)
  */
 template <typename value_t>
-int getUniquelabels(rmm::device_uvector<value_t>& unique, value_t* y, size_t n, cudaStream_t stream)
+int getUniquelabels(
+  bool dry_run, rmm::device_uvector<value_t>& unique, value_t* y, size_t n, cudaStream_t stream)
```
lowener, line 44: Why add the `dry_run` as a first argument, and not at the end, defaulted to false?
This would spare having a second version of that function just below

achirkin: This is to be very explicit in the detail namespace. At some point we should gradually refactor raft to always pass `raft::resources` handle everywhere and remove the need to have this argument at all.

cpp/tests/CMakeLists.txt
```diff
@@ -338,6 +338,8 @@ if(BUILD_TESTS)
     util/popc.cu
     util/pow2_utils.cu
     util/reduction.cu
+    util/dry_run_resources.cpp
```
bdice: Sort this list.

bdice (approved):
One request for CMake, no comments on CI. Approved.

PR #2964, opened 2026-02-23
Drop `thrust::zip_iterator` in favor of `cuda::zip_iterator`

miscco:


aamijar:
Hi @miscco, just bumping this

PR #2968, opened 2026-02-27
Unify memory resources

achirkin:
Use `cuda::mr::any_synchronous_resource` for host, pinned, and managed resource types and give the user explicit control for host, pinned, and managed resources.

#### New
  - `raft::resource::managed_memory_resource` and `raft::resource::pinned_memory_resource` are passed to managed and pinned mdarrays during construction via corresponding container policies. This allows the user to replace/modify these resources, for example, to add logging or memory pooling.
  - `raft::mr::get_default_host_resource` and `raft::mr::set_default_host_resource` can be used by the user to alter the default host resource the same way. It is not stored in `raft::resources` handle like the other two for two reasons:
    1. To mirror rmm default device resource getter/setter
    2. To avoid breaking the `raft::make_host_mdarray` overloads that do not take `raft::resources` as an argument (many instances across raft and cuvs).

#### Changed

 - Use `raft::mr::host_resource_ref` and `raft::mr::host_device_resource_ref` for the non-owning semantics (defined as `cuda::mr::synchronous_resource_ref` with appropriate access attributes)
 - Use `raft::host_resource` and `raft::host_device_resource` for owning semantics (defined as `cuda::mr::any_synchronous_resource` with appropriate access attributes)

With these changes, raft fully switches to `cuda::mr` types for host and host-device resources, while still using `rmm` types for device async resources. Changing the latter would break a lot of cuVS and is not needed - `rmm` will eventually fully converge to `cuda::mr` anyway.

#### Breaking changes
  - Rename container policies
  - Reuse of a single `host_container` for the three types of resources.
  - Switch to using `cuda::mr::any_synchronous_resource` from `std::pmr::memory_resource`

The effect of this changes should be limited, because the policies are hidden behind the mdarray templates and synonyms and the  `std::pmr::memory_resource` was introduced recently and haven't been used much.

### DIFF
--- a/cpp/include/raft/mr/mmap_memory_resource.hpp
+++ b/cpp/include/raft/mr/mmap_memory_resource.hpp
@@ -101,6 +101,7 @@ class mmap_memory_resource {
       ptr = mmap_verbose(bytes, prot, flags, -1, 0);
     }
     if (flags_ & kMmapRequestHugePages) {
+      // Find a page-aligned subrange of the allocated memory to madvise
       auto madvize_start = raft::round_up_safe(reinterpret_cast<uintptr_t>(ptr), kHugePageSize);
       auto madvize_end =
         raft::round_down_safe(reinterpret_cast<uintptr_t>(ptr) + bytes, kHugePageSize);


cpp/include/raft/mr/managed_memory_resource.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+namespace raft::mr {
+
+/**
+ * @brief A cuda::mr::synchronous_resource backed by cudaMallocManaged / cudaFree.
+ *
+ * Provides CUDA unified (managed) memory accessible from both host and device.
+ * Uses synchronous allocation (no stream). Binds to raft::mr::host_device_resource_ref.
+ */
+class managed_memory_resource {
```
bdice: This is implemented in CCCL already. Please do not introduce a new implementation of this since one already exists.

https://nvidia.github.io/cccl/unstable/libcudacxx/runtime/memory_pools.html#cuda-managed-memory-pool
https://nvidia.github.io/cccl/unstable/libcudacxx/runtime/legacy_resources.html#cuda-mr-legacy-managed-memory-resource

Use `cuda::mr::legacy_managed_memory_resource` on CUDA 12 and `cuda::managed_memory_pool` on CUDA 13 (it's considerably faster). Maybe write a factory that returns the correct resource type for your CUDA version.

achirkin:
Testing the breaking changes:
 - cuvs https://github.com/rapidsai/cuvs/pull/1882
 - cuml https://github.com/rapidsai/cuml/pull/7856

PR #2972, opened 2026-03-04
Use the new reference-based RMM API to set RMM pools in `device_resources_snmg`

viclafargue:
This PR updates the `device_resources_snmg` utility so that it uses the new reference-based RMM API to set RMM pools.


jinsolp (approved):
Thanks @viclafargue , LGTM

PR #2975, opened 2026-03-05
add no_libucx matrix option in dependencies.yaml

trxcllnt:
This option allows generating dependencies without `libucx` in the dependencies list, which is something we have to do for NVAIE/DLFW builds.


jameslamb (approved):
Totally support this, thanks for doing it.

PR #2993, opened 2026-04-02
Docs improvements

### LINKED ISSUE #2992: Docs function signature code blocks style
The function signature in the docs is hard to read. For example:

https://docs.rapids.ai/api/raft/stable/cpp_api/linalg_map_reduce/#_CPPv4I0000000E19coalesced_reductionvRKN4raft9resourcesEN4raft18device_matrix_viewIK11InValueType7IdxType12LayoutPolicyEEN4raft18device_vector_viewI12OutValueType7IdxTypeEE12OutValueTypeb10MainLambda12ReduceLambda11FinalLambda

Instead, can we make it look like cuvs docs like this:
https://docs.rapids.ai/api/cuvs/stable/cpp_api/cluster_kmeans/#_CPPv43fitRKN4raft9resourcesERKN4cuvs7cluster6kmeans6paramsEN4raft18device_matrix_viewIKfiEENSt8optionalIN4raft18device_vector_viewIKfiEEEEN4raft18device_matrix_viewIfiEEN4raft16host_scalar_viewIfEEN4raft16host_scalar_viewIiEE

aamijar:
Resolves #2992.

1. Change the sphinx theme to `nvidia_sphinx_theme` just like cuvs. This theme has better function signature formatting for the API reference.
2. Keep the rapids branding for the favicon and top left corner logo
3. Add higher quality rapids branding images
4. Copy https://github.com/rapidsai/cuvs/pull/1377 overload grouping from cuvs.

# Old
old_docs

# New
new_docs


lowener (commented):
I agree, the function signature are much more readable now.
Can you keep the "RAFT" in the top left corner in capital letters?
Also small remark but there is probably a case missing in the `collapse_overloads.js` because if you go to C++ API -> Sparse page there are still overloads present

aamijar:
The right sidebar grouping should be fixed now and the RAFT capitalization in the top left. Not sure how to handle the grouping on the TOC pages like when you just click on sparse.

lowener (approved):
LGTM

docs/source/_static/collapse_overloads.js
```diff
@@ -0,0 +1,68 @@
...
+    const seen = new Set();
+    links.forEach(link => {
+      let text = link.textContent.trim();
+      let norm = text.replace(/\(\)$/, ""); // strip trailing ()
+
+      if (seen.has(norm)) {
+        // hide duplicate
+        link.parentElement.style.display = "none";
+      } else {
+        seen.add(norm);
```
bdice, line 18: It’s unexpected to me that this set is mutable, given that it’s declared const. But my JS is rusty…

docs/source/conf.py
```diff
@@ -63,7 +63,7 @@
 master_doc = "index"
 
 # General information about the project.
-project = "raft"
+project = "RAFT"
 copyright = "2023-2026, NVIDIA Corporation"
```
bdice: Let’s make this self-updating:

https://github.com/rapidsai/rmm/blob/1ffd9de9f48f674a3112ec4aa227e5bbedd4ccd1/docs/conf.py#L26

bdice (approved):

docs/source/_static/collapse_overloads.js
aamijar: I haven't used JS either, but after looking it up `const` in JS means that you cannot reassign the label to other data, but you are allowed to mutate the existing data.

docs/source/conf.py
aamijar: Addressed in https://github.com/rapidsai/raft/pull/2993/commits/721c762181d9dc00110d5216c3053f03797990f0

Issue #2705, opened 2025-06-16
[FEA] Lanczos Solver SM algorithm


aamijar:
Follow up to #2628. During the #2628 work the SM tests did not consistently pass, so it was omitted from that PR. For example one of the eigenvalues did not match in the output.
The cupy implementation also does not implement SM as they too were having issues with tests cases passing.

PR #2999, opened 2026-04-09
Add Randomized SVDs

Intron7:
This PR adds randomized SVDs to raft based on (Halko et al. 2009) and (Tomás et al. 2024). I also added the possibility for a very limited linear operator. This one is C++ and might be useful for sparse PCA in cuml. It's tested in the C++ layer but not exposed in Python. This PR mimics the implementation I did for [rapids-singlecell](https://github.com/scverse/rapids-singlecell/blob/main/src/rapids_singlecell/preprocessing/_sparse_pca/_block_lanczos.py)

sparse svd benchmarks (k=50, n_oversamples=10)
GPU: RTX PRO 6000 Blackwell. raft = best of 3; sklearn/scipy = single run.

```
rows       nnz      raft GPU   sklearn CPU   scipy ARPACK   sk/raft   scipy/raft
50k        101M     0.27s      7.38s         18.69s         27.3×     69.2×
200k       400M     1.06s      25.61s        62.49s         24.1×     58.8×
500k       1.02B    2.75s      64.97s        153.40s        23.6×     55.8×
982k       2.01B    5.48s      126.16s       307.04s        23.0×     56.1×
```

~23-27x over sklearn, ~56-69x over scipy ARPACK. speedup is remarkably stable across scales — raft stays sub-linear while CPU methods scale linearly with nnz. at ~1M cells raft finishes in 5.5s vs 2 min (sklearn) / 5 min (scipy).


python/pylibraft/pylibraft/sparse/linalg/svds.pyx
```diff
@@ -0,0 +1,205 @@
...
+
+    # Extract CSR arrays and ensure int32 indices
+    indptr = A.indptr
+    indices = A.indices
+    data = A.data
+
+    IndexType = indptr.dtype
+    if IndexType == np.int64:
+        indptr = indptr.astype(np.int32)
+        indices = indices.astype(np.int32)
```
viclafargue, line 146: Isn't there an integer overflow risk here? I think we should at least warn the user that the indices will be converted.

cpp/include/raft/sparse/solver/svds_config.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * @brief Configuration parameters for the sparse randomized SVD solver
+ *
+ * @tparam ValueTypeT Data type for values (float or double)
+ */
+template <typename ValueTypeT>
+struct sparse_svd_config {
+  /** @brief Number of singular values/vectors to compute */
+  int n_components;
```
viclafargue: We should maybe set a default for `n_components`. A C++ user that forgets to specify a value would face an undefined behavior. Setting it to 0 would allow it to be catched immediately by parameters validation.

python/pylibraft/pylibraft/sparse/linalg/svds.pyx
```diff
@@ -0,0 +1,205 @@
...
+        namespace "raft::sparse::solver" nogil:
+
+    cdef cppclass sparse_svd_config[ValueTypeT]:
+        int n_components
+        int n_oversamples
+        int n_power_iters
+        optional[uint64_t] seed
+
+cdef sparse_svd_config[float] config_float
+cdef sparse_svd_config[double] config_double
```
viclafargue: Using module-level global config is not thread-safe. Please instantiate them in the function where they are needed.

cpp/include/raft/sparse/solver/randomized_svds.cuh
```diff
@@ -0,0 +1,96 @@
...
+ * @param A input sparse CSR matrix of shape (m, n)
+ * @param singular_values output singular values of shape (n_components,) in descending order
+ * @param U output left singular vectors of shape (m, n_components), col-major
+ * @param Vt output right singular vectors of shape (n_components, n), col-major
+ */
+template <typename ValueTypeT, typename NNZTypeT>
+void sparse_randomized_svd(
+  raft::resources const& handle,
+  sparse_svd_config<ValueTypeT> const& config,
+  raft::device_csr_matrix_view<ValueTypeT, int, int, NNZTypeT> A,
```
viclafargue:
```suggestion
  raft::device_csr_matrix_view<const ValueTypeT, int, int, NNZTypeT> A,
```

The input matrix should be `const`.

Also for the doc, we sometimes use the paradigm `@param [in]`, `@param [inout]`, `@param [out]`. `@param [in]` is supposed to be usable with `const` data.

cpp/include/raft/sparse/solver/detail/randomized_svds.cuh
```diff
@@ -0,0 +1,209 @@
...
+                         Omega.data_handle(),
+                         static_cast<std::size_t>(n) * block_size,
+                         ValueTypeT(0),
+                         ValueTypeT(1));
+    op.apply(handle,
+             raft::make_device_matrix_view<const ValueTypeT, uint32_t, raft::col_major>(
+               Omega.data_handle(), n, block_size),
+             Y.view());
+  }  // Omega freed here
+  cholesky_qr2(handle, Y.view());
```
viclafargue: We should maybe check the return value of `cholesky_qr2` calls and emit a warning when there's a fallback to standard QR.

python/pylibraft/pylibraft/sparse/linalg/svds.pyx
```diff
@@ -0,0 +1,205 @@
...
+    """
+    Compute the largest ``k`` singular values and corresponding singular
+    vectors of a sparse matrix using randomized SVD.
+
+    Computes the truncated SVD: ``A ~ U @ diag(S) @ Vt``.
+
+    Args:
+        A (cupyx.scipy.sparse.csr_matrix): Sparse CSR matrix of shape
+            ``(m, n)``. Must be of type
+            :class:`cupyx.scipy.sparse.csr_matrix`.
```
viclafargue, line 82: We should check that `A` is indeed a sparse CSR matrix and raise an error message otherwise.

cpp/include/raft/sparse/solver/detail/randomized_svds.cuh
```diff
@@ -0,0 +1,209 @@
...
+               "singular_values must have size n_components");
+  RAFT_EXPECTS(U.extent(0) == static_cast<uint32_t>(m) && U.extent(1) == static_cast<uint32_t>(k),
+               "U must have shape (m, n_components)");
+  RAFT_EXPECTS(
+    Vt.extent(0) == static_cast<uint32_t>(k) && Vt.extent(1) == static_cast<uint32_t>(n),
+    "Vt must have shape (n_components, n)");
+
+  auto stream = raft::resource::get_cuda_stream(handle);
+
+  int block_size = std::min(k + p, std::min(m, n));
```
viclafargue: We should avoid silently clamping as it affects approximation quality. Please add a warning when `n_components + n_oversamples < min(n_rows, n_features)`.

cpp/include/raft/sparse/solver/detail/randomized_svds.cuh
```diff
@@ -0,0 +1,209 @@
...
+  RAFT_EXPECTS(block_size >= k, "block_size (n_components + n_oversamples) must be >= n_components");
+
+  // Initialize RNG
+  uint64_t seed = config.seed.value_or(0);
```
viclafargue:
```suggestion
  uint64_t seed = config.seed.value_or(std::random_device{}());
```

Setting `seed` to `std::nullopt` is still deterministic and do not produce different runs. We should fix this.

python/pylibraft/pylibraft/sparse/linalg/svds.pyx
```diff
@@ -0,0 +1,205 @@
...
+
+    .. note::
+        This function uses randomized SVD (Halko et al. 2009) with
+        CholeskyQR2 orthogonalization (Tomas et al. 2024) for efficient
+        GPU execution.
+
+    """
+
+    if A is None:
+        raise Exception("'A' cannot be None!")
```
viclafargue: Please use a `ValueError` or `TypeError` for Python convention consistency.

cpp/include/raft/sparse/solver/randomized_svds.cuh
```diff
@@ -0,0 +1,96 @@
...
+  raft::device_vector_view<ValueTypeT, uint32_t> singular_values,
+  raft::device_matrix_view<ValueTypeT, uint32_t, raft::col_major> U,
+  raft::device_matrix_view<ValueTypeT, uint32_t, raft::col_major> Vt)
```
viclafargue: Just an idea, but it may be interesting to check if it would be possible to save memory and compute by making `U` and `Vt` optional as a user may not always need them (see [svd_qr implementation](https://github.com/rapidsai/raft/blob/034e73ecf22e1869efc985e50950974859e95d57/cpp/include/raft/linalg/svd.cuh#L198-L199)).

cpp/include/raft_runtime/solver/randomized_svds.hpp
```diff
@@ -0,0 +1,49 @@
...
+namespace raft::runtime::solver {
+
+/**
+ * @defgroup sparse_randomized_svd_runtime Sparse Randomized SVD Runtime API
+ * @{
+ */
+
+void sparse_randomized_svd_float(
+  const raft::resources& handle,
+  raft::sparse::solver::sparse_svd_config<float> config,
```
viclafargue:
```suggestion
  const raft::sparse::solver::sparse_svd_config<float>& config,
```

cpp/include/raft_runtime/solver/randomized_svds.hpp
```diff
@@ -0,0 +1,49 @@
...
+  int n_rows,
+  int n_cols,
+  int nnz,
+  raft::device_vector_view<float, uint32_t> singular_values,
+  raft::device_matrix_view<float, uint32_t, raft::col_major> U,
+  raft::device_matrix_view<float, uint32_t, raft::col_major> Vt);
+
+void sparse_randomized_svd_double(
+  const raft::resources& handle,
+  raft::sparse::solver::sparse_svd_config<double> config,
```
viclafargue:
```suggestion
  const raft::sparse::solver::sparse_svd_config<double>& config,
```

cpp/include/raft/sparse/solver/svds_types.hpp
```diff
@@ -0,0 +1,111 @@
...
+  csr_linear_operator(raft::device_csr_matrix_view<ValueTypeT, int, int, NNZTypeT> A, int m, int n)
+    : A_(raft::make_device_csr_matrix_view<const ValueTypeT, int, int, NNZTypeT>(
+        A.get_elements().data(), A.structure_view())),
+      m_(m),
+      n_(n)
+  {
+  }
+
+  int rows() const { return m_; }
+  int cols() const { return n_; }
```
viclafargue:
```suggestion
  int rows() const { return A_.structure_view().get_n_rows(); }
  int cols() const { return A_.structure_view().get_n_cols(); }
```

cpp/include/raft/sparse/solver/svds_types.hpp
```diff
@@ -0,0 +1,111 @@
...
+ *       is currently limited to int-indexed CSR matrices.
+ *
+ * @tparam ValueTypeT Data type of matrix values
+ * @tparam NNZTypeT Type for number of non-zeros
+ */
+template <typename ValueTypeT, typename NNZTypeT = int>
+struct csr_linear_operator {
+  raft::device_csr_matrix_view<const ValueTypeT, int, int, NNZTypeT> A_;
+  int m_;
+  int n_;
```
viclafargue: `A_` already stores the dimensions of the matrix, storing `m_` and `n_` here is redundant, adds complexity and introduces the possibility of bugs. The number of rows and cols can be derived directly from `A_`. Also `A_` should probably be a private member.

cpp/include/raft/sparse/solver/randomized_svds.cuh
```diff
@@ -0,0 +1,96 @@
...
+ *   [1] Halko, Martinsson, Tropp (2009) "Finding structure with randomness"
+ *       https://arxiv.org/abs/0909.4061
+ *   [2] Tomás, Quintana-Ortí, Anzt (2024) "Fast Truncated SVD of Sparse and Dense Matrices
+ *       on Graphics Processors" https://arxiv.org/abs/2403.06218
+ */
+template <typename ValueTypeT, typename OperatorT>
+void sparse_randomized_svd(
+  raft::resources const& handle,
+  sparse_svd_config<ValueTypeT> const& config,
+  OperatorT const& op,
```
viclafargue: Even though the `csr_linear_operator` struct is certainly useful, the user should not ever need to construct this exotic type. The public function should expose an interface with a `raft::device_csr_matrix_view` and the `csr_linear_operator` utility should be built inside of the function.

cpp/include/raft/sparse/solver/detail/cholesky_qr.cuh
```diff
@@ -0,0 +1,163 @@
...
+                        m,
+                        k,
+                        W.data(),
+                        potrf_workspace.data(),
+                        potrf_workspace_size,
+                        dev_info.data())) {
+    // Fallback to standard QR
+    rmm::device_uvector<ValueTypeT> Q_copy(m * k, stream);
+    raft::copy(Q_copy.data(), Q.data_handle(), m * k, stream);
+    raft::linalg::qrGetQ(handle, Q_copy.data(), Q.data_handle(), m, k, stream);
```
viclafargue:
```suggestion
    raft::linalg::qrGetQ(handle, Q.data_handle(), Q.data_handle(), m, k, stream);
```

`qrGetQ` already does a copy internally. Using `Q.data_handle()` twice would allow the operation to work inplace (even the copy could be avoided as `src==dst`). To double check though.

Additionally, `m * k` has an integer overflow risk.

cpp/include/raft/sparse/solver/detail/cholesky_qr.cuh
```diff
@@ -0,0 +1,163 @@
...
```
viclafargue: Similar comment here.

viclafargue (commented):
Thanks! Here are some comments.

cpp/include/raft/sparse/solver/detail/cholesky_qr.cuh
```diff
@@ -0,0 +1,159 @@
...
+                                                    m,
+                                                    &one,
+                                                    Q,
+                                                    m,
+                                                    Q,
+                                                    m,
+                                                    &zero,
+                                                    W,
+                                                    k,
+                                                    stream));
```
achirkin: Please use public `raft::linalg:gemm`

cpp/include/raft/sparse/solver/detail/cholesky_qr.cuh
```diff
@@ -0,0 +1,159 @@
...
+ * providing ~3x speedup over standard Householder QR.
+ *
+ * If Cholesky factorization fails (input is rank-deficient), falls back to standard QR.
+ *
+ * @param handle raft resources handle
+ * @param Q matrix to orthogonalize of shape (m, k), col-major, modified in-place
+ * @return true if CholeskyQR2 succeeded, false if fell back to QR
+ */
+template <typename ValueTypeT>
+bool cholesky_qr2(raft::resources const& handle,
```
achirkin, line 109: Would it be possible to reuse partially or in full, `raft/linalg
/qr.cuh` or `raft/linalg/cholesky_r1_update.cuh` here?

achirkin (changes requested):
Thanks for the PR! Couple notes about usage of raft primitives below

cpp/include/raft/sparse/solver/detail/cholesky_qr.cuh
Intron7: Switched the API

cpp/include/raft/sparse/solver/detail/cholesky_qr.cuh
Intron7: `cholesky_r1_update` does incremental rank-1 updates to an existing factorization, while `CholeskyQR` needs a full batch Cholesky (`potrf`) of the Gram matrix — different operation. `qr.cuh` (Householder QR) is already used as the fallback; the whole point of `CholeskyQR` is to replace it with the faster GEMM+Cholesky+TRSM pipeline. The remaining detail calls (`cusolverDnpotrf`, `cublastrsm`) don't have public wrappers in raft.

cpp/src/raft_runtime/solver/randomized_svds.cuh
```diff
@@ -0,0 +1,28 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <raft/sparse/solver/randomized_svds.cuh>
+
+#define FUNC_DEF(ValueType, Suffix)                                                  \
+  void sparse_randomized_svd_##Suffix(                                               \
```
aamijar: No need to have a separate function with a suffix here. You can just have the same function name with float and double overloads.

cpp/include/raft_runtime/solver/randomized_svds.hpp
```diff
@@ -0,0 +1,47 @@
...
+#include <cstdint>
+
+namespace raft::runtime::solver {
+
+/**
+ * @defgroup sparse_randomized_svd_runtime Sparse Randomized SVD Runtime API
+ * @{
+ */
+
+void sparse_randomized_svd_float(const raft::resources& handle,
```
aamijar: You can remove the suffix here.
Also you can use macros with FUNC_DECL for more concise declarations. See https://github.com/rapidsai/raft/blob/main/cpp/include/raft_runtime/solver/lanczos.hpp

cpp/include/raft/sparse/solver/svds_types.hpp
```diff
@@ -0,0 +1,30 @@
...
+ *                device_matrix_view<const T, uint32_t, col_major> in,
+ *                device_matrix_view<T, uint32_t, col_major> out) const;
+ *   - void apply_transpose(raft::resources const& handle,
+ *                          device_matrix_view<const T, uint32_t, col_major> in,
+ *                          device_matrix_view<T, uint32_t, col_major> out) const;
+ * @{
+ */
+/** @} */
+
+}  // namespace raft::sparse::solver
```
aamijar: Is there a reason to add this empty types file?

cpp/include/raft/sparse/solver/detail/csr_linear_operator.cuh
```diff
@@ -0,0 +1,89 @@
...
+ *       is currently limited to int-indexed CSR matrices.
+ *
+ * @tparam ValueTypeT Data type of matrix values
+ * @tparam NNZTypeT Type for number of non-zeros
+ */
+template <typename ValueTypeT, typename NNZTypeT = int>
+struct csr_linear_operator {
+  /**
+   * @brief Construct from a const CSR matrix view
+   */
```
aamijar, line 33: Great to see a LinearOperator implementation! It could be useful in the lanczos solver as well (https://github.com/rapidsai/raft/issues/2705). Is this one similar to https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.LinearOperator.html

I see this is in a detail file, but it might be nice to have a public interface.

cpp/include/raft/sparse/solver/randomized_svds.cuh
```diff
@@ -0,0 +1,108 @@
...
+ *   [2] Tomás, Quintana-Ortí, Anzt (2024) "Fast Truncated SVD of Sparse and Dense Matrices
+ *       on Graphics Processors" https://arxiv.org/abs/2403.06218
+ */
+template <typename ValueTypeT, typename OperatorT>
+void sparse_randomized_svd(raft::resources const& handle,
+                           sparse_svd_config<ValueTypeT> const& config,
+                           OperatorT const& op,
+                           raft::device_vector_view<ValueTypeT, uint32_t> singular_values,
+                           raft::device_matrix_view<ValueTypeT, uint32_t, raft::col_major> U,
+                           raft::device_matrix_view<ValueTypeT, uint32_t, raft::col_major> Vt)
```
aamijar, line 53: Is there any reason we aren't supporting int64 types?
I'm bringing this up because we've had to update algorithms like spectral embedding from int32 to int64 and it required a lot of refactoring later down the line.
In cuML and cuVS, we are generally preferring to only instantiate int64 index types to avoid binary size increases, which also sometimes requires updating downstream raft primitives.

aamijar (changes requested):
Thanks @Intron7 for the great work! It would be great to see your benchmarks results in the PR description!

I've left my review below:

cpp/include/raft/sparse/solver/randomized_svds.cuh
```diff
@@ -0,0 +1,108 @@
...
+{
+  detail::csr_linear_operator<ValueTypeT, NNZTypeT> op(A);
+  detail::sparse_randomized_svd(handle, config, op, singular_values, U, Vt);
+}
+
+/**
+ * @brief Compute truncated SVD of a sparse CSR matrix using randomized algorithm.
+ *
+ * Overload accepting a mutable CSR matrix view (implicitly converted to const).
```
aamijar: Do we need this overload at all? If `sparse_randomized_svd` doesn't modify `A` then we should just have a const API.

cpp/include/raft/sparse/solver/detail/csr_linear_operator.cuh
Intron7: This is not a full Linear Operator. Its just a very limited version for raft to do PCA with the SVDs

cpp/include/raft/sparse/solver/randomized_svds.cuh
Intron7: Do you mean int64 for indptr and indices or data?

Intron7: Currently blocked by raft::sparse::linalg::spmm, whose public signature hard-codes int for CSR indptr/indices. Once that's generalized to int64 (separate follow-up PR), adding int64 instantiations here is straightforward.

aamijar: Okay thanks for the clarification. If int64 isn't a must have for your use case then I say it's fine to do in a follow up when needed. Changing indptr and indices to int64 also will require more memory.

Intron7:
sparse svd benchmarks (k=50, n_oversamples=10)
GPU: RTX PRO 6000 Blackwell. raft = best of 3; sklearn/scipy = single run.

```
rows       nnz      raft GPU   sklearn CPU   scipy ARPACK   sk/raft   scipy/raft
50k        101M     0.27s      7.38s         18.69s         27.3×     69.2×
200k       400M     1.06s      25.61s        62.49s         24.1×     58.8×
500k       1.02B    2.75s      64.97s        153.40s        23.6×     55.8×
982k       2.01B    5.48s      126.16s       307.04s        23.0×     56.1×
```

~23-27x over sklearn, ~56-69x over scipy ARPACK. speedup is remarkably stable across scales — raft stays sub-linear while CPU methods scale linearly with nnz. at ~1M cells raft finishes in 5.5s vs 2 min (sklearn) / 5 min (scipy).

python/pylibraft/pylibraft/sparse/linalg/svds.pyx
Intron7: added in an error

aamijar (approved):
Thanks @Intron7!

cpp/include/raft/sparse/solver/detail/randomized_svds.cuh
```diff
@@ -0,0 +1,256 @@
...
+                       k,
+                       CUBLAS_OP_N,
+                       CUBLAS_OP_N,
+                       one,
+                       zero,
+                       stream);
+  }
+
+  // Step 9: Truncate S, optionally compute Vt
+  raft::copy(singular_values.data_handle(), S_full.data_handle(), k, stream);
```
achirkin, line 231: A nitpick
```suggestion
  raft::copy(handle, singular_values, raft::make_const_mdspan(S_full));
```

achirkin (approved):

cpp/include/raft/sparse/solver/detail/randomized_svds.cuh
```diff
@@ -0,0 +1,256 @@
...
+                      Z.data_handle(),
+                      n,
+                      block_size,
+                      S_full.data_handle(),
+                      U_bt_ptr,
+                      Vt_bt_ptr,
+                      true,  // transpose right vectors: Vt_bt -> V_bt (block_size x block_size)
+                      true,  // generate left vectors
+                      true,  // generate right vectors
+                      stream);
```
viclafargue: Do we need to allocate `U_bt`/`Vt_bt` scratches when not requested? We could maybe avoid this and run `svdQR` with nullptr and generate vectors set to false?

cpp/include/raft/sparse/solver/randomized_svds.cuh
```diff
@@ -0,0 +1,117 @@
...
+#include <type_traits>
+
+namespace raft::sparse::solver {
+
+namespace detail {
+// Helper alias used to put the optional U/Vt parameters in a non-deduced context
+// so template argument deduction picks up ValueTypeT from earlier parameters and
+// implicit conversion from `device_matrix_view` to `std::optional<...>` works.
+template <typename T>
+using nondeduced_optional_matrix_view_t = std::optional<std::type_identity_t<T>>;
```
viclafargue: It looks like `std::type_identity_t` is a C++ 20 feature. Should we not update [the CMakeLists.txt file](https://github.com/rapidsai/raft/blob/c38237f7b17bf9b4d52fb416d298573dada40bf8/cpp/CMakeLists.txt#L194) accordingly? cc @divyegala

python/pylibraft/pylibraft/sparse/linalg/svds.pyx
```diff
@@ -0,0 +1,284 @@
...
+    # `int` (int32) for nnz / indptr / indices, so any overflow on conversion
+    # cannot be recovered downstream — error out before the cast.
+    indptr = A.indptr
+    indices = A.indices
+    data = A.data
+
+    INT32_MAX = (1 << 31) - 1
+
+    IndexType = indptr.dtype
+    if IndexType == np.int64:
```
viclafargue: It looks like index dtype handling is driven only by `indptr.dtype`. That should usually be fine for CuPy CSR matrices, since indptr and indices normally share the same dtype, but it would be safer to validate or convert them independently.

viclafargue (approved):
Thanks! LGTM, minor comments.

cpp/include/raft/sparse/solver/randomized_svds.cuh
```diff
@@ -0,0 +1,117 @@
...
+}  // namespace detail
+
+/**
+ * @defgroup sparse_randomized_svd Sparse Randomized SVD
```
lowener, line 31: Add the newly created groups to doxygen (C++ and Python side)

cpp/include/raft/sparse/solver/svds_config.hpp
```diff
@@ -0,0 +1,35 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+#include <cstdint>
+#include <optional>
+
+namespace raft::sparse::solver {
+
```
lowener, line 12:
```
@addtogroup sparse_randomized_svd
@{
...
@}
```

aamijar:
This is targeting main, should it be targeting 26.06? Also it says this: "Waiting on code owner review from rapidsai/raft-cmake-codeowners and/or rapidsai/raft-python-codeowners."

jameslamb (approved):
Giving `ci-codeowners` / `packaging-codeowners` approval (CMake changes look non-controversial).

cjnolet (approved):

PR #3000, opened 2026-04-10
Fix gbench 1.9.5

aamijar:
https://github.com/rapidsai/rapids-cmake/pull/992 broke the raft gbench build. This PR addresses that and unblocks CI.

google-benchmark 1.9.5 requires passing in a unique pointer instead of a raw pointer for `RegisterBenchmarkInternal`


bdice (approved):
Thanks. I think I fixed this locally and lost track of filing a PR.

achirkin (approved):
Thanks, LGTM

PR #3002, opened 2026-04-14
[REVIEW] Add all-to-all, scatter, and scatterv functions to raft::comms_t

seunghwak:
`ncclAlltoAll`, `ncclScatter`, and `ncclGather` was added in NCCL 2.28.3.

We previously performed all-to-all using multi-cast and gather was performed using send/receive. `raft::comms_t` lacks scatter and scatterv.

This PR adds alltoall (`ncclAlltoAll` wrapper), scatter (`ncclScatter` wrapper), and scatterv (implemented using send/receive).

This PR also updates gather to directly call `ncclGather` (instead of emulating gather using send/receive).


cpp/include/raft/comms/detail/test.hpp
```diff
@@ -20,12 +20,52 @@ namespace raft {
...
+  rmm::device_uvector<int> recv_d(communicator.get_size(), stream);
+
+  RAFT_CUDA_TRY(cudaMemcpyAsync(temp_d.data(),
+                                sends.data(),
+                                sizeof(int) * communicator.get_size(),
+                                cudaMemcpyHostToDevice,
+                                stream));
+
+  communicator.alltoall(temp_d.data(), recv_d.data(), 1, stream);
+  communicator.sync_stream(stream);
```
tarang-jain: why do we need to sync the communicator here? The operations after this are ordered on the same stream, right?

cpp/include/raft/comms/detail/test.hpp
```diff
@@ -20,12 +20,52 @@ namespace raft {
...
+  comms_t const& communicator = resource::get_comms(handle);
+
+  std::vector<int> sends(communicator.get_size(), communicator.get_rank());
+
+  cudaStream_t stream = resource::get_cuda_stream(handle);
+
+  rmm::device_uvector<int> temp_d(communicator.get_size(), stream);
+  rmm::device_uvector<int> recv_d(communicator.get_size(), stream);
+
+  RAFT_CUDA_TRY(cudaMemcpyAsync(temp_d.data(),
```
tarang-jain: Please use raft::copy instead of direct calls to the CUDA function

cpp/include/raft/comms/detail/test.hpp
```diff
@@ -159,12 +199,98 @@ bool test_collective_allgather(raft::resources const& handle, int root)
...
+  cudaStream_t stream = resource::get_cuda_stream(handle);
+
+  rmm::device_uvector<int> temp_d(communicator.get_rank() == root ? communicator.get_size() : 0,
+                                  stream);
+  rmm::device_scalar<int> recv_d(stream);
+
+  if (communicator.get_rank() == root) {
+    std::vector<int> sends(communicator.get_size(), communicator.get_rank());
+    std::fill(sends.begin(), sends.end(), root);
+    RAFT_CUDA_TRY(cudaMemcpyAsync(
```
tarang-jain: again use raft::copy here

cpp/include/raft/comms/detail/test.hpp
```diff
@@ -159,12 +199,98 @@ bool test_collective_allgather(raft::resources const& handle, int root)
...
+      temp_d.data(), sends.data(), sizeof(int) * sends.size(), cudaMemcpyHostToDevice, stream));
+  }
+
+  communicator.scatter(
+    communicator.get_rank() == root ? temp_d.data() : nullptr, recv_d.data(), 1, root, stream);
+  communicator.sync_stream(stream);
+
+  int temp_h = -1;  // Verify more than one byte is being sent
+  RAFT_CUDA_TRY(
+    cudaMemcpyAsync(&temp_h, recv_d.data(), sizeof(int), cudaMemcpyDeviceToHost, stream));
```
tarang-jain: here as well. raft::copy

cpp/include/raft/comms/detail/test.hpp
```diff
@@ -159,12 +199,98 @@ bool test_collective_allgather(raft::resources const& handle, int root)
...
+
+  communicator.scatterv(
+    communicator.get_rank() == root ? temp_d.data() : nullptr,
+    recv_d.data(),
+    communicator.get_rank() == root ? sendcounts.data() : static_cast<size_t*>(nullptr),
+    communicator.get_rank() == root ? displacements.data() : static_cast<size_t*>(nullptr),
+    recv_d.size(),
+    root,
+    stream);
+  communicator.sync_stream(stream);
```
tarang-jain: Again here -- is there a specific reason why you sync the whole stream after the multi-gpu operation?

tarang-jain (commented):
Thanks! These seem to be much needed. A few comments from my end.

cpp/include/raft/comms/detail/test.hpp
seunghwak: Thanks, I used cudaMemcpyAsync to be consistent with the rest of the file but we should better replace cudaMemcpyAsync with raft calls in the entire file.

I replaced cudaMemcpyAsync with raft copy calls (raft::update_host & update_device).

cpp/include/raft/comms/detail/test.hpp
seunghwak: Yes, many tests in this file have unnecessary stream synchronizations. Just followed the convention but deleted all of them in this file (excluding the necessary ones after moving data from device to host).

cpp/include/raft/comms/detail/test.hpp
seunghwak: Done.

cpp/include/raft/comms/detail/test.hpp
seunghwak: Done.

cpp/include/raft/comms/detail/test.hpp
seunghwak: No, just following the convention in this file. No need, deleted all the stream synchronization after communication functions.

tarang-jain (approved):

cjnolet (approved):

PR #2996, opened 2026-04-03
Migrate RMM usage to CCCL MR design

bdice:
## Summary
- Remove `device_memory_resource` base class usage, de-template all resource and adaptor types, replace pointer-based per-device resource APIs with ref-based equivalents
- Part of rapidsai/rmm#2011. Migration guide: rapidsai/rmm#2344.
- Supersedes #2917 and #2920

Depends on https://github.com/rapidsai/rmm/pull/2361.
Depends on https://github.com/rapidsai/ucxx/pull/636.

### Core resource infrastructure
- **`device_memory_resource.hpp`**: Remove `any_resource_bridge` (which inherited from `rmm::mr::device_memory_resource`), remove all `shared_ptr<device_memory_resource>` constructor overloads, consolidate to `any_resource`-only path
- **`device_resources.hpp`**: Remove deprecated constructor taking `shared_ptr<device_memory_resource>`, update `get_workspace_resource()` return type (de-templated `limiting_resource_adaptor`)
- **`device_resources_snmg.hpp`**: Remove stale include, de-template `pool_memory_resource`
- **`handle.hpp`**: Remove deprecated constructors taking `shared_ptr<device_memory_resource>`
- **`device_resources_manager.hpp`**: Retype `workspace_mrs` vector from `shared_ptr<device_memory_resource>` to `raft::mr::device_resource`, update `set_workspace_memory_resource()` signature accordingly, de-template `pool_mr_` to `optional<pool_memory_resource>`, remove `dynamic_cast` for upstream type detection, replace `get/set_current_device_resource()` with `_ref` variants

### Memory tracking
- **`memory_tracking_resources.hpp`**: Remove `device_tracking_bridge` (inherited from `device_memory_resource`), use `set_current_device_resource_ref()` directly

### Call sites using `get_workspace_resource()` → `get_workspace_resource_ref()`
- `select_k-inl.cuh`, `select_radix.cuh`, `select_warpsort.cuh`, `sparse/select_k-inl.cuh`, `bitmap_to_csr.cuh`, `bitset_to_csr.cuh`

### Benchmarks
- **`benchmark.hpp`**: De-template `pool_memory_resource`, use `any_resource` for RAII restore
- **`gather.cu`**, **`subsample.cu`**: Same pattern

### Tests
- **`handle.cpp`**: Dereference `limiting_resource_adaptor*` for `device_buffer` constructor
- **`device_resources_manager.cpp`**: Remove workspace-related test code for removed APIs
[...]

PR #3004, opened 2026-04-24
Enhanced support for memory-tracking raft resources

achirkin:
1. Change the host memory resource to have the same owning semantics as the device memory resources as of https://github.com/rapidsai/raft/pull/2996
2. Add a workaround to  `statistics_adaptor.hpp` to compile via nvcc (see code comment)
3. Add `memory_stats_resources`. In constrast to `memory_tracking_resources` this doesn't stream the memory usage online, but only reports the overall usage metrics.

breaking change: moved raft/util/memory_tracking_resources.hpp -> raft/core/memory_tracking_resources.hpp


cpp/include/raft/mr/statistics_adaptor.hpp
```diff
@@ -72,6 +72,32 @@ class statistics_adaptor : public cuda::forward_property<statistics_adaptor<Upst
   {
   }
 
+  // NVCC injects __host__ __device__ on std::shared_ptr special members,
+  // which makes the *implicit* or *defaulted* special members __host__
+  // __device__ too.  That conflicts with Upstream types whose special
+  // members are __host__ only (e.g. rmm::device_async_resource_ref).
+  // User-defined bodies (not = default) force plain __host__ execution space.
```
bdice, line 82: Good find. Are there changes you would suggest for RMM or CCCL?

bdice (approved):

cpp/include/raft/mr/statistics_adaptor.hpp
achirkin: Thanks! I think, CCCL/rrm are fine, because the `cuda::mr::shared_resource` defines the copy/move constructors explicitly.

cpp/include/raft/core/memory_stats_resources.hpp
```diff
@@ -0,0 +1,237 @@
+/*
```
cjnolet, line 1: Why wouldn't we put this in raft/core/resources instead of raft/util? It would be nice to put all of these in the same place, given RAFT Is more than just memory resources. That wy users can find them easily. I think we should do the same with the above instead of putting them in `mr`.

achirkin: I follow the example of the `memory_tracking_resources` here. But that was introduced recentely too (26.04), so we can change both not risking breaking things too much.
Just to clarify the logic of my choice for these two: I've put them into `util`  folder, because they are not a part of a "normal" algorithm flow but rather utilities to analyize/profile the memory-related resource usage. It's not a strong prerefence though. Would like me to move them both to the core folder (and mark the PR as breaking)?

NB the current state of raft (main + PR):
  - `raft/core/resources/` - individual resources, such as cuda stream, cublas, memory resources
  - `raft/core/` - raft::resources itself (+memory_tracking_resources, memory_stats_resources, dry_run_resources if we decide so)
  - `raft/mr/` - CCCL/rmm compatible memory resources
  - `raft/util/` - current of memory_tracking_resources, memory_stats_resources, dry_run_resources

cjnolet: The layout makes sense once you describe the reasoning, but I just fear people aren't going to be able to find them in these locations because the logic behind them isn't immediately obvious. I agree w/ using `mr` namespace for the rmm compatible resources (vs the `raft::resources`), but I think these resources might be easier to find if we also put them in `mr`. Maybe we could do a combination of both and put them in `raft/core/resource/util`?

achirkin: Hmm, the unfortunate naming confuses me again :) By "these resources" you mean `raft::memory_tracking_resources`, `raft::memory_stats_resources`, `raft::dry_run_resources`? I think, it's reasonable to put them in `raft/core/resource/util`, but still not ideal.
These three are not "resources" like the one can find in `raft/core/resource` folder; they are flavors of `raft::resources` handle type itself. Hence I was considering to either put them in `raft/util` or in `raft/core` (alongside `raft::resources` handle), but the latter is a bit overcrowded already.

I also opened a separate PR to expose `raft::memory_tracking_resources` https://github.com/rapidsai/cuvs/pull/2073 . It supports the above positioning logic: in other languages, we don't need to distinguish between the handle types - we just provide another constructor function to create the new utility resource handle in lieu of the original resources handle.

cpp/include/raft/core/memory_stats_resources.hpp
```diff
@@ -0,0 +1,237 @@
...
+
+    // --- Managed ---
+    {
+      mr::statistics_adaptor<mr::host_device_resource_ref> sa{managed_ref};
+      managed_stats_ = sa.get_stats();
+      resource::set_managed_memory_resource(*this, std::move(sa));
+    }
+
+    // --- Device (global) ---
+    // Invalidate the cached thrust policy (the resource_ref it captured
```
tfeher, line 209: Do we need to invalidate this again during destruction of the `memory_stats_resources`?

cpp/include/raft/core/memory_stats_resources.hpp
tfeher: Since `memory_stats_resources` is a subclass of `raft::resources` It makes sense to me to keep it in `raft/core` where `resources.hpp` is located.  But we have many other headers there which makes it harder to discover the new utilities. To me placing a `raft::resources` object into the `mr` folder would be somewhat confusing.

tfeher (commented):
Thanks Artem for the PR! Overall looks good, I have only one question regarding Thrust policy handling.

cpp/include/raft/core/memory_stats_resources.hpp
achirkin: No, unlike the device memory resource, the thrust policy is a local resource object, so it dies out with its owning `raft::resources` handle.

cpp/include/raft/core/memory_stats_resources.hpp
cjnolet: Maybe I should clarify- I'd like to see this file in raft/core/resource with the other items in the resource/resources APIs. It doesn't belong in raft/util.

cjnolet (dismissed):

cpp/include/raft/core/memory_stats_resources.hpp
achirkin: Ok, as per our offline discussion, moved both files to `raft/core`. Marking the PR 'breaking'.

achirkin:
Moving this to 26.08 as 26.06 deadline is missed

tfeher (approved):
Thanks Artem for the update, LGTM.

Issue #1799, opened 2023-09-01
[FEA] New `raft-ann-bench` NGC container for reproducible benchmarking


nwstephens:
Once a docker container for benchmarking RAFT is completed (#1797) we should make that container available on NGC.

Issue #1796, opened 2023-08-31
[BUG] neighbors::refine can potentially undo pre-filtering


achirkin:
When fixing https://github.com/rapidsai/raft/pull/1793 , I realized that the refinement operation can undo the filtering if the number of valid top-k-results is less than `k`.

So far, we mark invalid results with infinite distance in IVF-Flat and IVF-PQ. But the `neighbors::refine` can "refine" the distance back to a small value!
It's not a problem with the fused IVF-PQ similarity kernel, because the invalid results do not make it to the outputs; but the non-fused kernel only marks the bad values with inftinite distance and relies on `matrix::select_k` to filter out invalid results.
`matrix::select_k` should normally treat infinity as normal values, so the invalid key-value pairs may make it all the way to the refine operation. Moreover, we may get infinities due to low-precision float overflows; refining these is an expected behavior.

achirkin:
I think, a reasonably ok-ish solution is to add the filtering template to the refine operation, same as it is passed to ANN search kernels.

Issue #1795, opened 2023-08-31
[FEA] Pluggable logging


cjnolet:
We should be able to abstract our logging APIs in a way where we don't have to force a specific logging dependency on our users. It's possible this feature may need to propagate through other upstream RAPIDS libraries as well.

PR #3005, opened 2026-04-29
Predictable raft::resources

achirkin:
Make the `raft::resources` resource initialization semantics more predictable.

1. All resources still are still initialized lazily on-demand (*no change*), but behave as-if they were initialized during the `raft::resources` handle construction: all copies of the `raft::resources` handle point to the same resources (*not a breaking change, fixes the re-initialization issue*).
2. `set_resource` changes to non-const semantics (*breaking change*).
     **Before**: all `set_xxx` resource-updating calls were operating on const handle
     **Now**: all `set_xxx` resources require a non-const handle.

### Thread-safety

First and foremost, thread-safety of using a specific resource depends on that resource. Here we discuss the thread-safety of using `raft::resources` handle itself (get_resource and set_resource functions).

#### Accessing the same resource by const ref

Updates (set_resource) are not possible (a user should copy a handle and modify the new one). All concurrent get_resource calls are atomic and safe, even if they initialize the factories and resources under the hood. The worst can happen is the same resource being initialized concurrently in two threads but only one being stored in the handle (the other one is discarded).

#### Accessing the same resource by non-const ref

Using the same object by non-const ref from multiple threads is always unsafe.

#### Accessing copies of the same resource

The resources and factories are updated atomically. Modifying any resource doesn't propagate to the copies. Accessing a resource while another threads access or modifies the same resource via another handle is thus safe.

### Implementation details

The PR adds one more layer of indirection (one extra shared_ptr) to each resource, which may cause an extra runtime overhead. This is unavoidable if we want to allow lazy-initialized resources back-propagate across handles.
On the other hand, the PR removes the handle mutex in favor of C++20 `std::atomic<shared_ptr>`, which may reduce the runtime overheads a little bit.

### Breaking changes

  - All resource setters change the function signature - this is a big breaking change. However, all known use-sites already call the resources setters on non-const `raft::resources` handles.
  - `std::atomic<shared_ptr>` requires enforcing C++20 for all dependent libraries. An alternative would be to put a mutex per `raft::resource_cell`.

### DIFF
--- a/cpp/include/raft/core/resources.hpp
+++ b/cpp/include/raft/core/resources.hpp
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2022-2024, NVIDIA CORPORATION.
+ * SPDX-FileCopyrightText: Copyright (c) 2022-2026, NVIDIA CORPORATION.
  * SPDX-License-Identifier: Apache-2.0
  */
 #pragma once
@@ -9,9 +9,7 @@
 #include <raft/core/error.hpp>  // RAFT_EXPECTS
 #include <raft/core/logger.hpp>
 
-#include <algorithm>
-#include <mutex>
-#include <string>
+#include <memory>
 #include <vector>
 
 namespace raft {
@@ -25,6 +23,16 @@ namespace raft {
  * accessor functions can then register and load resources as needed in order
  * to keep its usage somewhat opaque to end-users.
  *
+ * Copies of a resources handle share the underlying resource_cell objects.
+ * Lazy initialization (via get_resource / ensure_default_factory) stores into the
+ * shared cell's atomics, so all copies see the update.  Explicit modification
+ * (via add_resource_factory) replaces the shared_ptr<resource_cell> in the local
+ * vector, isolating the change from other copies.
+ *
+ * Thread safety: concurrent const operations on the same handle are safe
+ * (inner-cell atomics).  Concurrent const + non-const on the same handle
+ * requires external synchronization (standard C++ rules).
+ *
  * @code{.cpp}
  * #include <raft/core/resources.hpp>
  * #include <raft/core/resource/cuda_stream.hpp>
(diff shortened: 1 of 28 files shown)


cpp/tests/core/handle.cpp
```diff
@@ -342,4 +342,74 @@ TEST(Raft, HandleAssign)
...
+
+TEST(Raft, ResourcesDefaultMovable)
+{
+  raft::resources a;
+  resource::get_cuda_stream(a);
+
+  // Move construct
+  raft::resources b(std::move(a));
+  ASSERT_TRUE(b.has_resource_factory(resource::resource_type::CUDA_STREAM_VIEW));
+
```
jamxia155, line 408: Would it make sense to have another block like
```
// check moved-from object
ASSERT_FALSE(a.has_resource_factory(resource::resource_type::CUDA_STREAM_VIEW));
```
and similarly, after b has been moved into c?

My concern is that if `raft::resources` lacked move constructor and assignment altogether, this test would still pass as the construction and assignment can fall back to copying.

cpp/include/raft/core/resources.hpp
```diff
@@ -71,59 +66,76 @@ class resources {
...
+    return cells_[resource_type]->factory.load() != nullptr;
   }
 
   /**
-   * @brief Register a resource_factory with the current instance.
-   * This will overwrite any existing resource factories.
+   * @brief Register a resource_factory with the current instance (explicit set).
+   *
+   * Creates a new resource_cell with the given factory.  Other copies of this
+   * handle continue to point at the old cell, so the change does not propagate.
```
jamxia155, line 77: Would it make sense to enforce that `add_resource_factory` is only used with copies of a handle? Perhaps we could add a flag `is_copy` in `resources` that is default-initialized to false but gets set to true by the copy constructor, then add a `RAFT_EXPECTS(is_copy)` in `add_resource_factory`. This way, there's no risk of accidentally overwriting the resources in the original handle:
- `ensure_resource_factory` for when you want to update the original handle (works on either original or copy).
- `add_resource_factory` for when you want to make an isolated change to a _copy_ of a handle (does not work on original).

achirkin: I think we shouldn't distinguish between the "original" and "copied" handles - it doesn't really matter who's the first. Also it's not possible to overwrite a resource from one copy to another: `add_resource_factory` simply replaces the outer `shared_ptr` to a one that points to a new `resource_cell` (NB: copy constructor of the `raft::resources` copies the contents of the `cells_` vector same as before; the outer shared_ptrs are not shared among resources handles).

cpp/tests/core/handle.cpp
achirkin: Hmm, moved-from should have valid but unspecified state, so technically we cannot rely on ASSERT_FALSE. Since we know our implementation is `std::vector<std::shared_ptr>`, it is indeed false (the pointers become empty after move), but... does it really matter for us?

jamxia155: I just thought it would ensure that a move, not a copy, has in fact taken place.

cpp/include/raft/core/resources.hpp
jamxia155: Okay it's probably an overreach to make that distinction. Thanks for elaborating.

achirkin:
Because of requiring C++20, this PR is blocked on completing the migration of cuVS to C++20 (see PRs [#1795](https://github.com/rapidsai/cuvs/pull/1795), [#1796](https://github.com/rapidsai/cuvs/pull/1796), [#1799](https://github.com/rapidsai/cuvs/pull/1799)).

tarang-jain:
Note that https://github.com/rapidsai/cuvs/pull/1796 is blocked. https://nvbugspro.nvidia.com/bug/5356084 is the existing issue for the 12.9 compiler issue. Nothing we can do as the fix was merged into 13.1.

achirkin:
The issues with C++20 in cuVS are resolved, cuVS CI passes against this PR!
cuml and cugraph PRs have some failures but I don't see how are they connected to my change.

achirkin:
Update: due to recent changes in cuvs, https://github.com/rapidsai/cuvs/pull/2103 needs to be merged ahead of this PR.

achirkin:
All cuvs/cuml/cugraph C++ tests pass; I cannot quite get the python tests in cuml and cugraph to pass because they use conda-shipped version of pylibraft (and I already messed up with cuml/cugraph build recipes to make it use my cuvs-test-branch).
@cjnolet shall we try to merge this? I don't foresee any problems, but I cannot guarantee that with 100% :)

cjnolet (approved):

mlubin:
This likely broke the cuOpt build: https://github.com/NVIDIA/cuopt/actions/runs/26093548396/job/76725507023?pr=1240#step:11:2065

PR #3006, opened 2026-04-29
Fix symbol export

vyasr:
This PR adds symbol visibility controls to raft to avoid exporting weak symbols that it shouldn't.

Contributes to https://github.com/rapidsai/build-infra/issues/53

### DIFF
--- a/cpp/include/raft/common/detail/scatter.cuh
+++ b/cpp/include/raft/common/detail/scatter.cuh
@@ -1,14 +1,16 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2019-2023, NVIDIA CORPORATION.
+ * SPDX-FileCopyrightText: Copyright (c) 2019-2026, NVIDIA CORPORATION.
  * SPDX-License-Identifier: Apache-2.0
  */
 
 #pragma once
 
+#include <raft/core/detail/macros.hpp>
 #include <raft/util/cuda_utils.cuh>
 #include <raft/util/vectorized.cuh>
 
-namespace raft::detail {
+namespace RAFT_EXPORT raft {
+namespace detail {
 
 template <typename DataT, int VecLen, typename Lambda, typename IdxT>
 RAFT_KERNEL scatterKernel(DataT* out, const DataT* in, const IdxT* idx, IdxT len, Lambda op)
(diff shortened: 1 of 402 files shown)


cpp/include/raft/core/detail/macros.hpp
```diff
@@ -75,14 +75,22 @@
...
+// Macros for controlling symbol visibility in shared libraries.
+// When CXX_VISIBILITY_PRESET is set to hidden (the default for raft_objs),
+// only symbols explicitly marked with RAFT_EXPORT will be visible in the
+// shared library. RAFT_HIDDEN can be used to explicitly mark symbols as
+// hidden (redundant with the default, but useful for clarity).
 //
 // Hidden visibility also ensures that the linker doesn't de-duplicate the
 // symbol across multiple `.so`. This allows multiple libraries to embed raft
-// without issue
-#define RAFT_HIDDEN_FUNCTION __attribute__((visibility("hidden")))
```
robertmaynard, line 85: We need to make `RAFT_WEAK_FUNCTION` consistent with these new names.

I am fine having macros with and without the `FUNCTION` postfix. So in total we would have:
```
RAFT_WEAK_FUNCTION
RAFT_HIDDEN_FUNCTION
RAFT_EXPORT_FUNCTION

RAFT_EXPORT
RAFT_HIDDEN
```

robertmaynard (dismissed):
I expect we need all installed RAFT headers to have the markup of `RAFT_EXPORT` otherwise you can't pass a raft handle from a library that use raft into raft itself.

cpp/include/raft/common/detail/scatter.cuh
```diff
@@ -1,14 +1,16 @@
...
  */
 
 #pragma once
 
+#include <raft/core/detail/macros.hpp>
 #include <raft/util/cuda_utils.cuh>
 #include <raft/util/vectorized.cuh>
 
-namespace raft::detail {
+namespace RAFT_EXPORT raft {
```
csadorf, line 12: Technically the `raft::detail` namespace is not considered part of the public API. I think we should roll this out in stages though and maybe drop that in a follow-up to minimize discruption.

divyegala (approved):

PR #3007, opened 2026-04-30
[REVIEW] Drop extra copy in `CudaRuntimeError`

jakirkham:
In `CudaRuntimeError`'s `__init__`, the assignment of `const char*` to `bytes` results in a copy as `bytes` must own its memory. After that the `bytes` object is `decode`d to a Python `str`. However it is possible to `decode` directly from `const char*` to `str` avoiding the copy to the temporary `bytes` object. Hence this code makes that change.


aamijar (approved):

divyegala (approved):

jakirkham:
Thanks Divye and Anupam! 🙏

PR #3012, opened 2026-05-08
Fix eigDC version check for syevd

lowener:
Previous work on that bug:
- #2332
- #2430

Currently the version of cudart is check at runtime instead of checking for cusolver version. This PR solves that bug that could arise in the cases where CUDA version is superior to 12.5 but the cusolver version used was older.


bdice (approved):

achirkin (approved):
Thanks!

PR #3019, opened 2026-05-16
fix(visibility): restrict RAFT_EXPORT to core/ and raft_runtime/ only

vyasr:
Restrict symbol exports from libraft.so to only the public API surface:

- **`include/raft/core/`** (non-detail): exported via `namespace RAFT_EXPORT raft`
- **`include/raft_runtime/`**: exported via `RAFT_EXPORT` on function declarations (lanczos_solver, rmat_rectangular_gen)
- **Everything else** (linalg, matrix, random, sparse, spectral, stats, util, label, solver, comms, mr, pmr, and all detail/ namespaces): hidden

With `CXX_VISIBILITY_PRESET hidden` on `raft_objs`, this ensures only the intended public symbols are visible in the shared library, preventing symbol interposition between libcuml.so and libcuvs.so for internal raft implementation details.

### Changes

1. **Remove `RAFT_EXPORT` from detail namespaces** — prevents internal detail symbols from being exported
2. **Remove `RAFT_EXPORT` from all non-core directories** — linalg, matrix, random, sparse, spectral, stats, util, label, solver, comms, mr, pmr no longer export symbols
3. **Remove `RAFT_EXPORT` from `raft_runtime::matrix::select_k`** — this function has no `.cu` source in `src/` and is not compiled into libraft.so (it's instantiated by downstream consumers like cuvs)

### Verification

- libraft.so builds and links cleanly
- cuvs compiles successfully against these changes (tested distance, brute_force, cagra, select_k, sparse_distance translation units)

Contributes to https://github.com/rapidsai/build-infra/issues/53

### DIFF
--- a/cpp/include/raft/common/detail/scatter.cuh
+++ b/cpp/include/raft/common/detail/scatter.cuh
@@ -9,7 +9,7 @@
 #include <raft/util/cuda_utils.cuh>
 #include <raft/util/vectorized.cuh>
 
-namespace RAFT_EXPORT raft {
+namespace raft {
 namespace detail {
 
 template <typename DataT, int VecLen, typename Lambda, typename IdxT>
@@ -41,4 +41,4 @@ void scatterImpl(
 }
 
 }  // namespace detail
-}  // namespace RAFT_EXPORT raft
+}  // namespace raft
--- a/cpp/include/raft/comms/comms_test.hpp
+++ b/cpp/include/raft/comms/comms_test.hpp
@@ -10,7 +10,7 @@
 #include <raft/core/detail/macros.hpp>
 #include <raft/core/resources.hpp>
 
-namespace RAFT_EXPORT raft {
+namespace raft {
 namespace comms {
 
 /**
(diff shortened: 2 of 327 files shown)


aamijar:
So we are reverting some of the changes from an earlier PR right? https://github.com/rapidsai/raft/pull/3006

vyasr:
> So we are reverting some of the changes from an earlier PR right? #3006

Sort of. The core change in #3006 is [hiding symbols by default](https://github.com/rapidsai/raft/blob/4cbbf00f9629f1f254e6cf2d7ec1e21b669ce764/cpp/CMakeLists.txt#L296). That behavior has not changed. However, in that first PR I was more conservative in what APIs were actually hidden by exporting lots of symbols. This PR is effectively going further by removing a lot of the explicit exports, which means that we'll be hiding even more symbols. In particular, we have removed export from all header-only parts of raft except for the core/ directory, which contains the types that are part of public APIs in other libraries and therefore need public visibility.

divyegala (approved):

PR #3034, opened 2026-05-21
Add sparse Lanczos SVD solver

Intron7:
As discussed previously with @cjnolet I'm also adding my Lanczos SVD solver for sparse CSR matrices. 

  This is the more precise sparse SVD path next to the existing randomized solver. The solver repeatedly applies `A @ v` and `A.T @ u` to build Krylov bases, computes the SVD of the small bidiagonal problem, uses the resulting Ritz vectors to identify converged singular triplets, locks those vectors, and restarts on the remaining unconverged part. It also uses full reorthogonalization and a final `A @ V` refinement step to improve singular values and left singular vectors.

Compared with randomized SVD, this is aimed at quality: clustered spectra, slow singular-value decay, near-rank-deficient inputs, and PCA workloads where ARPACK-like accuracy matters.

Ran the #2999 -style row sweep on the same singlecell dataset, k=50, n_oversamples=10, n_power_iters=2, best-of-3 GPU timings. GPU: RTX PRO 6000 Blackwell.

```


  ┌──────┬───────┬─────────────────┬──────────────┬─────────────────────────┬─────────────────────┬──────────────────┐
  │ rows │   nnz │ raft randomized │ raft Lanczos │          Lanczos / rand │ randomized residual │ Lanczos residual │
  ├──────┼───────┼─────────────────┼──────────────┼─────────────────────────┼─────────────────────┼──────────────────┤
  │  50k │  101M │          0.180s │       0.252s │            1.40x slower │            3.11e-02 │         1.55e-07 │
  │ 200k │  400M │          0.698s │       1.328s │            1.90x slower │            3.12e-02 │         1.57e-07 │
  │ 500k │ 1.02B │          1.776s │       1.763s │          basically tied │            3.06e-02 │         1.57e-07 │
  │ 982k │ 2.01B │          3.536s │       3.423s │ Lanczos slightly faster │            3.09e-02 │         1.56e-07 │
  └──────┴───────┴─────────────────┴──────────────┴─────────────────────────┴─────────────────────┴──────────────────┘
```
  Using the 2999 CPU baselines for context:
```
  ┌──────┬─────────────┬──────────────┬─────────────────────────────┬──────────────────────────┐
  │ rows │ sklearn CPU │ scipy ARPACK │ randomized speedup vs scipy │ Lanczos speedup vs scipy │
  ├──────┼─────────────┼──────────────┼─────────────────────────────┼──────────────────────────┤
  │  50k │       7.38s │       18.69s │                        104x │                      74x │
  │ 200k │      25.61s │       62.49s │                         90x │                      47x │
  │ 500k │      64.97s │      153.40s │                         86x │                      87x │
  │ 982k │     126.16s │      307.04s │                         87x │                      90x │
  └──────┴─────────────┴──────────────┴─────────────────────────────┴──────────────────────────┘
```


aamijar:
Hi @Intron7, I haven't looked at this PR closely yet, but one thing to think about is that we should try and reuse parts of the existing lanczos eigensolver as much as possible. They should have some things in common right?

Intron7:
@aamijar the algorithm is different even is the name is similar. The two paths share the Lanczos name but the kernels are different algorithms: the eigensolver builds a symmetric tridiagonal via a one-vector recurrence and ritz-solves with syevd; the SVD builds a bidiagonal via Golub-Kahan with two coupled bases (A @ v and Aᵀ @ u) and ritz-solves with gesvdj. Restart is also different, the SVD path locks converged singular triplets and restarts on the unconverged subspace.
The realistic shared surface is the reorthogonalization helpers (CGS2/MGS2) and the cublas wrapper calls. The existing lanczos.cuh does its reorthogonalization inline against its own single-vector layout, so factoring CGS2/MGS2 into a shared utility would require touching the existing eigensolver too. I'd rather land this PR as-is and do a separate refactor PR to extract a shared bidiag_reorth /lanczos_reorth utility if you want.

cpp/include/raft/sparse/solver/solver_types.hpp
```diff
@@ -39,4 +39,48 @@ struct sparse_svd_config {
 
 /** @} */
 
+/**
+ * @addtogroup sparse_lanczos_svd
+ * @{
+ */
```
cjnolet, line 45: To keep to our conventions, can you put this in a file called solver_types.hpp? Create it if it oesn't exist alredy.

cpp/include/raft/sparse/solver/lanczos_svds.cuh
```diff
@@ -0,0 +1,104 @@
...
+ * @param[out] U optional output left singular vectors of shape (m, n_components), col-major.
+ *             Pass `std::nullopt` to skip storing U.
+ * @param[out] Vt optional output right singular vectors of shape (n_components, n), col-major.
+ *             Pass `std::nullopt` to skip storing Vt.
+ */
+template <typename ValueTypeT, typename NNZTypeT>
+void sparse_lanczos_svd(
+  raft::resources const& handle,
+  sparse_lanczos_svd_config<ValueTypeT> const& config,
+  raft::device_csr_matrix_view<const ValueTypeT, int, int, NNZTypeT> A,
```
cjnolet, line 106: Using our c++ sparse APIs. Very nice!

divyegala:
@coderabbitai review

cpp/include/raft/sparse/solver/lanczos_svds.cuh
```diff
@@ -0,0 +1,104 @@
...
+
+#include <raft/core/device_csr_matrix.hpp>
+#include <raft/core/device_mdspan.hpp>
+#include <raft/core/resources.hpp>
+#include <raft/sparse/solver/detail/csr_linear_operator.cuh>
+#include <raft/sparse/solver/detail/lanczos_svds.cuh>
+#include <raft/sparse/solver/detail/svds_optional.hpp>
+
+#include <optional>
+
```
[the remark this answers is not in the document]
Intron7: I dont see this done anywhere else.

cpp/include/raft/sparse/solver/detail/lanczos_svds.cuh
```diff
@@ -0,0 +1,767 @@
...
+      {
+        common::nvtx::range<common::nvtx::domain::raft> subtract_scope(
+          "lanczos_svds::mgs2_subtract");
+        RAFT_CUBLAS_TRY(raft::linalg::detail::cublasaxpy(
+          cublas_handle, n_rows, coeff, q_j, 1, target, 1, stream));
+      }
+    }
+  }
+  RAFT_CUBLAS_TRY(cublasSetPointerMode(cublas_handle, CUBLAS_POINTER_MODE_HOST));
+}
```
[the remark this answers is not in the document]
Intron7: done

cpp/include/raft/sparse/solver/lanczos_svds.cuh
```diff
@@ -0,0 +1,104 @@
...
+ * @param[in] handle raft resources handle
+ * @param[in] config SVD configuration parameters
+ * @param[in] op linear operator representing the matrix to decompose
+ * @param[out] singular_values output singular values of shape (n_components,) in descending order
+ * @param[out] U optional output left singular vectors of shape (m, n_components), col-major.
+ *             Pass `std::nullopt` to skip storing U.
+ * @param[out] Vt optional output right singular vectors of shape (n_components, n), col-major.
+ *             Pass `std::nullopt` to skip storing Vt.
+ */
+template <typename ValueTypeT, typename OperatorT>
```
divyegala, line 64: Since `sparse_lanczos_svd` exposes a generic `OperatorT` overload, could we add a small test using a non-CSR operator? The current tests exercise the CSR convenience overload well, but this would protect the advertised implicit-operator API too.

cpp/include/raft/sparse/solver/lanczos_svds.cuh
```diff
@@ -0,0 +1,104 @@
...
+void sparse_lanczos_svd(
```
divyegala, line 65: The generic overload is matrix-format agnostic: it only requires `apply` and `apply_transpose`. Would it fit better as `raft::linalg::lanczos_svd` (dropping the `sparse_` prefix as it is implied either from namespace or input type), with `raft::sparse::solver::lanczos_svd` retained as the CSR convenience wrapper? That would avoid labeling the general operator path as sparse, while keeping the CSR API discoverable.

cpp/include/raft/sparse/solver/detail/lanczos_svds.cuh
```diff
@@ -0,0 +1,766 @@
...
+
+template <typename ValueTypeT>
+ValueTypeT vector_norm(raft::resources const& handle, ValueTypeT const* x, int n)
+{
+  common::nvtx::range<common::nvtx::domain::raft> scope("lanczos_svds::vector_norm");
+  auto cublas_handle = resource::get_cublas_handle(handle);
+  auto stream        = resource::get_cuda_stream(handle);
+  ValueTypeT result{};
+  RAFT_CUBLAS_TRY(raft::linalg::detail::cublasnrm2(cublas_handle, n, x, 1, &result, stream));
```
divyegala: Could this use public `raft::linalg::norm` instead of the detail-level cuBLAS wrapper?

cpp/include/raft/sparse/solver/detail/lanczos_svds.cuh
```diff
@@ -0,0 +1,766 @@
...
+  return result;
+}
+
+template <typename ValueTypeT>
+void scale_vector(raft::resources const& handle, ValueTypeT* x, int n, ValueTypeT alpha)
+{
+  common::nvtx::range<common::nvtx::domain::raft> scope("lanczos_svds::scale_vector");
+  auto cublas_handle = resource::get_cublas_handle(handle);
+  auto stream        = resource::get_cuda_stream(handle);
+  RAFT_CUBLAS_TRY(raft::linalg::detail::cublasscal(cublas_handle, n, &alpha, x, 1, stream));
```
divyegala: Could this use public `raft::linalg::multiply_scalar` instead of the detail-level cuBLAS wrapper?

cpp/include/raft/sparse/solver/detail/lanczos_svds.cuh
```diff
@@ -0,0 +1,766 @@
...
+}
+
+template <typename ValueTypeT>
+void axpy(
+  raft::resources const& handle, int n, ValueTypeT alpha, ValueTypeT const* x, ValueTypeT* y)
+{
+  common::nvtx::range<common::nvtx::domain::raft> scope("lanczos_svds::axpy");
+  auto cublas_handle = resource::get_cublas_handle(handle);
+  auto stream        = resource::get_cuda_stream(handle);
+  RAFT_CUBLAS_TRY(raft::linalg::detail::cublasaxpy(cublas_handle, n, &alpha, x, 1, y, 1, stream));
```
divyegala: Could this use public `raft::linalg::axpy` instead of the detail-level cuBLAS wrapper?

cpp/include/raft/sparse/solver/detail/lanczos_svds.cuh
```diff
@@ -0,0 +1,766 @@
...
+  auto cublas_handle = resource::get_cublas_handle(handle);
+  auto stream        = resource::get_cuda_stream(handle);
+
+  ValueTypeT one     = ValueTypeT(1);
+  ValueTypeT zero    = ValueTypeT(0);
+  ValueTypeT neg_one = ValueTypeT(-1);
+  for (int pass = 0; pass < 2; ++pass) {
+    {
+      common::nvtx::range<common::nvtx::domain::raft> project_scope("lanczos_svds::cgs2_project");
+      RAFT_CUBLAS_TRY(raft::linalg::detail::cublasgemv(cublas_handle,
```
divyegala: Could this use public `raft::linalg::gemv` instead of the detail-level cuBLAS wrapper?

cpp/include/raft/sparse/solver/detail/lanczos_svds.cuh
```diff
@@ -0,0 +1,766 @@
...
+                                                       target,
+                                                       1,
+                                                       &zero,
+                                                       coeffs,
+                                                       1,
+                                                       stream));
+    }
+    {
+      common::nvtx::range<common::nvtx::domain::raft> subtract_scope("lanczos_svds::cgs2_subtract");
+      RAFT_CUBLAS_TRY(raft::linalg::detail::cublasgemv(cublas_handle,
```
divyegala: Could this use public `raft::linalg::gemv` instead of the detail-level cuBLAS wrapper?

cpp/include/raft/sparse/solver/detail/lanczos_svds.cuh
```diff
@@ -0,0 +1,766 @@
...
+                        int n_valid,
+                        ValueTypeT* coeffs)
+{
+  common::nvtx::range<common::nvtx::domain::raft> scope("lanczos_svds::mgs2_orthogonalize");
+  if (n_valid <= 0) { return; }
+
+  auto cublas_handle = resource::get_cublas_handle(handle);
+  auto stream        = resource::get_cuda_stream(handle);
+
+  raft::linalg::detail::cublas_device_pointer_mode<true> pointer_mode_guard(cublas_handle);
```
divyegala: Could we use the public device-scalar `raft::linalg::dot` and `raft::linalg::axpy` APIs here and remove the manual pointer-mode guard?

cpp/include/raft/sparse/solver/detail/lanczos_svds.cuh
```diff
@@ -0,0 +1,766 @@
...
+  for (int pass = 0; pass < 2; ++pass) {
+    for (int j = 0; j < n_valid; ++j) {
+      auto const* q_j = basis + static_cast<std::size_t>(j) * n_rows;
+      auto* coeff     = coeffs + j;
+      {
+        common::nvtx::range<common::nvtx::domain::raft> dot_scope("lanczos_svds::mgs2_dot");
+        RAFT_CUBLAS_TRY(
+          raft::linalg::detail::cublasdot(cublas_handle, n_rows, q_j, 1, target, 1, coeff, stream));
```
divyegala: Could this use public `raft::linalg::dot` instead of the detail-level cuBLAS wrapper?

cpp/include/raft/sparse/solver/detail/lanczos_svds.cuh
```diff
@@ -0,0 +1,766 @@
...
+      }
+      {
+        common::nvtx::range<common::nvtx::domain::raft> negate_scope("lanczos_svds::mgs2_negate");
+        negate_scalar_kernel<<<1, 1, 0, stream>>>(coeff);
+        RAFT_CUDA_TRY(cudaPeekAtLastError());
+      }
+      {
+        common::nvtx::range<common::nvtx::domain::raft> subtract_scope(
+          "lanczos_svds::mgs2_subtract");
+        RAFT_CUBLAS_TRY(raft::linalg::detail::cublasaxpy(
```
divyegala: Could this use public `raft::linalg::axpy` instead of the detail-level cuBLAS wrapper?

cpp/include/raft/sparse/solver/detail/lanczos_svds.cuh
```diff
@@ -0,0 +1,766 @@
...
+                           raft::random::RngState& rng_state,
+                           ValueTypeT* ortho_coeffs,
+                           bool use_mgs2,
+                           std::vector<ValueTypeT>& alphas,
+                           std::vector<ValueTypeT>& betas)
+{
+  common::nvtx::range<common::nvtx::domain::raft> scope("lanczos_svds::bidiagonalize");
+  int m          = op.rows();
+  int n          = op.cols();
+  ValueTypeT eps = ValueTypeT(1e-9);
```
divyegala: [AI-reviewed] The fixed `1e-9` breakdown threshold makes the solver non-scale-invariant: sufficiently small but nonzero matrices can be treated as zero. Could this criterion be made scale- and precision-aware?

cpp/include/raft/sparse/solver/detail/lanczos_svds.cuh
```diff
@@ -0,0 +1,766 @@
...
+  auto stream = resource::get_cuda_stream(handle);
+  auto d_coeffs =
+    raft::make_device_vector<ValueTypeT, uint32_t>(handle, static_cast<uint32_t>(ncv));
+  raft::update_device(d_coeffs.data_handle(), coeffs.data(), ncv, stream);
+
+  auto cublas_handle   = resource::get_cublas_handle(handle);
+  ValueTypeT const one = ValueTypeT(1);
+  ValueTypeT zero      = ValueTypeT(0);
+  auto* V_segment      = V_full.data_handle() + static_cast<std::size_t>(n_locked) * n;
+  RAFT_CUBLAS_TRY(raft::linalg::detail::cublasgemv(cublas_handle,
```
divyegala: Could this use public `raft::linalg::gemv` instead of the detail-level cuBLAS wrapper?

cpp/include/raft/sparse/solver/detail/lanczos_svds.cuh
```diff
@@ -0,0 +1,766 @@
...
+    ValueTypeT max_s     = h_s.empty() || h_s[0] <= ValueTypeT(0) ? ValueTypeT(1) : h_s[0];
+    int check_components = std::min(remaining + 2, active_ncv);
+
+    std::vector<int> selected;
+    {
+      common::nvtx::range<common::nvtx::domain::raft> scope("lanczos_svds::select_converged");
+      for (int i = 0; i < check_components; ++i) {
+        auto residual = betas[active_ncv] *
+                        std::abs(h_P[static_cast<std::size_t>(i) * active_ncv + active_ncv - 1]);
+        if (residual < config.tolerance * max_s) { selected.push_back(i); }
```
divyegala: [AI-reviewed] `tolerance = 0` is accepted, but the strict `<` comparison means even an exact zero residual can never converge. Is that intentional? If zero tolerance is supported, should this be `<=`?

cpp/include/raft/sparse/solver/detail/lanczos_svds.cuh
```diff
@@ -0,0 +1,766 @@
...
+          ValueTypeT sum = 0;
+          for (int row = 0; row < k_mix; ++row) {
+            sum += h_Qt[static_cast<std::size_t>(col) * active_ncv + row];
+          }
+          restart_coeffs[col] = sum;
+        }
+      }
+    }
+
+    compute_restart_vector(
```
divyegala, line 779: [AI-reviewed] `restart_coeffs` are coordinates in the active Krylov `V` basis starting at the pre-lock `n_locked`, but `compute_ritz_vectors` overwrites that segment and `n_locked` is incremented before `compute_restart_vector` runs. On partial convergence this forms the restart vector from a shifted/overwritten basis. Could we compute and preserve the restart vector before locking, or retain the original active basis until after restart construction?

cpp/include/raft/sparse/solver/lanczos_svds.cuh
divyegala, line 106: The generic API hard-codes uint32_t for all output and temporary views, while the operator shape and CSR row/column indices are int. Is signed 32-bit indexing an intentional constraint? If so, could we document and validate it explicitly; otherwise, should the index type be templated as in the existing sparse Lanczos eigensolver?

cpp/include/raft/sparse/solver/svds_config.hpp
divyegala, line 1: `svds_config.hpp` was a public header. Removing it may break downstream code that includes it directly. We could keep it as a forwarding header to `solver_types.hpp` and deprecate it later, unless you have already verified that this header is not included downstream in cuVS or cuML?

divyegala (commented):
Thank you for the PR! For the logical bits, I used AI's help and tagged those as `[AI-reviewed]`. If you find the comments invalid, please feel free to discard them.

cpp/include/raft/sparse/solver/lanczos_svds.cuh
Intron7: Done. Added `LanczosGenericOperator{F,D}.DenseOperator` which drives the generic overload with a dense col-major operator implemented via `raft::linalg::gemm` (only `rows()/cols()/apply/apply_transpose`, no CSR involved). The test also covers the `std::nullopt` U/Vt path and checks the singular values match the full computation.

cpp/include/raft/sparse/solver/lanczos_svds.cuh
Intron7: I'd prefer to keep both overloads in `raft::sparse::solver` for now: `sparse_randomized_svd` (#2999) exposes the exact same operator concept from the same namespace, and the two solvers are deliberate siblings (shared `csr_linear_operator`, shared config header, shared runtime/pylibraft plumbing). Moving only the Lanczos generic overload to `raft::linalg` would split that concept across namespaces. If we want a `raft::linalg::lanczos_svd`, I think the right shape is a follow-up that promotes the operator concept and both solvers' generic overloads together.

cpp/include/raft/sparse/solver/detail/lanczos_svds.cuh
Intron7: Done

cpp/include/raft/sparse/solver/detail/lanczos_svds.cuh
Intron7: Done

cpp/include/raft/sparse/solver/detail/lanczos_svds.cuh
Intron7: Done

cpp/include/raft/sparse/solver/detail/lanczos_svds.cuh
Intron7: Done

cpp/include/raft/sparse/solver/detail/lanczos_svds.cuh
Intron7: Done

cpp/include/raft/sparse/solver/detail/lanczos_svds.cuh
Intron7: Done

cpp/include/raft/sparse/solver/detail/lanczos_svds.cuh
Intron7: Done

cpp/include/raft/sparse/solver/detail/lanczos_svds.cuh
Intron7: Done

cpp/include/raft/sparse/solver/detail/lanczos_svds.cuh
Intron7: Done

cpp/include/raft/sparse/solver/svds_config.hpp
Intron7: Fair point — restored `svds_config.hpp` as a forwarding header that includes `solver_types.hpp` and emits a `#pragma message` deprecation warning, so any downstream includes keep compiling while we migrate them.

cpp/include/raft/sparse/solver/detail/lanczos_svds.cuh
Intron7: You're right, changed to `<=` so an exactly-zero residual converges under`tolerance = 0`.

cpp/include/raft/sparse/solver/detail/lanczos_svds.cuh
Intron7: It's an intentional constraint, matching sparse_randomized_svd: the operator reports its shape as int (the CSR path is additionally bound to int indptr/indices by the cuSPARSE SpMM wrapper), and all device views use uint32_t extents. I've documented this explicitly on both overloads (m, n < 2^31). Templating the index type like the existing eigensolver could be a follow-up if a >2^31 use case shows up, but I'd rather keep the two SVD solvers' APIs consistent for now.

cpp/include/raft/sparse/solver/lanczos_svds.cuh
Intron7: It's an intentional constraint, matching `sparse_randomized_svd`: the operator reports its shape as `int` (the CSR path is additionally bound to `int` indptr/indices by the cuSPARSE SpMM wrapper), and all device views use `uint32_t` extents. I've documented this explicitly on both overloads (m, n < 2^31). Templating the index type like the existing eigensolver could be a follow-up if a >2^31 use case shows up, but I'd rather keep the two SVD solvers' APIs consistent for now.

cpp/include/raft/sparse/solver/detail/lanczos_svds.cuh
Intron7: Good catch — the shifted/overwritten basis read is real. I tested constructing the restart vector from the intact pre-lock basis, but it caused silent correctness regressions: even with the shipped k + 10 ncv floor, the single-direction version returned the wrong top-k spectrum on 64/128 seeds of a geometric spectrum, while the current restart was correct on 128/128. A mixed pre-lock restart also failed cases that the current form passed.

The investigation did uncover a missing detail from rapids-singlecell: it allocates U_full and V_full with cp.zeros. When d > 1 Ritz vectors are locked together, the shifted restart window extends d - 1 columns beyond the basis written by the current bidiagonalization. RSC therefore reads deterministic zeros there, whereas RAFT’s uninitialized allocation could read arbitrary pooled-memory contents. I’ve restored that zero-initialization while retaining the restart behavior that performed best and matches RSC.

The sweep also exposed a separate porting regression in ncv: rapids-singlecell enforces ncv = max(ncv, k + 10) even for user-supplied values and uses a 2.5 * k default for the large-matrix/small-k branch. RAFT had reduced these to max(ncv, k) and approximately 2.2 * k. I restored both reference settings and documented the effective floor. With that floor, the current restart completed every tested production-reachable family correctly: 128/128 geometric, 64/64 linear, 64/64 repeated-spectrum, and 32/32 larger-k runs. This does not itself fix the shifted-tail read—the zero-initialization does—but it restores the convergence margin assumed by the reference algorithm.

divyegala: This is my agent's response:
```
However, this does not mathematically establish that the shifted restart basis is correct—it preserves the reference implementation’s behavior. I would accept it provisionally, but ask for a committed regression test that forces partial multi-vector locking and verifies the returned top-k spectrum/residuals.
```

divyegala (approved):
LGTM, thanks so much! (pending one response from my agent on AI reviewed comment)

PR #3036, opened 2026-05-26
Native row-major PCA

aamijar:
Currently PCA only supports col-major which is not ideal for consumers passing in row-major datasets.
This PR exposes row-major APIs for PCA and calls the corresponding row-major raft primitives based on templated row or col layout.

This PR is backward compatible with cuml and cuvs.


cpp/include/raft/linalg/detail/pca.cuh
```diff
@@ -134,19 +146,30 @@ void pca_fit(raft::resources const& handle,
...
 
+  // The eigendecomposition of the (symmetric) covariance matrix naturally produces a
+  // col-major components buffer. For row-major output we accumulate into a temporary
+  // and physically transpose at the end.
+  auto components_col_storage = raft::make_device_matrix<math_t, idx_t, raft::col_major>(
+    handle, input_row_major ? n_components : idx_t(0), input_row_major ? n_cols : idx_t(0));
+  math_t* components_col_data =
+    input_row_major ? components_col_storage.data_handle() : components.data_handle();
+  auto components_col_view = raft::make_device_matrix_view<math_t, idx_t, raft::col_major>(
+    components_col_data, n_components, n_cols);
```
divyegala: I found this confusing to read:



```suggestion
  std::optional<raft::device_matrix<math_t, idx_t, raft::col_major>> components_col_storage;
   if constexpr (input_row_major) {
     components_col_storage = raft::make_device_matrix_view<...)(...);
   }
  math_t* components_col_data =
    input_row_major ? components_col_storage->data_handle() : components.data_handle();
  auto components_col_view = raft::make_device_matrix_view<math_t, idx_t, raft::col_major>(
    components_col_data, n_components, n_cols);
```

cpp/include/raft/linalg/detail/pca.cuh
```diff
@@ -200,49 +244,69 @@ void pca_transform(raft::resources const& handle,
   rmm::device_uvector<math_t> components_copy{components_len, stream};
   raft::copy(components_copy.data(), components.data_handle(), components_len, stream);
 
+  auto components_copy_view = raft::make_device_matrix_view<math_t, idx_t, LayoutPolicy>(
+    components_copy.data(), n_components, n_cols);
```
divyegala: Why explicitly here instead of inline in the diff before this PR?

cpp/include/raft/linalg/detail/pca.cuh
```diff
@@ -257,49 +321,55 @@ void pca_inverse_transform(raft::resources const& handle,
...
```
divyegala: Same as above

cpp/tests/linalg/pca.cu
```diff
@@ -332,4 +332,115 @@ INSTANTIATE_TEST_CASE_P(PcaTests, PcaTestDataVecF, ::testing::ValuesIn(inputsf2)
...
+    std::vector<T> host_row(n_rows * n_cols);
+    raft::update_host(host_col.data(), col_major_src, n_rows * n_cols, stream);
+    RAFT_CUDA_TRY(cudaStreamSynchronize(stream));
+    for (int i = 0; i < n_rows; ++i) {
+      for (int j = 0; j < n_cols; ++j) {
+        host_row[i * n_cols + j] = host_col[j * n_rows + i];
+      }
+    }
+    raft::update_device(row_major_dst, host_row.data(), n_rows * n_cols, stream);
+  }
```
divyegala: Why is this needed? You're generating 1-dimensional random data. It can be interpreted as row or col major.

cpp/include/raft/linalg/detail/pca.cuh
```diff
@@ -134,19 +146,30 @@ void pca_fit(raft::resources const& handle,
...
+    mu.data_handle(), input.data_handle(), n_cols, n_rows, false, stream);
 
   auto len = static_cast<std::size_t>(n_cols * n_cols);
   rmm::device_uvector<math_t> cov(len, stream);
 
-  raft::stats::cov<false>(
+  raft::stats::cov<input_row_major>(
     handle, cov.data(), input.data_handle(), mu.data_handle(), n_cols, n_rows, true, true, stream);
 
+  // The eigendecomposition of the (symmetric) covariance matrix naturally produces a
```
cjnolet: But since covar matrix is symmetric, it shouldn't matter if it's column or row layout, right? Why do we need to transpose at all?

aamijar:
Hi, all the comments should be addressed now in the recent batch of commits.

cpp/include/raft/linalg/detail/pca.cuh
```diff
@@ -49,19 +55,26 @@ void trunc_comp_exp_vars(raft::resources const& handle,
...
+    raft::make_device_matrix_view<math_t, idx_t, LayoutPolicy>(
       components_all.data(), n_cols, n_cols),
     raft::make_device_vector_view<math_t, idx_t>(explained_var_all.data(), n_cols));
-  raft::matrix::trunc_zero_origin(
-    handle,
-    raft::make_device_matrix_view<const math_t, idx_t, raft::col_major>(
-      components_all.data(), n_cols, n_cols),
-    raft::make_device_matrix_view<math_t, idx_t, raft::col_major>(
-      components.data_handle(), n_components, n_cols));
+  if constexpr (is_row_major) {
```
divyegala, line 65: Do you not need to call `trunc_zero_origin` in this branch?

divyegala:
@coderabbitai review

cpp/include/raft/linalg/detail/pca.cuh
aamijar: It is doing the same thing as `trunc_zero_origin` except using a simple `raft::copy`.

The goal is to copy a slice of some rows from the src `n_cols x n_cols` matrix. 
If src is row-major we can just get the prefix `n_components x n_cols` from memory.
If src is col-major then we need to copy some elements, skip some, and the copy some more etc. That is what `trunc_zero_origin` is doing.

divyegala (approved):

lowener (approved):
LGTM

PR #3040, opened 2026-05-26
Revert NCCL comms updates that require NCCL 2.28 or later.

seunghwak:
In https://github.com/rapidsai/raft/pull/3002, we updated raft::comms_t to use ncclGather/ncclScatter/ncclAlltoAll functions added in NCCL 2.28.3.

But for various dependency issues, we currently can't increase the minimum NCCL version to 2.28. So, we are reverting the change for now.


seunghwak:
/ok to test [b325685](https://github.com/rapidsai/raft/pull/3040/commits/b3256854ef40bffc1da9dca7be3afaf9d97fb818)

cjnolet (approved):

PR #3041, opened 2026-05-27
SelectK race condition fix

### LINKED ISSUE #2997: [CI] Flaky SelectKCsrTest/SelectKCsrTest_float_int.Result/39
This test has been flaky for quite some time now. It is probably time to get to the bottom of it.
https://github.com/rapidsai/raft/actions/runs/24067374184/job/70262381782#step:12:1858

aamijar:
Resolves https://github.com/rapidsai/raft/issues/2997, a longstanding flaky test failure.

There is a synchronization race where the output buffer histogram may not be fully populated when it is read by the last block. 
The blocks signal completion via `finished_block_cnt`, but does not guarantee all threads are complete with the `filter_and_historgram` function. 
So we need to use `__syncthreads()` to guarantee all threads within the block are done before incrementing `finished_block_cnt`.

### DIFF
--- a/cpp/include/raft/matrix/detail/select_radix.cuh
+++ b/cpp/include/raft/matrix/detail/select_radix.cuh
@@ -748,15 +748,8 @@ RAFT_KERNEL radix_kernel(const T* in,
                                              pass,
                                              early_stop);
   __threadfence();
-  // Ensure every thread in this block has finished its writes (and the
-  // corresponding __threadfence) before any block signals completion via
-  // atomicInc. Otherwise warp 0's thread 0 can race ahead of lagging warps
-  // that are still issuing atomicAdds inside filter_and_histogram (e.g. the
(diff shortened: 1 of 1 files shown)


lowener (commented):
Were you able to reproduce the bug and verify that with this fix the test is now passing consistently?

aamijar:
Hi @lowener, yes I was able to reproduce the issue and verify that this fix passes the tests repeatedly.

lowener (approved):
LGTM

qwertyforce:
Hi, could this fix be included in the next release? 
I believe this race condition can lead to illegal memory access error in cuvs

Current RAFT version I checked (v26.06.00) does not include it, while main does.

divyegala:
Hi @qwertyforce yes, this fix will be included in 26.08 release for RAFT. As cuVS pins to the same version as RAFT, you will also need to update to cuVS 26.08 to get this fix. Please let us know if you have other questions!

PR #3042, opened 2026-05-27
chore(ci): use explicit secrets where possible

gforsyth:
Followup to #3009 -- some of the secret handling was added in later PRs.  Adding it in to `raft` now.

This is also part of rapidsai/build-planning#275

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,5 +1,4 @@
 name: build
-
 on:
   push:
     branches:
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yaml
```diff
@@ -1,5 +1,4 @@
 name: build
-
```
gforsyth, line 2: The whitespace changes are a result of using `yq` to bulk-edit the workflow files.  The loss of the whitespace isn't my first choice, but I think it's a reasonable trade-off for what `yq` allows us to do, especially when making bulk edits across repositories.

PR #3050, opened 2026-06-08
fix(warp_reduce): add explicit add_op overload to resolve CUB template ambiguity on CUDA 13.2+

zbrad:
## Summary

On CUDA 13.2 (SM 121, DGX Spark), IVF-PQ builds fail with an ambiguous template instantiation error. When CUB scan kernels call `warpReduce(val, raft::add_op{})`, both `raft::warpReduce<T, ReduceLambda>` and `cub::detail::scan::warpReduce<Tp, ScanOpT&>` match, producing a compile error.

**Fix:** Add an explicit non-template overload in `cpp/include/raft/util/reduction.cuh`:

```cpp
template <typename T>
DI T warpReduce(T val, raft::add_op reduce_op)
```

The explicit overload is preferred by the compiler over the generic `ReduceLambda` overload, resolving the ambiguity without changing any existing behavior.

- `cpp/include/raft/util/reduction.cuh` — explicit `raft::add_op` overload for `warpReduce`
- `cpp/tests/util/reduction.cu` — regression test (`WARP_REDUCE_WITH_ADD_OP`) to prevent future regressions

## Repro / context

Observed on DGX Spark (SM 121) with CUDA 13.2. The upstream CI does not test CUDA 13.2 / SM 121; the new test compiles and passes on all CUDA versions but will catch regressions if CUDA 13.2 support is added to CI.

Error seen without fix:
```
error: more than one instance of overloaded function "warpReduce" matches the argument list
```

## Test plan

- [x] `WARP_REDUCE_WITH_ADD_OP` regression test added in `cpp/tests/util/reduction.cu`
- [x] Verified fix compiles and tests pass on CUDA 13.2 (SM 121, DGX Spark)
- [x] No changes to existing `warpReduce` behavior — explicit overload only activates for `raft::add_op`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


cpp/tests/util/reduction.cu
```diff
@@ -31,6 +31,15 @@ RAFT_KERNEL test_reduction_kernel(const int* input, int* reduction_res, ReduceLa
   if (threadIdx.x == 0) { reduction_res[0] = th_val; }
 }
 
+// regression: warpReduce(val, raft::add_op{}) must not cause ambiguity with CUB on CUDA 13.2+
+RAFT_KERNEL test_warp_reduce_with_add_op_kernel(const int* input, int* reduction_res)
+{
+  assert(gridDim.x == 1);
+  int th_val = input[threadIdx.x];
+  th_val     = raft::warpReduce(th_val, raft::add_op{});
```
achirkin, line 39: Does this really cause an ambiguity without the extra overload in `util/reduction.cuh`? It's called here with raft namespace, so I doubt CUB overload is ever picked up here.

achirkin (changes requested):
Thank you for reporting the issue! Could you please provide the reference/example where the bug is triggered? Normally, I'd assume the fix should be on the user side - just call the function with explicit namespace.

cpp/tests/util/reduction.cu
zbrad: It showed up for me when trying to do a full source rebuild of cuvs on the dgx spark.   When finally debugging my build failure, it traced down to raft, but only showed up when building for arm64.

zbrad: Oh, I have the cuvs regression test for it that I'm submitting to cuvs, it's at cpp/tests/regression/warp_reduce_add_op.cu

divyegala:
@zbrad can you provide steps to reproduce this bug? I have a DGX Spark, and I have been building CUDA 13.2 without any failures.

zbrad:
some other folks had asked for more background, so I went back and re-created the original failure from compiling cuvs.  I've attached the doc and the repro.
[warp reduce ambiguity doc](https://github.com/zbrad/cuvs/blob/fix/regression-tests/cpp/tests/regression/warp_reduce_ambiguity.md)
[repro](https://github.com/zbrad/cuvs/blob/fix/regression-tests/cpp/tests/regression/repro_warp_reduce_ambiguity.cu)

achirkin:
Thanks for the reproducer documentation! I feel uneasy about both suggested workarounds:
- raft extra overload: it doesn't protect us from someone else failing exactly the same way on another device operation.
- cuvs swap to thrust::plus: because we have a best practice guideline to use the raft operations where possible.

Maybe we could make the raft's warpReduce template itself a bit more restrictive so it wouldn't get in the way of thrust+cub primitives?..

achirkin:
I'm closing this issue, because the fix doesn't belong to raft: we cannot keep adding specializations here for each new operation type. I let's try to address this issue at the call site (that is in cuvs).

PR #1585, opened 2023-06-09
This PR adds support to __half and nb_bfloat16 to myAtomicReduce

Kh4ster:

PR #3056, opened 2026-06-23
Add suppport for different datatype in stats::mean

### LINKED ISSUE #2625: [FEA] Support fp16 input type in stats::mean()
Support fp16 operations for `raft::statt::mean()`

fp16 support where the intermediate results are accumulated in fp32, to avoid potential overflow.

**Additional context**

This issue came up in https://github.com/rapidsai/cuvs/pull/716, and resulted in the following workaround

https://github.com/rapidsai/cuvs/pull/716/files#diff-15fe794d70734cd604f7fda31ee8c2240839939c1c811df6cceb71752a0c35c8R242

lowener:
This PR enables the support of two different data types for stats::mean. It will be used in https://github.com/rapidsai/cuvs/pull/2250
Add support for half in strided dataset. Enabled thanks to this PR: https://github.com/rapidsai/raft/pull/1585

Closes #2625

### DIFF
--- a/cpp/include/raft/linalg/strided_reduction.cuh
+++ b/cpp/include/raft/linalg/strided_reduction.cuh
@@ -69,7 +69,7 @@ void stridedReduction(OutType* dots,
   // other cases, because coalescedReduction supports arbitrary types.
   if constexpr (std::is_same_v<OutType, float> || std::is_same_v<OutType, double> ||
                 std::is_same_v<OutType, int> || std::is_same_v<OutType, long long> ||
-                std::is_same_v<OutType, unsigned long long>) {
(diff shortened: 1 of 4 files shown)


aamijar (approved):
Thanks @lowener, it looks good to me. From my understanding the main change is in the reduction function where we do the cast when reading the data, then do the add and divide ops.

So is the main purpose to calculate the mean in higher precision to avoid floating point errors?

lowener:
Yes, the reduction function cast before doing the mean operation, so for example FP16 data can be accumulated in FP32.
Yes that's the main purpose, but for me it was just to allow more flexibility on input and output type, without using an intermediate buffer.

achirkin (approved):
LGTM, I've recently encountered a problem this PR will address too: calculating the average graph degree (which is uint32_t).

PR #3060, opened 2026-06-24
Add missing `<atomic>` include in `interruptible.hpp`

legrosbuffle:
A future libc++ version removes transitive includes of the standard `<atomic>` header and this results in compilation errors in code using the library without an explicit inclusion.

```
raft/core/interruptible.hpp:265:8: error: no type named 'atomic_flag' in namespace 'std'
  265 |   std::atomic_flag continue_;
```


achirkin (approved):
Thanks for reporting! LGTM

PR #3075, opened 2026-07-06
Make permute deterministic given a seed

says1117:
## Summary
`raft::random::permute` drew the coefficients of its affine permutation
from the unseeded global C `rand()`, so results were not reproducible
across calls. This surfaced in cuML as rapidsai/cuml#7871:
`make_regression(..., shuffle=True)` returned different data on repeated
calls with the same `random_state`.

## Changes
- Add a `seed` parameter to `permute` that derives the permutation
  coefficients from a local `std::mt19937_64`, replacing the global `rand()`.
- Thread the existing `make_regression` seed through both the sample and
  feature shuffles (a distinct derived seed each, so the two shuffles stay
  independent).
- The public `permute` overloads default `seed = 0`, so existing callers
  are unaffected.
- Add `PermSeedTest` verifying the same seed yields identical permutations
  and different seeds yield different ones.

## Testing
All 667 tests in `RANDOM_TEST` pass locally (RTX 2060, CUDA arch 75).

Relates to rapidsai/cuml#7871

### DIFF
--- a/cpp/include/raft/random/detail/make_regression.cuh
+++ b/cpp/include/raft/random/detail/make_regression.cuh
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2019-2026, NVIDIA CORPORATION.
+ * SPDX-FileCopyrightText: Copyright (c) 2019-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
 
@@ -24,6 +24,7 @@
 #include <rmm/device_uvector.hpp>
 
 #include <algorithm>
+#include <random>
 
 namespace raft {
 namespace random {
@@ -240,8 +241,9 @@ void make_regression_caller(raft::resources const& handle,
   }
 
   if (shuffle) {
-    // creates shared generator. pulls 2 random numbers from
-    // internal sequence to generate a and b, remembers it gave out those numbers
+    // One generator shared by both shuffles below: each consumes distinct draws
+    // from its sequence, so the sample and feature permutations stay independent
+    // while remaining reproducible for a given seed.
     std::mt19937_64 gen(seed);
 
     rmm::device_uvector<DataT> tmp_out(n_rows * n_cols, stream);
@@ -251,14 +253,8 @@ void make_regression_caller(raft::resources const& handle,
     constexpr IdxT Nthreads = 256;
 
     // Shuffle the samples from out to tmp_out
-    raft::random::detail::permute<DataT, IdxT, IdxT>(perms_samples.data(),
-                                                     tmp_out.data(),
-                                                     out,
-                                                     n_cols,
-                                                     n_rows,
-                                                     true,
-                                                     stream,
-                                                     gen);  // now passes generator rather than seed
+    raft::random::detail::permute<DataT, IdxT, IdxT>(
+      perms_samples.data(), tmp_out.data(), out, n_cols, n_rows, true, stream, gen);
     IdxT nblks_rows = raft::ceildiv<IdxT>(n_rows, Nthreads);
     _gather2d_kernel<<<nblks_rows, Nthreads, 0, stream>>>(
       values, _values, perms_samples.data(), n_rows, n_targets);
--- a/cpp/include/raft/random/detail/make_regression.cuh
+++ b/cpp/include/raft/random/detail/make_regression.cuh
@@ -248,7 +248,7 @@ void make_regression_caller(raft::resources const& handle,
 
     // Shuffle the samples from out to tmp_out
     raft::random::permute<DataT, IdxT, IdxT>(
-      perms_samples.data(), tmp_out.data(), out, n_cols, n_rows, true, stream);
+      perms_samples.data(), tmp_out.data(), out, n_cols, n_rows, true, stream, seed);
     IdxT nblks_rows = raft::ceildiv<IdxT>(n_rows, Nthreads);
     _gather2d_kernel<<<nblks_rows, Nthreads, 0, stream>>>(
       values, _values, perms_samples.data(), n_rows, n_targets);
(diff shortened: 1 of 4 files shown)


says1117:
This is a non-breaking bug fix (the new permute seed parameter is defaulted, so the public API stays backward-compatible). Could a maintainer add the bug and non-breaking labels? Thanks!

viclafargue:
cc @vinaydes

cpp/include/raft/random/detail/make_regression.cuh
```diff
@@ -248,15 +248,22 @@ void make_regression_caller(raft::resources const& handle,
...
     raft::random::permute<DataT, IdxT, IdxT>(
-      perms_features.data(), out, tmp_out.data(), n_rows, n_cols, false, stream);
+      perms_features.data(),
+      out,
+      tmp_out.data(),
+      n_rows,
+      n_cols,
+      false,
+      stream,
+      seed + 1);  // different derived seed for feature shuffle keeps sample and feature independent
```
achirkin: Maybe it would be a bit cleaner to pass a single generator by reference to the two functions here in place of the integer seed?

cpp/include/raft/random/permute.cuh
```diff
@@ -90,7 +90,8 @@ template <typename InputOutputValueType, typename IntType, typename IdxType, typ
 void permute(raft::resources const& handle,
              raft::device_matrix_view<const InputOutputValueType, IdxType, Layout> in,
              std::optional<raft::device_vector_view<IntType, IdxType>> permsOut,
-             std::optional<raft::device_matrix_view<InputOutputValueType, IdxType, Layout>> out)
+             std::optional<raft::device_matrix_view<InputOutputValueType, IdxType, Layout>> out,
+             uint64_t seed = 0ULL)
```
achirkin: I think, this is a "slightly breaking" change: the previous behavior was non-deterministic and the new behavior is deterministic. Shall we pass the seed as `std::optional<>` instead (`uint64_t` or an optional generator type by value) and default to `rand()` producing the seed non-deterministically?

achirkin (changes requested):
Thanks for the PR! It's a clear improvement indeed, yet I have couple minor suggestions.

cpp/include/raft/random/detail/make_regression.cuh
says1117: Thanks for the suggestion! I'll build one std::mt19937_64 in make_regression_caller and pass it by reference to both permute calls instead of deriving seed/seed + 1.

cpp/include/raft/random/permute.cuh
says1117: I'll change it to std::optional<uint64_t> seed = std::nullopt, falling back to rand() when unset so existing callers keep their current non-deterministic behavior. Will push an update soon.

says1117:
Pushed an update addressing the review feedback:
- `make_regression_caller` now builds a single `std::mt19937_64` generator
  and passes it by reference to both the row-shuffle and column-shuffle
  `permute` calls, rather than deriving `seed`/`seed + 1` separately.
- `seed` is now `std::optional<uint64_t>` (default `std::nullopt`) across all
  three `permute` overloads. When unset it falls back to `rand()`, so
  existing callers keep their previous non-deterministic behavior; passing a
  seed opts into reproducible shuffles.
- Also addressed the CodeRabbit findings: the `std::nullopt`-convenience
  overload now accepts and forwards `seed` too (previously it was stuck on
  the default with no way to opt in), added missing `@param[in] seed`
  Doxygen to all three overloads, and replaced a hardcoded `float` in
  `PermSeedTest`'s `TEST_P` body with the `test_data_type` alias to match
  the sibling fixtures.
Verified: `RANDOM_TEST` builds clean and all 667 tests pass.

vinaydes:
Thanks @viclafargue for tagging me.

Thanks @says1117 for this PR. I also have been working on overhauling the RAFT permute. Unfortunately, the affine transformation (`ax + b`) used by permute is not a good strategy. It suffers from biasing. A permute can be viewed as a bijective (thus invertible) function that maps index `i` to `F(i)`. The bijectivity ensures no collision. Ideally we would want this function to behave as a keyed random function. We can measure the "randomness" of this mapping function by looking at the _avalanche_ property, much like hash functions. In short it measures the probability of changing `j` th output bit when `i` th input bit is flipped for all `i`s and `j`s. An ideal function would have each probability close to 0.5. The affine transform performs very poorly on avalanche metric. Following is the heat-map of avalanche probabilities for N = 2^24

bias_lcg64_n24_heatmap

I used Feistel network based hashing to get near perfect avalanche. See more in my PR https://github.com/NVIDIA/raft/pull/3079

cpp/include/raft/random/permute.cuh
```diff
@@ -120,13 +122,17 @@ void permute(raft::resources const& handle,
...
     const IdxType D = in.extent(1);
-    detail::permute<InputOutputValueType, IntType, IdxType>(permsOut_ptr,
-                                                            out_ptr,
-                                                            in.data_handle(),
-                                                            D,
-                                                            N,
-                                                            is_row_major,
-                                                            resource::get_cuda_stream(handle));
+    std::mt19937_64 gen(seed.has_value() ? *seed
+                                         : rand());  // use the given seed, else a random one
```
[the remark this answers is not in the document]
says1117: Pushed a fix. The unseeded fallback now uses std::random_device{}() instead of rand(), so it no longer depends on shared global state across concurrent calls. Also factored the seed-or-random logic into a helper inside permute_impl rather than duplicating it across both overloads, and added a double instantiation of PermSeedTest to match the existing PermTest/PermMdspanTest pairs.

says1117:
Thanks for the detailed writeup and the heatmap @vinaydes, that's a much deeper fix than what I was going for here. This PR was purely aimed at closing the determinism gap flagged in [rapidsai/cuml#7871](https://github.com/rapidsai/cuml/issues/7871) (same seed to same shuffle), not at fixing the affine transform's statistical quality, so it looks like #3079 supersedes what I'm doing here once it lands. Happy to either: (a) keep this as a narrow, low-risk fix to unblock cuml now while #3079 is still in progress, or (b) close this in favor of [#3079](https://github.com/NVIDIA/raft/pull/3079) if you'd rather consolidate. No strong preference either way. Let me know which you'd prefer.

vinaydes:
@says1117, thank you again for your PR and for the time and effort you put into this contribution. For now, we’ve decided to move forward with [PR #3079](https://github.com/NVIDIA/raft/pull/3079). We truly appreciate your enthusiasm and hope you’ll continue contributing in the future.

says1117:
@vinaydes absolutely! thank you! :)

PR #3077, opened 2026-07-08
Restore tracked device resources on their original GPU

fallintoplace:
## Summary
- restore `memory_stats_resources` and `memory_tracking_resources` device memory resources to the CUDA device they wrapped at construction time
- switch the destructor restore path to `rmm::mr::set_per_device_resource(...)` using the captured device id
- add guarded 2-GPU regressions that destroy each wrapper while a different CUDA device is current

## Why
Both wrappers used `rmm::mr::set_current_device_resource(...)` during teardown. In multi-GPU flows that writes the saved resource into whichever CUDA device is current at destruction time, which can leave the construction device pointing at the tracking adaptor and install the wrong resource on another GPU.

## Impact
This keeps per-device RMM state stable across wrapper teardown and closes a nasty multi-GPU correctness hole that could otherwise lead to wrong-device allocators being reused.

## Validation
- `git diff --check`
- attempted `PARALLEL_LEVEL=8 ./build.sh tests -n --limit-tests=CORE_TEST`
- local build is currently blocked by missing `nvcc` / `CUDAToolkit_ROOT`, so the new tests were not run here

### DIFF
--- a/cpp/include/raft/core/memory_stats_resources.hpp
+++ b/cpp/include/raft/core/memory_stats_resources.hpp
@@ -4,7 +4,7 @@
  */
 #pragma once
 
-#include <raft/core/device_setter.hpp>
+#include <raft/core/resource/device_id.hpp>
 #include <raft/core/resource/device_memory_resource.hpp>
 #include <raft/core/resource/managed_memory_resource.hpp>
 #include <raft/core/resource/pinned_memory_resource.hpp>
@@ -77,9 +77,9 @@ class memory_stats_resources : public resources {
  public:
   explicit memory_stats_resources(const resources& existing)
     : resources(existing),
-      device_id_(device_setter::get_current_device()),
       old_host_(mr::get_default_host_resource()),
-      old_device_(rmm::mr::get_current_device_resource_ref())
+      old_device_(rmm::mr::get_per_device_resource_ref(
+        rmm::cuda_device_id{resource::get_device_id(*this)}))
   {
     init();
   }
@@ -87,7 +87,8 @@ class memory_stats_resources : public resources {
   ~memory_stats_resources() override
   {
     mr::set_default_host_resource(old_host_);
-    rmm::mr::set_per_device_resource(rmm::cuda_device_id{device_id_}, std::move(old_device_));
+    rmm::mr::set_per_device_resource(rmm::cuda_device_id{resource::get_device_id(*this)},
+                                     std::move(old_device_));
   }
 
   memory_stats_resources(memory_stats_resources const&)            = delete;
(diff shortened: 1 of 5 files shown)


cpp/include/raft/core/memory_stats_resources.hpp
```diff
@@ -85,7 +87,7 @@ class memory_stats_resources : public resources {
   ~memory_stats_resources() override
   {
     mr::set_default_host_resource(old_host_);
-    rmm::mr::set_current_device_resource(std::move(old_device_));
+    rmm::mr::set_per_device_resource(rmm::cuda_device_id{device_id_}, std::move(old_device_));
```
achirkin: `core/resource/device_id.hpp` is the source of truth for the GPU device associated with the handle:
```suggestion
    rmm::mr::set_per_device_resource(resource::get_device_id(*this), std::move(old_device_));
```

cpp/include/raft/core/memory_stats_resources.hpp
```diff
@@ -76,6 +77,7 @@ class memory_stats_resources : public resources {
  public:
   explicit memory_stats_resources(const resources& existing)
     : resources(existing),
+      device_id_(device_setter::get_current_device()),
       old_host_(mr::get_default_host_resource()),
       old_device_(rmm::mr::get_current_device_resource_ref())
```
achirkin: Calling `resource::get_device_id()` on the current handle ensures the device id resource is instantiated and won't mutate until the resource is destroyed.
```suggestion
      old_device_(rmm::mr::get_per_device_resource_ref(resource::get_device_id(existing)))
```

achirkin (changes requested):
Thank you for uncovering this! The memory tracking/statistics APIs are rather new and haven't been properly tested in multi-gpu environment yet.

In the context of `raft::resources` handle we already have access to a "device id" resource (`core/resource/device_id.hpp`) stored per resources handle. Therefore, there's no need for an extra member variable duplicating this information (see below). Please adjust both files to use this resource instead.

fallintoplace:
If something comes up, let me know.

achirkin (approved):
Thanks for the updates! LGTM

achirkin:
@fallintoplace could you please apply the style fixes? `pre-commit run --all-files`

fallintoplace:
@achirkin Sorry, can you please check again?

cpp/include/raft/core/memory_stats_resources.hpp
```diff
@@ -77,15 +78,17 @@ class memory_stats_resources : public resources {
...
     init();
   }
 
   ~memory_stats_resources() override
   {
     mr::set_default_host_resource(old_host_);
-    rmm::mr::set_current_device_resource(std::move(old_device_));
+    rmm::mr::set_per_device_resource(rmm::cuda_device_id{resource::get_device_id(*this)},
+                                     std::move(old_device_));
   }
```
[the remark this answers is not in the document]
fallintoplace: I still need to resolve this.

PR #3078, opened 2026-07-08
Make cuBLASLt descriptor wrappers move-safe

fallintoplace:
## Summary
- make the cuBLASLt matrix layout and matmul descriptor wrappers move-safe
- transfer descriptor ownership explicitly on move construction and move assignment
- make moved-from wrappers null-safe on destruction
- add focused tests that verify each wrapper transfers its raw handle and nulls the source on move

## Why
These wrappers own raw `cublasLt*` descriptors and destroy them in their destructors, but their move constructor and move assignment operator were defaulted. That meant moving a wrapper copied the raw handle into the destination without clearing the source, so both objects believed they owned the same descriptor.

## Impact
This closes a double-destroy hazard in any path that moves or move-assigns these wrappers, including code that caches or returns compound matmul descriptor objects.

## Root cause
The wrapper types were acting like RAII owners while still using raw-pointer default move semantics.

## Validation
- `git diff --check`
- attempted `PARALLEL_LEVEL=8 ./build.sh tests -n --limit-tests=LINALG_TEST`
- local build is blocked here because `nvcc` is not installed and CMake cannot resolve `CUDAToolkit_ROOT`


cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp
```diff
@@ -108,12 +109,22 @@ struct cublastlt_matrix_layout {
...
+  inline cublastlt_matrix_layout(cublastlt_matrix_layout&& other) noexcept
+    : res(std::exchange(other.res, nullptr))
+  {
+  }
+  inline auto operator=(cublastlt_matrix_layout&& other) noexcept -> cublastlt_matrix_layout&
+  {
+    if (this != &other) {
+      if (res != nullptr) { RAFT_CUBLAS_TRY_NO_THROW(cublasLtMatrixLayoutDestroy(res)); }
+      res = std::exchange(other.res, nullptr);
+    }
```
achirkin: Rather than having multiple `cublasLtMatrixLayoutDestroy` uses, perhaps, a cleaner solution would be to use `std::swap(this->res, other.res)` and rely on the other one being correctly destroyed?

achirkin (commented):
Thanks for your contribution! The change looks reasonable. However, before we move forward, could you please tell what is the original bug/problem being addressed? These wrapper types are in the detail namespace, that is not supposed to be used outside of the raft implementation details.

fallintoplace:
A small repro is:
```
auto src = cublastlt_matmul_desc{CUBLAS_COMPUTE_32F, CUDA_R_32F};
auto raw = static_cast<cublasLtMatmulDesc_t>(src);
auto dst = std::move(src);

EXPECT_EQ(static_cast<cublasLtMatmulDesc_t>(src), nullptr);
EXPECT_EQ(static_cast<cublasLtMatmulDesc_t>(dst), raw);
```
On main, the first expectation fails because the defaulted move leaves both objects holding the same handle. When they go out of scope, both destructors try to destroy it.
The same move can happen internally because cublastlt_matmul_desc::for_matmul() and matmul_desc::create() return named local objects. Building with -fno-elide-constructors forces that path.
I also changed move assignment to use std::swap as suggested and updated the tests accordingly.

divyegala (approved):

PR #3079, opened 2026-07-09
[FEA] permute: replace LCG with keyed Feistel permutation, add deterministic key API

vinaydes:
**UPDATE on 27-July-2026:** I have update the PR to use [CCCL](https://github.com/NVIDIA/cccl) instead of hand coding the permutation after discussion with @sleeepyjack and @RAMitchell. Therefore I have scratched out the part of description that is no longer relevant or valid

---

The previous permute() implementation used a linear-congruential generator (LCG) -- `(a * tid + b) % N` -- seeded from `rand()` to map output indices to input indices. This had two significant problems:

- Non-reproducible: `rand()` is seeded by global state, so results varied between runs and between callers. There was no way to request the same permutation twice.
- LCG is not a very good randomizer causing biases

This PR reaplces the LCG with the `shuffle_iterator` from NVIDIA CCCL.

~~This PR replaces the LCG with a keyed Feistel permutation built from a 4-round network whose round function is a reduced version of MurmurHash3 32-bit finalizer (`fmix32`). The domain is extended to the next power of two and cycle walking is used to map results back into `[0, N)` without any modulo bias.~~

**API Changes**

Both the legacy pointer API and the mdspan API gain a required uint64_t key parameter. The same (key, N) pair always produces the same permutation, giving full reproducibility.

```C++
// legacy pointer API
raft::random::permute(perms, out, in, D, N, rowMajor, stream, key);

// mdspan API
raft::random::permute(handle, in, permsOut, out, key);
```
Callers that only want the permutation index array (no data shuffle) can pass out = nullptr; this now dispatches to a dedicated `permsOnlyKernel` with `ITEMS_PER_THREAD = 8` for higher throughput on large index arrays.

**Implementation Details**

`detail/permute.cuh`

- ~~`fmix32(h)`: Reudced version of MurmurHash3 finalizer; used as the Feistel round function.~~
- ~~`kperm_params`: host-precomputed schedule (split widths `a_bits`/`b_bits`, per-round key prefixes). Computed once per `permute()` call on the host and passed as a kernel argument, so all threads share it without extra loads.~~
- ~~`make_kperm_params(N, key)`: builds the key schedule.~~
- ~~`kperm_mix_nbits<WordType>(m, p)`: 4-round unrolled Feistel step.~~
- ~~`kperm_index<IdxType>(idx, p)`: cycle-walking wrapper~~
- `permsOnlyKernel`: new kernel invoked when `out == nullptr`; processes 8 indices per thread for better occupancy on pure index-generation workloads.

`detail/make_regression.cuh`

Derives two distinct keys from the caller-supplied seed so the sample shuffle and the feature shuffle get independent permutations:

```C++
const uint64_t samples_key  = seed;
const uint64_t features_key = seed ^ 0x9e3779b97f4a7c15ULL;
```
**Tests**

- All existing `PermTest` and `PermMdspanTest` cases updated to pass `params.seed` as the key
- New `SeedDiversity` test: 16384 threads each generate a permutation keyed by `base_seed + tid + 1` and count index coincidences with the reference permutation. The test asserts that the total match rate is below 5%, verifying that adjacent keys produce statistically independent permutations.

**Benchmarks**

- Existing permute<float/double> benchmarks pass a fixed key (123456ULL).
- New permute_perms_only<int/uint32_t> benchmarks measure the pure index-generation path (`out = nullptr`) at sizes up to 1 GiB.

~~**Avalanche Property**~~
~~A permutation of a finite set is by definition a bijection: a function that is both injective (no two inputs map to the same output) and surjective (every output is hit). A keyed hash function that is itself a bijection for every key is therefore directly usable as a permutation generator: apply it to every index in [0, N) and the results are a rearrangement with no duplicates or gaps. The Feistel construction guarantees bijectivity structurally -- each round is an invertible XOR-and-mask, so the full network is invertible regardless of what the round function does internally. The quality of the resulting permutation is measured by the avalanche property: flipping any single input bit should flip each output bit with probability close to 1/2, ensuring that nearby keys produce statistically independent permutations. I evaluated the avalanche behavior of this Feistel construction across all domain widths from 2 to 32 bits and found it satisfactory throughout.~~


cpp/include/raft/random/detail/permute.cuh
```diff
@@ -12,25 +12,146 @@
...
+  p.N = N;
+  if (N <= 1) return p;  // identity domain; remaining fields go unused
+
+  // n = ceil(log2(N)): smallest n with 2^n >= N (the cycle-walking domain).
+  int n         = 0;
+  uint64_t pow2 = 1;
+  while (pow2 < N) {
+    pow2 <<= 1;
+    ++n;
+  }
```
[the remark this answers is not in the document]
vinaydes: Done in latest push.

vinaydes:
Addressed all the @coderabbitai review comments.

cpp/include/raft/random/permute.cuh
```diff
@@ -192,7 +274,7 @@ void permute(IntType* perms,
              bool rowMajor,
              cudaStream_t stream)
 {
-  detail::permute<Type, IntType, IdxType, TPB>(perms, out, in, D, N, rowMajor, stream);
+  detail::permute<Type, IntType, IdxType, TPB>(perms, out, in, D, N, rowMajor, stream, uint64_t{0});
 }
```
achirkin: Maybe we'd just pass rand() as the explicit key to the new overload that requires it to preserve the expected behavior?

vinaydes: Okay, are you suggesting something like this?
```C++
detail::permute<Type, IntType, IdxType, TPB>(perms, out, in, D, N, rowMajor, stream, uint64_t{rand()});
```

vinaydes: Done.

cpp/include/raft/random/detail/permute.cuh
```diff
@@ -85,50 +143,87 @@ struct permute_impl_t {
...
              const Type* in,
              IntType D,
              IntType N,
              bool rowMajor,
-             cudaStream_t stream)
+             cudaStream_t stream,
+             uint64_t key)
 {
-  auto nblks = raft::ceildiv(N, (IntType)TPB);
+  if (N <= 0) { return; }
```
sleeepyjack:
```suggestion
  if (N <= 0 || (perms == nullptr && out == nullptr)) { return; }
```

When `out == nullptr`, this always launches `permsOnlyKernel`, which unconditionally writes through `perms`. The raw-pointer API permits both `perms` and `out` to be null.

cpp/include/raft/random/detail/permute.cuh
```diff
@@ -85,50 +143,87 @@ struct permute_impl_t {
...
+
+  cuda::shuffle_iterator shuffled_indices{cuda::random_bijection{N, cuda::std::minstd_rand{key}}};
```
sleeepyjack: 32-bit `minstd_rand` has modulus `2,147,483,647`, so many distinct `uint64_t` keys initialize the same generator state. `philox4x64` use 64-bit internal state and will give you better randomness (at the cost of being slower). Did you choose `minstd_rand` as a trade-off for better performance?

cpp/include/raft/random/detail/permute.cuh
vinaydes: That's a good catch. Thanks.

cpp/include/raft/random/detail/permute.cuh
vinaydes: I was just using the example given in the CCCL documentation. I missed that it only takes 32-bit key. I replaced it with SplitMix64 now which should be suffice for our purpose here. Thanks.

vinaydes:
I made some mistake while rebasing, that’s why it initiated several approval requests (for example @KyleFromNVIDIA). I am not sure they are needed any more.

cpp/include/raft/random/detail/permute.cuh
vinaydes: I am keeping the 32-bit RNG in the unit test for seed diversity because quality there does not matter much.

sleeepyjack (approved):

divyegala:
@vinaydes I removed unneeded reviewers, could you please resolve conflicts? Once @achirkin approves, this PR will have necessary codeowner approvals needed to merge it.

vinaydes:
> @vinaydes I removed unneeded reviewers, could you please resolve conflicts? Once @achirkin approves, this PR will have necessary codeowner approvals needed to merge it.

Thanks @divyegala.

Conflicts should be removed with the latest push. @achirkin, you can go ahead with the review now. Thanks.

achirkin (approved):

vinaydes:
Thanks @achirkin 
@divyegala The PR is ready to be merged.

PR #3084, opened 2026-07-16
Remove orphaned raft::runtime::matrix::select_k declaration

### LINKED ISSUE #3028: Move select_k out of raft_runtime
The [`select_k` header](https://github.com/rapidsai/raft/blob/main/cpp/include/raft_runtime/matrix/select_k.hpp) is in the `raft_runtime` directory, but it is not part of [the precompiled libraft library](https://github.com/rapidsai/raft/blob/main/cpp/CMakeLists.txt#L275) and the [header file](https://github.com/rapidsai/raft/blob/main/cpp/include/raft/matrix/select_k.cuh) requires CUDA to compile ([transitively via `select_k-inl.cuh`](https://github.com/rapidsai/raft/blob/main/cpp/include/raft/matrix/detail/select_k.cuh)). Based on the current intended scope of `raft_runtime`, `select_k` should be moved out of that directory.

says1117:
Closes #3028. The header declared raft::runtime::matrix::select_k but had no corresponding implementation anywhere in the repo, was not listed as a CMake build source, and produced no linkable symbols in libraft. Traced to PR #2498 (implementation removed) and PR #2885 (later raft_runtime cleanup that missed this file). The working select_k implementation remains at raft/matrix/select_k.cuh.


divyegala (approved):
Thank you for the PR!

Issue #3021, opened 2026-05-17
[CI] Lanczos SM tests fail cuda 13.2


aamijar:
Observed here:

https://github.com/rapidsai/raft/actions/runs/25983993593
https://github.com/rapidsai/raft/actions/runs/25983959550
https://github.com/rapidsai/raft/actions/runs/25955468286
https://github.com/rapidsai/raft/actions/runs/25955311798

https://github.com/rapidsai/raft/actions/runs/25972693380/job/76351882397?pr=3019#step:12:748

Most prevalent on a100, arm64

```
[----------] 1 test from LanczosTests/LanczosTestD_SM
[ RUN      ] LanczosTests/LanczosTestD_SM.Result/0
/tmp/conda-bld-output/bld/rattler-build_libraft-headers-only/work/cpp/tests/sparse/solver/lanczos.cu:351: Failure
Value of: raft::devArrMatch<ValueType>(eigenvalues.data_handle(), expected_eigenvalues.data_handle(), params.n_components, raft::CompareApprox<ValueType>(1e-5), stream)
  Actual: false (actual=0.013678239999999999 != expected=0.013691776200317895 @1)
Expected: true
```

divyegala:
@aamijar what is the absolute diff in all the failures that you are tracking at the moment?

aamijar:
Hi @divyegala, it is just this one failure in the issue description
```
Actual: false (actual=0.013678239999999999 != expected=0.013691776200317895 @1)

Abs diff = 0.0000135362
```

So we are above threshold by a tiny amount 1.3536e-5 > 1.0e-5

divyegala:
@aamijar and in the other links you posted? What is the error diff?

aamijar:
@divyegala, I think all the errors are the same. Same actual and expected values.

cjnolet:
Anupam,

The results are very close. I would suggest temporarily increasing the
threshold by just enough to get this test to pass. That will at least
unblock CI while still rendering the test useful.

After that, I would suggest testing based on the characteristics /
constraints of knowing the outputs are eigenpairs, instead of hard coding
them.

divyegala:
@aamijar when you temporarily increase the threshold as Corey suggested, could you also bring nightly check tolerance back to 14 days?

Issue #2758, opened 2025-07-23
[BUG] Eigen solver fails to converge to a solution


jnke2016:
**Describe the bug**
cuGraph spectral clustering which leverages `raft::spectral::matrix::sparse_matrix_t` returns poor clustering result. This appears to be due to the eigen solver's inability to converge to a solution.

**Steps/Code to reproduce bug**
```
#include <rmm/device_uvector.hpp>
#include <rmm/device_vector.hpp>
#include <raft/core/handle.hpp>
#include <raft/spectral/partition.cuh>
#include <raft/core/resources.hpp>
#include <raft/spectral/eigen_solvers.cuh>
#include <raft/spectral/matrix_wrappers.hpp>

#include <iostream>
#include <string>



int main(int argc, char** argv)
{

    raft::handle_t handle;

    using vertex_t    = int32_t;
    using edge_t      = int32_t;
    using weight_t    = float;

    std::vector<vertex_t> h_row_offsets = {0, 3, 5, 8, 11, 13, 15};
    std::vector<vertex_t> h_col_indices = {1, 2, 0, 2, 0, 1, 3, 2, 4, 5, 3, 5, 3, 4};
    //std::vector<weight_t> h_values = {1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1};
    //std::vector<weight_t> h_values = {0.1, 0.2, 0.1, 1.2, 0.2, 1.2, 2.3, 2.3, 3.4, 3.5, 3.4, 4.5, 3.5, 4.5};
    std::vector<weight_t> h_values = {0.1, 0.2, 0.1, 1.2, 0.2, 1.2, 2.3, 2.3, 3.4, 3.5, 3.4, 4.5, 3.5, 4.5};


    rmm::device_uvector<vertex_t> d_row_offsets(h_row_offsets.size(), handle.get_stream());
    rmm::device_uvector<vertex_t> d_col_indices(h_col_indices.size(), handle.get_stream());
    rmm::device_uvector<weight_t> d_values(h_values.size(), handle.get_stream());

    raft::update_device(d_row_offsets.data(),
                        h_row_offsets.data(),
                        h_row_offsets.size(),
                        handle.get_stream());

    raft::update_device(d_col_indices.data(),
                        h_col_indices.data(),
                        h_col_indices.size(),
                        handle.get_stream());

    raft::update_device(d_values.data(),
                        h_values.data(),
                        h_values.size(),
                        handle.get_stream());

[...]

PR #3086, opened 2026-07-20
Replace hardcoded Lanczos eigenvalue goldens with property-based checks

### LINKED ISSUE #2519: [BUG] Make new lanczos gtests more robust to varying cuda runtime versions
Release 24.12 brings with it a new and improved lanczos solver, but there are some gtests which are failing because the expected results have been hardcoded while other tests use generated data. The hard-coded data is causing issues, and we should remove it, however we need to make sure the test coverage doesn't change significantly when we remove it.

says1117:
Fixes #2519.

## The problem

The Lanczos gtests checked computed eigenvalues against hardcoded arrays (`devArrMatch` + `CompareApprox(1e-5)`). That's fragile: floating-point addition isn't associative, so cuSPARSE/cuBLAS can sum things in a different order depending on CUDA version and GPU architecture, and the last few digits of an eigenvalue shift. #3021 hit this on A100/ARM64 — `0.01367824` vs `0.01369178`, a difference just outside the tolerance, even though the solve was correct.

## The fix

Instead of comparing against a fixed answer, check that what came back is actually a valid eigenpair:

- **Residual**: `||A*v - λv||` should be tiny. This is the same thing the solver's own `tolerance` config is defined to converge on, so it's not an arbitrary check - it's the solver's own success criterion.
- **Rayleigh quotient**: `λ ≈ (vᵀAv)/(vᵀv)`. Added this because the residual alone gets loose on matrices with a large norm, and can miss a small eigenvalue being slightly wrong. The Rayleigh quotient pins it down directly.
- **Unit norm**: eigenvectors should have length 1.
- **Orthogonality**: distinct eigenvectors shouldn't overlap. This one isn't in the old test at all - it catches a specific Lanczos failure mode where the solver returns the same eigenpair twice ("ghost duplicates"), which the residual check alone can't see.
- **Ordering**: eigenvalues come back sorted ascending, which I confirmed against all 11 of the old golden arrays before deleting them.

`rmat_lanczos_tests` already had the matrix-vector-multiply helper needed for the residual check sitting there unused; I just started using it. `lanczos_tests` needed it added. Eigenvectors were already being computed by every test, just never checked - no solver changes needed.

Left the reproducibility check alone (it runs the solve twice with the same seed and diffs the results exactly - that's a determinism check, not a golden-value comparison, so it isn't affected by the cross-platform issue).

## Picking the tolerance

I didn't want to just guess a number, so I instrumented the checks, ran all 11 test fixtures, and looked at what the residuals actually came out to:

| fixture | residual ÷ (‖A‖·ε) | orthogonality ÷ ε | \|λ − Rayleigh\| ÷ (‖A‖·ε) |
|---|---|---|---|
| double, n=100, ‖A‖=8.2 | 2.7 | 2.7 | 1.7 |
| float, n=100, ‖A‖=8.2 | 0.7 | 2.5 | 0.9 |
| double, n=100, ‖A‖=10.5 | 14.2 | 2.1 | 18.8 |
| float, n=100, ‖A‖=10.5 | 8.6 | 0.4 | 4.9 |
| float, n=4096, ‖A‖=298 | 4.8 | 18.9 | 2.0 |

Two things stood out. First, the residual scales with `‖A‖ · machine epsilon`, not with the matrix size `n` - the small fixtures (n=100) and the big RMAT one (n=4096) land in the same range. That mattered a lot (see below). Second, orthogonality and unit-norm error are both just small multiples of epsilon.

So the tolerances ended up being:

\`\`\`
matrix_tol = 500  × max(‖A‖, 1) × ε     — residual, Rayleigh, ordering
unit_tol   = 1000 × ε                    — unit norm, orthogonality
\`\`\`

Every measured value sits at 0.04%–3.8% of its tolerance - comfortable margin for cross-platform noise, without being so loose the check stops meaning anything.

## Making sure I didn't just weaken the test

I proved these checks can actually fail, by corrupting a solver's output on purpose and confirming the test catches it:

- Scale every eigenvalue by 1% → caught by the residual and Rayleigh checks, on all 11 tests.
- Zero out an eigenvector → caught immediately by the unit-norm check.
- Copy eigenpair 0 into eigenpair 1's slot (still a perfectly valid, unit-norm, low-residual eigenpair — just a duplicate) → only the orthogonality check catches this. Everything else passes. This is exactly why that check earns its place.

Worth being honest about: my first attempt at this got it wrong. I'd included an `n` factor in the residual tolerance, and it turned out `RmatLanczosTestF` still passed even with every eigenvalue scaled by 1% - a real regression from the old test, which would've caught that instantly. Digging into the data above is what showed the residual doesn't actually scale with `n`, and once I fixed that (and added the Rayleigh check), the same corruption test failed cleanly across the board, including on the hardest case - the smallest eigenvalue in the RMAT set, where a 1% error is only 0.08 in absolute terms.

## Verification

Built and ran on an RTX 2060 SUPER (sm_75), CUDA 13.3, WSL2:

\`\`\`
$ cmake --build cpp/build -j2 --target SOLVERS_TEST
[2/2] Linking CXX executable gtests/SOLVERS_TEST

$ ./cpp/build/gtests/SOLVERS_TEST --gtest_filter='*Lanczos*'
[==========] 11 tests from 11 test suites ran. (162255 ms total)
[  PASSED  ] 11 tests.
\`\`\`

`pre-commit` passes clean.

## For reviewers

- The tolerance is calibrated on one GPU/CUDA combo. The margins (26x–2300x) should hold up across platforms, but I'd genuinely appreciate someone running this on an A100 or ARM64 box, since that's the exact combination that surfaced the original bug in #3021.
- Checked for overlap with #3034 (adds a Lanczos SVD solver) - it's a different file and still unmerged, so no conflict.
- vyasr flagged on #2519 that this might mask a real convergence bug (see #2758), not just a flaky comparison. I didn't see any evidence of that here - every residual came in 1–2 orders of magnitude under tolerance - but I could only test on this one GPU/CUDA version, so I can't rule it out.

Refs #2519, #3021.


says1117:
Test files aren't in Doxygen scope per cpp/REVIEW_GUIDELINES.md

aamijar:
Hi @says1117, it looks like the lanczos cpp tests aren't passing with this change.

says1117:
CI surfaced that LanczosTestD_SM fails on H100/V100 while passing on A100/L4/RTX. 

Root cause: the smallest-magnitude eigenvector is ill-conditioned and its residual varies by orders of magnitude across CUDA versions/architectures, but the eigenvalue is correct on every platform (the Rayleigh quotient vᵀAv/vᵀv agrees to full tolerance everywhere), matching the SM concerns in #2705/#2758.

Rather than loosen everything, I split the checks: eigenvalue accuracy (Rayleigh + ordering) stays tight for all modes, and only the SM eigenvector bounds (residual/norm/orthogonality) are relaxed. Non-SM and all eigenvalue tolerances are unchanged. I confirmed a 1% eigenvalue perturbation on SM is still caught by the Rayleigh check, so the SM test isn't vacuous. Flagging in case the underlying SM eigenvector conditioning warrants separate solver investigation.

says1117:
@cjnolet  thanks for trying to kick this off! Looks like that /ok to test didn't include the commit SHA. Here's the current one if it helps: 24353160872fdd703b779dbac9e0b6c9d9e3d558

says1117:
@divyegala Just following up. This has had a passing CI run and CodeRabbit's automated review since Jul 23 with no outstanding actionable comments, but it's still waiting on a human review or approval.

divyegala:
Hi @says1117 thanks for following up! @aamijar could you please review this PR?

aamijar (changes requested):
Thanks for the contribution @says1117! 
My main concern is that we aren't validating the `which` parameter. How do we know if the eigenvalues returned are `SA, SM, LA, LM` ?

cpp/tests/sparse/solver/lanczos.cu
```diff
@@ -68,9 +73,201 @@ struct rmat_lanczos_inputs {
...
+ *   eigenvector matrix
+ * @param[in] n dimension of the (square) matrix
+ * @param[in] n_components number of eigenpairs to validate
+ * @param[in] frobenius_norm Frobenius norm of A, used to scale tolerances
+ * @param[in] which which eigenvalues were requested; selects the tight or
+ *   relaxed eigenvector tolerance (SM is relaxed)
+ */
+template <typename IndexType, typename ValueType>
+void expect_valid_eigenpairs(raft::resources const& handle,
+                             raft::spectral::matrix::sparse_matrix_t<IndexType, ValueType> const& A,
```
aamijar: Another smaller detail: we would prefer to use the newer device_csr_matrix_view type here instead

says1117:
Thanks for the review, @aamijar!

**`which` validation.** Added `compute_full_spectrum()`, which densifies the matrix and runs `raft::linalg::eig_dc` (cuSOLVER's dense symmetric eigensolver, independent of Lanczos and computed fresh rather than hardcoded), plus `expect_correct_selection()` to check the returned eigenvalues against the correct SA/SM/LA/LM subset of that reference. Verified it actually discriminates by feeding it a deliberately wrong criterion (SA output checked against LA) and confirming it fails.

**`device_csr_matrix_view`.** Switched as suggested, which also let me drop the separate `sparse_matrix_t` construction at each call site.

**CodeRabbit nitpicks.** Documented the dense-eigensolve cost at RMAT's scale (n=4096) and used the `eig_dc` mdspan wrapper.

**Bounds guard.** Added an `ASSERT_LE` so `n_components` exceeding the matrix dimension fails cleanly instead of running off the end of the iterators.

divyegala:
Hi @says1117 , sorry for the late response. Could you please resolve the conflicts in your PR? We would like to get this PR merged and appreciate your contribution!

says1117:
Hello @divyegala ! I believe this should resolve it. Thank you :)

aamijar (approved):
Thank you for addressing my comments @says1117!

divyegala (approved):

PR #3094, opened 2026-07-28
Use cuda::std::numeric_limits in LAP kernels instead of passing infinity

says1117:
## Description

The LAP (Linear Assignment Problem) Hungarian-algorithm kernels took the sentinel
"infinity" value — `std::numeric_limits<weight_t>::max()` — as a host-computed
kernel argument. This was guarded by `FIXME` comments noting it should switch to
`cuda::std::numeric_limits<weight_t>::max()` once CUDA 10.2 was the baseline.

That baseline has long been met, so this PR resolves the `FIXME`s by computing the
limit on-device instead of threading it through as a parameter.

## Changes

- Drop the `weight_t infinity` parameter from `kernel_rowReduction`,
  `kernel_columnReduction`, `kernel_dualUpdate_1`, and `kernel_dualUpdate_2`.
- Use `cuda::std::numeric_limits<weight_t>::max()` directly inside those kernels
  (for the running-minimum initializers and the `d_sp_min[spid] < infinity` guard).
- Remove the corresponding host-side `std::numeric_limits<weight_t>::max()`
  arguments at the call sites in `lap_functions.cuh`.
- Add `#include <cuda/std/limits>` and delete the four resolved `FIXME` comments.

No functional or numerical behavior change — the sentinel value is identical; it is
now sourced on the device rather than passed from the host.

## Testing

No new tests are required; this is a non-functional refactor covered by the existing
suite. Verified locally:

- `SOLVERS_TEST` builds clean.
- All 6 `Raft.Hungarian*` LAP tests pass (`IntFloat`, `IntDouble`, `IntLong`,
  `LongFloat`, `LongDouble`, `LongLong`).


says1117:
Could a maintainer please add the required labels when you get a chance? This is a non-functional refactor (identical sentinel value, just computed on-device), so `improvement` + `non-breaking`. Thanks!

divyegala (approved):

PR #3095, opened 2026-07-28
Use cuda::std::bit_cast in stats minmax instead of hand-rolled helper

says1117:
## Description

`raft::stats::detail` (in `stats/detail/minmax.cuh`) carried a local, `memcpy`-based
`bit_cast` helper with a `TODO` to switch to the standard facility "once we adopt
C++20 or libcu++ adds it." libcu++ now ships `cuda::std::bit_cast` (`<cuda/std/bit>`),
which works in host and device code, so this PR uses it directly and removes the
hand-rolled helper.

## Changes

- Delete the local `bit_cast` template and its `TODO`.
- Add `#include <cuda/std/bit>`.
- Use `cuda::std::bit_cast` at the four `encode`/`decode` call sites.

No functional or numerical behavior change — `cuda::std::bit_cast` is a drop-in
replacement for the previous `memcpy`-based reinterpretation (same size/trivially-
copyable requirements), now sourced from libcu++.

## Testing

No new tests are required; this is a non-functional refactor covered by the existing
suite. Verified locally:

- `STATS_TEST` builds clean.
- All 56 `MinMaxTests` cases pass (28 `MinMaxTestF` + 28 `MinMaxTestD`).


divyegala (approved):
Thanks for the cleanup!

PR #3098, opened 2026-07-29
Mitigate cuBLASLt 13.6 GEMM bug for inputs greater than 2^31

### LINKED ISSUE #3097: [BUG] RAFT GEMM fails at a 2^31-element input span with cuBLASLt 13.6
`raft::linalg::gemm` fails with `CUBLAS_STATUS_EXECUTION_FAILED` when the nominal input span reaches exactly `2^31` elements under `cuBLASLt 13.6`. The adjacent below-boundary case succeeds.

The reproducer uses only RAFT and CUDA runtime allocations. It does not depend on a downstream RAPIDS library or Python package.

## Boundary results

| Case | Nominal input span | Selected algorithm | Result |
|---|---:|---:|---|
| `134,217,727 × 16` | `2^31 - 16` elements | 13, option 62 | Pass |
| `134,217,728 × 16` | exactly `2^31` elements | 68 | `CUBLAS_STATUS_EXECUTION_FAILED` |

The GEMM descriptors reported by `CUBLASLT_LOG_LEVEL=5` for the failing case are:

```text
A: type=R_32F rows=2 cols=134217728 ld=16
B: type=R_32F rows=1 cols=2 ld=1
C: type=R_32F rows=134217728 cols=1 ld=134217728
D: type=R_32F rows=134217728 cols=1 ld=134217728
computeType=COMPUTE_32F
transa=OP_T
transb=OP_T
algoId=68
workSpaceSizeInBytes=0
```

At `134,217,727` rows, `cuBLASLt 13.6` selects algorithm 13 with `customOption=62`, and the operation succeeds. At `134,217,728` rows, the heuristic selects algorithm 68, which is returned as supported but fails during `cublasLtMatmul`.

The exact power-of-two boundary suggests a 32-bit span/index limitation in algorithm 68 or its heuristic eligibility logic.

## Environment

- Image: `rapidsai/base@sha256:7d3527573dcfec08e119774118f9edfaa96b542bf0f58d168ea883bd7356df03`
- `libraft`: `26.08.00a59`, build `cuda13_260729_eec9ba69`
- RAFT revision: `eec9ba69be3e7735306bd7779204d2265bd53f37`
- `libcublas`: `13.6.0.2`
- `cuda-cudart`: `13.3.29`
- GPU: NVIDIA RTX PRO 6000 Blackwell Workstation Edition
- Driver: `595.84`
- OS: Ubuntu 24.04 x86_64

The reproducer requires approximately 9 GiB of available device memory.

## Steps to reproduce

1. Save the source below as `reproducer.cu`.

<details>
<summary>RAFT-only reproducer source</summary>

```cpp
#include <raft/linalg/gemm.cuh>

#include <cuda_runtime_api.h>

#include <cstdint>
#include <cstdlib>
#include <exception>
#include <iostream>
#include <limits>
#include <stdexcept>
#include <string>

namespace {

void check_cuda(cudaError_t status, char const* operation)
{
  if (status != cudaSuccess) {
    throw std::runtime_error(std::string{operation} + ": " +
                             cudaGetErrorString(status));
  }
}

}  // namespace

int main(int argc, char** argv)
{
  constexpr std::int64_t default_n_samples = 134217728;
  constexpr int lda                       = 16;
  constexpr int n_informative             = 2;
  constexpr int n_targets                 = 1;

  const auto n_samples =
    argc > 1 ? static_cast<std::int64_t>(std::stoll(argv[1]))
             : default_n_samples;

  if (n_samples <= 0 || n_samples > std::numeric_limits<int>::max()) {
    std::cerr << "n_samples must be in [1, INT_MAX]\n";
    return EXIT_FAILURE;
  }

  const auto a_elements =
    static_cast<std::uint64_t>(n_samples) * lda;

  std::cout << "n_samples=" << n_samples
            << ", lda=" << lda
            << ", nominal_A_elements=" << a_elements << '\n';

  float* a = nullptr;
  float* b = nullptr;
  float* c = nullptr;

  try {
    check_cuda(
      cudaMalloc(reinterpret_cast<void**>(&a),
                 a_elements * sizeof(float)),
      "cudaMalloc(A)");

    check_cuda(
      cudaMalloc(reinterpret_cast<void**>(&b),
                 n_informative * sizeof(float)),
      "cudaMalloc(B)");

    check_cuda(
      cudaMalloc(reinterpret_cast<void**>(&c),
                 n_samples * sizeof(float)),
      "cudaMalloc(C)");

    raft::resources resources;
    const auto stream = raft::resource::get_cuda_stream(resources);

    check_cuda(
      cudaMemsetAsync(a, 0, a_elements * sizeof(float), stream.value()),
      "cudaMemsetAsync(A)");

    check_cuda(
      cudaMemsetAsync(b, 0, n_informative * sizeof(float), stream.value()),
      "cudaMemsetAsync(B)");

    check_cuda(
      cudaMemsetAsync(c, 0, n_samples * sizeof(float), stream.value()),
      "cudaMemsetAsync(C)");

    const float alpha = 1.0f;
    const float beta  = 0.0f;

    // Exact layout:
    //
    // A: rows=2, cols=n_samples, ld=16, OP_T
    // B: rows=1, cols=2,         ld=1,  OP_T
    // C: rows=n_samples, cols=1, ld=n_samples
    raft::linalg::gemm(resources,
                       true,
                       true,
                       static_cast<int>(n_samples),
                       n_targets,
                       n_informative,
                       &alpha,
                       a,
                       lda,
                       b,
                       n_targets,
                       &beta,
                       c,
                       static_cast<int>(n_samples),
                       stream.value());

    check_cuda(
      cudaStreamSynchronize(stream.value()),
      "cudaStreamSynchronize");

    std::cout << "PASS\n";
  } catch (std::exception const& error) {
    std::cerr << "FAIL: " << error.what() << '\n';
    cudaFree(c);
    cudaFree(b);
    cudaFree(a);
    return EXIT_FAILURE;
  }

  check_cuda(cudaFree(c), "cudaFree(C)");
  check_cuda(cudaFree(b), "cudaFree(B)");
  check_cuda(cudaFree(a), "cudaFree(A)");

  return EXIT_SUCCESS;
}
```

</details>

2. Build and run both boundary cases in the affected image:

```bash
docker run --rm --gpus all --user root \
  -v "$PWD/reproducer.cu:/tmp/reproducer.cu:ro" \
  --entrypoint bash \
  rapidsai/base@sha256:7d3527573dcfec08e119774118f9edfaa96b542bf0f58d168ea883bd7356df03 \
  -lc '
    set -eux

    apt-get update
    apt-get install -y --no-install-recommends g++

    nvcc -std=c++20 -O2 \
      /tmp/reproducer.cu \
      -o /tmp/reproducer \
      -ccbin /usr/bin/g++ \
      -I/opt/conda/targets/x86_64-linux/include/cccl \
      -I/opt/conda/include \
      -L/opt/conda/lib \
      -Xlinker -rpath \
      -Xlinker /opt/conda/lib \
      -lraft \
      -lrmm \
      -lcublasLt \
      -lcublas \
      -lcudart

    CUBLASLT_LOG_LEVEL=5 /tmp/reproducer 134217727
    CUBLASLT_LOG_LEVEL=5 /tmp/reproducer 134217728
  '
```

## Actual behavior

```text
n_samples=134217727, lda=16, nominal_A_elements=2147483632
PASS

n_samples=134217728, lda=16, nominal_A_elements=2147483648
FAIL: cuBLAS error encountered at:
.../raft/linalg/detail/cublaslt_wrappers.hpp line=299:
call='cublasLtMatmul(...)',
Reason=13:CUBLAS_STATUS_EXECUTION_FAILED
```

## Expected behavior

Both valid GEMM configurations should succeed. If algorithm 68 cannot support this layout, it should not be returned by `cublasLtMatmulAlgoGetHeuristic`.

## Additional controls

The same exact-boundary operation succeeds with `cuBLASLt 13.1.1`. That version selects algorithm 13 with `customOption=14`.

RAFT currently requests one heuristic result and executes it without an alternative-algorithm fallback: 
https://github.com/NVIDIA/raft/blob/eec9ba69be3e7735306bd7779204d2265bd53f37/cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp#L192-L201

The primary defect appears to be in `cuBLASLt 13.6`, but RAFT could potentially mitigate it by requesting alternative algorithms, falling back to classic cuBLAS for large physical spans, or chunking the operation.

divyegala:
Closes https://github.com/NVIDIA/raft/issues/3097

### DIFF
--- a/cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp
+++ b/cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp
@@ -101,15 +101,15 @@ struct matmul_key_hash {
 };
 
 /**
- * cuBLASLt 13.6 selects algorithm 68 for this FP32 layout once A's physical span reaches
- * 2^31 elements, but that algorithm fails during execution. Prefer the validated alternatives.
+ * cuBLASLt 13.6 may select algorithm 68 once A's physical span reaches 2^31 elements, but that
+ * algorithm fails during execution for FP32. Prefer the validated alternatives.
  */
 inline auto needs_cublaslt_13_6_workaround(const matmul_key_t& args, std::size_t version) noexcept
   -> bool
 {
   constexpr uint64_t max_safe_span = (uint64_t{1} << 31) - 1;
-  return version == 130600 && args.trans_a && args.trans_b && args.n == 1 && args.k == 2 &&
-         args.lda == 16 && args.ldb == 1 && args.ldc == args.m && args.m > max_safe_span / args.lda;
+  const auto a_columns             = args.trans_a ? args.m : args.k;
+  return version == 130600 && args.lda != 0 && a_columns > max_safe_span / args.lda;
 }
 
 inline auto get_cublaslt_algorithm_id(const cublasLtMatmulHeuristicResult_t& heuristic) -> int
--- a/cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp
+++ b/cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2024-2026, NVIDIA CORPORATION.
+ * SPDX-FileCopyrightText: Copyright (c) 2024-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
 #pragma once
@@ -18,6 +18,7 @@
 
 #include <cublasLt.h>
 
+#include <array>
 #include <type_traits>
 
 namespace raft {
@@ -98,6 +99,29 @@ struct matmul_key_hash {
     return x.m * x.n * x.k + x.lda * x.ldb * x.ldc + size_t{x.trans_a} + size_t{x.trans_b} * 2;
   }
 };
+
+/**
+ * cuBLASLt 13.6 selects algorithm 68 for this FP32 layout once A's physical span reaches
+ * 2^31 elements, but that algorithm fails during execution. Prefer the validated alternatives.
+ */
+inline auto needs_cublaslt_13_6_workaround(const matmul_key_t& args, std::size_t version) noexcept
+  -> bool
+{
+  constexpr uint64_t max_safe_span = (uint64_t{1} << 31) - 1;
+  return version == 130600 && args.trans_a && args.trans_b && args.n == 1 && args.k == 2 &&
+         args.lda == 16 && args.ldb == 1 && args.ldc == args.m && args.m > max_safe_span / args.lda;
+}
+
+inline auto get_cublaslt_algorithm_id(const cublasLtMatmulHeuristicResult_t& heuristic) -> int
+{
+  int algorithm_id{};
+  std::size_t bytes_written{};
+  RAFT_CUBLAS_TRY(cublasLtMatmulAlgoConfigGetAttribute(
+    &heuristic.algo, CUBLASLT_ALGO_CONFIG_ID, &algorithm_id, sizeof(algorithm_id), &bytes_written));
+  return algorithm_id;
+}
+
+inline constexpr std::array<int, 2> cublaslt_13_6_fallback_algorithms{13, 16};
 
 /** Descriptor for a column-major cublasLt matrix. */
 struct cublastlt_matrix_layout {
@@ -186,6 +210,16 @@ struct matmul_desc {
       cublastlt_matrix_layout::for_matmul<A>(!(args.trans_a), args.m, args.k, args.lda),
       cublastlt_matrix_layout::for_matmul<B>(!(args.trans_b), args.k, args.n, args.ldb),
       cublastlt_matrix_layout::for_matmul<C>(true, args.m, args.n, args.ldc)};
+
+    bool use_cublaslt_13_6_workaround = false;
+    if constexpr (std::is_same_v<S, float> && std::is_same_v<A, float> &&
+                  std::is_same_v<B, float> && std::is_same_v<C, float>) {
+      use_cublaslt_13_6_workaround = needs_cublaslt_13_6_workaround(args, cublasLtGetVersion());
+    }
+
+    constexpr int max_heuristic_results = 8;
+    std::array<cublasLtMatmulHeuristicResult_t, max_heuristic_results> heuristic_results{};
+    const int requested_results = use_cublaslt_13_6_workaround ? max_heuristic_results : 1;
     int algo_count;
     cublasLtMatmulPreference_t preference;
     RAFT_CUBLAS_TRY(cublasLtMatmulPreferenceCreate(&preference));
@@ -196,10 +230,29 @@ struct matmul_desc {
                                                    r.c,
                                                    r.c,
                                                    preference,
-                                                   1,
-                                                   &r.heuristics,
+                                                   requested_results,
+                                                   heuristic_results.data(),
                                                    &algo_count));
     RAFT_CUBLAS_TRY(cublasLtMatmulPreferenceDestroy(preference));
+
+    RAFT_EXPECTS(algo_count > 0, "cuBLASLt did not return a matmul algorithm");
+    if (!use_cublaslt_13_6_workaround) {
+      r.heuristics = heuristic_results.front();
+      return r;
+    }
+
+    for (const auto preferred_algorithm : cublaslt_13_6_fallback_algorithms) {
+      for (int i = 0; i < algo_count; ++i) {
+        const auto& candidate = heuristic_results[i];
+        if (candidate.state == CUBLAS_STATUS_SUCCESS && candidate.workspaceSize == 0 &&
+            get_cublaslt_algorithm_id(candidate) == preferred_algorithm) {
+          r.heuristics = candidate;
+          return r;
+        }
+      }
+    }
+
+    RAFT_FAIL("cuBLASLt 13.6 did not return algorithm 13 or 16 for the affected large FP32 GEMM");
     return r;
   }
 };
--- a/cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp
+++ b/cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp
@@ -102,8 +102,8 @@ struct matmul_key_hash {
 };
 
 /**
- * cuBLASLt 13.6 may select algorithm 68 once A's physical span reaches 2^31 elements. That
- * algorithm fails during execution for FP32, so select the next ranked heuristic instead.
+ * cuBLASLt 13.6 and later may select algorithm 68 once A's physical span reaches 2^31 elements.
+ * That algorithm fails during execution for FP32, so select the next ranked heuristic instead.
  */
 inline auto needs_cublaslt_13_6_workaround(const matmul_key_t& args,
                                            std::size_t version,
@@ -115,7 +115,7 @@ inline auto needs_cublaslt_13_6_workaround(const matmul_key_t& args,
   const bool is_affected_architecture =
     (device_major == 10 && device_minor == 0) ||
     (device_major == 12 && (device_minor == 0 || device_minor == 1));
-  return version == 130600 && is_affected_architecture && args.lda != 0 &&
+  return version >= 130600 && is_affected_architecture && args.lda != 0 &&
          a_columns > max_safe_span / args.lda;
 }
 
--- a/cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp
+++ b/cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp
@@ -101,8 +101,8 @@ struct matmul_key_hash {
 };
 
 /**
- * cuBLASLt 13.6 may select algorithm 68 once A's physical span reaches 2^31 elements, but that
- * algorithm fails during execution for FP32. Prefer the validated alternatives.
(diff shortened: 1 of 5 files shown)


cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp
```diff
@@ -99,6 +100,29 @@ struct matmul_key_hash {
...
+/**
+ * cuBLASLt 13.6 may select algorithm 68 once A's physical span reaches 2^31 elements, but that
+ * algorithm fails during execution for FP32. Prefer the validated alternatives.
+ */
+inline auto needs_cublaslt_13_6_workaround(const matmul_key_t& args, std::size_t version) noexcept
+  -> bool
+{
+  constexpr uint64_t max_safe_span = (uint64_t{1} << 31) - 1;
+  const auto a_columns             = args.trans_a ? args.m : args.k;
+  return version == 130600 && args.lda != 0 && a_columns > max_safe_span / args.lda;
```
csadorf: Did you test other versions? Are we confident that only this very specific version is affected? What about future versions?

cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp
```diff
@@ -186,6 +210,16 @@ struct matmul_desc {
...
       cublastlt_matrix_layout::for_matmul<B>(!(args.trans_b), args.k, args.n, args.ldb),
       cublastlt_matrix_layout::for_matmul<C>(true, args.m, args.n, args.ldc)};
+
+    bool use_cublaslt_13_6_workaround = false;
+    if constexpr (std::is_same_v<S, float> && std::is_same_v<A, float> &&
+                  std::is_same_v<B, float> && std::is_same_v<C, float>) {
+      use_cublaslt_13_6_workaround = needs_cublaslt_13_6_workaround(args, cublasLtGetVersion());
+    }
+
+    constexpr int max_heuristic_results = 8;
```
csadorf: Is eight an empirically determined number? Is it guaranteed to give us enough options?

cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp
```diff
@@ -99,6 +100,29 @@ struct matmul_key_hash {
...
+inline auto get_cublaslt_algorithm_id(const cublasLtMatmulHeuristicResult_t& heuristic) -> int
+{
+  int algorithm_id{};
+  std::size_t bytes_written{};
+  RAFT_CUBLAS_TRY(cublasLtMatmulAlgoConfigGetAttribute(
+    &heuristic.algo, CUBLASLT_ALGO_CONFIG_ID, &algorithm_id, sizeof(algorithm_id), &bytes_written));
+  return algorithm_id;
+}
+
+inline constexpr std::array<int, 2> cublaslt_13_6_fallback_algorithms{13, 16};
```
csadorf: Would it make sense to maybe just exclude the known faulty algos rather than selecting 13, 16, specifically?

cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp
```diff
@@ -99,6 +100,29 @@ struct matmul_key_hash {
   }
 };
 
+/**
+ * cuBLASLt 13.6 may select algorithm 68 once A's physical span reaches 2^31 elements, but that
+ * algorithm fails during execution for FP32. Prefer the validated alternatives.
+ */
+inline auto needs_cublaslt_13_6_workaround(const matmul_key_t& args, std::size_t version) noexcept
```
csadorf: I'd recommend checking whether the work-around is needed on all archs or only sm_120/sm_121.

csadorf (commented):
Have you considered constraining our dependency?

cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp
divyegala: This will be a bit tough to verify this close to the release.

cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp
divyegala: Fair, in case the algorithm changes.

cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp
divyegala: No, but the tests should help with that if it fails in a future version. If you prefer being defensive, then I can add a >= check instead?

cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp
divyegala: No, just a large enough number. The thought process is that for size exceeding 2**31 on cublas 13.6, if algorithm 68 is returned as the best by the heuristic search then skip it and try the next one. So I'll just request 2 plans and disallow 68 if it's the top one.

cpp/tests/linalg/gemm_large.cpp
```diff
@@ -0,0 +1,78 @@
...
+  constexpr std::array<std::int64_t, 3> sample_counts{134217727, 134217728, 134217729};
+  constexpr auto max_samples = sample_counts.back();
+  constexpr auto a_elements  = max_samples * lda;
+
+  raft::resources resources;
+  const auto stream = raft::resource::get_cuda_stream(resources);
+
+  auto a = raft::make_device_vector<float, std::int64_t>(resources, a_elements);
+  auto b = raft::make_device_vector<float, std::int64_t>(resources, informative);
+  auto c = raft::make_device_vector<float, std::int64_t>(resources, max_samples);
```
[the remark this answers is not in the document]
divyegala: The test requests 100% of the GPU in the CMake, and all modern GPUs exceed the capacity requested here.

cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp
csadorf: I think we might have to be defensive here and actually set 130600 as a lower bound unless we get some kind of assurance that this will be fixed in the next patch release.

cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp
csadorf: I like that more.

cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp
csadorf: Yes, but I think we should at least check sm_90 since restricting the algorithm might lead to a performance regression.

divyegala: Good point, let me skip the heuristic for non SM120/121 and CI tests should give us an indication for sm80 and sm90.

cpp/tests/linalg/gemm_basic.cpp
```diff
@@ -163,4 +163,56 @@ TEST(Raft, GemmPointerModeDeviceAlpha) { test_gemm_pointer_mode_device(true, fal
...
+  constexpr std::size_t affected_version = 130600;
+  const detail::matmul_key_t at_boundary{134217728, 1, 2, 16, 1, 134217728, true, true};
+
+  EXPECT_TRUE(detail::needs_cublaslt_13_6_workaround(at_boundary, affected_version, 12, 0));
+  EXPECT_TRUE(detail::needs_cublaslt_13_6_workaround(at_boundary, affected_version, 12, 1));
+  EXPECT_FALSE(detail::needs_cublaslt_13_6_workaround(at_boundary, affected_version, 7, 5));
+  EXPECT_FALSE(detail::needs_cublaslt_13_6_workaround(at_boundary, affected_version, 8, 0));
+  EXPECT_FALSE(detail::needs_cublaslt_13_6_workaround(at_boundary, affected_version, 8, 9));
+  EXPECT_FALSE(detail::needs_cublaslt_13_6_workaround(at_boundary, affected_version, 9, 0));
+}
```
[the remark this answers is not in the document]
divyegala: There's no SM 122

cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp
```diff
@@ -186,6 +213,18 @@ struct matmul_desc {
...
+    if constexpr (std::is_same_v<S, float> && std::is_same_v<A, float> &&
+                  std::is_same_v<B, float> && std::is_same_v<C, float>) {
+      const auto& device_properties = resource::get_device_properties(res);
+      use_cublaslt_13_6_workaround  = needs_cublaslt_13_6_workaround(
+        args, cublasLtGetVersion(), device_properties.major, device_properties.minor);
+    }
+
+    constexpr int workaround_heuristic_results = 2;
+    std::array<cublasLtMatmulHeuristicResult_t, workaround_heuristic_results> heuristic_results{};
+    const int requested_results = use_cublaslt_13_6_workaround ? workaround_heuristic_results : 1;
```
dantegd, line 229: Is it guaranteed or documented that these affected GEMMs always have a usable non-68 algorithm in the top two heuristic results?

dantegd (approved):
One quick question, besides it lgtm

cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp
divyegala: No, but this only skips 68 if it's at the top of the heuristics. If nothing but 68 is returned we will have to fail.

gforsyth (approved):

cjnolet (approved):

PR #3106, opened 2026-08-10
[REVIEW] Addition of `gemm_strided_batched`

landrumb:
This PR adds the wrapper for strided and batched matrix multiplication with cublasLt requested by @achirkin in the reviews for https://github.com/NVIDIA/cuvs/pull/2352 , where this operation is used for batched row-wise dot products. TF32 support is included. 

The associated test covers most of the functionality besides eg transposed inputs, but can be expanded if reviewer(s) prefer.


cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp
```diff
@@ -348,6 +368,70 @@ struct coef_wrapper<true, S> {
...
+
+  auto a_layout = cublastlt_matrix_layout::for_strided_batched_matmul<A>(
+    !trans_a, m, k, lda, batch_count, stride_a);
+  auto b_layout = cublastlt_matrix_layout::for_strided_batched_matmul<B>(
+    !trans_b, k, n, ldb, batch_count, stride_b);
+  auto c_layout =
+    cublastlt_matrix_layout::for_strided_batched_matmul<C>(true, m, n, ldc, batch_count, stride_c);
+
+  auto stream = resource::get_cuda_stream(res);
+  coef_wrapper<DevicePointerMode, S> coefficients(alpha, beta, stream);
```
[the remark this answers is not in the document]
landrumb: I'm pretty sure this validation is done by cuBLASLt and would therefore be redundant.

cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp
```diff
@@ -348,6 +368,70 @@ struct coef_wrapper<true, S> {
...
+                                 b_layout,
+                                 coefficients.beta,
+                                 c_ptr,
+                                 c_layout,
+                                 c_ptr,
+                                 c_layout,
+                                 nullptr,
+                                 nullptr,
+                                 0,
+                                 stream));
```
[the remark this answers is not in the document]
landrumb: This nullptr arg for algo is supported.

cpp/include/raft/linalg/gemm.cuh
```diff
@@ -283,6 +283,81 @@ void gemm(raft::resources const& res,
...
+  int64_t stride_a,
+  const B_t* b,
+  uint64_t ldb,
+  int64_t stride_b,
+  const S_t* beta,
+  C_t* c,
+  uint64_t ldc,
+  int64_t stride_c,
+  int32_t batch_count,
+  cublasComputeType_t compute_type = detail::get_matmul_type<S_t, A_t, B_t, C_t>())
```
achirkin: Please rewrite the public API to use mdspans - we should not add new legacy overloads to public functions.
Also note, we can encode the `_strided_` part as in the layout policy -  it doesn't need to be a part of the name; I'd suggest we make it here generic and pass the stride = matrix size if the layout is not strided.
```suggestion
template <typename ValueType,
          typename IndexType,
          typename LayoutPolicyX,
          typename LayoutPolicyY,
          typename LayoutPolicyZ,
          typename ScalarIdxType  = std::uint32_t,
          typename ScalarViewType = raft::host_scalar_view<ValueType, ScalarIdxType>,
          typename                = std::enable_if_t<std::disjunction_v<
                           std::is_same<ScalarViewType, raft::host_scalar_view<ValueType, ScalarIdxType>>,
                           std::is_same<ScalarViewType, raft::device_scalar_view<ValueType, ScalarIdxType>>>>>
void gemm_batched(raft::resources const& res,
          raft::device_mdspan<ValueType, raft::extent_3d<IndexType>, LayoutPolicyX> x,
          raft::device_mdspan<ValueType, raft::extent_3d<IndexType>, LayoutPolicyY> y,
          raft::device_mdspan<ValueType, raft::extent_3d<IndexType>, LayoutPolicyZ> z,
          std::optional<ScalarViewType> alpha = std::nullopt,
          std::optional<ScalarViewType> beta  = std::nullopt,
          cublasComputeType_t compute_type_override = detail::get_matmul_type<ValueType, ValueType, ValueType, ValueType>()))
```

achirkin: Also I like the way you suggest to optionally pass the cublas compute type override - looks like a good compromise between the user-friendliness and flexibility.
However, maybe we could sacrifice a bit of brevity add replace the optional default argument with two overloads (with/without argument) - to avoid `detail` namespace from appearing in the public headers?

achirkin (changes requested):
Thanks for implementing this! Let's refine the API (see below)

cpp/include/raft/linalg/gemm.cuh
landrumb: Haha yeah needed that because I use TF32 in the other PR (iirc because you suggested it in one of your comments). Switched to overload in https://github.com/NVIDIA/raft/pull/3106/commits/406a785095a0e5e074265582864168b6e2a0c79f

cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp
```diff
@@ -348,6 +368,70 @@ struct coef_wrapper<true, S> {
...
+                            int64_t stride_b,
+                            const S* beta,
+                            C* c_ptr,
+                            uint64_t ldc,
+                            int64_t stride_c,
+                            int32_t batch_count,
+                            cublasComputeType_t compute_type)
+{
+  common::nvtx::range<common::nvtx::domain::raft> batch_scope(
+    "linalg::matmul_strided_batched(m = %d, n = %d, k = %d, batch_count = %d)",
```
achirkin: nitpick: please align the nvtx annotation to either `linalg::detail::matmul_strided_batched` or `linalg::matmul_batched`.

achirkin (approved):
Looks very nice, thanks!
Seems like there's an invalid cublas config in the test, otherwise I think this is ready to merge.

PR #3109, opened 2026-08-11
Fix lanczos launch grid

achirkin:
Fix accidentally swapped block size and grid size arguments.

https://github.com/NVIDIA/raft/pull/3104#discussion_r3730249841


divyegala (commented):
Could you please add a regression test?

divyegala (approved):

aamijar (approved):
Good catch!

PR #3110, opened 2026-08-12
add-batched-linalg for cuvs

Intron7:
This adds batched linalg function for cuvs consumption in GMM @cjnolet


achirkin (changes requested):
Ok, sure! Then, this PR requires a bit of rework, because the only public API it exposes (batched gemm) is already implemented in https://github.com/NVIDIA/raft/pull/3106

PR #3116, opened 2026-08-13
fix: remove unused C compiler from libraft outputs

andrewwhitecdw:
This PR removes the unused C compiler dependency from the `libraft`, `libraft-static`, and `libraft-headers-only` outputs in `conda/recipes/libraft/recipe.yaml`. The install scripts for these outputs only run `cmake --install`, so they do not need a compiler at install time.

Per maintainer feedback, the separate PRs for `libraft` (#3114) and `libraft-static` (#3115) have been closed and their changes are incorporated here so the shared verification question can be answered once.

## Changes
- `conda/recipes/libraft/recipe.yaml`: removed `${{ compiler("c") }}` from the `build` requirements of `libraft`, `libraft-static`, and `libraft-headers-only`.

## Testing
Covered by existing conda recipe build paths; no executable code changed. I will follow up with local recipe-build verification without the compilers metapackage as requested.

### DIFF
--- a/cpp/include/raft/common/nccl_macros.hpp
+++ b/cpp/include/raft/common/nccl_macros.hpp
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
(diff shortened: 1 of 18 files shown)


gforsyth (changes requested):
Thanks for putting this in @andrewwhitecdw -- before we kick off a full CI build-and-test, can you verify locally that you can build the new recipe without the compilers metapackage available?

Copying and pasting the relevant terminal output is sufficient proof.  Thanks!

PR #3118, opened 2026-08-18
Performance improvements to sparse MST solver

alexfallin:
## Summary

Replaces the internals of `raft::sparse::solver::mst` with a solver based on [ECL-MST, SC'23](https://dl.acm.org/doi/10.1145/3581784.3607093). The public API is unchanged. Ties break deterministically on (weight order key, edge index) instead of randomized weight alteration, which both fixes float-precision issues and removes the alteration cost.

Two main features:
1. **Drop-in replacement** `detail/mst_kernels.cuh`, `detail/mst_solver_inl.cuh` rewritten, `MST_solver` reduced `detail/mst_utils.cuh` deleted (its only reference was in the old kernels)
2. **New capability**: 64-bit `vertex_t`/`edge_t` support (uniform 64-bit and mixed 32-bit vertex / 64-bit edge), validated to 2.25B edges on H200 and 4.29B edges on ~250GB GPU memory (only real limiting factor is GPU memory capacity now, edges more expensive than vertices due to the worklists). When the extra capacity isn't required, the 32-bit instantiations keep the faster performance.

## Algorithm overview

ECL-MST is an edge-parallel approach. Setup adds each undirected edge into a worklist (an optional two-phase filter first solves over the light edges when the average degree makes it worthwhile). Each round then runs three edge-parallel steps over the worklist:

1. **Min-selection**: every edge whose endpoints have different parents (analogous to colors) proposes itself to both endpoint components via `atomicMin` on a per-component slot, using a packed **(weight order key, edge index)** value the index half makes ties deterministic, which is what lets this PR delete weight alteration.
2. **Select + join**: an edge that won at least one of its endpoint slots is in the MST. The two components are merged in a lock-free union-find (join by `atomicCAS` on parent pointers).
3. **Compact**: components are flattened, and surviving edges (endpoints still in different components) are compacted into the opposite worklist. Rounds repeat until no edge survives.

There is now a narrow and wide path. The **narrow path** (4-byte weight and edge index) packs the key into
64 bits and uses hardware `atomicMin`. This is the path that all the existing downstream implementations (cuGraph, cuML, cuVS) would use.

Changes made for this PR beyond the published algorithm:
- **A wide path** for the type combinations the paper's packing cannot represent (8-byte weights and/or 8-byte edge indices) using a 128-bit (key, index) min-selection, implemented as a single-pass 128-bit CAS loop on >sm_89 and a portable two-pass min (weight key, then edge index among key-ties) when <=sm_89. Both produce the same results as each other and as the narrow path. When available, the 128-bit CAS is faster, so it's selected at compile time if the compute capability is there.
- **64-bit indexing throughout** enabling the >2B-edge graphs in the validation section. For speed, 32-bit instantiations keep 32-bit counters.
- **Path halving in the union-find** to avoid a possible extreme runtime case. Published ECL-MST's find is vulnerable to a quadratic complexity on large equal-weight tie graphs. This did cost some perf but was negligible when compared to a multi-minute hang.
- **Single-instruction inline-PTX word accessors** for the intentionally racy `parent[]` reads/writes, pinning them against compiler tearing.
- **RAFT API semantics preserved on top of the algorithm** resume from prior colors, bounded rounds, optional output symmetrization, and MSF colors.

## Correctness fixes over the current solver

1. **Float alteration underflow**: the tie-breaking perturbation (bounded by min-weight-gap/2) rounds to zero at weight magnitudes ~1e7. On the SC'23 suite at float: 3 graphs throw the precision guard, 2 return silently wrong results past the guard (I actually observed this when testing back when I was doing ECL-MST, I just didn't know the reason until now).
2. **Thread-index overflow at v > 2^26**: `kernel_min_edge_per_vertex` (launched `<<<v, 32>>>`) computes its thread id in 32-bit arithmetic. Above 2^26 vertices (int) / 2^27 (int64) vertices silently stop participating and a truncated forest is returned past the guards.

## Performance (H200, best-of-9 for both, interleaved, w/ correctness validated)
Geomean vs current solver on the SC'23 suite: **6.25x float** (over the 13 graphs the current solver completes correctly), **7.33x double** (all 18).
| Graph | V | E | old float (ms) | new float (ms) | float speedup | old double (ms) | new double (ms) | double speedup |
|---|--:|--:|--:|--:|--:|--:|--:|--:|
| internet | 124.7K | 387K | 2.770 | 0.296 | 9.36x | 2.827 | 0.327 | 8.65x |
| USA-road-d.NY | 264K | 730K | 5.098 | 0.378 | 13.49x | 5.208 | 0.453 | 11.50x |
| rmat16.sym | 65.5K | 968K | 2.381 | 0.537 | 4.43x | 2.572 | 0.572 | 4.50x |
| citationCiteseer | 268K | 2.3M | 4.246 | 0.728 | 5.83x | 4.494 | 0.903 | 4.98x |
| 2d-2e20.sym | 1.05M | 4.2M | 12.396 | 1.343 | 9.23x | 12.718 | 1.503 | 8.46x |
| amazon0601 | 403K | 4.9M | 5.395 | 0.931 | 5.79x | 5.850 | 1.074 | 5.45x |
| as-skitter | 1.7M | 22.2M | 22.247 | 2.819 | 7.89x | 25.727 | 3.201 | 8.04x |
| in-2004 | 1.4M | 27.2M | 20.765 | 2.843 | 7.30x | 25.228 | 3.433 | 7.35x |
| coPapersDBLP | 540K | 30.5M | 9.190 | 2.570 | 3.58x | 10.888 | 2.958 | 3.68x |
| cit-Patents | 3.8M | 33.0M | 36.457 | 6.450 | 5.65x | 38.737 | 7.375 | 5.25x |
| USA-road-d.USA | 23.9M | 57.7M | threw[^1] | 10.166 | — | 259.236 | 12.699 | 20.41x |
| rmat22.sym | 4.2M | 65.7M | wrong[^2] | 7.928 | — | 40.865 | 9.020 | 4.53x |
| r4-2e23.sym | 8.4M | 67.1M | 147.712 | 14.981 | 9.86x | 154.212 | 16.869 | 9.14x |
| soc-LiveJournal1 | 4.8M | 85.7M | 48.275 | 9.816 | 4.92x | 54.698 | 11.014 | 4.97x |
| delaunay_n24 | 16.8M | 100.7M | threw[^1] | 15.604 | — | 194.258 | 18.624 | 10.43x |
| europe_osm | 50.9M | 108.1M | threw[^1] | 18.389 | — | 1055.941 | 22.129 | 47.72x |
| kron_g500-logn21 | 2.1M | 182.1M | 91.920 | 41.049 | 2.24x | 130.872 | 46.258 | 2.83x |
| uk-2002 | 18.5M | 523.6M | wrong[^2] | 48.219 | — | 311.016 | 58.048 | 5.36x |
| **geomean** | | | | | **6.25x** (13 graphs) | | | **7.33x** (18 graphs) |

[^1]: current solver throws its "loss in precision" guard at float precision.
[^2]: current solver completes at float but returns a non-minimal/invalid forest (excluded from the float geomean; times against wrong results are not comparable).

H200 (sm_90), CUDA 13.3, best-of-9 per graph, H2D transfer excluded, both
binaries built from one harness source in the same session and executed
interleaved per graph; every run validated against a Kruskal oracle.
Colors are FNV-checksum identical to the current solver on every run
where it is correct.

## Determinism contract
The forest **edge set, edge count, total weight, and colors are deterministic** run-to-run, across architectures, and across the CAS128/two-pass implementations. The **order** of edges within the output COO is unspecified. Note: from the float domain the order key distinguishes -0.0 < +0.0 because of the sign bit.

## Observable behavior changes (possible downstream issues?)
- Tie-breaking is deterministic instead of randomized. Downstream snapshot tests (e.g. dendrograms) could notice
- cuVS issue. PR: https://github.com/NVIDIA/cuvs/pull/2478
- Output COO ordering differs from old solver

## Notes for reviewers

- **Toolkit compatibility**: the wide worklist entry is `longlong4_16a` on CUDA >= 13.0 and plain `longlong4` on older toolkits (the `_16a` aligned vector types do not exist before 13.0, and `longlong4` is deprecated from 13.0)
- **Kernel launches** use the new `raft::launch_kernel` dispatcher (#3104)
- **Inline-PTX word helpers**: I wanted to force non-tearing accesses that wouldn't bypass the L1 like an atomic would. Measured atomics in the find at +27-215% runtime (pretty large perf hit)
- **NaN weights** order deterministically by bit pattern
- **Two-pass wide path CI coverage**: sm_90+ selects the single-pass 128-bit CAS implementation at compile time, so once CI hardware is all sm_90+ the portable two-pass path is never built. `-DRAFT_MST_FORCE_TWOPASS` compiles it anywhere if it needs to be tested on 90+ hardware
- Input CSR must be symmetric (already implicit but not enforced in existing solver)
- Resume (`initialize_colors = false`) requires colors from a previous solve (documented precondition; malformed seeds can hang the union-find
- The e/2+1-entry worklists dominate memory. I found that to actually use memory near device capacity, I had to preallocate the RMM pool. Default pool growth strands some memory preventing full use of the device memory

## Validation summary

- gtests: original 7 fixtures kept; 8 added (Kruskal-exact float ~1e7 regression, symmetrize pairing, resume == one-shot, disconnected colors, uniform-weight path pathology (path compression cause), double weights, integer weights, and the 64-bit/mixed/unsigned instantiations)
- Differential fuzzing: 50k+ iterations, 13 graph families x 9 weight regimes, exact edge-set/colors equality vs a tie-break-matched Kruskal oracle, including 3-way instantiation cross-checks and resume paths
- compute-sanitizer memcheck/racecheck/initcheck: clean across gtests, all instantiations, both wide implementations
- Scale: correct at e = INT_MAX-67 (int32), 2.25B edges (int64, H200), 4.29B edges / 1.35B vertices (~250GB device)


cpp/include/raft/sparse/solver/detail/mst_kernels.cuh
```diff
@@ -1,324 +1,472 @@
...
+}
+
+template <typename T>
+RAFT_DEVICE_INLINE_FUNCTION T mst_atomic_add(T* addr, T val)
+{
+  if constexpr (sizeof(T) == 4) {
+    return static_cast<T>(
+      atomicAdd(reinterpret_cast<unsigned int*>(addr), static_cast<unsigned int>(val)));
+  } else {
+    return static_cast<T>(atomicAdd(reinterpret_cast<mst_ull*>(addr), static_cast<mst_ull>(val)));
```
achirkin, line 108: Please replace with primitives from https://github.com/NVIDIA/raft/blob/main/cpp/include/raft/util/device_atomics.cuh (which you can extend if needed)

cpp/include/raft/sparse/solver/detail/mst_kernels.cuh
```diff
@@ -1,324 +1,472 @@
...
-      atomicMin(&min_edge_color[self_color], min_edge_weight[0]);
-    }
+template <typename T>
+RAFT_DEVICE_INLINE_FUNCTION void mst_word_store(T* addr, T val)
+{
+  if constexpr (sizeof(T) == 4) {
+    asm volatile("st.b32 [%0], %1;" ::"l"(addr), "r"(static_cast<unsigned int>(val)) : "memory");
+  } else {
+    asm volatile("st.b64 [%0], %1;" ::"l"(addr), "l"(static_cast<unsigned long long>(val))
+                 : "memory");
```
achirkin, line 136: Please check if you can use the primitives from https://github.com/NVIDIA/raft/blob/main/cpp/include/raft/util/device_loads_stores.cuh (also can be extended if needed).

cpp/include/raft/sparse/solver/detail/mst_solver_inl.cuh
```diff
@@ -6,449 +6,371 @@
...
+  const int vblocks =
+    static_cast<int>((static_cast<size_t>(v) + mst_block_size - 1) / mst_block_size);
+  if (initialize_colors) {
+    thrust::sequence(rmm::exec_policy(stream), parent.begin(), parent.end());
+  } else {
+    raft::launch_kernel(
+      stream, vblocks, mst_block_size, mst_init_parent_kernel<vertex_t>, v, color, parent.data());
+  }
+  RAFT_CUDA_TRY(cudaMemsetAsync(minv_raw.data(), 0xFF, minv_bytes, stream));
+  RAFT_CUDA_TRY(cudaMemsetAsync(in_mst.data(), 0, e * sizeof(bool), stream));
```
achirkin, line 101: We have recently merged the dry run PR https://github.com/NVIDIA/raft/pull/2961 . Once you merge the latest main and resolve the conflicts, the coderabbitai bot will likely complain that the dry run compliance is broken here: an unguarded cuda work is performed in thrust::sequence, in the kernel, and in cudaMemSetAsync. 
I'd suggest to pass the `raft::resources` handle to all launch_kernel calls here and replace the thust::sequence and cudaMemsetAsync with raft::linalg::map_offset(handle, ...) and aft::linalg::map(handle, ...).

The rule of thumb: try to always call functions, which take `raft::resources` handle as the first argument - these are always dry run compliant.

cpp/include/raft/sparse/solver/detail/mst_solver_inl.cuh
```diff
@@ -6,449 +6,371 @@
...
+  key_t thr_key = std::numeric_limits<key_t>::max();
+  bool filtered = false;
+  if (iterations == 0 && e / v >= filter_min_avg_degree) {
+    const int ns = static_cast<int>(std::min<edge_t>(e, max_samples));
+    rmm::device_uvector<key_t> keys_d(ns, stream);
+    raft::launch_kernel(
+      stream, 1, 32, mst_sample_keys_kernel<edge_t, weight_t>, ns, e, weights, keys_d.data());
+    key_t keys[max_samples];
+    raft::update_host(keys, keys_d.data(), ns, stream);
+    RAFT_CUDA_TRY(cudaStreamSynchronize(stream));
```
achirkin, line 118:
```suggestion
    raft::resource::sync_stream(handle);
```

cpp/include/raft/sparse/solver/detail/mst_solver_inl.cuh
```diff
@@ -6,449 +6,371 @@
...
+                                                edge_t const* offsets,
+                                                vertex_t const* indices,
+                                                weight_t const* weights,
+                                                vertex_t const v,
+                                                edge_t const e,
+                                                vertex_t* color,
+                                                cudaStream_t stream,
+                                                bool symmetrize_output,
+                                                bool initialize_colors,
+                                                int iterations)
```
achirkin, line 46: It is unfortunate that our API here accepts the stream argument separately from the handle argument. We try to gradually cleanup this - always use the stream that is stored in the handle (`raft::resource::get_cuda_stream(handle)`); this gets somewhat stronger enforced since the recent dry run PR.

To workaround the problem in this function, I'd suggest to copy the handle and set the given stream to the copy. Then use the new handle throughout the function:
```C++
auto res = handle;
raft::resource::set_cuda_stream(res, stream);
```

achirkin (changes requested):
Thanks for the PR. It's exciting to see such a perf boost getting merged into raft!

I have a few comments on the usage of raft API at this time. A big refactoring PR https://github.com/NVIDIA/raft/pull/2961 was merged since you opened this one - merging that will need some non-trivial updates.

cpp/include/raft/sparse/solver/detail/mst_solver_inl.cuh
divyegala: We could also make a breaking change

alexfallin: So, if I'm understanding right, the two options would be:

1. Change the API to take the handle only (the breaking change), or
2. Keep the API non-breaking and do the suggested `auto res = handle;
raft::resource::set_cuda_stream(res, stream);`

Which would y'all prefer?

divyegala: I'm okay with 1 as you already have downstream PRs open!

alexfallin: I kinda went in-between for this one. The existing cuVS pr is more of a correctness thing than a full API overhaul so I think it might be better to not do full breaking. I just overloaded the existing API to have the new cleaned up one and the old one but with a deprecation warning. If you want the full breaking change, I can do that, just figured this is a bit lower friction.

cpp/include/raft/sparse/solver/detail/mst_solver_inl.cuh
alexfallin: Done

cpp/include/raft/sparse/solver/detail/mst_solver_inl.cuh
alexfallin: Done as you suggested. Also added the dry-run stuff which predicts a real solve's peak bc every allocation here is sized from v/e/flags.

cpp/include/raft/sparse/solver/detail/mst_kernels.cuh
alexfallin: Done, via "extend if needed". The header now has `raft::ldg_ca` / `raft::stg_wb` which is single-instruction 4/8-byte global access with the default .ca/.wb cache policy. The existing `ldg` wasn't a substitute because it bypasses the L1. The union-find takes advantage of the L1 and stale reads are benign. I measured L1-bypassing instructions at 27-215% slower. There's a comment explaining why for anyone who comes by down the road and thinks it's weird. Essentially, I just needed a non-tearing load and store but none of the other stuff that tends to come along with something like an atomic.

PR #3123, opened 2026-08-25
Use l-values to initialize RMM constructors

divyegala:
https://github.com/rapidsai/rmm/pull/2527 disallowed using r-values for aysnchronous constructors, so this PR fixes an existing bug.


cpp/tests/linalg/dot.cu
```diff
@@ -63,7 +63,8 @@ class DotTest : public ::testing::TestWithParam<DotInputs<T>> {
     uniform(handle, r, x.data(), x_len, T(-1.0), T(1.0));
     uniform(handle, r, y.data(), y_len, T(-1.0), T(1.0));
 
-    rmm::device_scalar<T> ref(0, resource::get_cuda_stream(handle));
+    T zero = 0;
+    rmm::device_scalar<T> ref(zero, resource::get_cuda_stream(handle));
```
bdice, line 67: This is a special case where you can use a different pattern if you want.
```suggestion
    rmm::device_scalar<T> ref(resource::get_cuda_stream(handle));
    ref.set_value_to_zero_async(resource::get_cuda_stream(handle));
```

cjnolet (approved):

cpp/tests/linalg/dot.cu
```diff
@@ -77,7 +78,7 @@ class DotTest : public ::testing::TestWithParam<DotInputs<T>> {
     raft::update_host(&ref_output, ref.data(), 1, stream);
 
     // Test out both the device and host api's
-    rmm::device_scalar<T> out(0, resource::get_cuda_stream(handle));
+    rmm::device_scalar<T> out(zero, resource::get_cuda_stream(handle));
```
bdice, line 81: If you adopt the above, adopt this too:
```suggestion
    rmm::device_scalar<T> out(resource::get_cuda_stream(handle));
    out.set_value_to_zero_async(resource::get_cuda_stream(handle));
```

bdice (approved):
LGTM. I shared one alternative pattern that you can adopt if you wish. Up to you.

cpp/tests/util/device_atomics.cu
```diff
@@ -38,7 +38,8 @@ TEST(Raft, AtomicIncWarp)
   rmm::cuda_stream_pool pool{1};
   auto s = pool.get_stream();
 
-  rmm::device_scalar<int> counter{0, s};
+  int zero = 0;
+  rmm::device_scalar<int> counter{zero, s};
```
wence-, line 42: These changes fix the compile error, but, from a cursory look (unless raft::launch_kernel synchronises the stream) do not fix the latent bug. The stream still needs to be synced after the copy is queued up

divyegala: Device to Host memory is synchronous https://docs.nvidia.com/cuda/cuda-runtime-api/api-sync-behavior.html#api-sync-behavior__memcpy-async

But yeah, the stream sync should probably be explicit so that we don't rely on hidden mechanisms.

wence-: No, "might" be synchronous if you're doing memcpy async (or, in rmm) batch memcpy async

divyegala: You are right, I misread it. Thanks!

PR #3134, opened 2026-09-03
Fix striding in mdspan 1d copies

divyegala:
I accidentally stumbled upon this bug, while trying to convert a 4096 1D mdspan from `int32` to `int64`. This approach leads to `0 + 1 + … + 4095 = 8,386,560` iterations on the device, leading to incredibly degenerate performance.

`raft::copy` uses a CUDA kernel for type-converting mdspan copies. That kernel converts each flat thread index into an mdspan coordinate.

For a 1-D vector of four elements, the old code finds element `3` by repeatedly subtracting the stride:

```text
increment = 3, stride = 1

3 → 2 → 1 → 0   # three loop iterations
```

Copying all four elements performs `0 + 1 + 2 + 3 = 6` iterations just to recover coordinates. For an `N`-element vector, this grows to roughly `N² / 2`.

Replace the repeated-subtraction loop with division and multiplication:

```text
index = increment / stride
increment -= index * stride
```

This makes coordinate calculation constant-time per element while preserving the existing general mdspan copy behavior. 

In general, the previous subtraction approach is leads to degenerate performance in N-D matrices when one or few dimensions are far larger than others.

### DIFF
--- a/cpp/include/raft/core/detail/copy.hpp
+++ b/cpp/include/raft/core/detail/copy.hpp
@@ -303,12 +303,10 @@ __device__ auto increment_indices(IdxType* indices,
       }
     }(i);
 
-    auto cur_index = IdxType{};
-
-    while (cur_index < md.extent(real_index) - 1 && increment >= index_strides[real_index]) {
-      increment -= index_strides[real_index];
-      ++cur_index;
-    }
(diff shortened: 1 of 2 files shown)


achirkin (approved):
LGTM

PR #3136, opened 2026-09-05
Adopt CUDA stream compatibility accessors

bdice:
## Summary

This PR helps RAFT adopt `cuda::stream_ref`. It is a non-breaking subset of the changes in #3129 to ease migration.

Use the `get()` and `sync()` compatibility aliases added in [RMM #2537](https://github.com/rapidsai/rmm/pull/2537). These spellings are shared by `rmm::cuda_stream_view` and `cuda::stream_ref`.

This preserves existing stream types and public APIs while extracting mechanical accessor updates from the broader [stream migration](https://github.com/rapidsai/build-planning/issues/318). It is independently buildable without [RMM #2372](https://github.com/rapidsai/rmm/pull/2372) and leaves the migration PR focused on actual type and signature changes.

This updates raw CUDA, library, kernel-launch, and legacy API boundaries throughout RAFT and uses the CUDA stream umbrella header where stream definitions are required. Existing `rmm::cuda_stream_view` return types remain unchanged; their migration stays in [RAFT #3129](https://github.com/NVIDIA/raft/pull/3129).


bdice:
Validated this preparation branch against RMM `upstream/main`: the local `-j0` build completed and focused select-k tests passed. The dependent migration PR #3129 has been rebased onto this branch.

bdice:
CI exposed benchmark targets that the default local build did not compile. The preparation branch now adds the missing raw-stream accessors at those benchmark boundaries (`f356e946`). Validation:

- `build-raft -j0` passes against the current RMM dependency
- all primitive benchmark targets pass with `ninja -C cpp/build -j0` after configuring `bench-prims` against RMM `main`
- focused pre-commit hooks pass with copyright-only churn excluded

bdice:
Extracted the review follow-up into this preparation stack as commit 51a23751: raw `cudaStream_t` member initialization, strided-reduction native-handle forwarding, and the debug-only `RAFT_CHECK_CUDA` boundary. `build-raft -j0` passes against RMM `upstream/main`.

bdice:
One additional extraction: the benchmark `.sync()` compatibility change now also lives here at e40580b4. All primitive benchmark targets compile with `ninja -C cpp/build -j0`; migration #3129 is rebased to ae7eee06 with the same validated source tree.

divyegala (approved):

PR #3139, opened 2026-09-08
CI: Use isolated build for all wheel packages

robertmaynard:
Replace pip wheel with isolated python -m build --wheel in the shared wheel-build helper used by every wheel build job


vyasr (approved):
Needs copyright updates then good to go.

msarahan (approved):

PR #3149, opened 2026-09-10
Add stream pool configuration to `device_resources_snmg`

viclafargue:
Adds `device_resources_snmg::set_stream_pool()` to configure a CUDA stream pool on every managed GPU. Sorry for opening this PR during burndown, but this should target `release/26.10`.


tarang-jain (approved):

cpp/include/raft/core/device_resources_snmg.hpp
```diff
@@ -113,6 +116,24 @@ class device_resources_snmg : public device_resources {
     RAFT_CUDA_TRY(cudaSetDevice(main_gpu_id_));
   }
 
+  /**
+   * @brief Set a CUDA stream pool on all GPUs of the multi-GPU world
+   *
```
tfeher, line 121: Note that the single GPU version states the following in the docstring, might make sense to add that here as well
> Note that this will overwrite an existing stream pool on the res.

tfeher (approved):
Looks good to me. Just a small comment, but that need not to hold off the PR.

